PR #416, opened 2014-12-20
Optimize the fig dev env/build

### LINKED ISSUE #344: Use persistent data container for panoptes fig container
Try and get panoptes container building into a persistent data volume so the installed gems are persisted between container launces.

camallen:
Closes #344 - not ready to merge just yet, can someone check the README instructions make sense and work before merging?

So i tried using a persistent data volume beneath panoptes but `fig up` would recreate the containers on each launch, whic was no better than the existing setup with `fig up --no-recreate`. The solution I've settled on is to switch in a development dockerfile (with dev/test gems) before `fig build` to ensure all gems are installed. Then just manually migrate and seed the db before using `fig up`. This alleviates the need to install gems / migrate the DB on each run and we can recreate containers each time with minimal work to get started.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,6 @@ RUN apt-get update && apt-get -y upgrade && \
     bundle install --without test development && \
     jbundle install --without test development
 
-RUN touch /firstrun
-
 ADD ./ /rails_app
 
 EXPOSE 80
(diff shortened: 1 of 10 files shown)


camallen:
Closing in favour of #445

Issue #375, opened 2014-12-08
Cannot create Workflow


mkosmala:
I simplified down the example payload to Create Workflow to this, but still cannot create it. I get an Error 422 "The change that you want was rejected." I created the project (id 60) I am trying to add the Workflow to, but I suspect there's some missing permission info somewhere.
```python
    values = """
    {
        "workflows": {
            "display_name": "Do some round stuff",
            "tasks":  {
                "roundness": {
                    "type": "single",
                    "question": "How round is it?",
                    "answers": [
                        {"value": "very", "label": "Very"},
                        {"value": "sorta", "label": "In between"},
                        {"value": "not", "label": "Cigar shaped"}
                    ],
                    "next": null
                }
            },
            "primary_language": "en-us",
            "links": {
                "project": "60"
            }
        }
    }"""
```
I'm sending the POST to:
```
https://panoptes-staging.zooniverse.org/api/workflow
```

edpaget:
Closed by #418

Issue #370, opened 2014-12-08
Cannot view group member's membership state


mkosmala:
It's not clear to me whether as an 'active' member of my group whether or not I should be able to at least see what the other membership types of the other members of my group are. I think it makes sense to allow that. Currently, if I request information about the membership of someone (other than myself) in my group, I get an empty payload back.

edpaget:
Closed by #418

PR #418, opened 2015-01-02
Better Permissions System

### LINKED ISSUE #381: Cannot create Subject
When I try to create a Subject, I get Error 404 with message body {"errors":[{"message":"Couldn't find resource"}]}

Here's what I am sending:
```python
    {
        "subjects": {
            "locations": {
                "standard": "image/jpeg"
            },
            "metadata": {
                "Site": "Harvard Forest",
                "Vegetation": "Deciduous Broadleaf Forest",
                "Latitude": "42.5378",
                "Longitude": "-72.1715",
                "Elevation": "340",
                "Camrea": "StarDot NetCam SC"
            },
            "links": {
                "project": "63"
            }
        }
    }
```
to
```
https://panoptes-staging.zooniverse.org/api/subjects
```

edpaget:
Hey all, I've reworked how the permissions system operates making it much simpler with less Arel hackery. 

Basically, instead of having multiple ways to determine if a user is allowed a access a project, (Is the user the owner, does the user have the right roles, is the user part of a user group that owns the project, etc), now all access is done through user groups and being a "owner" is just another type of role. 

Pros of this approach:
* Much simpler code paths
* Reduced the amount of code
* No using private ActiveRecord APIs
* Minimal API change (the only change is not having to a post an "owner" link for Subjects)
* No longer tied to Rails 4.1.7
* Closes #381, #375, #370 

Cons:
* Union queries may be expensive. I know a bunch can be replaced with a Union All query
* I think it's less simple conceptually
* Project and Collection name validation could be subject to a race condition


edpaget:
Travis is building our project with the wrong version of JRuby (JRuby 9000 instead of 1.7). I'm trying to figure out why that's happening.

edpaget:
ugh https://github.com/travis-ci/travis-ci/issues/3067

edpaget:
img_20150105_132139 I diagrammed the before/after.

camallen:
Nice work @edpaget, 'll have to look at this in greater detail when i'm back from holidays. Let me know if this is urgent and blocking other work.

edpaget:
No rush, but it does make other changes significantly easier as I found in #427. Also if it diverges too much from master the rebase is going to be really fun to handle.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM zooniverse/ruby:jruby-1.7.18
+FROM zooniverse/ruby:2.1.5
```
camallen: Any reason for switching to MRI ruby on deploy?

app/controllers/api/v1/classifications_controller.rb
```diff
@@ -16,32 +15,43 @@ class Api::V1::ClassificationsController < Api::ApiController
 
   private
 
+  def access_denied(exception)
```
camallen: Perhaps refactor to something like:

``` ruby
def access_denied(exception)
  if %w(update destroy).include?(action_name) && completed_classification?
    error = StandardError.new(completed_error)
    not_authorized(error) and return
  end
  not_found(exception)
end

def completed_classification?
  Classification.created_by(api_user.user)
    where(id: resource_ids, completed: true)
    .exists?
end
```

app/controllers/api/v1/classifications_controller.rb
```diff
@@ -66,4 +76,12 @@ def create_params
   def update_params
     params.require(:classifications).permit(create_update_params)
   end
+
+  def completed_error
+    if resource_ids.is_a?(Array)
+      "Classifications with ids='#{resource_ids.join(',')}' are complete"
```
camallen: Perhaps use `controller_name` instead of manually naming the controller model

Dockerfile
edpaget: Whoops I was trying to chase down an MRI specific bug yesterday. I'll
revert.

spec/models/access_control_list_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+RSpec.describe AccessControlList, :type => :model do
```
camallen: Remove this if we're testing the ACL indirectly

edpaget:
@camallen I think I fixed all the orphans ACLs.

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -15,10 +12,18 @@ def show
 
   private
 
+  def context
```
camallen: seems the `context` method is defined twice in this class

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -15,10 +12,18 @@ def show
...
+    else
+      {}
+    end
+  end
+
   def create_response(workflow)
     serializer.resource({},
-                        resource_scope(workflow),
-                        languages: [workflow.primary_language])
+                        resource_scope(workflow))
```
camallen: Any reason why the context isn't on the create_response?

edpaget: Actually I think these methods can be entirely removed since they no longer have class specific behaviour.

lib/json_api_controller/updatable_resource.rb
```diff
@@ -11,38 +11,42 @@ module UpdatableResource
     end
 
     def update
-      ActiveRecord::Base.transaction do
-        build_resource_for_update(update_params)
-        controlled_resource.save!
+      resource_class.transaction do
```
camallen: Should we protect against  side effects in nested transactions like create? `resource_class.transaction(requires_new: true)`

lib/json_api_controller/creatable_resource.rb
```diff
@@ -23,9 +30,29 @@ def build_resource_for_create(create_params)
       if block_given?
         yield create_params, link_params
       end
-      @controlled_resource = resource_class.new(create_params)
-      link_params.try(:each) { |k,v| update_relation(k,v) }
-      controlled_resource
+      resource = resource_class.new(create_params)
+
+      link_params.try(:each) do |k,v|
+        #http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one
```
camallen: I don't think these comments have too much value here.

app/models/classification.rb
```diff
@@ -17,26 +14,32 @@ class Classification < ActiveRecord::Base
...
+  scope :complete, -> { where(completed: true) }
+
+  def self.scope_for(action, actor, opts={})
+    case action
+    when :show, :index
+      query = joins(:project).merge(Project.scope_for(:update, actor))
+        .union_all(joins(:user_group).merge(actor.user_groups))
+        .union_all(actor.classifications)
+      # Workaround Broken Bind Value Assignment in Subqueries in Rails 4.1
+      # This is fixed in Rails 4.2 when we're able to to migrate to that
```
camallen: Are we still stuck on rails 4.1 with the PR, sounded like removing hte AREL hackery freed us to bump rails?

app/models/concerns/kafka_event.rb
```diff
@@ -14,6 +14,7 @@ def kafka_event(kafka_key, topic: 'events', attributes: [], links: [])
     end
 
     def publish_to_kafka(topic, key, *ids)
+      #where(id: ids).explain
```
camallen: clean up explain

app/models/classification.rb
edpaget: When Rails 4.2.1 is out we should be good https://github.com/rails/rails/issues/18414

app/models/owner_name.rb
```diff
@@ -1,16 +0,0 @@
-class OwnerName < ActiveRecord::Base
```
camallen: This model is still in the schema.

app/models/user.rb
```diff
@@ -91,6 +111,17 @@ def email_changed?
     authorizations.blank?
   end
 
+  def build_identity_group
+    raise StandardError, "Identity Group Exists" if identity_group
```
camallen: Should we use a custom error here? Also it looks like this is never tested in the specs (removing this line and all pass).

camallen:
@edpaget this looks good! 

However, merging this will break the current panoptes data. Either we migrate our way around it (ACL's and IdentityGroups for all users) or we just blow it all away and re-seed from the front end and an updated Galaxy Zoo migration script.

edpaget:
I think we should probably just blow everything up and redo it.

camallen:
Feel free to merge and recreate the DB.

edpaget:
Will do just addressing some of the remaining comments.

app/models/user.rb
edpaget: There really shoudn't be a way for build_identity_group to be invoked twice, so I didn't think it needed a custom error class. I will add a spec for it though.

PR #428, opened 2015-01-08
Create subject sets from collections

### LINKED ISSUE #297: Import Collections as SubjectSets

edpaget:
Close #297

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -14,7 +14,6 @@ module Api
     rescue_from Api::UnauthorizedTokenError,             with: :not_authenticated
     rescue_from Api::UnsupportedMediaType,               with: :unsupported_media_type
     rescue_from ControlControl::AccessDenied,            with: :not_authorized
-    rescue_from ActiveRecord::StatementInvalid,          with: :bad_query
     rescue_from Api::PatchResourceError,                 with: :unprocessable_entity
     rescue_from Api::UserSeenSubjectIdError,             with: :unprocessable_entity
     rescue_from ActionController::UnpermittedParameters, with: :unprocessable_entity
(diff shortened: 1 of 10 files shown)


app/controllers/api/api_controller.rb
```diff
@@ -14,7 +14,6 @@ class ApiController < ApplicationController
     rescue_from Api::UnauthorizedTokenError,             with: :not_authenticated
     rescue_from Api::UnsupportedMediaType,               with: :unsupported_media_type
     rescue_from ControlControl::AccessDenied,            with: :not_authorized
-    rescue_from ActiveRecord::StatementInvalid,          with: :bad_query
```
camallen: I assume Json Schema Validator is checking id formats now so we don't need to rescue this one explictly?

app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -5,6 +5,14 @@ class Api::V1::SubjectSetsController < Api::ApiController
...
   schema_type :json_schema
 
-  allowed_params :create
-  allowed_params :update
+  protected
+
+  def build_resource_for_create(create_params)
+    super do |_, link_params|
+      if collection_id = link_params.delete("collection")
+        link_params["subjects"] = Collection.scope_for(:show, api_user)
```
camallen: Will this return nil or raise and error if the `api_user` has no scope to `view` the collection?

app/controllers/api/api_controller.rb
edpaget: The schema is just checking types. The other thing checks if its an
alphabet string or number.

app/controllers/api/v1/subject_sets_controller.rb
edpaget: I'll fix this, it should probably just be following the normal `link_to` code path.

edpaget: Couldn't figure out a good way to make it use the same code, but it's now a find that will return the same 404 error that a link failure would. (I think #418 changes the specific error code).

PR #430, opened 2015-01-09
Add support for Conditional Requests

### LINKED ISSUE #296: Optimistic Locking using Versions
We should add locking for resources that can be collaboratively edited.

edpaget:
Using Last-Modified. Closes #296

### DIFF
--- a/lib/json_api_controller/indexable_resource.rb
+++ b/lib/json_api_controller/indexable_resource.rb
@@ -1,6 +1,7 @@
 module JsonApiController
   module IndexableResource
     def index
+      fresh_when last_modified: visible_scope.maximum(:updated_at)
       render json_api: serializer.page(params, visible_scope, context)
     end
   end
--- a/lib/json_api_controller/precondition_check.rb
+++ b/lib/json_api_controller/precondition_check.rb
@@ -10,9 +10,7 @@ module JsonApiController
     
     def precondition
       case action_name
-      when "show"
-        request.headers["If-Modified-Since"]
-      else
+      when "update", "destroy"
         request.headers["If-Unmodified-Since"]
       end
     end
--- a/lib/json_api_controller/precondition_check.rb
+++ b/lib/json_api_controller/precondition_check.rb
@@ -1,0 +1,28 @@
+module JsonApiController
+  module PreconditionCheck
+    def precondition_check
+      if !precondition
+        raise PreconditionNotPresent, precondition_error_msg
+      elsif precondition_fails?
+        raise PreconditionFailed
+      end
+    end
+    
+    def precondition
+      case action_name
+      when "show"
+        request.headers["If-Modified-Since"]
+      else
+        request.headers["If-Unmodified-Since"]
+      end
+    end
+
+    def precondition_fails?
+      !resource_class.where("updated_at <= ?", precondition).exists?(params[:id])
+    end
+
+    def precondition_error_msg
+      "Request requires #{precondition} header to be present"
+    end
+  end
+end
--- a/apiary.apib
+++ b/apiary.apib
@@ -19,6 +19,15 @@ follow a common convention.
 Panoptes API provides an OAuth 2.0 authentication mechanism. (TODO:
 Describe Authentication Workflow).
 
+# Headers
+Panotpes API sometime requires specific headers to be sent as part of requests.
+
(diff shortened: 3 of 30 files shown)


lib/json_api_controller/precondition_check.rb
```diff
@@ -0,0 +1,28 @@
...
+      if !precondition
+        raise PreconditionNotPresent, precondition_error_msg
+      elsif precondition_fails?
+        raise PreconditionFailed
+      end
+    end
+    
+    def precondition
+      case action_name
+      when "show"
```
camallen: Are we checking the modified since header timestamp for show get requests anywhere? By the look it we're only doing this for update / delete requests.

lib/json_api_controller/precondition_check.rb
```diff
@@ -0,0 +1,28 @@
...
+        request.headers["If-Modified-Since"]
+      else
+        request.headers["If-Unmodified-Since"]
```
camallen: Is the header being used consistent as per [here](https://github.com/zooniverse/Panoptes/issues/296#issuecomment-69229832)? Do we need to update the issue to reflect which header should be set client side?

lib/json_api_controller/indexable_resource.rb
```diff
@@ -1,6 +1,7 @@
 module JsonApiController
   module IndexableResource
     def index
+      fresh_when last_modified: visible_scope.maximum(:updated_at)
```
camallen: Nice idea re knowing what has changed in the index collection.

camallen:
Looks good - I think this will work!

lib/json_api_controller/precondition_check.rb
edpaget: We could. I think it'd be a little hostile for API exploration, which is why I didn't require them. I can also get rid of this condition since its not being used.

lib/json_api_controller/precondition_check.rb
edpaget: No it's not the right header in the issue, I hadn't finished reading the precondition spec when I pitched it. I documented the correct header in Apiary.

edpaget:
Also this should probably actually lock the row on when updating/deleting.

edpaget:
This now actually locks the record when updating for models most likely to updated by multiple users.

lib/json_api_controller/precondition_check.rb
camallen: Sorry, I meant are we using the show action when checking the header preconditions? It seemed like we are only checking preconditions on the update / delete actions, so the show is not needed.

lib/json_api_controller/precondition_check.rb
camallen: Can you update the other issue so @brian-c knows which header to use (in case he doesn't read the apiary docs).

camallen:
@edpaget - it seems Rails's Optimisitic locking is only good in a [single process](http://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html):
>This locking mechanism will function inside a single Ruby process. To make it work across all web requests, the recommended approach is to add lock_version as a hidden field to your form.

I couldn't find much info about enabling this across processes, any thoughts about making it work across multple API nodes?

edpaget:
I don't think that's quite correct. I think when you have optimistic locking enabled, it should issue a `UPDATE .. SET .. WITH lock_version = 1` if another processes writes to the table before that statement the `lock_version` column will get incremented and the update will fail. I'm not sure what exactly that documentation means.

lib/json_api_controller/precondition_check.rb
edpaget: I'll take it out.

lib/json_api_controller/precondition_check.rb
edpaget: Done.

edpaget:
Yeah that text isn't present in the [Rails Guide](http://guides.rubyonrails.org/active_record_querying.html#optimistic-locking). I think it's talking about another scenario that's taken care of by our Conditional headers. 

The Docs are talking about a more traditional Rails situation where two users make a request against `/resource/1/edit` and then one submits the form before the other. The first submitter's edits get made, then the second submitter's edit happen through though their form had out of date data. You can include the lock_version hidden field in the form to prevent that kind of error from happening. In our case that's handled by the 'If-Unmodified-Since` header.

camallen:
That makes sense @edpaget, this looks good to merge.

Issue #212, opened 2014-10-20
Project and Workflow JSON blobs


edpaget:
I'd like to start working on the versioning issues #206 and #47, but I'm still not sure what those resources will end up looking like. 

I think the blobs in question are:
- Project
  - Pages
  - Example Guide(?)
- Workflow
  - Tasks

I just need to know what they'll generally be like so I can extract strings for the translations features. @brian-c do you have any examples of what will be in them?

PR #441, opened 2015-01-20
Visit by class name instead of key

edpaget:
Visit by class so we can be more flexible in defining workflows
Maybe revisit (heh) when the format is more locked down

This should fix the 500 errors @brian-c was seeing.

### DIFF
--- a/lib/tasks_visitor.rb
+++ b/lib/tasks_visitor.rb
@@ -1,13 +1,13 @@
 class TasksVisitor
   def visit(node, path=[], key=nil)
     method = :"visit_#{key}"
-    method = respond_to?(method, true) ? method : :visit_hash
+    method = respond_to?(method, true) ? method : :"visit_#{node.class.to_s.gsub('::', '_')}"
     send(method, node, path)
   end
 
   private
 
-  def array_node(n, path)
+  def visit_Array(n, path)
     n.each_with_index.map { |sub_node, i| visit(sub_node, path + [i]) }
   end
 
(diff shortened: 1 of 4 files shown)


lib/tasks_visitor.rb
```diff
@@ -15,17 +15,19 @@ def noop(n, p)
...
   end
 
-  def visit_hash(n, path)
+  def visit_Hash(n, path)
     n.each do |k, v|
       n[k] = visit(v, path + [k], k)
     end
   end
+  
+  alias :visit_ActiveSupport_HashWithIndifferentAccess :visit_Hash
```
camallen: Is this alias list exhaustive of the expected types? If not what about using `method_missing` to dispatch the relevant class method?

brian-c:
I didn't realize the back end was actually reading anything in the workflow's tasks object. This is still a bit in flux since I've started work on the editor. Can we not care about what's in that until it's finalized?

camallen:
@brian-c, we're reading the worfklow contents for create / update AND when serialising the workflow api responses. The original work was based on https://github.com/zooniverse/Panoptes/issues/212#issuecomment-59781884.

lib/tasks_visitor.rb
edpaget: That should be an exhaustive list of types its possible to get back from
json.

edpaget:
I want to make sure the translation stuff works right away. This code
should make sure that it didn't care what's in the workflow now (as opposed
to relying on the json object's key names as it was before) and just looks
for specific fields it knows to translate.

PR #444, opened 2015-01-21
Update to newest Cellect and MRI ruby

edpaget:


edpaget:
Awesome jruby-1.7 series can't handle the new keyword argument syntax. I'll set aside some time in the future to figure out what's breaking in jruby 9000 and see if we can use that.

PR #445, opened 2015-01-21
Better fig

edpaget:
Supersedes #416. Adds my [fig_rake](https://github.com/edpaget/fig_rake) gem to manage interaction with containers, which makes stuff a lot simpler. You can almost treat Panoptes running in fig exactly the same as if it was running locally. 

I did make some changes to the Dockerfile that we should think about, namely installing all dependencies during the build instead of excluding the test and development ones. I'm not sure it really matters having them installed in production. Thoughts?


edpaget:
I should also note that I'm using FigRake's monkey patching capabilities. We could rework it to use the `frake` and `frails` commands instead.

config/database.yml.hudson
```diff
@@ -3,19 +3,17 @@ default: &default
...
   <<: *default
-  host: postgres
   database: panoptes_development
 
 test:
-  adapter: postgresql
-  database: travis_ci_test
-  username: postgres
+  <<: *default
+  database: panoptes_development
```
camallen: Keep the test DB explicitly different, `panoptes_test`.

README.md
```diff
@@ -40,90 +40,34 @@ An easy way to get the full Panoptes stack running (see `fig.yml` to dig into th
...
 
-Finally there are some helper scripts to get access to a console, bash shell etc. **Note:** these commands build a new run container
-* To get a rails console `scripts/fig/rails_console_panoptes.sh`
-  + **Note:** you can override the RAILS_ENV by passing a valid argument, just make sure you've setup the DB for it!
-* To get a bash console `scripts/fig/run_cmd_panoptes.sh bash`
-* You can also attach a bash process to the running container, e.g. `docker exec -it panoptes_panoptes_1 bash`
-  + Assuming the 'panoptes_panoptes_1' container is running, use `fig ps` or `docker ps` to check.
-
-**Note:** if you've ever built a Panoptes docker container before you should just run `fig up` instead of the `./scripts/fig/up_panoptes.sh` to ensure the previously built container is not re-used. After rebuilding you should be good to use `./scripts/fig/up_panoptes.sh` script to use the re-created containers.
+If you've added new gems you'll need to rebuild the docker image via the command in step 4.
```
camallen: Step 4 doesn't reference a fig build step. Best to make and explicit reference here.

README.md
```diff
@@ -40,90 +40,34 @@ An easy way to get the full Panoptes stack running (see `fig.yml` to dig into th
...
   * [Windows](http://docs.docker.com/installation/windows/) - Boot2Docker
 
-#### Installation
+* [fig](http://fig.sh)
 
-1. Ensure your repo directory starts with a lowercase letter - you may need to move `/Panoptes` to `/panoptes`.
+#### Usage
 
-2. Prepare your fig development environment config files. You should only have to do this before the first boot. **Note:** The fig docker environment uses linked docker containers, so your Postgres and Zookeeper hosts urls need to refer to these containers.
+1. Run `rake configure FIG_RAKE=off`
```
camallen: Should we put more context about what these steps are trying to achieve. E.g. "1. Setup the application configuration files, run `rake configure FIG_RAKE=off`"

And likewise for steps 2 and 3.

camallen:
I'm digging on this, my only concern is building the docker image with test and dev gems which may not really matter apart from time constraints on build. 

I'm testing it out on my machine to see how well it fits normal dev practises as well.

edpaget:
Sorry about that database mistake. I've update that config file as well as adding more explanations to the README steps.

edpaget:
@camallen re: build times. They get installed as part of the Docker image build which doesn't have the same timeout constraints as the AMI build. I also think the dominating time factor in new builds is that AWS sometimes takes hours to actually make the image.

README.md
```diff
@@ -40,90 +40,33 @@ An easy way to get the full Panoptes stack running (see `fig.yml` to dig into th
...
   * [Windows](http://docs.docker.com/installation/windows/) - Boot2Docker
 
-#### Installation
+* [fig](http://fig.sh)
 
-1. Ensure your repo directory starts with a lowercase letter - you may need to move `/Panoptes` to `/panoptes`.
+#### Usage
 
-2. Prepare your fig development environment config files. You should only have to do this before the first boot. **Note:** The fig docker environment uses linked docker containers, so your Postgres and Zookeeper hosts urls need to refer to these containers.
+1. Setup the application configuration files by running `rake configure FIG_RAKE=off`
```
camallen: What do you think about reversing the behaviour of FIG_RAKE, i.e. having it to be off by default and forcing the user to understand and opt in? Our use is only for dev and we could easily set and env variable to activate it in dev mode.

camallen:
I've reworked the dockerfile to allow faster gem modifications and docker image rebuilds https://github.com/edpaget/Panoptes/pull/11

README.md
edpaget: I think that's fine. I'm going to allow this variable to set from a `.fig_rake` file as well.

edpaget:
I merged your pull request.

README.md
camallen: By this I meant reversing the behaviour of the gem for all monkey patched commands, a la [here](https://github.com/edpaget/fig_rake/blob/0387991954ff88db057041812b79157c1258b24a/lib/fig_rake/rails.rb#L31)

``` ruby
define_method method do
  if ENV['FIG_RAKE'] == 'on'
    run_in_fig(argv.unshift(method))
  else
    m.bind(self).call
  end
end
```

edpaget:
I bumped the version of Fig Rake.

PR #452, opened 2015-01-26
Make GET methods Public for User and User Groups

### LINKED ISSUE #451: Not authorized to get a project's owner link
This should probably be public.

edpaget:
Closes #451


app/controllers/api/v1/groups_controller.rb
```diff
@@ -1,8 +1,9 @@
 class Api::V1::GroupsController < Api::ApiController
   include JsonApiController
 
-  doorkeeper_for :index, :create, :show, scopes: [:public]
```
camallen: Are we exposing the collection of user groups to not logged in users?

edpaget: Yes. UserGroup admins can set the visibility of their group (by setting visible_to in this verison or private: true after #418), so if they don't want their groups to be visible they have the option of hiding them.

PR #458, opened 2015-01-28
Allow Subject Ids to be sent as a link to Classification

### LINKED ISSUE #449: Post to /api/classifications errors
With this data, "found unpermitted parameters: subject":

```
{
  "classifications":{
    "links":{
      "project":"73",
      "workflow":"44",
      "subject":"365314"
    },
    "metadata":{
      "started_at":"2015-01-26T11:37:54.805Z",
      "user_agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2270.0 Safari/537.36",
      "user_language":"en",
      "workflow_version":"446.353",
      "finished_at":"2015-01-26T11:37:58.991Z"
    },
    "annotations":[
      {
        "present":false
      }
    ],
    "complete":true
  }
}
```

edpaget:
Closes #449

### DIFF
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -36,10 +36,23 @@ class Api::V1::ClassificationsController < Api::ApiController
     classification
   end
 
+  def sms_ids_from_links(link_params)
+    sms = if ids = link_params.delete(:set_member_subjects)
+      SetMemberSubject.select(:id).find(ids)
+    elsif ids = link_params.delete(:subjects)
+      SetMemberSubject.joins(:subject_set)
+        .where(subject_sets: {workflow_id: link_params[:workflow]})
+        .select(:id)
+        .find_by(subject_id: ids)
+    end
+    sms.try(:map, &:id) || [sms.id]
+  end
+
   def build_resource_for_create(create_params)
     classification = super(create_params) do |create_params, link_params|
       link_params[:user] = api_user.user
       create_params[:user_ip] = request_ip
+      create_params[:set_member_subject_ids] = sms_ids_from_links(link_params)
     end
     lifecycle(:create, classification)
     classification
@@ -69,7 +82,8 @@ class Api::V1::ClassificationsController < Api::ApiController
   def create_params
     param_set = create_update_params | [ links: [:project,
                                                  :workflow,
-                                                 :set_member_subject] ]
+                                                 set_member_subjects: [],
+                                                 subjects: []] ]
     params.require(:classifications).permit(param_set)
   end
 
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -4,7 +4,6 @@ class SetMemberSubject < ActiveRecord::Base
   
   belongs_to :subject_set, counter_cache: true, touch: true
   belongs_to :subject
-  has_many :classifications
 
   enum state: [:active, :inactive, :retired]
 
--- a/app/models/user_subject_queue.rb
+++ b/app/models/user_subject_queue.rb
@@ -8,8 +8,8 @@ class UserSubjectQueue < ActiveRecord::Base
 
   can_through_parent :workflow, :update, :destroy, :update_links, :destroy_links
 
-  delegate :add_subject, to: :set_member_subjects
-  delegate :remove_subject, to: :set_member_subjects
+  delegate :add_subjects, to: :set_member_subjects
+  delegate :remove_subjects, to: :set_member_subjects
 
   def self.scope_for(action, groups, opts={})
     case action
@@ -22,17 +22,17 @@ class UserSubjectQueue < ActiveRecord::Base
 
   def self.enqueue_subject_for_user(user: nil, workflow: nil, set_member_subject: nil)
     ues = find_or_create_by!(user: user, workflow: workflow)
-    ues.add_subject(set_member_subject)
+    ues.add_subjects(set_member_subject)
   end
 
-  def self.dequeue_subject_for_user(user: nil, workflow: nil, set_member_subject: nil)
+  def self.dequeue_subjects_for_user(user: nil, workflow: nil, set_member_subject_ids: nil)
     ues = find_by!(user: user, workflow: workflow)
-    ues.remove_subject(set_member_subject)
+    ues.remove_subjects(set_member_subject_ids)
     ues.destroy if ues.set_member_subject_ids.empty?
   end
 
-  def self.is_subject_queued?(user: nil, workflow: nil, set_member_subject: nil)
-    where.any(set_member_subject_ids: set_member_subject.id)
+  def self.are_subjects_queued?(user: nil, workflow: nil, set_member_subject_ids: nil)
+    where.overlap(set_member_subject_ids: set_member_subject_ids)
       .exists?(user: user, workflow: workflow)
   end
   
--- a/db/migrate/20150128160822_add_subject_ids_and_sms_ids_to_classifications.rb
+++ b/db/migrate/20150128160822_add_subject_ids_and_sms_ids_to_classifications.rb
@@ -1,0 +1,6 @@
+class AddSubjectIdsAndSmsIdsToClassifications < ActiveRecord::Migration
+  def change
+    remove_reference :classifications, :set_member_subject, index: true
+    add_column :classifications, :set_member_subject_ids, :integer, array: true, default: []
+  end
+end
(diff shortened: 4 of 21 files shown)


app/controllers/api/v1/classifications_controller.rb
```diff
@@ -36,10 +36,23 @@ def build_resource_for_update(update_params)
...
   end
 
+  def sms_ids_from_links(link_params)
+    sms = if ids = link_params.delete(:set_member_subjects)
+      SetMemberSubject.select(:id).find(ids)
+    elsif ids = link_params.delete(:subjects)
+      SetMemberSubject.joins(:subject_set)
+        .where(subject_sets: {workflow_id: link_params[:workflow]})
+        .select(:id)
+        .find_by(subject_id: ids)
```
camallen: Perhaps use pluck if we're just interested in the ids to avoid converting the DB records to AR models.

``` ruby
SetMemberSubject.joins(:subject_set)
   .where(subject_sets: {workflow_id: link_params[:workflow]})
   .where(subject_id: ids)
   .pluck(:id)
```

app/controllers/api/v1/classifications_controller.rb
```diff
@@ -36,10 +36,23 @@ def build_resource_for_update(update_params)
     classification
   end
 
+  def sms_ids_from_links(link_params)
+    sms = if ids = link_params.delete(:set_member_subjects)
+      SetMemberSubject.select(:id).find(ids)
```
camallen: likewise here `SetMemberSubject.where(id: ids).pluck(:id)`
then you wont have to try the array on line 48.

app/models/set_member_subject.rb
```diff
@@ -4,7 +4,6 @@ class SetMemberSubject < ActiveRecord::Base
   
   belongs_to :subject_set, counter_cache: true, touch: true
   belongs_to :subject
-  has_many :classifications
```
camallen: I think the main issue with losing this relation is that it had the classification counter cache on it. The plan was to use this to cache the number of raw classification for a SMS that was then to be used in simple retirement (first version). If we lost the counter cache then we'll have to query the classification table (possibly very large) and possibly costly as time goes on. 

We can just keep this relation and get all `set_member_subjects` for a `subject_set` and then sum the classification counts. Thoughts?

Finally if we do remove it, then the _count field in the schema should go to.

edpaget: I'm thinking of removing all the classification counter caches for
classifications and using the Kafka stream to calculate them instead.

The main reason for removing it is needing to handle multi subject
classifications which can't be supported through the normal rails relations
without another join table.

app/controllers/api/v1/classifications_controller.rb
edpaget: We need to raise a not found error if a subject id is sent that does match
the database. I can pluck and then test the result against the number we
were expecting. I was just using find since it did that automatically.

app/models/set_member_subject.rb
edpaget: Also processsing from Kafka saves us a bunch of INSERTs on each classification we save.

camallen: We'll need these counts for simple retirement in Stargazing. I was thinking background workers that would allow over classification based on incoming rates.

When you say multi-subject classifications, what do you mean?

We could use kafka service consumer to calculate counts. Where do we store them?

My statement before was incorrect re sum (on two levels), i was assuming a similar counter feature that doesn't without a belongs to.

app/controllers/api/v1/classifications_controller.rb
camallen: Meh - if we get the raise for free it really depends on the object creation time then after DB retrieval. I'll it up to you.

app/models/user_subject_queue.rb
```diff
@@ -22,17 +22,17 @@ def self.scope_for(action, groups, opts={})
 
   def self.enqueue_subject_for_user(user: nil, workflow: nil, set_member_subject: nil)
     ues = find_or_create_by!(user: user, workflow: workflow)
-    ues.add_subject(set_member_subject)
+    ues.add_subjects(set_member_subject)
   end
 
-  def self.dequeue_subject_for_user(user: nil, workflow: nil, set_member_subject: nil)
+  def self.dequeue_subject_for_user(user: nil, workflow: nil, set_member_subject_ids: nil)
```
camallen: Pluralise the method name?

app/models/set_member_subject.rb
edpaget: We need to support Pairwise classifications like the Sunspotters project at some point. This would get us out ahead like that. 

And yeah we could just add a task in the background worker that updates a counter and checks the retirement rules. I think moving towards using Kafka and Storm or Spark for most of our background tasks will make that migration easier in the future when have to. 

For this I'd probably make a quick Storm topology and store the counts in Redis or DynamoDB get something up and running quickly. The topology could also check retirement conditions.

app/controllers/api/v1/classifications_controller.rb
edpaget: Cool I'll leave as is then. We can always revisit if this proves to be a bottleneck.

camallen: I very much doubt it!

app/models/user_subject_queue.rb
```diff
@@ -22,17 +22,17 @@ def self.scope_for(action, groups, opts={})
 
   def self.enqueue_subject_for_user(user: nil, workflow: nil, set_member_subject: nil)
```
camallen: Thought: I assume the front end will only really know about subjects here so when submitting a subject to the queue, it'll probably be similar to the classification submitting subject_ids.

app/models/set_member_subject.rb
camallen: Do you think we'd have a storm topology ready and load tested for stargazing (March 20th so really late feb early march)? It adds a little overhead to the infrastructure. 

I assume the storm spout would be the kafka consumer for 'classification' topic? How do you handle storm failures and starting from the correct topic offset to ensure we get correct counts?

I think the most basic working version of this is a background worker (classification lifecycle) that manually inc's all the classification counts for the subject SMS's for that workflow.

camallen:
Looks good, the only thing is to figure out the simplest way to retire subjects for stargazing.

app/models/set_member_subject.rb
edpaget: Storm/Kafka integration is part of the Storm project, and it automatically handles kafka failures and offset tracking. 

Let me knock out a demo of the storm topology (I think it would take less than a day), and we can figure it out from there.

db/migrate/20150128160822_add_subject_ids_and_sms_ids_to_classifications.rb
```diff
@@ -0,0 +1,6 @@
+class AddSubjectIdsAndSmsIdsToClassifications < ActiveRecord::Migration
+  def change
+    remove_reference :classifications, :set_member_subject, index: true
+    add_column :classifications, :set_member_subject_ids, :integer, array: true, default: []
```
camallen: Remove the `classifications_count` column on the `set_member_subjects` if we're not using the counters.

app/models/set_member_subject.rb
camallen: Sounds good.

app/models/user_subject_queue.rb
edpaget: Let's keep this in mind for the future. The administrative functions in a our clients will definitely need to know about and handle set_member_subjects. I didn't think it made sense to make a client that was only interested in submitting classifications worry about them.

camallen: makes sense.

db/migrate/20150128160822_add_subject_ids_and_sms_ids_to_classifications.rb
edpaget: This is done

camallen: https://github.com/edpaget/Panoptes/blob/subjects-in-classifications/db/schema.rb#L176

edpaget: Forgot actually save the file. Now it fixed for real.

PR #459, opened 2015-01-29
Update migration script to remove subject owner

edpaget:
Just forgot to remove the Subject owner.


camallen:
#461 will supercede this. I've got a migration PR to come through to set the uploading user id.

PR #461, opened 2015-01-29
track which user uploads a subject

camallen:
Store the specific user id without a relation as subjects may outlive their users. Useful if we need to audit which user uploaded a subject. 

Not using an AR relation here just storing the user id as a subject may well outlive a user account.


edpaget:
Cool are you waiting for this to do that migration? Do you think we'd be able to get #458 as well as I rebase?

camallen:
Yeah - i'm getting this through and then finalising the migration script. Looking at #458 now.

edpaget:
Great it should be rebased now.

Issue #466, opened 2015-01-30
Private project can be indexed and shown


camallen:
Projects with explicit private field can still be indexed and shown to auth / non-auth users.

Example: http://demo.zooniverse.org/panoptes-front-end/#/projects

Code:
https://github.com/zooniverse/Panoptes/blob/c7afefee4bc54ebd14803e2c9e1ca630141315e5/app/models/project.rb#L25

PR #467, opened 2015-01-30
ensure private field is set by the client

### LINKED ISSUE #465: Private not being set on projects by default
Projects being created don't have private field set to nil, which means they fail the show / index visible_scope test.

- [x] A projects private field should be configurable from the front end. 
- [ ] It should default to true.
- [x] It should be a required param. 
- [x] Serialise the private field in the response.
- [x] update docs accordingly.

camallen:
closes #465 - I think the client should explicitly set false / true with the front end defaulting to true (private).

### DIFF
--- a/app/schemas/project_create_schema.rb
+++ b/app/schemas/project_create_schema.rb
@@ -2,6 +2,7 @@ class ProjectCreateSchema < JsonSchema
   schema do
     type "object"
     description "A Project"
+    required "name", "description", "primary_language", "private"
     additional_properties false
 
     property "display_name" do
--- a/app/schemas/project_create_schema.rb
+++ b/app/schemas/project_create_schema.rb
@@ -19,6 +19,10 @@ class ProjectCreateSchema < JsonSchema
       description "Two character ISO 638 language code, optionally include two character ISO 3166-1 alpha-2 country code seperated by a hyphen for specific locale. ie 'en', 'zh-tw', 'es_MX'"
     end
 
+    property "private" do
+      type "boolean"
+    end
+
     property "title" do
       type "string"
       description "Translatable name for the project"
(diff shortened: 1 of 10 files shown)


app/schemas/project_create_schema.rb
```diff
@@ -19,6 +19,10 @@ class ProjectCreateSchema < JsonSchema
       description "Two character ISO 638 language code, optionally include two character ISO 3166-1 alpha-2 country code seperated by a hyphen for specific locale. ie 'en', 'zh-tw', 'es_MX'"
     end
 
+    property "private" do
+      type "boolean"
```
edpaget: Can we make this field required in the schema if we're going to require it to create a project? See https://github.com/camallen/Panoptes/blob/projects_private_fix/app/schemas/subject_set_create_schema.rb#L5

edpaget:
Does this resolve #465 and #466?

camallen:
This resolves #465 only. #466 seemed like non-intuitive behaviour to me.

edpaget:
What do you mean sorry?

On Fri, Jan 30, 2015 at 2:15 PM, Campbell Allen <notifications@github.com>
wrote:

> This resolves #465 <https://github.com/zooniverse/Panoptes/issues/465>
> only. #466 <https://github.com/zooniverse/Panoptes/issues/466> seemed
> like non-intuitive behaviour to me.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zooniverse/Panoptes/pull/467#issuecomment-72263464>.
>



-- 
Ed Paget

camallen:
I manually set the first 3 projects that @brian-c made after rebuilding the DB to private: true. But they were then visible to non-authenticated user (when private is nil no projects are listed as per https://github.com/zooniverse/Panoptes-Front-End/issues/88).

It seems counter intuitive to me based on the public_scope and the roles defined here: https://github.com/zooniverse/Panoptes/blob/c7afefee4bc54ebd14803e2c9e1ca630141315e5/app/models/project.rb#L25

Thoughts?

edpaget:
That makes sense. I'll write a test case and work on it.

app/schemas/project_create_schema.rb
camallen: done

PR #470, opened 2015-01-30
Make sure only active users are indexable

edpaget:
This should fix that Honeybadger error you were seeing. Also why don't enums work right with JRuby :(

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -51,7 +51,7 @@ class User < ActiveRecord::Base
   def self.scope_for(action, user, opts={})
     case action
     when :show, :index
-      all
+      active
     else
       where(id: user.id)
     end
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -51,7 +51,7 @@ def memberships_for(action, klass)
   def self.scope_for(action, user, opts={})
     case action
     when :show, :index
-      all
+      where(activated_state: 0)
```
camallen: Can you try and use the activatable scopes here: https://github.com/zooniverse/Panoptes/blob/master/app/models/concerns/activatable.rb#L6

They need to change to:
``` ruby
scope :active, -> { where(activated_state: activated_states[:active]) }
scope :disabled, -> { where(activated_state: activated_states[:inactive]) }
```

Does this still get used in the controller actions? https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/users_controller.rb#L34

edpaget: I think I did it one better those scope declarations were already provided
by the enum declaration.

PR #473, opened 2015-01-31
ignore user model association link ids

### LINKED ISSUE #471: Why do we include classifications in the serialized user response?
Firstly why do we include the classification sets of a user on their index route?

Also it seems Restpack is querying the includes even when not actually using them. This will be pretty slow for basic index routes on users (especially as classifications come in).

https://github.com/zooniverse/Panoptes/blob/671ba191fb30193de8104b442a009714990f7a61/app/serializers/user_serializer.rb#L4

Index route (no include params) result set:
screen shot 2015-01-30 at 22 37 01

Test Log DB queries:
screen shot 2015-01-30 at 22 36 24

camallen:
Closes #471 

Avoid costly queries to possibly large table / index scans. I don't think we need the actual instance ID links on the user instances to traverse the API. Instead we can use the non-instance links element for URL traversing to related models.

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -29,4 +29,8 @@ class UserSerializer
     @context[:requester] && @context[:requester].logged_in? &&
       (@context[:requester].is_admin? || @model.id == @context[:requester].id)
   end
+
+  def add_links(model, data)
+    data[:links] = {}
+  end
 end
--- a/spec/controllers/api/v1/users_controller_spec.rb
+++ b/spec/controllers/api/v1/users_controller_spec.rb
@@ -71,6 +71,23 @@ describe Api::V1::UsersController, type: :controller do
         end
       end
     end
+
+    describe "overridden serialiser instance assocation links" do
+
+      before(:each) do
+        user = users.first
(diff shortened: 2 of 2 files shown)


edpaget:
@brian-c does this work for you? Are you using the ids returns for the user at all?

brian-c:
Do I understand correctly that instead of each user resource having a`{links: {classifications: [/* potentially thousands of IDs */]}}` there'll just be a top-level `{links: {'users.classifications': {href: '/classifications?user_id={users.id}'}}}`?

That sounds fine to me.

edpaget:
Almost. Cam's PR removes all the linked ids for a user and only leaves the top level links.

camallen:
Yeah this will remove all associations on each serialised user instance in the response. The top level links provide enough information to traverse the links. 

This is a hack and won't work if the user model needs to serialize simple relations (not has_many) as well. Long term if we need both simple links and ignoring costly has_many links then i'll need to modify restpack to take directives about which links can be ignored when serialising each instance.

brian-c:
Yeah this should be fine. The user will re-fetch every link every time it's requested but we can make optimizations when we need them.

PR #518, opened 2015-02-10
Squash databse migrations into a single script

edpaget:
This just tidies the migrations directory. Since we're not actually migrating any preexisting data yet, we don't need to have 100+ files. This will make it easier to see and find migrations that actually do migrate data when we need to handle that.


camallen:
@edpaget out of interest did you use a gem for this?

edpaget:
No all you have to do is copy the `db/schema.rb` file.

camallen:
cool.

edpaget:
Yeah I just realized that when I was trying to find a migration I had just generated. I think we should regularly squash migrations until we're writing ones that need to be mindful of pre-existing data.

camallen:
will do.

Issue #512, opened 2015-02-07
Top-level links for classifications


brian-c:
Here's a weird question: What's a reasonable way for me to get the top-level links of the `classifications` type? Since the client creates classifications, but generally never retrieves any, there's no obvious opportunity to get this until I save one. But when passing around a client-created classification in the client it'd be nice to be able to get linked things instead of passing them along too.

I could make a request for /classifications?this_matches=nothing to get back an empty list with the top-level `links` key, but that seems hacky (and is currently not allowed anyway).

As for the links themselves, it appears `workflow` isn't in there. Subjects are referred to by `set_member_subjects`; are we still exposing that?

camallen:
Conforming to the idea of HATEOS the only way is to get a representation of the resouce. You could request only one classification via `/api/classifications?page_size=1`and then harvest the links. As you said you can get the resource after a valid create. A more hacky solution you could create a fake, incomplete classification and then delete it after, that should have the top level links on it.

Classifications store the `set_member_subjects to link them to the project / workflow. The `set_member_subject` links between the `subject_set` and the `subject`. 

What's the use case for passing a client-created, non-persisted classification around?

brian-c:
GET /api/classifications is only allowed when logged in. I'm fine with this solution otherwise.

I get what set_member_subjects is, but I thought we agreed (can't find the issue now) it was a back-end implementation detail that wouldn't be exposed to the client. The classification doesn't care what subject_set its subject is a member of. Docs need updating if that's not the case.

In the client, a classification is _created_ by the project's Classify page and then passed into a more generic Classifier component to do all the work. Given any classification resource, the Classifier should be able to find the other resources (workflow and subject) it needs to build the classification interface.

camallen:
`set_member_subject` is an implementation detail. We should probably revisit the classification representation. Originally the classification representation was for use in aggregation, however that use case requires the project, workflow, subject set and subject links. I'll confirm with Greg what he needs.

edpaget:
We could add a `subject_ids` field to the Classification model and not accept set_member_subjects in the response. We would have to change the UserSeenSubject model to use subject_ids instead of set_member_subject ids and revisit our Cellect adapter unless we wanted to take the extra database hit to retrieve the SetMemberSubject ids after a classification.

camallen:
Why don't we store all the required info on save and just serialize the representation as is, I.e. take the db hits in the background workers instead of when serializing. 

[...]

PR #519, opened 2015-02-10
Subjects in classifications

edpaget:
Address Brian's concerns from #512. Hide set_memebr_subjects entirely.
I'll need to update the cellect_panotpes driver to use only subjects

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -294,8 +294,8 @@ A classifications has the following attributes:
 - gold_standard
 - annotations
 
-Annotations is an array of maps of the form `{ "question_key":
-"question answer" }`. Metadata contains additional information about a
+Annotations is an array of maps of the form `{ "task": "task_key",
+"value": "question answer" }`. Metadata contains additional information about a
 classification including:
 
 - started_at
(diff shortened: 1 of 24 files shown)


edpaget:
Merge after #518

camallen:
@ed - can you rebase? Also what role do the set member subjects play now? Just placeholders for counters? How / when will they get updated?

edpaget:
Rebased. The Set Member Subjects are going to act as a Many-To-Many join table for Subject -> SubjectSet, with two additional columns for the classification count and their active state. The only clients that will need to touch them are ones that want to retire subjects based on some external criteria.

camallen:
I guess this depends on load testing / usage patterns but I think we should still use the active state on `set_member_subjects` when loading via cellect to keep the memory footprint smaller across all our cellect nodes.

camallen:
Also do you want to remove this migration and just keep the schema or leave it and squash it on the next migration?

edpaget:
Yeah I was intending to do that too. I'm redoing the cellect adapter to use ActiveRecord instead of just sql strings, so we can benefit from prepared/cached statements.

edpaget:
Good point I'll squash it.

camallen:
Nice!

edpaget:
I also need to update the docs.

Issue #526, opened 2015-02-11
Can't get project roles unless logged in


brian-c:
This works as described in the API docs, but I think users with a roles on a project should be public. Any reason they're not?

Also, the error message I get is `"ApiErrors::NotLoggedIn"`. Should that resolve to a value?

camallen:
Not sure I follow, if you're not logged in then what user can be used to determine the roles?

Re the error message, resolve to what value? Do you want a better format?

brian-c:
I want a list of roles for the project, not a user. I want to know the owners, the talk mods, the translators, etc.

That error just looks like Ruby code that was supposed to have been called, I thought it might be a bug. It's fine as-is if it's not.

camallen:
Just banging up a PR to expose this info on the `/project_roles?project_id=zx` route. This won't include private projects but non-private ones will have their roles exposed with the PR.

I've added a nicer message for login, "You must be logged in to access this resource." - sound ok?

brian-c:
Beautiful, thanks. Will private projects have their roles exposed to their owners?

camallen:
What url params are you using for filtering? It's not synced with the docs for project and user filtering from my read.

camallen:
Yeah private projects will have their roles exposed to owner, collaborator, tester, translator, scientist and moderators.

brian-c:
I'm not actually using this yet. When I do I'll just want to know the users with roles on a given project, so I imagine I'll just be using the `project_id` param.

Does it actually make more sense to think of user roles as links from a project? Sorta like `set_member_subject`, the client doesn't care about the relationships themselves, just the resources those relationships bring together.

GET /projects/5

```
links: {
  translators: {
    type: 'users',
    ids: ['4', '3']
  },
  moderators: {
    type: 'users',
    ids: ['4', '2']
  }
}
```

camallen:
Yeah that should already work via, `/projects/ID?include=project_roles`. However testing shows it's failing with an error for me. You can try it on staging.

brian-c:
Yeah, 500 over here. Maybe related to #523?

camallen:
[...]

PR #528, opened 2015-02-11
expose public project roles 

camallen:
partly solves #526 @edpaget can you have a look?

Filtering on the polymorphic access control list resource assocation requires the use of `resource_id` instead of the preferred `project_id`.


edpaget:
Can we still use `project_id` as the param and just change it to `resource_id` in the controller. That param name is an implementation detail and I'd prefer not to expose it in case we want to change the implementation (ie non-SQL datastore) in the future.

camallen:
Yeah i didn't like exposing the param at all. I'll flip it in the controller.

Issue #530, opened 2015-02-13
Inverse polymorphic relations (project -> project_roles) href representation


camallen:
The representation is wrong for the inverse polymorphic resource id in the hrefs (see below):

{"project_roles"=>[{"id"=>"6", "roles"=>["owner"], "links"=>{"user_group"=>"6", "project"=>"3", "owner"=>{"id"=>"3", "to_s"=>"users", "href"=>"users"}}}],
 "links"=>{"project_roles.project"=>{"type"=>"projects", "href"=>"/projects/{project_roles.project}"}},
 "meta"=>
  {"project_roles"=>
    {"page"=>1,
     "first_href"=>"/project_roles?resource_id=3",
     "previous_href"=>nil,
     "next_href"=>nil,
     "last_href"=>"/project_roles?resource_id=3"}}}

camallen:
This isn't going to be a problem for stargazing so i'm bumping to v0.0.7.

PR #531, opened 2015-02-13
ensure the specialized serializers loaded on include by restpack

### LINKED ISSUE #523: A project's project_roles link is wrong
The top-level `projects` type's `project_roles` link is `{href: "/?project_id={projects.id}", type: "project_roles"}`

The href should be `"/project_roles/?project_id={projects.id}"`.

### LINKED ISSUE #526: Can't get project roles unless logged in

### LINKED ISSUE #527: Ensure all project serializable includes work
Linked to #526, add this to `projects_controller_spec.rb`

```ruby 
        context "when the serializer models are known", :focus do
          let(:included_models) do
            %w(workflows subject_sets project_contents project_roles)
          end
          let(:includes) { included_models.join(',') }

          it "should include the relations in the response as linked" do
            expect(json_response['linked'].keys).to match_array(included_models)
          end
        end
```

camallen:
Fixes #527, closes #526, closes #523 

#530 will depend on this work. Also this relies on an updated version of restpack serializer, https://github.com/edpaget/restpack_serializer/pull/3 so it'll fail till that become available via the dev branch.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,10 +8,10 @@ GIT
 
 GIT
   remote: git://github.com/edpaget/restpack_serializer.git
-  revision: 13f5d936c689a00f220cc78e73c3f8eb14d3d516
+  revision: 9a7cdb2777006c7cd05c512859f5c9451248ca8d
   branch: dev
   specs:
-    restpack_serializer (0.5.7.pre1)
+    restpack_serializer (0.5.9)
       activerecord (>= 4.0.3, < 5.0)
       activesupport (>= 4.0.3, < 5.0)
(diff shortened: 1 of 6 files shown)


edpaget:
I am having some trouble with the Docker build I will get this in once I short it out.

camallen:
I had some problems with the postgtes image. I had to manually hack it a bit to work. Not sure if you have the same error but let me know if you do.

PR #536, opened 2015-02-16
Add retirement schemes for subjects

### LINKED ISSUE #298: Subject retirement rules
We'll need to support setting simple retirement rules for subject sets.

### LINKED ISSUE #487: Deactivate subjects
Worker script to set subject to inactive and remove from Cellect set

edpaget:
Closes #487. Closes #298

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1994,6 +1994,8 @@ A SubjectSet has the following attributes
 
 - id
 - display_name
+- metadata
+- retirement
 - set_member_subjects_count
 - created_at
 - updated_at
@@ -2031,6 +2033,9 @@ All attributes except display_name are set by the API
                     "display_name": "Weird Looking Galaxies",
                     "metadata": {
                         "category": "things"
+                    },
+                    "retirement": {
+                        "criteria": "classification_count"
                     },
                     "created_at": "2014-02-13T10:11:34Z",
                     "updated_at": "2014-02-13T10:11:34Z",
@@ -2204,6 +2209,9 @@ Subject Sets are returned as an array under *subject_sets*.
                     "metadata": {
                         "category": "things"
                     },
+                    "retirement": {
+                        "criteria": "classification_count"
+                    },
                     "created_at": "2014-02-13T10:11:34Z",
                     "updated_at": "2014-02-13T10:11:34Z",
                     "set_member_subject_count": 100,
@@ -2216,6 +2224,9 @@ Subject Sets are returned as an array under *subject_sets*.
                     "display_name": "Boring Looking Galaxies",
                     "metadata": {
                         "category": "things"
+                    },
+                    "retirement": {
+                        "criteria": "classification_count"
                     },
                     "created_at": "2014-02-13T10:11:34Z",
                     "updated_at": "2014-02-13T10:11:34Z",
@@ -2255,6 +2266,16 @@ Instead of a list of subjects a SubjectSet may include a link to a
 Collection which will import the collection's subjects into a new
 SubjectSet.
 
+A SubjectSet may also include a _retirement_ object with a
+_criteria_ key and an _options_ key. _criteria_ describes the strategy
+Panoptes will use to decide when to retire subjects. Currently, this
+may only be 'classification\_count' which retires subjects after a
+target number of classifications are reached. _options_ describes the
+parameters for the strategy. The 'classification\_count' strategy
+accepts _count_ as an option with an integer number of classifications
+for each subject. If _retirement_ is left blank Panoptes defaults to
+the 'classification\_count' strategy with 15 classifications per subject.
+
 + Request
 
     + Headers
@@ -2269,6 +2290,9 @@ SubjectSet.
                     "display_name": "A Group of Interesting Subjects",
                     "metadata": {
                         "category": "things"
+                    },
+                    "retirement": {
+                        "criteria": "classification_count"
                     },
                     "links": {
                         "project": "43",
--- a/app/models/subject_set.rb
+++ b/app/models/subject_set.rb
@@ -10,6 +10,16 @@ class SubjectSet < ActiveRecord::Base
 
   validates_presence_of :project
 
+  DEFAULT_CRITERIA = 'classification_count'
+  DEFAULT_OPTS = { 'count' => 15 }
+
+  validate do |set|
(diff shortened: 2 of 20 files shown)


app/models/subject_set.rb
```diff
@@ -21,4 +28,17 @@ class SubjectSet < ActiveRecord::Base
...
   end
+
+  def retire_member?(sms)
+    retirement_scheme.retire?(sms)
+  end
+
+  def retirement_scheme
+    case retirement.fetch('criteria', 'classification_count')
+    when 'classification_count'
+      RetirementSchemes::ClassificationCount.new(*retirement.fetch('options', { 'count' => 15 }).values_at('count'))
```
camallen: Perhaps move 15 to a constant?

lib/classification_lifecycle.rb
```diff
@@ -14,6 +14,12 @@ def queue(action)
       raise ClassificationNotPersisted.new(message)
     end
     ClassificationWorker.perform_async(classification.id, action)
+    if classification.complete?
+      classification.subject_ids.each do |sid|
+        ClassificationCountWorker.perform_async(sid, classification.workflow_id)
+        RetirementWorker.perform_async(sid, classification.workflow_id)
```
camallen: We can't guarantee ordering of worker execution here so if a classification that is 1 off retirement happens then the retirement worker may execute before the counter bump and we will end up with an extra classification. What about just chaining the `RetirementWorker` from the `ClassificationCountWorker` after bumping the count?

lib/classification_lifecycle.rb
```diff
@@ -14,6 +14,12 @@ def queue(action)
       raise ClassificationNotPersisted.new(message)
     end
     ClassificationWorker.perform_async(classification.id, action)
+    if classification.complete?
```
camallen: Why are you checking for complete before bumping the counters?

camallen: My bad - completed here is referencing the state of the classification not the retirement state of subject counts. Ignore me.

app/models/subject_set.rb
```diff
@@ -10,6 +10,13 @@ class SubjectSet < ActiveRecord::Base
 
   validates_presence_of :project
 
+  validate do |set|
+    criteria = %w(classification_count)
+    unless set.retirement.empty? || criteria.include?(set.retirement['criteria'])
+      set.errors.add(:"retirement.criteria", "Retirement criterea must be one of #{criteria.join(', ')}")
```
camallen: criteria spelling in error message.

apiary.apib
```diff
@@ -2255,6 +2266,16 @@ Instead of a list of subjects a SubjectSet may include a link to a
 Collection which will import the collection's subjects into a new
 SubjectSet.
 
+A SubjectSet may also include a _retirement_ object with a
+_criteria_ key and an _options_ key. _criteria_ describes the strategy
+Panoptes will use to decide when to retire subjects. Currently, this
+may only be 'classification\_count' which retires subjects after a
+target number of classifications are reached. _options_ describes the
+parameters for the stratgy. The 'classification\_count' strategy
```
camallen: strategy spelling.

lib/classification_lifecycle.rb
edpaget: I wanted them to get going at the same time to make sure subjects were retired as fast as possible. What if the RetirementWorker compares classification_count + 1?

camallen: Yeah it'd work but it couples the execution of the code to the way we fire off workers, e.g. in the future we run retirement workers as a periodic task and not tied to the incoming classifications (this is a stupid case but it demonstrates the context coupling). Why not just have one worker to call the counter code and retirement together but keep the two classes (remove them as workers and just call them from the single worker)?

Under heavy loads I don't think we can avoid some oversampling based on the way the client loads the subjects and buffers the limit set. 

Also, this will retire the subject but won't remove it from the cellect set for a user, so we would probably overclassify a little anyway under heavy load. We need to call [this](https://github.com/zooniverse/Cellect/blob/4b5341b8fb7dced6e9a8a6cda50cf6c804471595/lib/cellect/client/connection.rb#L23) in the retirement worker to get it out of the active set for users.

edpaget: Sure I'll change it to that. I hope in the future we don't have these tied to the rails code anyway.

camallen: Yeah this should suffice as a first attempt and what we need for stargazing. 

Also are you going to add the cellect call if it's retiring as well?

edpaget:
I think everything should have been addressed now. I'm running into the adler and I'm not quite sure tests are passing. I'll check as soon as I'm in.

app/workers/retirement_worker.rb
```diff
@@ -0,0 +1,16 @@
...
+  sidekiq_options queue: :high, retry: 3
+  
+  def perform(subject_id, workflow_id)
+    sms = SetMemberSubject.by_subject_workflow(subject_id, workflow_id)
+    if sms.retire?
+      sms.retired!
+      Cellect::Client.connection
+        .remove_subject(subject_id,
+                        workflow_id: workflow_id,
+                        group_id: sms.subject_set_id)
```
camallen: Don't we want to retire all sms's for this workflow in the case where the subject exists across subject_sets for the same workflow?

edpaget: Yes we do I'll change it to accommodate that.

In app/workers/retirement_worker.rb
<https://github.com/zooniverse/Panoptes/pull/536#discussion_r24824852>:

> @@ -0,0 +1,16 @@
> +class RetirementWorker
> +  include Sidekiq::Worker
> +
> +  sidekiq_options queue: :high, retry: 3
> +
> +  def perform(subject_id, workflow_id)
> +    sms = SetMemberSubject.by_subject_workflow(subject_id, workflow_id)
> +    if sms.retire?
> +      sms.retired!
> +      Cellect::Client.connection
> +        .remove_subject(subject_id,
> +                        workflow_id: workflow_id,
> +                        group_id: sms.subject_set_id)

Don't we want to retire all sms's for this workflow in the case where the
subject exists across subject_sets for the same workflow?

—
Reply to this email directly or view it on GitHub
<https://github.com/zooniverse/Panoptes/pull/536/files#r24824852>.

PR #539, opened 2015-02-17
serialize a workflow that is missing translatable content

camallen:
Fix workflow 500'ing when there is no content_association.


edpaget:
Workflows should always one content association though. We should make sure that's happening.

camallen:
Right - Happens by default when we build via the controllers but I don't think it's enforced at the model layer, I'll add it in.

PR #548, opened 2015-02-18
Fix preconditions

edpaget:
This converts our precondition strategy to use ETags instead of the Last-Modified method.

### DIFF
--- a/lib/subject_selector.rb
+++ b/lib/subject_selector.rb
@@ -1,27 +1,29 @@
 class SubjectSelector
   class MissingParameter < StandardError; end
 
-  attr_reader :user, :params
+  attr_reader :user, :params 
 
-  def initialize(user, params, scope, host)
-    @user, @params, @scope, @host = user, params, scope, host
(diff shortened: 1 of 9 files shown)


lib/subject_selector.rb
```diff
@@ -1,27 +1,29 @@
...
-    @user, @params, @scope, @host = user, params, scope, host
+  def initialize(user, params, scope, host, controller)
+    @user, @params, @scope, @host, @controller = user, params, scope, host, controller
   end
 
   def create_response
     case params[:sort]
     when 'cellect'
-      cellect_subjects
+      @controller.render json_api: SubjectSerializer.page(params, cellect_subjects)
```
camallen: Any specific reason for passing the controller context into the selector? I assume it's because you wanted to switch the response based on the params[:sort] values but couldn't we do this in the controller (as the params[:sort] value is there) and seperate out the concerns here? I.e. let the selector choose the subjects and the controller render a response.

PR #570, opened 2015-02-20
Using display_name for visible name of all resources

### LINKED ISSUE #40: What's a User?
Currently a user signs up with:

- Username
- Password
- Real name (for recognition in papers, etc.)
- "Send me beta announcements"

In #38 @edpaget suggested a separate key for URL. That gives us three "name" properties. Can we use the login for the URL, encoding what needs encoding when calling the API? URLs in the browser should render okay: [#/ß®ï@ñ-¢](#/ß®ï@ñ-¢).

### LINKED ISSUE #544: Filter owner name by display_name
Linked to https://github.com/zooniverse/Panoptes-Front-End/pull/105

The owner filter should check against display_name: 
https://github.com/zooniverse/Panoptes/blob/634a981d7e517bcda271c8c975e205a8a9a09be2/app/controllers/concerns/filter_by_owner.rb#L10

edpaget:
Remove login attribute from user model. Restrict user names to not allow
whitespace or the characters ($,@).

Closes #544, Closes #40


edpaget:
This will require user sign_up, sign_in actions to use 'dispaly_name' as the param instead of login. We could shim in support for continuing to use login as the parm if we felt like it.

app/models/user_group.rb
```diff
@@ -76,8 +77,8 @@ def identity?
   private
   
   def downcase_case_insensitive_fields
-    if name
-      self.name = name.downcase
+    unless name
```
camallen: Perhaps combine the two conditionals here? `if name.nil? && display_name`

app/models/user.rb
```diff
@@ -24,9 +24,8 @@ class User < ActiveRecord::Base
   has_many :project_roles, through: :identity_group
   has_many :collection_roles, through: :identity_group
 
-  validates :login, presence: true, uniqueness: true
-  validates_presence_of :display_name
-  validates_uniqueness_of :display_name, case_sensitive: false
+  validates :display_name, presence: true, uniqueness: { case_sensitive: false },
+            format: { without: /\$|@|\s+/ }
```
camallen: Should we extract the shared User/UserGroup display_name regex to a common constant to decrease the chance of them diverging?

lib/user_info_scrubber.rb
```diff
@@ -20,7 +20,7 @@ def self.scrub_personal_info!(user)
   def self.scrub_details
     begin
```
camallen: The retries here were to avoid a possibly duplicate on the display_name unique index. Add `@user.update_columns(email: nil)` to the scrub_personal_info! and remove all the `display_name` cruft now it's not needed.

spec/lib/user_info_scrubber_spec.rb
```diff
@@ -17,14 +17,9 @@ def scrub_user_details(user)(user)
       end
 
       it "should replace their display name" do
```
camallen: should not replace their display name?

camallen:
Looks good.

Issue #572, opened 2015-02-23
JSON Parse Error


astopy:
The following seems to be a result of #564. I'm not sure what JSON it's referring to...

    => Notice: server is listening on all interfaces (0.0.0.0). Consider using 127.0.0.1 (--binding option)
    => Ctrl-C to shutdown server
    log4j:WARN No appenders could be found for logger (org.apache.zookeeper.ZooKeeper).
    log4j:WARN Please initialize the log4j system properly.
    Exiting
    JSON::ParserError: A JSON text must at least contain two octets!
                              initialize at json/ext/Parser.java:175
                                     new at json/ext/Parser.java:151
                                   parse at /usr/local/lib/ruby/gems/shared/gems/json-1.8.2-java/lib/json/common.rb:155
                               send_ping at /usr/local/lib/ruby/gems/shared/gems/honeybadger-2.0.6/lib/honeybadger/config.rb:309
                                    ping at /usr/local/lib/ruby/gems/shared/gems/honeybadger-2.0.6/lib/honeybadger/config.rb:246
                                   start at /usr/local/lib/ruby/gems/shared/gems/honeybadger-2.0.6/lib/honeybadger/agent.rb:69
                                   start at /usr/local/lib/ruby/gems/shared/gems/honeybadger-2.0.6/lib/honeybadger.rb:30
                                 Railtie at /usr/local/lib/ruby/gems/shared/gems/honeybadger-2.0.6/lib/honeybadger/init/rails.rb:14
                           instance_exec at org/jruby/RubyBasicObject.java:1562
                                     run at /usr/local/lib/ruby/gems/shared/gems/railties-4.1.9/lib/rails/initializable.rb:30
                        run_initializers at /usr/local/lib/ruby/gems/shared/gems/railties-4.1.9/lib/rails/initializable.rb:55
                              tsort_each at /usr/local/lib/ruby/1.9/tsort.rb:150
       each_strongly_connected_component at /usr/local/lib/ruby/1.9/tsort.rb:183
    each_strongly_connected_component_from at /usr/local/lib/ruby/1.9/tsort.rb:219
       each_strongly_connected_component at /usr/local/lib/ruby/1.9/tsort.rb:182
                                    each at org/jruby/RubyArray.java:1613
       each_strongly_connected_component at /usr/local/lib/ruby/1.9/tsort.rb:180
                              tsort_each at /usr/local/lib/ruby/1.9/tsort.rb:148
[...]

PR #573, opened 2015-02-23
remove honey badger config init loader

camallen:
honeybadger now looks for a config/honeybadger.yml file by default.


edpaget:
Did this fix #572 ?

camallen:
@astopy - did that fix the honeybadger error in the running docker container?

PR #574, opened 2015-02-23
Allow admin overrides of all permission scopes

### LINKED ISSUE #559: Allow admin users to see all classifications through API

edpaget:
Closes #559


edpaget:
This allows a user who has the admin flag set on their account to operate against the api in admin-mode by including a `?admin=any_value_here` parameter in their request. Otherwise the an admin user's api request continue to behave normally.

PR #577, opened 2015-02-23
Allow filtering of classifications by subject_id

### LINKED ISSUE #558: Retrieve Classifications by subject id.

edpaget:
Also only allow users to retrieve their incomplete classifications.

Closes #558

### DIFF
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -6,6 +6,8 @@ class Api::V1::ClassificationsController < Api::ApiController
   
   schema_type :json_schema
 
+  before_action :filter_by_subject_id, only: :index
+
   rescue_from RoleControl::AccessDenied, with: :access_denied
 
   def create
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/classifications_controller.rb
```diff
@@ -18,6 +20,13 @@ def update
 
   private
 
+  def filter_by_subject_id
+    if subject_ids = params.delete(:subject_id).try(:split, ',')
```
camallen: Should we test if the returned array from split is empty instead of just truthy here? I.e a `{ params[:subject_id] => "" }` would result in an empty array for subject_ids due to `p "test" if "".split(",")`.

edpaget: Yeah we should probably do that. I'm going to change that [here](https://github.com/zooniverse/Panoptes/blob/master/app/controllers/concerns/filter_by_owner.rb#L9) too.

PR #581, opened 2015-02-24
ensure passwords match for the sha1 hash algorithm

camallen:
possible security issue where we don't compare hashed values == for older sha1 haah users (we don't have any)


edpaget:
good catch

PR #590, opened 2015-02-25
add firebase JWT auth token to /api/me route

### LINKED ISSUE #587: Firebase custom auth token
https://www.firebase.com/docs/web/guide/login/custom.html

It'd be good to have this is sooner than later, since I basically have no idea what I'm doing.

Ideally, the user could get a `firebase_auth_token` property, but I can just make another authenticated request for one if that's easier.

@camallen has my Firebase secret, or just ask.

camallen:
closes #587


spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -144,6 +145,35 @@
...
+    end
+
+    context "when the requesting user is the resource owner" do
+      it "should have a firebase auth token for the user" do
+        get :me
+        expect(response_fb_token).to eq(jwt_token)
+      end
+    end
+
+    context "when the requesting user is not the resource owner" do
```
edpaget: Do you want to move this test to the user#show describe block? The /me route can't return a user that's not the resource owner.

camallen: will do - i felt a little dirty writing it.

Issue #593, opened 2015-02-25
Failed jobs in sidekiq


camallen:
Seems that sometimes classification workers are too quick for the app, even when testing on `persisted?` [here](https://github.com/zooniverse/Panoptes/blob/d1525c4cc7ee18308f5c5cf6c0b80444ccd1145c/lib/classification_lifecycle.rb#L12).

![sidekiq_failed](https://cloud.githubusercontent.com/assets/295329/6374505/430bb59a-bd09-11e4-95e3-e36f21652526.jpg)

I'm not 100% but [.persisted?](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-persisted-3F) might be getting set `after_save` but before `after_commit` [callbacks](http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html).

Some details: http://thechangelog.com/earn-a-sidekiq-blackbelt/

We can just delay processing for a small time offset (5 seconds?) to ensure the transaction is committed in the DB. Also the number of retries should be extended to the max as we don't want to lose data.

https://github.com/mperham/sidekiq/wiki/Delayed-extensions#activerecord
https://github.com/mperham/sidekiq/blob/7351c3947f310b7fa2f18fdad424491c70daebbd/lib/sidekiq/extensions/active_record.rb#L27

edpaget:
I can't see the sidekiq status page at the moment, but do you know what error they're failing with?

camallen:
Set your account to be admin and you'll be able to see it.

Workers fail without a log unless they are stuck in a backoff retry and you can see them in the UI. So this specific failure retried again and finished but it still records the failure. If they fail beyond the retries count they go into the dead queue.

edpaget:
@camallen I'm going to setup sidekiq so errors get sent to Honeybadger. https://github.com/mperham/sidekiq/wiki/Error-Handling

camallen:
Great idea @edpaget !

edpaget:
@camallen if your suspicion about the after_save/after_commit stuff is right maybe we can solve it by moving the queuing action outside the transaction block? https://github.com/zooniverse/Panoptes/blob/master/lib/json_api_controller/creatable_resource.rb#L11

camallen:
Yep - we could. We could just move it to the classifications controller too.
https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/classifications_controller.rb#L14

camallen:
[...]

PR #594, opened 2015-02-25
retry classification processing for as long as we can

camallen:
partly solves #593


edpaget:
I think we do want to limit these at some point when we figure out more about what #593 is caused by

camallen:
Sidekiq does 25 by default: https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry

PR #596, opened 2015-02-25
Add Retired Subject count to subject sets

### LINKED ISSUE #565: Expose project completeness somehow
I'd like to be able to tell when a project is complete. I'd imagine each subject_set could have a `"retired_count": 123` kind of field. Not sure if we've already implemented anything like this.

I'm going to use this to display a message on the project home page and direct folks to the Results page, if that exists, instead of the Classify page.

I'm also planning on keeping classification possible (with a "we're done" message) by not fetching subjects through Cellect when a project is complete. Will that work?

edpaget:
Closes #565

### DIFF
--- a/app/models/concerns/subject_counts.rb
+++ b/app/models/concerns/subject_counts.rb
@@ -2,6 +2,10 @@ module SubjectCounts
   include ActiveSupport::Concern
 
   def subjects_count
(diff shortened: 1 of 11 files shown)


edpaget:
You'll just be able to compare the retired subjects vs. the total to test if a project is done

PR #597, opened 2015-02-26
Catch JSON error raised in JRuby

edpaget:
JRuby seems to raise a JSON::ParserError instead of one raised by MRI,
this just expands our middleware to catch both


camallen:
yeah - that was me running an incorrectly recorded jmeter script via blazemeter. Nice to not have it 500.

edpaget:
Ah cool I wondered what that was coming from.

camallen:
FYI - I'm trying to add comments to honeybadger errors.

PR #609, opened 2015-03-04
allow aggregations to be updated after creation

camallen:
aggregation will be an interative process that runs at different time scales, e.g. hourly best effort vs nightly full re-run, etc.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -5164,3 +5164,216 @@ creation.
 + Response 201
 
     [CollectionPreference][]
+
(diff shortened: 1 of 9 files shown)


edpaget:
So the whole reason I didn't originally support updating Aggregations make sure old ones we're never overwritten. I thought we wouldn't want to keep all the old results around. We should definitely version them if we're going to allow them to updated.

camallen:
Ok - i'll add versions to the aggregation results.

PR #613, opened 2015-03-05
Add emergency deployment mode to start.sh

astopy:


scripts/docker/start.sh
```diff
@@ -10,6 +10,13 @@ fi
 if [ "$RAILS_ENV" == "development" ]; then
   exec foreman start
 else
+  USER_DATA=$(curl --fail http://169.254.169.254/latest/user-data || echo "")
+
+  if [ "$USER_DATA" == "EMERGENCY_MODE" ]
+  then
+    git pull
```
camallen: Should we use a `git pull --rebase origin master` here? 
``` bash
ubuntu@ip-10-1-1-123:~$ docker exec -it panoptes bash
root@ed2cf3cede01:/rails_app# git remote -v
origin	https://github.com/zooniverse/Panoptes.git (fetch)
origin	https://github.com/zooniverse/Panoptes.git (push)
```

camallen: `git pull` should be fine since there is only 1 origin from the docker build `git clone` and it allows the branch that docker built off to be pulled from.

PR #614, opened 2015-03-05
Add support for login in through ZooHome database

### LINKED ISSUE #483: Integrate logins and signups with zoo home
Ensure logins and signups are not isolated in Panoptes and are usable across the zooniverse.

edpaget:
Closes #483


edpaget:
@alianos- when this is merged, just make sure to not have any "zooniverse_home" databases defined in your database.yml and things should continue to work normally.

spec/controllers/registrations_controller_spec.rb
```diff
@@ -48,6 +48,10 @@
           expect(subject).to receive(:clean_up_passwords)
           post :create, user: user_attributes
         end
+
+        it "should create a new ZooniverseUser" do
+          expect(ZooniverseUser.where(login: login).first).to_not be_nil
```
camallen: Database cleaner isn't cleaning the sqllite db and some records are being left around, hence this passes without a post request to create a resource.

lib/zooniverse_user.rb
```diff
@@ -1,4 +1,77 @@
...
+    self.crypted_password = Sha1Encryption.encrypt(plain_password, salt: password_salt)
+  end
+
+  def authenticate(plain_password)
+    crypted_password == Sha1Encryption.encrypt(plain_password, salt: password_salt) ? self : nil
+  end
+
+  def import
+    user = User.find_or_initialize_by(zooniverse_id: id.to_s) do |u|
+      u.display_name = login
```
camallen: just set this once on line 53

lib/zooniverse_user.rb
```diff
@@ -1,4 +1,77 @@
...
+      where(login: users)
+    end.find_each(&:import)
+  end
+
+  def self.authenticate(username, password)
+    find_by(login: username).try(:authenticate, password)
+  end
+
+  def self.from_user(user)
+    zu = find_or_create_by(login: user.display_name) do |u|
```
camallen: find_or_create_by with a save just below would be 2 persistence calls. Maybe just find_or_initialize_by and then save!

app/controllers/registrations_controller.rb
```diff
@@ -24,6 +27,12 @@ def build_resource(sign_up_params)
     resource.build_identity_group
   end
 
+  def build_zoo_user(resource)
```
camallen: can we call this create_zoo_user?

app/controllers/registrations_controller.rb
```diff
@@ -13,6 +15,7 @@ def create
   def create_from_json
     build_resource(sign_up_params)
     resource_saved = resource.save
+    build_zoo_user(resource)
```
camallen: Building zoo user here after we've created a local user account fails the condition where the zoo user has a uniq, duplicate email / login condition. Add this to registrations spec under '#create', 'with_valid_attributes':

``` ruby
context "when the user already exists in zooniverse_home" do
  let!(:existing_zoo_user) { create(:zooniverse_user, login: display_name) }
    
  it "should not create the panotpes user" do
    expect { post :create, user: user_attributes }.not_to change{ User.count }
  end
end
```

spec/lib/zooniverse_user_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        expect(zu.import.display_name).to eq(zu.login)
+      end
+    end
+
+    context 'when the User has been imported' do
+      it 'should update the User' do
+        user = create(:user, migrated: true, build_zoo_user: true)
+        zu = ZooniverseUser.find(user.zooniverse_id.to_i)
+        zu.update!(login: "a new login")
+        expect(zu.import.display_name).to eq("a new login")
```
camallen: do you want to test the `user.display_name == "a new login"` here?

lib/zooniverse_user.rb
edpaget: I can't I need it to exist to build the identity group. I can rewrite so the fields aren't set in the block and instead it can test if the resource already exists and then assign those fields.

camallen: no probs

spec/lib/zooniverse_user_spec.rb
edpaget: `zu.import` returns the User record

camallen: excellent

app/controllers/passwords_controller.rb
```diff
@@ -36,6 +39,12 @@ def respond_to_request(action_status)
       render status: response_status, json: {}
     end
 
+    def update_zooniverse_user(resource)
+      if resource.errors.empty? && ZooHomeConfiguration.use_zoo_home?
+        ZooniverseUser.find_from_user(resource).password = resource_params[:password]
```
camallen: Will this persist the password changes to zoo home?

edpaget: Good point, I removed the automatic save from that method.

PR #620, opened 2015-03-09
allow existing panotpes user accounts to be updated

camallen:
some real zoo_home_users have setup accounts on panoptes that share the login/display_name. Allow these accounts to be updated from zoo home so we don't have to remove dup account and orphpaning any existing linked instances e.g. projects, etc.

### DIFF
--- a/lib/zoo_user_warden_strategy.rb
+++ b/lib/zoo_user_warden_strategy.rb
@@ -4,11 +4,9 @@ Warden::Strategies.add(:zoo_user) do
   end
 
   def authenticate!
-    u = ZooniverseUser.authenticate(params['user']['display_name'], params['user']['password'])
-    if u.nil?
-      fail!("Could not log in")
-    else
-      success!(u.import)
-    end
+    zoo_home_user = ZooniverseUser.authenticate(params['user']['display_name'],
+                                                params['user']['password'])
+    imported_user = zoo_home_user.import if zoo_home_user
+    imported_user ? success!(imported_user) : fail!("Could not log in")
   end
 end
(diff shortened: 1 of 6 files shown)


lib/zoo_user_warden_strategy.rb
```diff
@@ -4,11 +4,10 @@ def valid?
   end
 
   def authenticate!
-    u = ZooniverseUser.authenticate(params['user']['display_name'], params['user']['password'])
-    if u.nil?
-      fail!("Could not log in")
-    else
-      success!(u.import)
-    end
+    imported_user = nil
```
edpaget: I'm pretty sure you don't need to set imported_user to nil here. In line 9 the assignment expression evaluates to nil when the if expression if false.

lib/zooniverse_user.rb
```diff
@@ -82,4 +69,28 @@ def simple_token
...
+      u.migrated = true
+      u.zooniverse_id = id.to_s
+      u.build_identity_group if new_account
+    end
+  end
+
+  def panoptes_user_account_exists?(user)
+    user.display_name = login &&
+    user.zooniverse_id.nil? &&
+    user.hash_func == 'bcrypt' &&
```
edpaget: If you're setting the encrypted password to the crypted password from Zoo Home this needs to be 'sha1'. It'll be switched back to bcrypt whenever we enable login through Panoptes.

camallen: I run this test before setting any attributes ala line 74 so on the first time an existing panoptes user account has the bcrypt hash func. This should detect any user that is already in panoptes via PFE / manual html registration and in zoo home but hasn't been migrated yet.

edpaget: Sorry I can read apparently. I mistook the == for an assignment.

PR #628, opened 2015-03-09
Fix destroy links routing

### LINKED ISSUE #623: DELETE /links routes uses wrong controller method

edpaget:
Closes #623

### DIFF
--- a/spec/support/updatable.rb
+++ b/spec/support/updatable.rb
@@ -59,7 +59,7 @@ shared_examples "is updatable" do
   end
 end
 
-shared_examples "has updatable links" do
+RSpec.shared_examples "has updatable links" do
   let(:old_ids) { resource.send(test_relation).map(&:id) }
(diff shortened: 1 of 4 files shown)


spec/support/updatable.rb
```diff
@@ -100,3 +100,28 @@ def self.extract_linked_resource_ids(linked_resource)
     expect(updated_resource.send(test_relation).map(&:id)).to include(*old_ids)
   end
 end
+
+RSpec.shared_examples "supports destroy_links" do
```
camallen: Are these getting called anywhere?

edpaget: No I can remove it if you want I was using it to test the user group method until I realized should be deleting members.

camallen: Can you clean it up then, best not to leave code lying around that should be removed.

PR #630, opened 2015-03-09
Set remember token in Zoo-Home authentication

edpaget:
I think the issue with what was happening in the Front-End specs may have been related to failing to handle the rememberable token in the Warden strategy. I'm still figuring out how to test it but it may be worth hitting it with the front-end specs to see if anything happens.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -29,16 +29,14 @@ class RegistrationsController < Devise::RegistrationsController
   end
 
   def create_zoo_user(resource)
-    if resource.valid? && ZooHomeConfiguration.use_zoo_home?
+    return true unless ZooHomeConfiguration.use_zoo_home?
+    if resource.valid?
       zu = ZooniverseUser.create_from_user(resource)
-      if zu.persisted?
-        true
-      else
-        zu.errors.each do |attr, errors|
-          resource.errors.add(attr, errors)
-        end
-        false
+      return true if zu.persisted?
+      zu.errors.each do |attr, errors|
+        resource.errors.add(attr, errors)
       end
+      false
     end
   end
 
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -24,8 +24,9 @@ Doorkeeper.configure do
   end
 
   resource_owner_from_credentials do
-    if params[:display_name] && u = User.find_for_database_authentication(display_name: params[:display_name])
-      valid_non_disabled_user = u.valid_password?(params[:password]) && !u.disabled?
+    if params[:display_name] && u = ZooniverseUser.authenticate(params[:display_name],
+                                                                params[:password]).try(:import)
+      valid_non_disabled_user = !u.disabled?
     else
       u = current_user
       valid_non_disabled_user = !u.blank? && !u.disabled?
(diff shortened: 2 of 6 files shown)


edpaget:
@camallen I added some tests that I think replicate the behaviour, but I can't seem to get them to fail even if I remove my changes, so I'm not sure they did anything. If it blocking work this may be a good place to start.

camallen:
I've tracked it down and will have a PR in shortly. It was paper trail and devise interacting to attempt to get the current user in before filters. More details to come.

edpaget:
Awesome it might worth merging this anyway just for the correct remember_me token handling. I'll remove the spec since it doesn't seem like it's providing any value.

app/controllers/registrations_controller.rb
```diff
@@ -37,6 +37,8 @@ def create_zoo_user(resource)
         end
         false
       end
+    else
+      !ZooHomeConfiguration.use_zoo_home?
```
camallen: Can you just return on the `use_zoo_home conditional` at the start?

config/initializers/doorkeeper.rb
```diff
@@ -24,8 +24,9 @@ def self.optional
   end
 
   resource_owner_from_credentials do
-    if params[:display_name] && u = User.find_for_database_authentication(display_name: params[:display_name])
-      valid_non_disabled_user = u.valid_password?(params[:password]) && !u.disabled?
+    if params[:display_name] && u = ZooniverseUser.authenticate(params[:display_name],
```
camallen: Wouldn't they have signed in via devise routes here and thus have a User account? So no need to auth against the Zoo Home account? 

Or would the `u.valid_password?` method be using the wrong hashing alg?

edpaget: No it would be testing against the User database, which may not be in sync with the ZooHome database (user just reset their password on ZooHome for example).

lib/zoo_user_warden_strategy.rb
```diff
@@ -1,12 +1,22 @@
...
                                                 params['user']['password'])
-    imported_user = zoo_home_user.import if zoo_home_user
-    imported_user ? success!(imported_user) : fail!("Could not log in")
+    if zoo_home_user && imported_user = zoo_home_user.import 
+      remember_me(imported_user)
+      imported_user.after_database_authentication
+      success!(imported_user)
+    end
+    
+    fail!("Invalid display_name or password.") unless zoo_home_user
```
camallen: Can we use the devise translation key here `I18n.t "devise.failure.invalid"`

edpaget: Yes I figured it didn't matter since we would delete all this code in about 2 weeks. Your PR has it so we should just resolve the merge in its favor.

config/initializers/doorkeeper.rb
camallen: Right - are we even using this doorkeeper auth flow?

edpaget: The common line is the the u = current_user below it which is how we do the session-base authentication for 'password' requests.

camallen: Right - I was forgetting where that was being triggered for doorkeepr / devise.

camallen:
Hey looks good - i assume this was rebased of the latest tests? If so i'm good to merge.

edpaget:
Yes it should be good to go.

PR #634, opened 2015-03-10
ensure zoo home auth strategy only runs when it should

camallen:
This fixes the panoptes-front-end failing api tests.

Paper trail adds before filters to rails's application controller that will call current_user to determine the whodunnit field. Devise controllers implement this method and were triggering the zoo_home warden authentication strategy when they shouldn't have. I've disabled paper trails in the application_controller base class and enabled it for the api end points. I could be more specific here and just enable it for the end points that are using versioning but it might add some surprises when people add new models and expect versioning to work.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -14,7 +14,7 @@ class RegistrationsController < Devise::RegistrationsController
 
   def create_from_json
     build_resource(sign_up_params)
-    if created_in_zoo_home = create_zoo_user(resource)
+    if create_zoo_user(resource)
       resource_saved = resource.save
     end
     yield resource if block_given?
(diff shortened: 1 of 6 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -14,7 +14,9 @@ def create
 
   def create_from_json
     build_resource(sign_up_params)
-    resource_saved = create_zoo_user(resource) && resource.save
+    if created_in_zoo_home = create_zoo_user(resource)
```
edpaget: Is created_in_zoo_home being used anywhere?

camallen: No - i'll remove it.

edpaget:
Awesome job tracking this down. It had me completely stumped. I'll merge once travis finishes.

PR #640, opened 2015-03-12
Psql selection

edpaget:
Not super happy with some bits of this (mostly how it does the retired/already seen marking), but I think it should work for Stargazing.

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -12,23 +12,17 @@ class Api::V1::SubjectsController < Api::ApiController
     case params[:sort]
     when 'cellect'
       render json_api: SubjectSerializer.page(params,
-                                              selector.cellect_subjects,
-                                              selector_context)
+                                              *selector.cellect_subjects)
     when 'queued'
       render json_api: SubjectSerializer.page(params,
-                                              selector.queued_subjects)
+                                              *selector.queued_subjects)
     else
       super
     end
   end
 
   private
-
-  def selector_context
-    { retired: workflow.finished?,
-      already_seen: api_user.has_finished?(workflow) }
-  end
-
+  
   def context
     case action_name
     when "update", "create"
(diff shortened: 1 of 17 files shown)


lib/postgresql_selection.rb
```diff
@@ -0,0 +1,71 @@
+class PostgresqlSelection
+  attr_reader :workflow, :user
+  
+  def initialize(workflow, user=nil)
+    @workflow, @user = workflow, user
+  end
+
+  def select(opts={})
+    limit = opts.fetch(:limit, 20).to_i
+    selection_statement = case
```
camallen: can you extract this to a method?

app/controllers/api/v1/subjects_controller.rb
```diff
@@ -21,6 +24,11 @@ def index
 
   private
 
+  def selector_context
+    { retired: workflow.finished?,
```
camallen: Is this being tested anywhere in the subjects_controller responses?

edpaget: I added some tests.

camallen:
Looks fine to merge.

In benchmarking mood and am wondering what sort of performance hit this might add to selection routes.

edpaget:
Yeah loading the workflow and doing the `has_finished?` and `finished?` methods will add some overhead. I think it could be avoided with some changes to the subject selector, but I'm not exactly sure.

app/controllers/api/v1/subjects_controller.rb
```diff
@@ -11,9 +11,12 @@ class Api::V1::SubjectsController < Api::ApiController
   def index
     case params[:sort]
     when 'cellect'
-      render json_api: SubjectSerializer.page(params, selector.cellect_subjects)
+      render json_api: SubjectSerializer.page(params,
+                                              selector.cellect_subjects,
+                                              selector_context)
```
camallen: Just had a thought re avoiding paying the cost with context methods. Can we move the `selector_context` into the selector object and switch it depending on whether cellect returned results? It's be nice not to check if the workflow is finished / a user has finished each time we want subjects.

PR #641, opened 2015-03-12
Switch to mri ruby for the API nodes

camallen:
This will run the API nodes on MRI. based on benchmarking on my dev machine MRI is ~3 times faster, I'm not 100% sure why but my first guess would be the jruby gems.

Dont merge untill the ruby:2.2.1 docker image is built, https://registry.hub.docker.com/u/zooniverse/ruby/builds_history/67460/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM zooniverse/ruby:jruby-1.7.18
+FROM zooniverse/ruby:2.2.1
 
 WORKDIR /rails_app
 


camallen:
**Don't merge this PR until i report back!** - I'm manually testing this via infrastructure launch commands.

edpaget:
@camallen do you want to update the travis.yml file as part of this PR to have MRI 2.2.1 be the one we test against and don't allow failures from?

camallen:
@edpaget should we just remove jruby-head from teh build matrix? it never passes and just takes forever.

edpaget:
Yeah that's probably fine.

edpaget:
Can you change the fig.yml too?

PR #642, opened 2015-03-12
bump honeybadger to 2+ since we're not on jruby

camallen:
Depends on #641

V 2+ backgrounds jobs automatically if it detects a background worker framework, e.g sidekiq. I've configured the production_config/honeybadger to support both new and old versions of the gem by default.


edpaget:
Are we still planning on running sidekiq on JRuby or is that on MRI as well?

camallen:
Good point, i'm not sure yet but switching the base image to MRI (still trying to get an image up for testing) will force sidekiq to use mri. We'd need to split sidekiq into it's own docker image / deploy setup to be deployed with panoptes staging.

PR #643, opened 2015-03-13
Add new relic instrumentation to cellect client

edpaget:


lib/cellect_client.rb
```diff
@@ -23,9 +23,12 @@ def self.get_subjects(session, workflow_id, user_id, group_id, limit)
       .request(:get_subjects, group_id: group_id, user_id: user_id, limit: limit)
   end
 
-  def set_client_params(params)
```
camallen: Any reason this method is removed?

edpaget: Yes it wasn't used. I'm not sure why I committed it in the first place.

PR #646, opened 2015-03-13
Generate random values before setting column constraint

edpaget:


db/migrate/20150309171224_add_random_to_set_member_subjects.rb
```diff
@@ -1,6 +1,8 @@
 class AddRandomToSetMemberSubjects < ActiveRecord::Migration
   def change
-    add_column :set_member_subjects, :random, :decimal, null: false
+    add_column :set_member_subjects, :random, :decimal
+    SetMemberSubject.find_each{ |sms| sms.update!(random: rand) }
```
camallen: On gotta i've been bit with before is model code in a migration. In the future if the model has been removed and this migration tries to run, it'll fail. You can just define the model you want in the migration and it will inherit from the schema (if the table still exists) or we can just write plain sql.

That said i'm pretty sure it won't be a problem.

camallen:
Looks fine to merge.

PR #654, opened 2015-03-16
Eager Load translations.

### LINKED ISSUE #647: Reduce Number of queries for loading a translation
It takes 2 queries per model to load the content for translatable models. This should be done with a preload.

edpaget:
Closes #647


edpaget:
Ugh looks like there's a problem with eager_loading and the count function in Kaminari. I'll be back on this later.

PR #659, opened 2015-03-16
ensure only production gems make it into the docker image

### LINKED ISSUE #651: Production env should not include test and development gems
SPRING is loading into the production env and so is FIG_RAKE.

We need to make sure these never make it in to production.

Fig rake is in config/boot and rakefile, needs to have an `if ENV['FIG_RAKE']` attached to the load.
Spring will only install if the gem is included. Remove the env from bundle install and it's removed.

camallen:
closes #651 - ensure gems like spring and fig_rake are no where near production / staging.

this may break some fig envs, just change the docker file to use bundle install and then it should work again. we need to migrate fig to docker-compose and use and alternate dockerfile for production deploys and dev work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ bin/**
 #ignore the api commit file
 public/commit_id.txt
 vendor/cache/
+vendor/bundle/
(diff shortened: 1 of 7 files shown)


edpaget:
closed via cc3d9bef8c5e1475726a0da001fe992426714111

PR #671, opened 2015-03-17
allow a user to update their zoo wide email preferences

camallen:
allow the user to update their panoptes global email preference from the front end


camallen:
This doesn't update zoo home with their subscription details.

PR #683, opened 2015-03-18
Handle ruby hash changes by bruteforcing an offset

edpaget:
THIS IS TERRIBLE REMOVE ASAP


Dockerfile
```diff
@@ -13,7 +13,7 @@ ADD ./Gemfile.lock /rails_app/
 ADD ./Jarfile /rails_app/
 ADD ./Jarfile.lock /rails_app/
 
-RUN bundle install --without development test
+RUN bundle install
```
camallen: don't do this, it adds spring and figrake back to prod env. do we need these gems installed?

edpaget: My bad I'll back it out.

Issue #292, opened 2014-11-17
Classifications and Subject Storage


edpaget:
Storing in Postgres is going to be slow.

camallen:
We'll probably need a GIN index on the 'set_member_subject_ids' field on the classifications table to filter classifications / subject for stargazing.

edpaget:
We've decided to use Cassandra for storage if its needed in the future, but we won't be using for stargazing.

camallen:
If greg wants to retrieve classifications by subject Ids then we will still probably want the index on the db: https://github.com/zooniverse/Panoptes/blob/829be6289731513dacc8ad7d5abb1f2ed0e6f8cb/db/schema.rb#L59

Might be costly long term so we could ditch it after stargazing.

edpaget:
I was still intending to save classifications for that query in some other way for greg to access.

camallen:
Interesting, I was planning running a read only rds replica with non-public api nodes so greg can run long running queries. Other solutions welcome!

ggdhines:
So support for JSON in Cassandra seems to be slightly limited - http://stackoverflow.com/questions/25026887/json-type-in-cassandra-data-model. You can get around that by using User Defined Types which (I think) make you spell out the exact details of everything about a JSON object in advance - definitely removes the flexibility. Also, the current Python/Cassandra connection (cqlegine - https://github.com/cqlengine/cqlengine) does not currently support UDTs - https://github.com/cqlengine/cqlengine/issues/317. They are working towards it but definitely not there yet. So for the mean time, JSONs in Cassandra are just going to be strings which means not being able to search/sort based on any fields in the JSON object.

camallen:
Thats not entirely correct, we can use maps to store arbitrary data (e.g. JSON):
http://www.datastax.comhttp://www.datastax.com/documentation/cql/3.1/cql/cql_reference/cql_data_types_c.html
An example: http://dba.stackexchange.com/questions/55165/cassandra-query-a-column-with-collection-type

The question really is what is our query (use-case) paths will be and if we should de-normalise the data into their own columns to improve query performance. 

We really need to evaluate and test some different datastores. Cassandra, Riak, HBase are the main contenders for me but we should look broadly like Couchbase, Mongodb, Redis and others.

ggdhines:
[...]

PR #686, opened 2015-03-18
remove the GIN index on subject ids

camallen:
Most likely fastupdate in batches was causing the db problems, http://www.postgresql.org/docs/9.1/static/gin-implementation.html


edpaget:
Should we replace it with a gist index?

edpaget:
Or not worry for now.

camallen:
I am thinking this is tied to #292 and we can probably solve this using different DB technology.

edpaget:
Sounds good.

PR #688, opened 2015-03-19
Add reload_cellect endpoint to workflow

### LINKED ISSUE #669: Reload cellect subjects end point via workflow
Add an endpoint for a project owner / admin to reload cellects workflow data, e.g. after batch import of subjects.

This can probably be bumped to after stargazing but if someone has time it'd be sweet to have this and a wrapper over [this](https://github.com/zooniverse/Cellect/blob/0793d08f82d7ad88c7b4d24a0f4eabc1bb2e2b3d/lib/cellect/client/connection.rb#L13) in the Panoptes CellectClient to reload a cellect workflow.

edpaget:
Closes #669

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -11,6 +11,11 @@ class Api::V1::WorkflowsController < Api::ApiController
     super
   end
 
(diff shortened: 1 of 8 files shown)


camallen:
Super nice!

PR #690, opened 2015-03-19
Add timeout add retry configuration to Cellect.

### LINKED ISSUE #687: Scaling panopes down casues large response times
hmm_blocking_io

It might be linked to cellect going offline and the api tries sending http get requests for subjects and waiting on the responses. Or it might not be blocking and something else is happening. I dont think api nodes shutting down would cause this spike in ruby time either. @edpaget thoughts?

2 cellect nodes went down
1st one:
2015 March 18 23:31:59 UTC
2015 March 18 23:33:01 UTC
2nd one:
2015 March 18 23:26:28 UTC
2015 March 18 23:27:30 UTC

3 API nodes went down
1:
2015 March 18 23:36:11 UTC
2015 March 18 23:37:15 UTC
2:
2015 March 18 23:35:40 UTC
2015 March 18 23:36:48 UTC
3:
2015 March 18 23:25:39 UTC
2015 March 18 23:26:42 UTC

edpaget:
Also call Thread.pass before retries

Close #687

### DIFF
--- a/spec/lib/cellect_client_spec.rb
+++ b/spec/lib/cellect_client_spec.rb
@@ -63,6 +63,16 @@ RSpec.describe CellectClient do
       
       CellectClient.load_user({ 1 => 'test.host' }, 1, 2)
     end
+
+    it 'should retry four times' do
+      counter = 0
+      allow(Cellect::Client.connection).to receive(:load_user) do
+        (counter += 1) < 4 ? raise(StandardError) : true
+      end
+
+      CellectClient.load_user({ 1 => 'test.host' }, 1, 2)
+      expect(counter).to eq(4)
+    end
   end
 
   describe "#remove_subject" do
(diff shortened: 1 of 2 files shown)


spec/lib/cellect_client_spec.rb
```diff
@@ -63,6 +63,17 @@ def raise_error_for(method, times=1)
       
       CellectClient.load_user({ 1 => 'test.host' }, 1, 2)
     end
+
+    it 'should retry four times' do
```
camallen: should this test that the counter actually makes it to 4?

camallen:
Looks good

spec/lib/cellect_client_spec.rb
edpaget: Added a test for that instaed

camallen:
be interesting to see how this works

PR #722, opened 2015-03-31
Add Oauth Application editing

edpaget:


camallen:
I assume this is to override the default behaviour on `/oauth/applications/`?

edpaget:
Yes it just makes sure an an owner and default scopes are set when creating an application.

Issue #702, opened 2015-03-23
Always return all fields for a project


mkosmala:
In retrieving a Project, it's "science case" text field is blank. For example, as part of the project 'Drawing', I get these fields:
```
'title': 'Drawing', 
'introduction': '', 
'created_at': '2015-03-06T19:11:47.984Z', 
'faq': '', 
'updated_at': '2015-03-06T19:11:47.984Z',
'classifications_count': 1, 
'guide': '', 
'private': True, 
'retired_subjects_count': 0, 
'avatar': None, 
'available_languages': ['en-us'], 
'education_content': '', 
'science_case': '', 
'id': '62', 
'description': 'draw me'
```

science_case is listed as '', but I can clearly see the science case on Panoptes-FrontEnd.

edpaget:
No point to try to save query time since the content is eager loaded anyway now.

edpaget:
This is the same as #718

edpaget:
Closed via #724

PR #724, opened 2015-03-31
Always return all content for projects on index

### LINKED ISSUE #718: Project attributes not set on Project creation (but ProjectContents attributes are)
When I create a project and specify the 'education_content','faq', 'introduction', 'results', and 'science_case', these continue to be listed as null when I retrieve the project itself. The 'display_name' and the 'description' get changed to the values I listed.

However, all these values ARE properly set in the associated ProjectContents. 

So why do these fields even exist at the Project level? Wouldn't it make more sense to just have a single 'introduction' (and 'education_content', 'faq', etc.), instead of one in the Project object (which isn't being updated) and one in the ProjectContents (which is being updated)?

edpaget:
Closes #718

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -28,7 +28,6 @@ class Api::V1::ProjectsController < Api::ApiController
                     :science_case,
                     :introduction]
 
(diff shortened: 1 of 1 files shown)


edpaget:
Closes #702

PR #731, opened 2015-04-01
Unique acls

edpaget:
Make sure ACLs are unique and make them deletable.

### DIFF
--- a/app/models/access_control_list.rb
+++ b/app/models/access_control_list.rb
@@ -3,6 +3,7 @@ class AccessControlList < ActiveRecord::Base
   belongs_to :resource, polymorphic: true
 
   validates_presence_of :user_group
+  validates_uniqueness_of :user_group, scope: [:resource_id, :resource_type], message: ". Roles have already been set for this user or group"
 
   def self.scope_for(action, groups, resource_type: nil)
     case action
@@ -12,5 +13,4 @@ class AccessControlList < ActiveRecord::Base
       where(resource: resource_type.scope_for(:update, groups))
     end
   end
-  
 end
--- a/spec/controllers/api/v1/project_contents_controller_spec.rb
+++ b/spec/controllers/api/v1/project_contents_controller_spec.rb
@@ -3,13 +3,6 @@ require 'spec_helper'
 RSpec.describe Api::V1::ProjectContentsController, type: :controller do
   let(:authorized_user) { create(:user) }
   let(:project) { create(:project_with_contents) }
-  let!(:acl) do
-    create(:access_control_list,
-           user_group: authorized_user.identity_group,
-           resource: project,
-           roles: ["translator"])
-  end
-
   let(:api_resource_name) { 'project_contents' }
   let(:api_resource_attributes) do
     %w(id title description science_case introduction guide language team_members faq result education_content)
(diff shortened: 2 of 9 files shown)


app/models/access_control_list.rb
```diff
@@ -3,6 +3,7 @@ class AccessControlList < ActiveRecord::Base
   belongs_to :resource, polymorphic: true
 
   validates_presence_of :user_group
+  validates_uniqueness_of :user_group, scope: [:resource_id, :resource_type], message: ". Roles have already been set for this user or group"
```
camallen: Is the period before roles on purpose?

spec/controllers/api/v1/project_contents_controller_spec.rb
```diff
@@ -24,6 +17,14 @@
   let(:primary_content) { project.primary_content }
 
   describe "#index" do
+    let!(:acl) do
```
camallen: Do you want to extract these out to a common let block or method?

app/models/access_control_list.rb
edpaget: Yes it makes the error message look like "Validation Failed: User groups. Roles.."

spec/controllers/api/v1/project_contents_controller_spec.rb
edpaget: They could be. I could have them at the top level since it causes a validation error in the #destroy tests.

spec/controllers/api/v1/project_roles_controller_spec.rb
```diff
@@ -94,6 +94,28 @@
...
+        post :create, create_params
+      end
+      
+      it 'should return 400' do
+        expect(response).to have_http_status(:bad_request)
+      end
+
+      it 'should have an error message' do
+        msg = json_response['errors'][0]['message']
+        expect(msg).to match(/Roles have already been set for this user or group/)
```
camallen: there is a [helper method for these](https://github.com/zooniverse/Panoptes/blob/5f85c5558245cb75073d5501a58af1a3da468dda/spec/support/controller_helpers.rb#L6)

spec/controllers/api/v1/workflow_contents_controller_spec.rb
```diff
@@ -22,6 +15,14 @@
   let(:primary_content) { workflow.primary_content }
 
   describe "#index" do
+    let!(:acl) do
```
camallen: again, perhaps extract. If they cause errors else where then i don't think it matters, just easier to modify a change in one spot (method perhaps) rather than multiple times.

app/models/access_control_list.rb
camallen: cool.

spec/controllers/api/v1/workflow_contents_controller_spec.rb
edpaget: Yeah I'd have to either extract them and have a conditional not create them for the #destroy tests or extract them into a method a have a before(:each) call that method in each of the describe blocks. I neither one really seems better to me.

spec/controllers/api/v1/project_roles_controller_spec.rb
edpaget: I don't think that helper does what I want though. Since I'd prefer to match to a regex instead of testing exact equality between the json strings.

camallen:
Looks fine to merge.

PR #739, opened 2015-04-06
Update rails jsonb

edpaget:
So a few notes about this update.

* You'll need to completely delete your local postgresql container and pull the new version of the postgresql image
* JRuby doesn't fully support 4.2 yet dude to ActiveRecord issues
* I only converted JSON columns that I thought were likely to need indexing for now.


edpaget:
Also don't update until we've updated our RDS instance

app/models/membership.rb
```diff
@@ -25,20 +25,17 @@ def self.joins_for
...
 
-
   def self.scope_for(action, user, opts={})
     return all if user.is_admin?
     roles, _ = parent_class.roles(action)
     accessible_groups = user.user_groups.where.overlap(memberships: {roles: roles})
-    query = where(user_group_id: accessible_groups)
-            .or(Membership.where(user_id: user.id))
-            .where(identity: false)
+    query = where(user_group_id: accessible_groups, identity: false)
```
camallen: Thoughts on adding a scope to do `identity: false` for these sorts of queries?

config/environments/test.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.configure do
+  ActiveSupport::Deprecation.silenced = true
```
camallen: was this a noise thing? I'd kinda like to know about these but not if they are super noisy.

edpaget: RestPack is unbelievably noisy in controller tests.

app/models/membership.rb
edpaget: I'll give it another try. 4.2 changed the default parenthesis rules a bit and it wasn't working right with the OR statements.

camallen:
looks good.

PR #747, opened 2015-04-20
Use queues

edpaget:
This is to experiment with using queues instead of Cellect for subject selection. No idea if it'll work or if it'll be too slow, but @camallen suggested at least briefly putting it in staging to see what happened.


camallen:
question about grouped selection [here](https://github.com/zooniverse/Panoptes/blob/a38442955fdedc337590627fce0e1e381d942c4e/lib/postgresql_selection.rb#L61) is this sampling from the supplied `group_id` when set?

lib/subject_selector.rb
```diff
@@ -3,47 +3,38 @@ class MissingParameter < StandardError; end
...
+    @user, @workflow, @params, @scope = user, workflow, params, scope
   end
 
   def queued_subjects
     raise workflow_id_error unless workflow
-    user_enqueued = UserSubjectQueue
-                    .find_by!(user: user.user, workflow_id: params[:workflow_id])
-    selected_subjects(user_enqueued.next_subjects(default_page_size))
-  end
+    raise group_id_error if needs_set_id?
```
camallen: Current cellect behaviour is to [select a group randomly](https://github.com/zooniverse/Cellect/blob/0793d08f82d7ad88c7b4d24a0f4eabc1bb2e2b3d/lib/cellect/server/grouped_workflow.rb#L26), do you think we should add this behaviour?

edpaget:
It looks like I don't have the grouped selection right. I'll fix that.

lib/subject_selector.rb
edpaget: Up to you. I'd kind of prefer the client to always decide the group, since randomly sampling from a SQL database is a pita.

camallen: @brian-c is there a use case a project with 1 workflow and many subject groups (like stargazing), where a subject is selected from a group randomly? I think this happens in other projects @eatyourgreens have you used this before?

edpaget: Well the client has the subject_set ids already from the workflow resource links. They can sample from those before sending the request.

camallen:
this looks good to go, are you going to fix this grouped selection in this PR? if not i'll get it merged.

edpaget:
Yeah I just fixed it locally, I'm fixing the spec to make sure it's actually testing grouped selection

lib/subject_selector.rb
brian-c: No problem for the client to do it.

eatyourgreens: @camallen do you mean something like the groups/random/subjects endpoint in Ouroboros? Science Gossip falls back to that if a group isn't chosen by the user. Should give back random subjects, without any regard to group ID. There's also the groups/subjects endpoint, which returns a set of subjects from a single random group. OWD might use that if no diary is chosen.

Sounds like this can be done in the client, anyway.

camallen: yep, if the client has the info then this is point is moot.

PR #755, opened 2015-04-27
Many-To-Many Workflows <-> Subject Sets

### LINKED ISSUE #721: Duplicate Subjset Sets for multiple Workflows
I have a Subject Set that I want to assign to two Workflows. In the docs for "Update subject set links" under "Workflow", it says it "Creates a copy of the subject set if it belongs do a different *project*." And under "Create a Workflow", it says "A SubjectSet that already belongs to another *workflow* will be duplicated when it is linked."

If I assign Subject Set A to Workflow X and then re-assign Subject Set A to Workflow Y, it removes the link between Subject Set A and Workflow X. (Rather than duplicating the Subject Set.) Is this expected behavior?

If I assign Subject Set A to Workflow X and then create Workflow Y with Subject Set A assigned, then the back-end will duplicate Subject Set A so that a copy is assigned to each of Workflow X and Workflow Y. Is that right? (I haven't tried this; it's my interpretation of the docs.)

Why duplicate Subject Sets on Workflow creation but not on assignment to Workflows?

I'm wondering because I am assigning Subject Sets to Workflows after they already exist. So I will have to manually duplicate my Subject Set to assign it to two Workflows. No big deal, but I want to make sure I'm understanding the API correctly.

### LINKED ISSUE #737: Expert subject set is not linked at workflow creation
Posting a new workflow only responds with a "project" link. The expert subject set is created, it's just not linked.

~~Also, it appears that the whatever subject set is first linked to a workflow automatically becomes its expert_subject_set. Kinda weird behavior, might not be a big deal.~~

### LINKED ISSUE #738: Un-linking a subject set from a workflow deletes the subject set
After DELETE https://panoptes-staging.zooniverse.org/api/workflows/608/links/subject_sets/1142
GET https://panoptes-staging.zooniverse.org/api/subject_sets/1142 returns a 404.

edpaget:
Closes #738, Closes #737, Closes #721


edpaget:
@camallen I need to make sure all the migrations apply cleanly to the database before this is merged.

app/models/set_member_subject.rb
```diff
@@ -1,51 +1,58 @@
...
-  def self.by_subject_workflow(subject, workflow)
-    joins(:subject_set)
-      .where(subject_id: subject, subject_sets: { workflow_id: workflow })
+  def self.by_subject_workflow(subject_id, workflow_id)
+    joins(subject_set: :workflows)
+      .where(subject_id: subject_id, workflows: {id: workflow_id })
   end
 
   def self.available(workflow, user)
     fields = '"set_member_subjects"."id", "set_member_subjects"."random"'
```
camallen: small refactor possible here, https://github.com/edpaget/Panoptes/pull/12

edpaget:
This is all sorted now @brian-c do you think anything will need to change on the front end? I think the only api difference will be that subject_sets have an array of workflows.

brian-c:
I'll check, but I don't think the front end ever needs a subject_set's workflows, so this should be good. Thanks.

db/schema.rb
```diff
@@ -11,16 +11,18 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150409130306) do
+ActiveRecord::Schema.define(version: 20150427204917) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "hstore"
```
camallen: when did we start using hstore again?

db/schema.rb
```diff
@@ -11,16 +11,18 @@
...
+  enable_extension "postgres_fdw"
```
camallen: and what are we using the foreign data wrapper for?

docker-compose.yml
```diff
@@ -4,6 +4,8 @@ postgres:
     - "PG_USER=panoptes"
     - "DB=panoptes_development"
     - "PASS=panoptes"
+  ports:
```
camallen: do you need this on by default, i run two db services both on port 5432 but one in docker. The docker one won't start if the port is already taken

db/schema.rb
edpaget: No where as far as I know. I'm not sure what enabled those

camallen: try and recreate the schema for the test env `bin/rake db:migrate:reset RAILS_ENV=test`

lib/gem_ext/rails/join_association.rb
```diff
@@ -0,0 +1,20 @@
+require 'active_record/associations/join_dependency/join_association'
+
+ActiveRecord::Associations::JoinDependency::JoinAssociation.class_eval do
+  def build_constraint(klass, table, key, foreign_table, foreign_key)
```
camallen: is this monkey patching the JoinAssociation class?

edpaget: Yes it has to have special behaviour to do an `IN` for the join instead of an `=`

docker-compose.yml
edpaget: Done

lib/gem_ext/rails/join_association.rb
camallen: cool - good to know.

lib/gem_ext/rails/reflection.rb
```diff
@@ -0,0 +1,59 @@
+require 'active_record/reflection'
+
+ActiveRecord::Reflection.module_eval do
+  def self.create(macro, name, scope, options, ar)
```
camallen: Are you going to try and get this into rails core?

edpaget: I'm not sure. I'm going to extract it into a gem for sure, but I don't know if they'd want it in Rails since it'll only work with Postgres.

lib/gem_ext/rails/join_association.rb
edpaget: or `= ANY` rather.

lib/gem_ext/rails/reflection.rb
camallen: Ah right, def a gem candidate

edpaget: Yeah I wish the way associations were hooked was a little more extensible so I didn't have to do as much monkey patching to make it work.

camallen:
@edpaget looks good to go, you happy for me to merge?

PR #757, opened 2015-04-28
Default queues

edpaget:
Also closes zooniverse/Panoptes-Front-End#229


app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -3,8 +3,32 @@ class Api::V1::SubjectSetsController < Api::ApiController
   resource_actions :default
   schema_type :json_schema
 
+  def create
+    super { |workflow| refresh_queue(workflow) }
```
camallen: i think the block param should be called subjet_set for this controller, more changes below too.

edpaget: Good point

camallen:
This looks good to go to me.

lib/array_association_scope.rb
```diff
@@ -4,13 +4,41 @@ def self.scope(association, connection)
...
   end
 
+  class BindSubstitution
+    def initialize(block)
+      @block = block
+    end
+
+    def bind_value(scope, column, value, alias_tracker)
+      substitute = alias_tracker.connection.substitute_at(column)
+      cast_type = ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Array
```
camallen: Wouldn't the default array column be the correct type?

edpaget: Sorry I don't understand

camallen: I'm wondering why there was a need to create a new Array type object when *I assume* the column that's being bound is an array already.

edpaget: AR assumes there be a `workflow.id = $1` comparison when loading the association, so the cast_type is for the workflow.id column which will be an integer. The bind param needs to be cast to an array of integers since we're doing a `workflow.id = ANY($1)` comparison.

camallen: Ah right, makes sense now. Thanks.

PR #760, opened 2015-04-29
Add configuration, approved, and beta to project

### LINKED ISSUE #455: Projects need a "public" state
@aliburchard, correct me if I'm wrong: I believe last time it was discussed we landed on "play -> private -> public -> promoted".

I don't think there are any fields on the project to store this state.

How about two fields, "live" and "public"; both default to false. You could in theory have public non-live projects, we just won't query for these.

Promoted projects could just be a list we maintain manually at first. I don't think that's a property of the project itself.

### LINKED ISSUE #754: Add Project Configuration
From zooniverse/Panoptes-Front-End#189

edpaget:
Closes #754
Closes #455

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -29,6 +29,7 @@ module Api
                 JsonApiController::NotLinkable,
                 JsonApiController::BadLinkParams,
                 Api::NoUserError,
+                Api::UnpermittedParameter,
                 RestPack::Serializer::InvalidInclude,    with: :unprocessable_entity
 
     prepend_before_action :require_login, only: [:create, :update, :destroy]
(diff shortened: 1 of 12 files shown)


edpaget:
@brian-c and anyone building projects. This changes the default index on Panoptes to only show approved projects.

brian-c:
Nice. Looks like users are allowed to put their own projects into beta. Should we have some validation before potentially wasting beta testers' time?

edpaget:
So when you do `GET /api/projects?beta=true` it'll still only list approved betas.

edpaget:
You'd have to do `GET /api/projects?approved=false` to see non-approved projects.

brian-c:
Ah hah, makes sense.

PR #766, opened 2015-04-30
add live mode field to project 

camallen:
addresses @aliburchard comments about live project https://github.com/zooniverse/Panoptes-Front-End/issues/240. Added a field to determine the status science mode status of a project that can be used to drive the UI and provide warnings to users about changing workflow configs, etc.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -17,7 +17,7 @@ class Project < ActiveRecord::Base
 
   accepts_nested_attributes_for :project_contents
 
-  validates_inclusion_of :private, :live, in: [true, false]
+  validates_inclusion_of :private, :live, in: [true, false], message: "must be true or false"
 
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -17,7 +17,7 @@ class Project < ActiveRecord::Base
 
   accepts_nested_attributes_for :project_contents
 
-  validates_inclusion_of :private, in: [true, false]
+  validates_inclusion_of :private, :live, in: [true, false]
 
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -7,11 +7,11 @@ class ProjectSerializer
              :title, :description, :guide, :team_members, :science_case,
              :introduction, :avatar, :background_image, :private, :faq, :result,
              :education_content, :retired_subjects_count, :avatar, :background_image,
-             :configuration, :beta, :approved
+             :configuration, :beta, :approved, :live
 
   can_include :workflows, :subject_sets, :owners, :project_contents,
               :project_roles
-  can_filter_by :display_name, :beta, :approved
+  can_filter_by :display_name, :beta, :approved, :live
 
   def title
     content[:title]
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -11,7 +11,7 @@ class ProjectSerializer
 
   can_include :workflows, :subject_sets, :owners, :project_contents,
               :project_roles
-  can_filter_by :display_name, :beta, :approved, :live
+  can_filter_by :display_name, :beta, :approved
 
   def title
     content[:title]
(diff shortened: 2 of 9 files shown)


edpaget:
Can this just go in Project configuration if we don't need to index it?

edpaget:
I never added these because it seems like purely a UI thing.

camallen:
I think it should be it's own field and not lumped into a project config at it may get removed by updates `{ live: true, option: 1 }` -> `{ option: 2, other_option: true}`. 

We need to capture the state of the project "launching" to live mode somewhere. And the UI can use this project field to determine the experience re warning the users not to change things.

app/models/project.rb
```diff
@@ -17,7 +17,7 @@ class Project < ActiveRecord::Base
 
   accepts_nested_attributes_for :project_contents
 
-  validates_inclusion_of :private, in: [true, false]
+  validates_inclusion_of :private, :live, in: [true, false]
```
edpaget: Can we remove this validation entirely? It'll always fail if you try to write something other than true or false to a not null boolean column.

db/migrate/20150430084746_add_live_field_to_projects.rb
```diff
@@ -0,0 +1,5 @@
+class AddLiveFieldToProjects < ActiveRecord::Migration
+  def change
+    add_column :projects, :live, :boolean, index: true, default: false
```
edpaget: Can you add a task updates all the existing projects to have live false?

app/serializers/project_serializer.rb
```diff
@@ -6,12 +6,12 @@ class ProjectSerializer
...
              :introduction, :avatar, :background_image, :private, :faq, :result,
-             :education_content, :retired_subjects_count, :avatar, :background_image,
-             :configuration, :beta, :approved
+             :education_content, :retired_subjects_count, :configuration,
+             :beta, :approved, :live
 
   can_include :workflows, :subject_sets, :owners, :project_contents,
               :project_roles
-  can_filter_by :display_name, :beta, :approved
+  can_filter_by :display_name, :beta, :approved, :live
```
edpaget: Why would we want to filter by live?

spec/factories/projects.rb
```diff
@@ -32,6 +34,14 @@
       private(true)
     end
 
+    factory :approved_project do
```
edpaget: This and the live_project factory seem unnecessary since you're already setting them to true in the default project factory.

edpaget:
What is a beta project that isn't live? What is an approved project that's not live?

app/serializers/project_serializer.rb
camallen: Added it in in case there's a use for it in the future, i.e. find all the projects that are in the live state, I should probably add and index on it if that's the case.

edpaget:
What do we do with an approved project that's not live or a beta project that's not live?

camallen:
The `live` project state is set when a project owner is happy with the state and launches to "science mode" as it was previously called. I can change the terminology if it's confusing. 

The `live` setting is used to drive the user experience somewhere in the project creation tool. Based on previous discussions with @aliburchard (and I may be wrong) there would be some UI tool to that translates to "I am happy with my project and am moving to a science mode where I understand the implications of changing my project state". This sets the `live` project state to true and then whenever they want to change the project, the UI is like, hey you're project is in "Science mode" you really don't want to change anything unless you really understand the implications.

A project that we need to review (our approved ones will be in live mode when we're finished with beta's, i.e. "launched") should be in a live state. I think this is a UI issue about ensuring a project ends up in the "science mode" state before being set to beta, etc.

edpaget:
So I'm happy to merge this once it's rebased and some of the comments have been addressed since it is just a single database field, but I still think it's mostly unnecessary.

edpaget:
So I'm happy to merge this since it's just one field (after comments have been addressed and its rebased), but I still think that we should just use the project configuration hash, or brian's suggestion, and not support indexing or filtering by this field.

aliburchard:
yes to everything @camallen said above about the "live" state. 

A small project that doesn't want Zooniverse approval can go directly from "playground" to "live." A project that wants Zooniverse approval goes from "playground" to "beta" to "live" (and "approved" I assume is the switch that gets flipped between beta and live)

app/models/project.rb
camallen: I've added null: false to the column definition however we still need the validation for useful error messages. If live gets set via `project.update_attributes(live: nil)` you will get a `ActiveRecord::StatementInvalid:PG::NotNullViolation: ERROR`

camallen:
@aliburchard i think a project owner can set the beta field to request a review by us for approval. I think in order to do this there project should have been active in "science mode". So something like "playground" to "live / science mode" to "beta" once they think it's producing useful research.

edpaget:
@camallen those two tests that failed should just be removed.

PR #767, opened 2015-04-30
remove unused vibility code

camallen:
seems this was used with pundit originally and just hung around.

Also there's a bunch of `visible_to` fields in the API docs which seems vestigial, see here: http://docs.panoptes.apiary.io/#reference/collection/collection. I'll remove them and update the PR if they should be removed.


edpaget:
Yeah those should be removed.

PR #771, opened 2015-04-30
add default values and indexes to projects

### LINKED ISSUE #769: Can't create project with approved=false
When I try to create a project with "approved": false, I get a 422 error message saying:
```Only Admins may Approve Projects```

### PR DESCRIPTION
**Pull request by camallen**
Closes #769 

Updated the approved and beta fields. Also it seems add_column index: true doesn't actually add the index.

### DIFF
--- a/db/migrate/20150430132128_update_projects_defaults_and_indexes.rb
+++ b/db/migrate/20150430132128_update_projects_defaults_and_indexes.rb
@@ -1,0 +1,13 @@
+class UpdateProjectsDefaultsAndIndexes < ActiveRecord::Migration
+  def up
+    change_column :projects, :approved, :boolean, default: false
+    add_index :projects, :approved
+    change_column :projects, :beta, :boolean, default: false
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: db/migrate/20150430132128_update_projects_defaults_and_indexes.rb [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+  def up
+    change_column :projects, :approved, :boolean, default: false
+    add_index :projects, :approved
+    change_column :projects, :beta, :boolean, default: false
+    add_index :projects, :beta
+  end
+
+  def down
+    change_column :projects, :approved, :boolean, default: nil
+    change_column :projects, :beta, :boolean, default: nil
```
edpaget: This is fine, but in the future you can have it all under a def change method using `change_column_default` like `change_column_default(:projects, :approved, false)`

camallen: yeah will do.

PR #772, opened 2015-04-30
Don't filter projects by default

### LINKED ISSUE #765: How can we fetch all a user's projects?
`/projects?owner=brian-testing` comes up blank, I'd guess because `/projects` excludes unapproved and beta projects. But `/projects?owner=brian-testing&approved=false` also comes back blank (bug?). And since I actually want any projects the user owns regardless of state, can I maybe query something like `?approved=*&beta=*`?

edpaget:
Closes #765

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -30,12 +30,6 @@ class Api::V1::ProjectsController < Api::ApiController
 
 
   before_action :add_owner_ids_to_filter_param!, only: :index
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -15,12 +15,14 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "hstore"
```
camallen: seems your rake db:migrate tasks adds some funk to the schema.

PR #775, opened 2015-04-30
Add dependent relationships to models

### LINKED ISSUE #652: Revisit destroyable vs deactivatable resources
Double check we've implemented the desired functionality. Are science mode projects (and their sub-resources) allowed to be deleted.

Currently only Users, Groups and Memberships and their sub-resources are deactivatable. All the rest are destroyable if you have rights.

edpaget:
When a project is destroyed its subject sets and associated models
should also be destroyed, but we should keep subjects around in case
they are being used by other projects. Classifications and workflows
should also remain so we can still redo the since if necessary.

Subjects and Workflows can still be explicitly destroyed on their own.

Closes #652

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -25,7 +25,7 @@ class User < ActiveRecord::Base
   has_many :collection_roles, through: :identity_group
   has_many :user_seen_subjects
 
-  has_many :subject_queues, dependent: :destroy
+  has_many :subject_queues
 
   validates :display_name, presence: true, uniqueness: { case_sensitive: false },
             format: { without: /\$|@|\s+/ }, unless: :migrated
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -25,7 +25,7 @@ class User < ActiveRecord::Base
   has_many :collection_roles, through: :identity_group
   has_many :user_seen_subjects
 
-  has_many :subject_queues, dependent: :destroy
```
camallen: Did you remove this because we don't actually delete users? I assume if we ever did then we'd clean up their queues.

edpaget: Yes its up to you. If we do want to set dependent relationships on the user model we should probably add Preferences and Memberships as well.

camallen: Yeah, when deleting on the console (maybe not us but private instances) i'd expect that stuff to get cleaned up.

edpaget: Cool will add that then.

PR #776, opened 2015-04-30
Has many filter

edpaget:


camallen:
Nice work @edpaget !

app/serializers/concerns/filter_has_many.rb
```diff
@@ -0,0 +1,31 @@
...
+  module ClassMethods
+    def page(params = {}, scope = nil, context = {})
+      filters = has_many_filterable_by.map do |filter|
+        filter << params.delete(filter[1])
+      end.delete_if do |filter|
+        filter[2].nil?
+      end
+
+      scope = filters.reduce(scope || self.model_class.all) do |query, filter|
+        scope.joins(filter[0]).where(filter[0] => {id: filter[2]})
```
edpaget: Crap I just realized this should query instead of scope

camallen: Doesn't the RP serializer page method use a scope [here](https://github.com/RestPack/restpack_serializer/blob/b1f3cb62de1af8cca11870aec4fbac5da6d901c1/lib/restpack_serializer/serializable/paging.rb#L5)?

edpaget: Yes, but I'm passing the scope into the reduce block as the `query` variable. How it works now will only allow filtering by one has many param (which is as many as any of the models have).

camallen: Do you want to revert or just fix in another PR?

PR #783, opened 2015-05-01
Set and Workflow relationship should be unique

### LINKED ISSUE #780: Remove dup workflow -> subject_set links
The join table between workflows & subject_sets has dups in it
>eg. workflow.subject_sets.map(&:id) 
=> [1770, 1770, 1770, 1770, 1770, 1769, 1770, 1770, 1770, 1770, 1770, 1770, 1770, 1769]

We should extract to a real model via a has_many :through and add uniq validations on a multi colunm index.

edpaget:
Closes #780

### DIFF
--- a/app/models/subject_set.rb
+++ b/app/models/subject_set.rb
@@ -3,7 +3,8 @@ class SubjectSet < ActiveRecord::Base
   include Linkable
 
   belongs_to :project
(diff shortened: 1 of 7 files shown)


camallen:
looks good, merge when it passes

PR #784, opened 2015-05-01
Use update_all to skip locking during retirement

### LINKED ISSUE #782: RetirementWorker stale object errors
Multiple retirement workers are being kicked off i think [here](https://github.com/zooniverse/Panoptes/blob/4a5318659165ab1450238b5b9bd595d1f2de25b1/app/workers/classification_count_worker.rb#L8) and are touching the `set_member_subejct` objects all at once. I think we should just put an incrementing delay on each one to offset their execution [ala here](https://github.com/mperham/sidekiq/wiki/Delayed-extensions#activerecord).

stale_worker_failure

edpaget:
Close #782


camallen:
looks good

PR #794, opened 2015-05-04
Add variation to non-logged queue

### LINKED ISSUE #779: add variation into selection from non-logged in user queues
sample `set_member_subject_ids`  before taking [here](https://github.com/zooniverse/Panoptes/blob/0777d11048152d6037293528c6fe1edc8aebb83f/app/models/subject_queue.rb#L87)

edpaget:
Closes #779

### DIFF
--- a/app/models/subject_queue.rb
+++ b/app/models/subject_queue.rb
@@ -79,6 +79,10 @@ class SubjectQueue < ActiveRecord::Base
   end
 
   def next_subjects(limit=10)
-    set_member_subject_ids[0..limit-1]
+    if user_id
+      set_member_subject_ids[0..limit-1]
+    else
+      set_member_subject_ids.sample(limit)
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/subject_queue.rb
```diff
@@ -79,6 +79,10 @@ def below_minimum?
   end
 
   def next_subjects(limit=10)
-    set_member_subject_ids[0..limit-1]
+    if user_id
+      set_member_subject_ids[0..limit-1]
```
camallen: Won't all logged in users get the same sort order based on the `create_for_user` copying the `logged_out_queue` on queue creation? So if there is a spike in traffic there might be an uneven distribution of classifications (dependant on the size of the spike).

Perhaps introducing some randomised sorting into the `set_member_subject_ids` when creating the queue would alleviate this and even out users effort.

edpaget: Sure we could do that. I don't think any of it really matters though since we're still retiring once a certain classification count is reached. I think it's probably a good think to directly user attention to few subjects at once as that's likely to produce more discussion on talk.

camallen: Good point about talk, i didn't think about that. Having more "voices" on subject threads in a time period is nice. Leave it as is and we'll see how it goes.

PR #795, opened 2015-05-04
Update queues based on their previous state

### LINKED ISSUE #788: ::dequeue and ::enqueue methods need to be lock free
It should do the same thing as the _for_all versions of those methods, using the built-in Postgres array functions.

edpaget:
Closes #788

### DIFF
--- a/app/models/subject_queue.rb
+++ b/app/models/subject_queue.rb
@@ -29,39 +29,32 @@ class SubjectQueue < ActiveRecord::Base
     end
   end
 
-  def self.reload(workflow, subject_ids, user: nil, set: nil)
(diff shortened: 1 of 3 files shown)


app/models/subject_queue.rb
```diff
@@ -74,6 +67,14 @@ def self.create_for_user(workflow, user, set: nil)
     return queue if queue.persisted?
   end
 
+  def self.dequeue_update(query, sms_ids)
+    query.update_all(["set_member_subject_ids = ARRAY(SELECT unnest(set_member_subject_ids) except SELECT unnest(array[?]))", sms_ids])
```
camallen: nice

PR #799, opened 2015-05-04
Create a new workflow version field and index it

### LINKED ISSUE #680: Index workflow version metadata
Long term request - could we maybe make workflow version a field in the classification table? That way it can be really easy to read in classifications from only the most recent version?

edpaget:
Closes #680

### DIFF
--- a/db/migrate/20150504185433_add_workflow_version_to_classification.rb
+++ b/db/migrate/20150504185433_add_workflow_version_to_classification.rb
@@ -1,0 +1,7 @@
+class AddWorkflowVersionToClassification < ActiveRecord::Migration
+  def change
+    add_column :classifications, :workflow_version, :text
+    Classification.update_all("workflow_version = metadata->>'workflow_version'")
+    add_index :classifications, :workflow_version
+  end
+end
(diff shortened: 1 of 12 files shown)


edpaget:
This failed a couple times because it was returning the csv export path instead of the subjects upload path in the subjects controller tests. I have no idea why that would happen, but this is a note to investigate tomorrow.

db/migrate/20150504185433_add_workflow_version_to_classification.rb
```diff
@@ -0,0 +1,7 @@
+class AddWorkflowVersionToClassification < ActiveRecord::Migration
+  def change
+    add_column :classifications, :workflow_version, :text
+    Classification.update_all("workflow_version = metadata->>'workflow_version'")
```
camallen: Should we remove the metadata[:workflow_version] field as well to have one source of this data?

camallen:
looks good, @edpaget i've issued a PR to fix the failing specs.

@ggdhines note for future work, this PR extracts the workflow version to it's own field from the metadata hash.

ggdhines:
sounds good

db/migrate/20150504185433_add_workflow_version_to_classification.rb
edpaget: I don't think it's worth doing. That field won't be created for new classifications, and since we can't do a bulk update to remove a JSONB field. It'll take days and days to finish that migration on the Production db.

camallen: cool

PR #803, opened 2015-05-05
Disable Deep Munging in Rails

### LINKED ISSUE #743: Empty workflow.tasks.task.answers array turns into `null`
Putting an empty array as the answers to a "single"-type task seems to respond with that value set to `null` instead.

Ask me if you need to see a demo.

Discovered by @chrislintott

edpaget:
Closes #743

Because most users of Rails are apparently unable or unwilling to check
params before passing arguments to AR#find, Rails converts empty arrays
in params to nil. This makes 0 sense to me, but whatever, it's easy to
turn off.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -15,6 +15,7 @@ module Panoptes
     config.autoload_paths += Dir[Rails.root.join('app', 'models', '**/')]
     config.autoload_paths += Dir[Rails.root.join('app', 'serializers', '**/')]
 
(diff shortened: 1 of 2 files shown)


camallen:
Yeah - i'm happy to turn this off for the moment. Looks like https://github.com/rails/rails/issues/13420 should end up sorting where an emprty `[]` param doesn't get converted to nil.

edpaget:
I'm not holding my breath since that issue was opened in 2013.

PR #808, opened 2015-05-05
Don't create an expert set by default

### LINKED ISSUE #805: Don't auto create an expert subject set
From https://github.com/zooniverse/Panoptes-Front-End/issues/243

edpaget:
Closes #805


camallen:
Looks like we don't allow expert sets to be [created via the api](https://github.com/zooniverse/Panoptes/blob/master/app/schemas/subject_set_create_schema.rb). Should we add this behaviour in? If so should only project owners / collabs be allowed to?

edpaget:
@camallen I've added that now too.

camallen:
looks good to merge if it passes.

PR #815, opened 2015-05-06
use the serialized response object to generate the etags

camallen:
Etags dont' include object associations in the response object so they wouldn't change even though some association content had (i.e. project and project_contents).

From here https://github.com/zooniverse/Panoptes-Front-End/issues/283

Chrome sends an `If-None-Match` with the `ETag` value from the first cached request, Firefox doesn't. Our api is responding with 200 but chrome / safari are interpreting the etag : in-none-match header and deciding it was a 304 response even though the content actually may have changed depending on how we generate the etag. For what it's worth ETag's seems to be set by rails even when I had perform_caching turned off. I couldn't figure out why they were being set in the response objects but could be linked to the if-none-match header being set.

This is a simple attempt to use the fully serialized response obj to generate the etag. It should work for the PUT | DELETE on a resource if it hasn't changed but we have to pay the cost of duplicate serialization of the object. It's something like this or we figure out how to generate the etag based on all the serialized object associations so responses have a different ETag and thus are 200's and not 304's.

**UP FOR DISCUSSION - NO NEED TO MERGE JUST YET**

### DIFF
--- a/lib/json_api_controller.rb
+++ b/lib/json_api_controller.rb
@@ -1,6 +1,6 @@
 module JsonApiController
   extend ActiveSupport::Concern
-  
+
   class BadLinkParams < StandardError; end
   class PreconditionNotPresent < StandardError; end
   class PreconditionFailed < StandardError; end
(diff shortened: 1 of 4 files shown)


edpaget:
So a way to do this would be to use modify the model's cache_key method. The way rails calculate etags by default is just by hashing a model's cache_key, which is usually just the updated_at field. See: https://github.com/rails/rails/blob/69eadfd35c384d8ce66ac49671798745a8ffa1f9/actionpack/lib/action_dispatch/http/cache.rb#L86-L87

We could add other parts of a model that invalidate its etag to its cache_key

camallen:
I'll look into using the cache key to include the associations

camallen:
Superseded by #825

Issue #696, opened 2015-03-21
Denormalize Recent Subjects to new table


camallen:
https://github.com/zooniverse/Panoptes/blob/master/app/controllers/concerns/recents.rb#L12

Looks like we are using it somewhere. Perhaps we can store these in a FIFO queue type data structure?

edpaget:
I can't find any info about whether ANY works with GIST indexes, but using that could also be an option. 

Otherwise I agree with denormalizing recents maybe related to #292

camallen:
@edpaget just wondering how you're going to achieve this? I had an idea about appending a per project set of recent classification IDs on the user project preferences model that gets built in the background like a FIFO queue. We could extract these ID's for a simple scope like `Classification.where(id: [recent_ids])`. Thoughts?

edpaget:
I was literally just gong to make it its own table. That just related classification_id to a subject_id. You'd be able to filter it by project and workflow

camallen:
So the recents will really just be a full history?

edpaget:
I think it has to be. I'd like to find a way to handle this without having to use a full table, but I'm not sure how.

camallen:
Sounds fine, i was thinking more along the lines of a [capped collection](http://docs.mongodb.org/manual/core/capped-collections/). 

Would this table be a more natural place to store a user's project : workflow classification counts?

edpaget:
I don't think so, the table is just going to be primary_key, classification_id, subject_id

Issue #425, opened 2015-01-06
Pluggable Subject Storage


edpaget:
Would make it easier for third parties to use Panoptes and reduce our AWS dependency. As in #422

edpaget:
This is done enough as far as I'm concerned if there's a use case for someone to have non-s3 storage it should be pretty easy for them to implement it.

PR #820, opened 2015-05-07
Generic media

edpaget:


edpaget:
This PR also starts to address #425 by adding an adapter for different storage types besides AWS. 

We'll need a replace our aws.yml config file in production/staging with a storage.yml:

```yaml
staging: 
  prefix: panoptes-uploads.zooniverse.org
  bucket: zooniverse-static
```

edpaget:
Also this is going to break recetns until I finish #696 which I'll make my next priority.

lib/subject_locations_extractor.rb
```diff
@@ -42,13 +42,7 @@ def panoptes_locations
...
   def s3_url(path, mime_type)
     if @context[:post_urls]
-      obj = ::Panoptes.subjects_bucket.objects[path]
-      obj.url_for(:write, {secure: true,
-                           content_type: mime_type,
-                           expires_in: 20.minutes.from_now,
-                           response_content_type: mime_type,
-                           acl: 'public-read'}).to_s
-    else
+   else
```
camallen: should this be `unless @context[]` ?

edpaget: This class isn't being used any more and should be deleted.

spec/factories/subjects.rb
```diff
@@ -21,9 +24,9 @@
       end
     end
 
-    factory :migrated_project_subject do
-      locations({standard: "http://www.galaxyzoo.org.s3.amazonaws.com/subjects/standard/1237679543502373086.jpg"})
-      migrated true
-    end
+    #factory :migrated_project_subject do
```
camallen: prob just remove this if we aren't supporting them

db/migrate/20150506195759_remove_avatar_background_from_project.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveAvatarBackgroundFromProject < ActiveRecord::Migration
+  def change
+    remove_column :projects, :avatar, :text
+    remove_column :projects, :background_image, :text
```
camallen: Are we migrating project avatars and background images?

db/migrate/20150506195817_remove_avatar_from_user.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveAvatarFromUser < ActiveRecord::Migration
+  def change
+    remove_column :users, :avatar, :text
```
camallen: Same, are we migrating these as well?

camallen:
@edpaget - This look good to me, really impressive work!

db/migrate/20150506195759_remove_avatar_background_from_project.rb
edpaget: Only the test stargazing projects have them. I figured it'd just be easier to reupload them.

db/migrate/20150506195817_remove_avatar_from_user.rb
edpaget: No one has them yet.

spec/factories/subjects.rb
edpaget: We can support them with the external_url in Medium, but they don't require and special handling or tests from the subject model.

db/migrate/20150506195759_remove_avatar_background_from_project.rb
camallen: http://demo.zooniverse.org/panoptes-front-end/#/projects/mkosmala/Season%20Spotter%20DETAILS%20Beta1

edpaget: Ah Cool I'll add a migration treating these as externally linked.

PR #825, opened 2015-05-08
allow the resource cache_key to include associations and methods

camallen:
Supersedes #815 

allow the resource's cache key to be created from association cache keys and selected method keys e.g. summary counters.

@edpaget I haven't specc'd this yet but wanted to run the idea by you. No need to merge just yet.

### DIFF
--- a/app/models/concerns/association_cache_key.rb
+++ b/app/models/concerns/association_cache_key.rb
@@ -1,0 +1,53 @@
+module AssociationCacheKey
+  extend ActiveSupport::Concern
+
+  included do
+    @included_associations = []
+    @included_resource_methods = []
+  end
+
+  module ClassMethods
+    def cache_by_association(*associations)
+      associations.each do |association|
+        if self.reflect_on_association(association)
+          @included_associations << association
+        end
+      end
+    end
+
+    def cache_by_resource_method(*resource_methods)
+      resource_methods.each do |method|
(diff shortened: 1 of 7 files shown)


edpaget:
That looks fine to me. I wonder a little bit if we should just hold the counts on the project model? 

Also this will invalidate the cache for every content change even if its for a language the user isn't requesting. I wonder if there's any reasonable way to cache and check for changes for only the requested language.

camallen:
Re counts, we could via background workers (workflow already stores it's retired counts) but i think the current sum methods should be ok as both the counts are stored on the respective models so the summary count query should be fairly quick.
`set_member_subject#set_member_subjects_count` `workflow#retired_set_member_subjects_count` 

The project subjects count couldn't use the `subject#project_id` link as these may not be actively linked to a projects active `subject_sets`. 

I'll have a look at deriving the project_contents association cache key from the desired language key. I don't think it's too much of a big deal to get the resource refreshed when ever it's associations change.

camallen:
@edpaget I think this is ready. I gave up on the idea of cache based on the requesting language. My first naive attempt was to change the DSL syntax to specify an instance state field on the model that the association lookup could reference. It got a bit crufty and I didn't like it too much. 

Are you happy to punt this into future plans if we find the cost of refreshing the resource is too much?

PR #826, opened 2015-05-09
Add management of jisc mail subscriptions

### LINKED ISSUE #748: Keep JISCMail in sync
We need to propagate subscriptions/unsubscriptions to JISCMail by emailing their listserv.

edpaget:
Closes #748

### DIFF
--- a/app/workers/subscribe_worker.rb
+++ b/app/workers/subscribe_worker.rb
@@ -1,0 +1,7 @@
+class SubscribeWorker
+  include Sidekiq::Worker
+
+  def perform(email, display_name)
+    JiscMailer.subscribe(email, display_name).deliver
+  end
+end
--- a/app/workers/subscribe_worker.rb
+++ b/app/workers/subscribe_worker.rb
@@ -2,6 +2,8 @@ class SubscribeWorker
   include Sidekiq::Worker
 
   def perform(email, display_name)
-    JiscMailer.subscribe(email, display_name).deliver
+    if Rails.env == 'production' || Rails.env == 'test'
+      JiscMailer.subscribe(email, display_name).deliver
+    end
   end
 end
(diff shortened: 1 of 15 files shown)


edpaget:
I think [this](https://github.com/edpaget/Panoptes/blob/jisc-mail-workers/app/controllers/api/v1/users_controller.rb#L31) may be some bug in active record. I'll try to track it down after the 21st.

app/workers/subscribe_worker.rb
```diff
@@ -0,0 +1,7 @@
+class SubscribeWorker
+  include Sidekiq::Worker
+
+  def perform(email, display_name)
+    JiscMailer.subscribe(email, display_name).deliver
```
camallen: Do we want to check which env we are in? I.e only run these in prod.

camallen:
looks good.

@astopy does Jisc mail require a certain email address / domain to sub / unsub these accounts? Pretty sure these will come from no-reply@zooniverse.org.

astopy:
no-reply@zooniverse.org <mailto:no-reply@zooniverse.org> will be fine I think. I’ll just need to add the right password to the production config.

app/workers/subscribe_worker.rb
edpaget: Sure I'll make that change.

app/workers/unsubscribe_worker.rb
```diff
@@ -0,0 +1,7 @@
+class UnsubscribeWorker
+  include Sidekiq::Worker
+
+  def perform(email)
+    JiscMailer.unsubscribe(email).deliver
```
camallen: And here too?

astopy:
The password/config should be set up correctly now, ready for when this is live.

PR #827, opened 2015-05-11
Add labelled urls to project resource.

### LINKED ISSUE #524: External URLs (blog, social) for a project(/user/group/subject?)
We should have a way to associate external URLs with a resource (zooniverse/Panoptes-Front-End#91). Projects for sure (blog, Facebook, Twitter, etc.), and probably users and groups, but maybe subjects too, for links to some external data source? I don't think we need to be super descriptive with them, just a list. The client can be responsible for identifying the service and picking an icon for recognized ones.

I'm thinking:

```
{
  "projects": {
    "urls": [
      "https://github.com/brian-c",
      "https://twitter.com/__brian_c__"
    ]
  }
}
```

edpaget:
Closes #524

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -26,7 +26,8 @@ class Api::V1::ProjectsController < Api::ApiController
                     :result,
                     :team_members,
                     :science_case,
-                    :introduction]
+                    :introduction,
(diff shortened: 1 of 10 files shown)


edpaget:
Doesn't support actually doing translations for the labels at the moment, but I will add that at a later date.

Issue #615, opened 2015-03-05
Migrate Zooniverse User Table to Panoptes


edpaget:
So the way I think would work best for handling this is having the Panoptes database replace Zooniverse Home as the authoritative database for usernames/passwords across the the zooniverse. 

Here's a list of things we'll need to do to handle the change:
* [x] Point Ouroboros's `ZooniverseUser` model at the Panoptes database
* [x] Connect the CAS Server to the Panoptes database for authenticating users
* [x] Write a script to migrate users into Panoptes
* [x] Write a script to migrate email subscriptions
* [x] Write a script to migrate classification counts

mrniaboc:
For what it's worth, here are the things I use the Zoo Home DB for at the moment:

- Building email lists for newsletters
- Generating lists of names for author pages
- Updating/maintaining the publications page
- The Zooniverse advent calendar
- Updating Zoo Home (obvs)

edpaget:
This is all good then. We just have to do the migration from ZooHome.

vrooje:
Does this mean the migration hasn't taken place yet? Am thinking ahead to the possible World Bank/Nepal project on Panoptes, which we'd want to email out to the volunteers and let them use their logins.

edpaget:
No we're planning to do it the week before the launch

camallen:
@edpaget are you working on this?

edpaget:
Yes I was planning to find a time next week were we could tolerate some downtime (or at least frozen account creation/password resets) while the migration happened'

camallen:
So have you written the email / count scripts? Also what's the plan with migrating classification counts, where will they live in Panoptes?

edpaget:
Nope. I only added them recently when I realized we'd need them. I'm going to add a classifications_count field the user project preferences model that users per-project classification counts will be saved at.

camallen:
Mind if I take a crack at them? Just been delving into the subscription table to figure out why new project counts weren't showing up on zoo home.

edpaget:
For sure. Do you want to add stuff to denormalize user-per-project classification counts too then?

camallen:
Do you mean add a `classifications_count` and `panoptes_project` field in `UserProjectPreferences`?

edpaget:
I don't know what the second field would be for...

camallen:
Marker to say it's a legacy project outside of panoptes, prob should be inverted if that's the case. Perhaps just legacy|migrated_project?

PR #833, opened 2015-05-12
migrate scripts for legacy projects, user email & activitiy count subscriptions

camallen:
linked to #615, should finish most of it.

@edpaget second commit, i removed the dependent destroy relation from the `*_preferences` belongs to as it should be the other way around. Can you just double check this.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -329,7 +329,7 @@ ActiveRecord::Schema.define(version: 20150512123559) do
     t.jsonb    "preferences",         default: {}
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.jsonb    "activity_count",      default: {}
+    t.integer  "activity_count"
   end
 
   add_index "user_project_preferences", ["project_id"], name: "index_user_project_preferences_on_project_id", using: :btree
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150512012101) do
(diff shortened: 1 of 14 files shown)


db/schema.rb
```diff
@@ -330,7 +329,7 @@
     t.jsonb    "preferences",         default: {}
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.string   "roles",               default: [], null: false, array: true
+    t.jsonb    "activity_count",      default: {}
```
edpaget: Why is activity count a json field? Is there anyway it can just be an integer field, so it can be updated with `UPDATE activity_count = activity_count + 1 WHERE id = 1`? That would prevent use from having to worry about any locking issues.

camallen: So it keeps per workflow counts as the content coming through via the events api [here](https://github.com/zooniverse/Zooniverse-Home/blob/773965dd8d775f941dc8a637cce6024f0bbbf55b/app/controllers/api/events_controller.rb#L25) is a hash that looks like this:
```
 activity_count: {"orchid"=>{"count"=>"2", "updated_at"=>"2015-05-07 13:52:04 UTC"}}>
```
Not sure what the best way of storing those per workflow counts are. Do you know of any project's subscription events in zoo home that would have different workflow counts?

edpaget: As far as I know, it's only used for the ribbon on https://www.zooniverse.org/projects/current. Some projects will have multiple workflow counts like Galaxy Zoo, but I don't think they're ever shown separately inside of zoo-home.

camallen: Ok - i'll dig into the zoo home data more tomorrow and see if it's the data is namespaced to the single workflow / project level and fix it.

PR #834, opened 2015-05-12
Remove subjects from queues

edpaget:
This _kind of_ handles updating queues to reflect changes in subject sets (being dissociated from workflows/having subjects added or removed). It should handle actually removing the items from queues just fine, but introduces a few problems. 

* When a SubjectSet is deleted. It will double queue subject removal through both the `subject_sets_workflows` and `set_member_subjects` callbacks. This shouldn't cause any errors, but may strain sidekiq for large queues. 
* It could empty a user's queue entirely. This was cause the user's initial request for subjects to fail, but will trigger a reload of their queue. This could be addressed by refreshing queues after deletion. 

Overall it may be worth thinking more about destroying related objects. Right now, destroying a project with two Subject Sets with 100k subjects in each them, where the subjects have all been classified at least once, will result in at least 600k Ruby objects being instantiated from the database during the request cycle.

### DIFF
--- a/app/workers/reload_queue_worker.rb
+++ b/app/workers/reload_queue_worker.rb
@@ -18,6 +18,7 @@ class ReloadQueueWorker
   def reload_subjects(set=nil)
     subjects = PostgresqlSelection.new(workflow, nil)
       .select(limit: SubjectQueue::DEFAULT_LENGTH, subject_set_id: set)
+      .compact
 
     SubjectQueue.reload(workflow, subjects, set: set)
   end
--- a/app/workers/reload_queue_worker.rb
+++ b/app/workers/reload_queue_worker.rb
@@ -13,6 +13,8 @@ class ReloadQueueWorker
     else
       reload_subjects
     end
+  rescue ActiveRecord::RecordNotFound
+    nil
   end
 
   def reload_subjects(set=nil)
--- a/app/workers/subject_queue_worker.rb
+++ b/app/workers/subject_queue_worker.rb
@@ -15,6 +15,8 @@ class SubjectQueueWorker
     else
       load_subjects
     end
+  rescue ActiveRecord::RecordNotFound
+    nil
   end
 
   private
--- a/app/workers/subject_queue_worker.rb
+++ b/app/workers/subject_queue_worker.rb
@@ -22,6 +22,7 @@ class SubjectQueueWorker
   def load_subjects(set=nil)
     subject_ids = PostgresqlSelection.new(workflow, user)
       .select(limit: limit, subject_set_id: set)
+      .compact
     unless subject_ids.empty?
       SubjectQueue.enqueue(workflow, subject_ids, user: user, set: set)
     end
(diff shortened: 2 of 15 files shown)


app/workers/reload_queue_worker.rb
```diff
@@ -13,11 +13,14 @@ def perform(workflow_id)
     else
       reload_subjects
     end
+  rescue ActiveRecord::RecordNotFound
```
camallen: Wondering here if a user:workflow queue had existed but the workflow was deleted before this event fired via sidekiq. If we should then cleanup any queues that may exist for that workflow?

app/workers/subject_queue_worker.rb
```diff
@@ -15,6 +15,8 @@ def perform(workflow_id, user=nil, limit=SubjectQueue::DEFAULT_LENGTH)
     else
       load_subjects
     end
+  rescue ActiveRecord::RecordNotFound
```
camallen: Same thing here -> Wondering here if a user:workflow queue had existed but the workflow was deleted before this event fired via sidekiq. If we should then cleanup any queues that may exist for that workflow?

edpaget: That should be handled https://github.com/zooniverse/Panoptes/blob/master/app/models/workflow.rb#L17

camallen: Got it.

camallen:
>It could empty a user's queue entirely. This was cause the user's initial request for subjects to fail, but will trigger a reload of their queue. This could be addressed by refreshing queues after deletion.

What's the drawback with scheduling a queue rebuild if it's below the threshold?

> When a SubjectSet is deleted. It will double queue subject removal through both the subject_sets_workflows and set_member_subjects callbacks. This shouldn't cause any errors, but may strain sidekiq for large queues.

I'm not so worried about sidekiq, the N+1 db requests and instantiating all those objects in the api nodes do though.

>Overall it may be worth thinking more about destroying related objects. Right now, destroying a project with two Subject Sets with 100k subjects in each them, where the subjects have all been classified at least once, will result in at least 600k Ruby objects being instantiated from the database during the request cycle. 

Good points, here. I assume (and i've checked again) that there is no way to know if the destroy context is via a callback or directly on the model instance in AR/rails? @parrish ?

Perhaps we can just extract all this associated deletion strategy to workers and get rid of dependent callbacks. So just remove the requested resource in the response cycle and queue deletions for later? Prob something to think about for future plans.

edpaget:
Doing some sort of delayed removal sounds reasonable to me, but we could hold off since stuff won't be too large to start given the user limits and stuff.

camallen:
I'm happy for this to go through - we should test this out as well when it's live.

camallen:
@edpaget did you want to add queue refreshing after deletions or do you think this would be better handled on a per requests basis by the user?

edpaget:
@camallen I'll take a look and see if there's a way that will work.

edpaget:
@camallen I think this should be all set

PR #835, opened 2015-05-13
Add limit on subjects uploaded from a user

### LINKED ISSUE #774: Hard limits for normal users?
Should we enforce limiting the number of projects a normal user can spin up? Also how many subjects and subject sets can they create for a project?

edpaget:
Adds a configuration file user_limits.yml with the form:
```
rails_env:
  subjects: 1000
```

Leaving it empty for an environment will cause it to not limit the
number of subjects.

Closes #774

### DIFF
--- a/config/user_limits.yml.hudson
+++ b/config/user_limits.yml.hudson
@@ -1,0 +1,5 @@
+development:
+
+test:
+  subjects: 100
+  projects: 100
--- a/config/user_limits.yml.hudson
+++ b/config/user_limits.yml.hudson
@@ -2,4 +2,3 @@ development:
 
 test:
   subjects: 100
-  projects: 100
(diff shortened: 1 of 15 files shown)


config/user_limits.yml.hudson
```diff
@@ -0,0 +1,5 @@
+development:
+
+test:
+  subjects: 100
+  projects: 100
```
camallen: do you want to remove projects since we're not limiting on it?

camallen:
looks good to merge.

config/user_limits.yml.hudson
edpaget: Ah yeah will do

brian-c:
Subjects are limited to 1000 per set or per project or per owner?

edpaget:
Per uploader.

PR #838, opened 2015-05-14
Return something on subject requests

edpaget:


app/models/user_seen_subject.rb
```diff
@@ -14,4 +14,8 @@ def add_subjects(subjects)
     subject_ids.concat(subjects)
     save!
   end
+
+  def already_seen?(ids)
```
camallen: can you change this method name to be not_seen or something? Looks like thats what it's actually doing.

Also seems the set intersection for large seen sets will be slower than just iterating over the classification subject ids and searching the seen set via `include?` https://gist.github.com/camallen/29a63c16b03bb9caa255

spec/lib/classification_lifecycle_spec.rb
```diff
@@ -88,37 +76,66 @@
...
-        subject.transact! { true }
       end
 
       it "should call the #publish_to_kafka method" do
         expect(subject).to receive(:publish_to_kafka).once
-        subject.transact!
+      end
+    end
+
+    context "when the classification subjects have not been seen" do
```
camallen: can you change the wording to have been seen before or something like it? Please excuse the the pedantry but I had trouble reading this.

camallen:
looks good

Issue #830, opened 2015-05-12
non-logged in user in same session can keep repeat classifying subjects already classified


mschwamb:
testing out [P4:Terrains](http://demo.zooniverse.org/panoptes-front-end/#/projects/mschwamb/Planet%20Four:%20Terrains/classify)  - there are only 6 images in the set and I can keep repeat classifying them on firefox with no stop. If I'm logged in it brings up the no-more subjects data message. Same issue occurs on Chrome.

mschwamb:
checked at a different ip address. Same result.

edpaget:
This is intended behaviour. We don't track subjects logged out users have seen.

camallen:
@mschwamb you might want to make an issue in the front-end to warn the user when they've already classified this subject for that session.

chrislintott:
@camallen But we are recording classifications, right? It's just that they're not repeated.

brian-c:
Signed-in users will get a warning when #571 is fixed and zooniverse/Panoptes-Front-End#135 is merged.

camallen:
All classifications are recorded. 

Brian, does the front end skip subjects that have already been seen in a session? Non logged in users may have this problem with small data sets and slow moving projects that don't retire quickly. 

Also are we prompting to sign up / log in after x classifications?

brian-c:
The front end doesn't skip any subjects. I could put the same warning as logged-in users get though. Is there any incentive to register in knowing when you've seen repeats?

I've got a PR open for prompting (zooniverse/Panoptes-Front-End#264). I'll ping @chrissnyder to merge it.

camallen:
@brian-c I thought all our front ends kept a list of project subjects that have been seen for a session to avoid people accidentally seeing duplicates (and getting upset that we've wasted their time).

Check out [this project](http://demo.zooniverse.org/panoptes-front-end/#/projects/klmasters_panoptes/VOLCROWE/classify) for an example, it only has 6 subjects so a non-logged in user will just loop through these without any warning.

PR #843, opened 2015-05-16
do not create user seens for anonymous classifications

camallen:
Fixes: https://app.honeybadger.io/projects/40595/faults/12807870 which is backing up in sidekiq with retires.

Do not attempt to track seen for anonymous classifications. Also always count towards retirement for anonymous users but don’t count towards retirement if a known user has seen them before.

@edpaget this does raise the question of anonymous users seeing the same subject in a session as per [here](https://github.com/zooniverse/Panoptes/issues/830#issuecomment-101422062) and if we should count these towards retirement as the two classifications are now not independent of each other. @ggdhines does this matter do you think? If we need to ignore these for retirement counts / aggregation then we'll need another field on the classification to track it.

### DIFF
--- a/app/workers/classification_worker.rb
+++ b/app/workers/classification_worker.rb
@@ -4,14 +4,16 @@ class ClassificationWorker
   sidekiq_options queue: :high
 
   def perform(id, action)
-    classification = ClassificationLifecycle.new(Classification.find(id))
+    classification_lifecycle = ClassificationLifecycle.new(Classification.find(id))
     case action
     when "update"
-      classification.transact!
+      classification_lifecycle.transact!
     when "create"
-      classification.transact! do
-        classification.subject_ids.each do |sid|
-          ClassificationCountWorker.perform_async(sid, classification.workflow.id)
+      classification_lifecycle.transact! do
+        if should_count_towards_retirement?
+          classification.subject_ids.each do |sid|
+            ClassificationCountWorker.perform_async(sid, classification.workflow.id)
+          end
         end
         create_project_preference
       end
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -18,12 +18,12 @@ class ClassificationLifecycle
 
   def transact!(&block)
     Classification.transaction do
-      unless seen_subjects.subjects_seen?(subject_ids)
+      if unseen_subjects_for_user?
         mark_expert_classifier
         update_seen_subjects
         dequeue_subject
-        instance_eval(&block) if block_given?
       end
+      instance_eval(&block) if block_given?
       publish_to_kafka
     end
   end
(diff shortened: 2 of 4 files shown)


ggdhines:
I would discard multiple classifications from the same IP address - doesn't guarantee a solution since a non-logged in user's IP address can change. And two people could be using the same IP address. So I personally would not count multiple classifications from the same IP address towards the retirement. Might be nice to look at how often the same IP address (or similar IP addresses) classify the same subject. Because of the issues I just mentioned, this wouldn't give us a perfect view of what happens with anonymous users but would probably give us a rough estimate - and hopefully the analysis would be "probably not a problem"

camallen:
@ggdhines - we always store classifications no matter what. 

I think we can't distinguish classifications from the same IP without capturing more information. I could see a case where 2 non-logged in volunteers are on a private NAT'd lan are sitting with their computers and one encourages the other to jump on and classify. They'd both have the same IP but they are different classifiers and the annotations are independent. 

Ideally when we show a user a subject they've already seen (and thus the annotations may not be independent) we can record that on the classification metadata perahps using something like `metadata: { ... seen_before: true }`. That way the aggregation engine / research teams  can choose to ignore them if they want to.

@ggdhines if you think we need to record this information for use in the aggregation engine then we'll need to add that field in on the front-end and ensure it's captured properly in the backend. See #830 and https://github.com/zooniverse/Panoptes-Front-End/issues/312 for details.

ggdhines:
Since subjects are randomly assigned, not sure how often the scenario you mention would lead to multiple classifications of the same subject by the same IP address. Would be something to double check.
When I'm reading in classifications, if the same IP address has already viewed a subject, I just ignore that new classification. Easy enough for me to do - I don't think we need an additional field for that.

camallen:
Ok - but I'm not sure that proxying the dup classifiers on the IP address is the best solution.

In small subject sets different non-logged in volunteers could end up classifying the same subject from the same IP address. There is some randomised selection from the non-logged in queue but if the queue is the same size as the selection page then it will end up the same list with a different sort order, see [here](https://github.com/zooniverse/Panoptes/blob/master/app/models/subject_queue.rb#L94). 

So for edge cases, if the subject set is small enough (small scale projects) OR if most subjects have been retired already (especially in small projects) then users could see the same subject twice in session and / or two independent non-logged in users from the same IP could classify the same subject.

ggdhines:
I hadn't been thinking about small scale projects. I think we might need to just wait and see one in action to see how much of a problem it could be. One possibility is that we could have a "switch" where we either include all classifications, even if they are from the same IP address, or we only consider the first one. We could run the analysis both ways and if there is a major difference we could present the people running the project with a warning, giving the two different results and explaining why there is a difference. Not sure what else we could do. But I'd really like to see some small scale projects in action first.

camallen:
Sounds good.

PR #845, opened 2015-05-17
ensure the user details migrate correctly

camallen:
Fixes: https://app.honeybadger.io/projects/40595/faults/11655067

I ran into this with my account, the password updated but hash function didn't as the user already existed as a panoptes user. Instead now just update all the user account details and only build the avatar and identity group when it’s a new account.

### DIFF
--- a/lib/zooniverse_user.rb
+++ b/lib/zooniverse_user.rb
@@ -86,30 +86,20 @@ class ZooniverseUser < ActiveRecord::Base
     SecureRandom.hex 64
   end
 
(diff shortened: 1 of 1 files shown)


edpaget:
How are we handling users like yours on staging? Just fixing their passwords when they run into problems?

camallen:
@edpaget - they just need the their hash algorithm update to SHA1. I thought about an update query for all migrated users with a bcyrpt hash and with an updated_at after the migration started should catch most of them.

PR #847, opened 2015-05-18
CSV exports

edpaget:
Adds support for kicking off a classifications_export through the the API.

### DIFF
--- a/spec/controllers/api/v1/media_controller_spec.rb
+++ b/spec/controllers/api/v1/media_controller_spec.rb
@@ -4,48 +4,148 @@ RSpec.describe Api::V1::MediaController, type: :controller do
   let(:authorized_user) { create(:user) }
   let(:api_resource_name){ 'media' }
   let(:resource_class) { Medium }
-  let(:api_resource_attributes) { %w(id src media_type) }
+  let(:api_resource_attributes) { %w(id src media_type content_type created_at) }
   let(:api_resource_links) { [] }
   let(:scopes) { %w(public medium) }
 
-  RSpec.shared_examples "has_many media" do |parent_name, media_type|
-    describe "#index" do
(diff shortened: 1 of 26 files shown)


spec/controllers/api/v1/media_controller_spec.rb
```diff
@@ -4,48 +4,148 @@
...
+          end
+
+          it_behaves_like "an api response"
+        end
+
+        context "when media does not exist" do
+          let!(:resources) { [] }
+
+          before(:each) do
+            parent.send(media_type).destroy
```
camallen: Any reason why we can't just stub find methods here?

spec/lib/media_storage_spec.rb
```diff
@@ -56,5 +56,19 @@
...
     end
+
+    describe "::put_path" do
+      it 'should call stored path on the adapter' do
+        expect(adapter).to receive(:put_file)
+        MediaStorage.put_file
+      end
+    end
+
+    describe "::put_path" do
```
camallen: I assume this is actually `::delete_file`

camallen:
Looks good

spec/controllers/api/v1/media_controller_spec.rb
edpaget: We could stub whatever the media type method is on the parent to return nil. Is there a reason you'd prefer that?

camallen: Wondering if this is adding time to destroy the resources when we could just stub and let the db cleaner wipe up after the spec.

edpaget: Yeah I think it would involve making a whole chain of stubs though, since some of of the instances of that destroy command are also doing an .exists? activerecord test.

PR #855, opened 2015-05-20
add metadata field to flag classifications for already seen subjects

camallen:
Closes https://github.com/zooniverse/Panoptes-Front-End/issues/359 based of discussion in #843 

This adds an optional `metadata#seen_before` boolean attribute that can be submitted if a volunteer (logged in & anonymous) has classified this subject before. @brian-c you ok with this solution if the subject has been seen before?

Happy to make changes, thought I'd get this into code for discussion.

### DIFF
--- a/app/models/classification.rb
+++ b/app/models/classification.rb
@@ -12,7 +12,7 @@ class Classification < ActiveRecord::Base
     :workflow, :annotations, :user_ip, :workflow_version
 
   validates :user, presence: true, if: :incomplete?
-  validate :metadata, :required_metadata_present
+  validate :metadata, :validate_metadata
   validate :validate_gold_standard
 
   scope :incomplete, -> { where(completed: false) }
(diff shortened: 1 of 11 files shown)


edpaget:
This looks fine to me. I'm a bit wary of relying on clients to be well-behaved, but it should only affect when retirement is triggered.

brian-c:
There's already an `already_seen` property on subjects, right? Is this different?

A subject's `metadata` is a direct dump from the manifest, so can we either rename that property or find another place for internal metadata?

camallen:
Yeah this enables the classification to store the `seen_before` state and not have to reference a user's seen set to understand the timing of the duplicate classifications. Also it allows anonymous users to mark that they've seen a subject in a session already.

camallen:
@edpaget, yeah it's totally client dependent. This is just another attribute that the aggregation engine / research teams can use to detect possible duplicate classifications for a user on a subject.

edpaget:
I'm good to merge if this idea seem all right to you @brian-c

brian-c:
No problem. When a back-end change requires front-end changes, would one of you open an issue there and provide the details so I make sure I'm doing things right? Thanks.

camallen:
@brian-c i'll do it now

PR #861, opened 2015-05-20
Add support for updating a password via json

### LINKED ISSUE #858: Support for JSON password resets and updates.
We need to be able to do reset requests and password updates from the front-end.

edpaget:
Closes #858

### DIFF
--- a/config/frontend_redirect.yml.hudson
+++ b/config/frontend_redirect.yml.hudson
@@ -1,0 +1,15 @@
+development:
+  base: "http://localhost:2727"
+  password_reset: "/#/reset-password"
+  
+test:
+  base: "http://localhost:2727"
+  password_reset: "/#/reset-password"
+  
+staging:
+  base: "http://demo.zooniverse.org/panoptes-front-end"
+  password_reset: "/#/reset-password"
+  
+production:
+  base: "https://panoptes.zooniverse.org"
+  password_reset: "/#/reset-password"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -25,6 +25,7 @@ Rails.application.routes.draw do
 
     get "/users/sign_up" => "registrations#new", as: :new_user_registration
     post "/users" => "registrations#create", as: :user_registration
+    put "/users" => "registrations#update"
   end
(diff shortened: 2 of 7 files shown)


config/frontend_redirect.yml.hudson
```diff
@@ -0,0 +1,15 @@
...
+test:
+  base: "http://localhost:2727"
+  password_reset: "/#/reset-password"
+  
+staging:
+  base: "http://demo.zooniverse.org/panoptes-front-end"
+  password_reset: "/#/reset-password"
+  
+production:
+  base: "https://panoptes.zooniverse.org/"
```
camallen: trailing slash will result in `//`

config/routes.rb
```diff
@@ -15,16 +15,13 @@
   end
 
   devise_for :users,
-    controllers: { omniauth_callbacks: 'omniauth_callbacks', passwords: 'passwords' },
-    skip: [ :sessions, :registrations ]
+    controllers: { omniauth_callbacks: 'omniauth_callbacks', passwords: 'passwords', registrations:  'registrations' },
```
camallen: Pretty sure we don't want to include all the devise registration controller functionality (specifically destroy) and instead just add custom routes to the `as :user` block to wire up to the registrations controller actions.

config/frontend_redirect.yml.hudson
edpaget: I don't think those will necessarily be used. I was including them as examples. We'll need to add a config file on s3 whenever we decide on deployment locationa

brian-c:
The email I get points to `https://panoptes-staging.zooniverse.org/users/password/edit?reset_password_token=foo`, should that still be the case with this merged?

edpaget:
Yeah it's a configuration option. You want it to go to https://demo.zooniverse.org/panoptes-front-end/#/reset-password for staging right? I'll add configuration files and restart.

brian-c:
**http://**, no SSL on staging, right?

And https://preview.zooniverse.org/panoptes/#/ for production for right now?

PR #895, opened 2015-05-27
Obfuscate private details in classification dumps

### LINKED ISSUE #894: obfuscate private classification dump info
- [x] usernames should be public by default
- [x] classification ips should be obfuscated with a time salted hash value

camallen:
Closes #894

### DIFF
--- a/lib/presenter/formatter_csv_classification.rb
+++ b/lib/presenter/formatter_csv_classification.rb
@@ -61,7 +61,7 @@ module Formatter
 
       def obfuscate_value(value)
         if obfuscate
-          "#{value}#{salt}".hash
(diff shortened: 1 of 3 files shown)


lib/presenter/formatter_csv_classification.rb
```diff
@@ -58,6 +58,14 @@ def annotations
       def expert
         classification.expert_classifier
       end
+
+      def obfuscate_value(value)
+        if obfuscate
+          "#{value}#{salt}".hash
```
edpaget: As far I can tell Ruby uses Murmur Hash for Object#hash. I think if we actually want to make sure this value is obfuscated we should at least use some sort of cryptographic hash like SHA-1 or -2.

PR #896, opened 2015-05-27
Dequeue and update user seen subjects in sync

edpaget:
Handle this bit of bookkeeping during the request/response cycle to make
sure seen subjects are never shown to users. Also apply the "uniq"
function in PostgreSQL to queues and seen lists to make sure they don't
contain duplicate data.

Enable the "intarray" extension in PostgreSQL, which is installed in
RDS, in order to make sure these array operations and faster and easier to understand.

This should Close zooniverse/Panoptes-Front-End#426 and zooniverse/Panoptes-Front-End#425

### DIFF
--- a/app/models/user_seen_subject.rb
+++ b/app/models/user_seen_subject.rb
@@ -5,14 +5,16 @@ class UserSeenSubject < ActiveRecord::Base
   validates_presence_of :user, :workflow
 
   def self.add_seen_subjects_for_user(user: nil, workflow: nil, subject_ids: nil)
-    uss = find_or_create_by!(user: user, workflow: workflow)
-    uss.add_subjects(subject_ids)
+    uss = where(user: user, workflow: workflow)
+    if uss.exists?
+      uss.add_subjects(subject_ids)
+    else
+      uss.create!(subject_ids: subject_ids)
+    end
   end
 
-  def add_subjects(subjects)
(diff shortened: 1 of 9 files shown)


app/models/user_seen_subject.rb
```diff
@@ -5,14 +5,16 @@ class UserSeenSubject < ActiveRecord::Base
...
+      uss.create!(subject_ids: subject_ids)
+    end
   end
 
-  def add_subjects(subjects)
-    subject_ids_will_change!
-    subject_ids.concat(subjects)
-    save!
+  def self.add_subjects(subjects)
+    update_all(["subject_ids = uniq(subject_ids + array[?])", subjects])
```
camallen: I'm slightly concerned that this method could be destructive if used on it's own without a scoped query. I.e. `UserSeenSubject.update_all(["subject_ids = uniq(subject_ids + array[?])", subjects])`.

Is it used anywhere else? If not can it be moved to where it's used in `#add_seen_subjects_for_user`.

camallen:
Looks good, just the one comment on the `::add_subjects` method

app/models/user_seen_subject.rb
edpaget: Sure I'll move it in there.

camallen:
good to merge

PR #898, opened 2015-05-27
Always return some subjects from subjects route

### LINKED ISSUE #882: Associating subject set with workflow appears to work, but doesn't
I've got subject set 8 loaded (382 subjects) and workflow 3 written for preview site [project 3](http://preview.zooniverse.org/panoptes/#/lab/3) (which I'm happy to add anyone to who wants it; it's still a private project), but when I try to classify in the project I get the message "Error: No subjects available for workflow 3". 

I've tried waiting >1 hour (nope) and un-ticking and re-ticking the subject set box in the workflow editor (nope). I don't get any additional errors in the console when I do this.

My console has errors, e.g. related to Zooniverse/Panoptes-front-end#409 (which I suspect I probably should have posted here instead; sorry) and Zooniverse/Panoptes-front-end#412, but nothing that is obviously related to this:

screen shot 2015-05-26 at 14 13 41

edpaget:
If a user has a blank queue make sure to respond by using the database's
random selection with at least a few subjects while their queue is
replenished.

This should Close #882, Close zooniverse/Panoptes-Front-End#402, and
Close zooniverse/Panoptes-Front-End#393

### DIFF
--- a/lib/postgresql_selection.rb
+++ b/lib/postgresql_selection.rb
@@ -10,38 +10,20 @@ class PostgresqlSelection
     limit = opts.fetch(:limit, 20).to_i
 
     results = []
-    enough_available = false
-    first_pass = true
-    until results.length >= limit do
-      if !enough_available && !first_pass
-        if limit < available_count
-          enough_available = true
-        else
-          results = available.all.shuffle.map(&:id)
-          break
-        end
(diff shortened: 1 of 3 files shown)


camallen:
looks good.

PR #900, opened 2015-05-28
Custom callback for creating user group slugs

edpaget:
Make sure no empty slugs are created and only update slugs if the
display_name actually changed


camallen:
looks good

PR #916, opened 2015-05-29
Select users to notify about classification dumps

### LINKED ISSUE #914: Add mail collaborator param to request classification dump
Linked to https://github.com/zooniverse/Panoptes-Front-End/issues/446#issuecomment-106904364

Email requester by default, with a query param option to mailing all collaborators.

edpaget:
Pass a metadata key "recipients" when a classification export is created
with user ids and only those users will receive email notifications when
the dump is ready.

If no key is passed the project owner will be notified.

Closes zooniverse/Panoptes-Front-End#446
Closes zooniverse/Panoptes-Front-End#399
Closes #914

### DIFF
--- a/app/controllers/api/v1/media_controller.rb
+++ b/app/controllers/api/v1/media_controller.rb
@@ -96,7 +96,7 @@ class Api::V1::MediaController < Api::ApiController
 
   def controlled_scope
     case media_name
-    when "classifications_exports"
+    when "classifications_export"
       :update
     else
       action_name.to_sym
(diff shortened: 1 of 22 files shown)


camallen:
Is the `metadata: { "recipients": "..." }` param  being set from the front end? If so looks like it'll need to be allowed through the controller params. 

Also looks like the projects create export is building a new medium every time. Should it try and update like the dump worker does if one already exists for a project?

edpaget:
Ah yeah I'll need to fix the metadata param. I was looking at the wrong create method for that. 

Also I was thinking about keeping a log of classification exports, which is why I was always creating a new medium and attaching it to that, but instead I'll have a project always keep around only it's most recent export and make it easier to link to it from the Front-End interface.

edpaget:
Ack just realized non-privileged users could get a link to the export by doing a /projects?include="classifications_export", I'll need to address that too.

edpaget:
This should be good to go.

PR #917, opened 2015-06-02
track the project id a user signs up for

### LINKED ISSUE #901: display_name should always be treated as case insensitive
We should store it with the supplied case but for searches and logins we should ignore the user name's case.

camallen:
closes #901 allow the project_id param to track which project a user is signing up for when registering.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,6 +2,9 @@ class User < ActiveRecord::Base
   include Activatable
   include Linkable
 
+  #on registration track which project the user signs up with
+  attr_accessor :project_id
+
   devise :database_authenticatable, :registerable,
     :recoverable, :rememberable, :trackable, :validatable,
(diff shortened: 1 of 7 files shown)


camallen:
@mrniaboc any desire to know which project a user signed up for after the fact? If so i'll have to persist this in the database.

mrniaboc:
Yes, I think that will be very useful to know.

app/models/user.rb
```diff
@@ -2,6 +2,9 @@ class User < ActiveRecord::Base
   include Activatable
   include Linkable
 
+  #on registration track which project the user signs up with
+  attr_accessor :project_id
```
edpaget: Are you not persisting this anywhere?

camallen: nope, was just going to pass it to the welcome emailer to modify the content there. I'll add a field to the users table to track it.

edpaget: :+1:

Issue #720, opened 2015-03-31
Types should be predictable.


brian-c:
A project's "introduction" value is `null` until it's manually defined. An introduction is a string. It should never be anything else. It should never be `null`.

This is an easy to fix in the client, but I don't think that's the right place to do it. Let me know if I'm wrong.

cc: zooniverse/Panoptes-Front-End#170

camallen:
So you want missing data values to be represented as empty based on their types? I think strings kind of make sense here, `introduction: ""` but what about numbers `count: 0`?

brian-c:
Yup, that'd be ideal.

camallen:
I'm thinking with my logic hat that missing (null) is different to the blank / empty values of a type and there is a semantic difference here.

camallen:
Specifically showing that a value isn't set by returning a null value versus an actual value that may be allowed to be blank based on type.

brian-c:
The client generally doesn't care if a value has been set by a human or not. Maybe this is just a matter of defining sensible default values.

Does something like `classifications_count` not start at 0?

camallen:
Our summary counters will most likely default to 0 (i'd have to double check) but some other integer fields may not be set. 

I'm just wondering if this is better dealt with client side on object initialisation (but you don't have type information) traded-off against a possible API use case that requires knowledge of set / unset field values. I'm leaning towards implementing what the client requires and setting blank values in the serialized responses, @edpaget thoughts?

edpaget:
I'm fine with setting blank values if that's easier for brian to deal with.

camallen:
Started to handle this in #921, https://github.com/camallen/Panoptes/commit/bfbf32dd912e081ae6b23a76248c4789814bd203

edpaget:
I think that'll work fine for now, but we'll probably end up having to reflect on the database column or declare types in the serializer in order to handle the full spectrum of potentially null-able values.

PR #921, opened 2015-06-02
add user subject limits

### LINKED ISSUE #910: Add user subject counts and hard limit number to user serialiser
Linked to https://github.com/zooniverse/Panoptes-Front-End/issues/436

Show a logged in user's subject count and hard limit in their serialised response.

camallen:
closes #910, add the user’s uploaded_subjects_count and max_subjects to the serialised user response.

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -47,7 +47,7 @@ class UserSerializer
   private
 
   def permitted_requester?
-    @perrmitted ||= @context[:include_private] || requester
+    @permitted ||= @context[:include_private] || requester
   end
 
   def requester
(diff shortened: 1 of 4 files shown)


app/serializers/user_serializer.rb
```diff
@@ -58,4 +66,9 @@ def show_firebase_token?
   def add_links(model, data)
     data[:links] = {}
   end
+
+  def permitted_value(value)
+    permitted_requester? ? value : ""
```
edpaget: With regards to #720, can set a different value depending on the attribute's type. I think we said 0 was okay for Integer fields as a 'blank' value.

edpaget: Also do you mind fixing my spelling error in the `permitted_requester?` method?

camallen: Yeah - thought about doing it properly and respect #720. Will do so now.

PR #933, opened 2015-06-03
Talk api worker

edpaget:
Adds a worker for creating project admins when a project is created.

### DIFF
--- a/lib/tasks/talk.rake
+++ b/lib/tasks/talk.rake
@@ -5,6 +5,8 @@ namespace :talk do
   desc "import owners and collaborators as admins of their talk instances"
   task :create_admins do
     client = TalkApiClient.new
-    Project.find_each(&:create_talk_admin)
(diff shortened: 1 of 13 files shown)


lib/tasks/talk.rake
```diff
@@ -0,0 +1,10 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+namespace :talk do
+  desc "import owners and collaborators as admins of their talk instances"
+  task :create_admins do
+    client = TalkApiClient.new
+    Project.find_each(&:create_talk_admin)
```
camallen: do you need to pass the client in as a param here?

camallen:
Looks good to merge if you're happy with the rake task.

lib/tasks/talk.rake
edpaget: Yup whoops

PR #946, opened 2015-06-04
Update project visibility flags

### LINKED ISSUE #941: Project visibility fields
For the interface described in zooniverse/Panoptes-Front-End#240, projects will need some additional fields:

- `private` (Already exists)
- `beta_requested` (Currently just `beta`, which is vague)
- `beta_approved` (Only writable by Zoo admin)
- `launch_requested`
- `launch_approved` (Only writable by Zoo admin)
- `live` (Exists; when changed to false, should also toggle `{beta,launch}_{requested,approved}` to false)

edpaget:
Make sure they're in line with zooniverse/Panoptes-Front-End#243

Closes #941

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -75,17 +75,21 @@ class Api::V1::ProjectsController < Api::ApiController
   end
 
   def build_resource_for_create(create_params)
-    admin_allowed(:approved, :redirect)
+    admin_allowed(:beta_approved, :launch_approved, :redirect)
     create_params[:project_contents] = [ProjectContent.new(content_from_params(create_params))]
     add_user_as_linked_owner(create_params)
     super(create_params)
   end
 
   def build_update_hash(update_params, id)
-    admin_allowed(:approved, :redirect)
+    admin_allowed(:beta_approved, :launch_approved, :redirect)
     content_update = content_from_params(update_params)
     unless content_update.blank?
       Project.find(id).primary_content.update!(content_update)
+    end
+    if update_params[:live] == false
+      update_params[:launch_approved] = false
+      update_params[:beta_approved] = false
     end
     super(update_params, id)
   end
--- a/app/schemas/project_create_schema.rb
+++ b/app/schemas/project_create_schema.rb
@@ -38,11 +38,19 @@ class ProjectCreateSchema < JsonSchema
       type "boolean"
     end
 
-    property "approved" do
+    property "launch_approved" do
       type "boolean"
     end
 
-    property "beta" do
+    property "beta_approved" do
+      type "boolean"
+    end
+
+    property "launch_requested" do
+      type "boolean"
+    end
+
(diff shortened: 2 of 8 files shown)


app/controllers/api/v1/projects_controller.rb
```diff
@@ -75,18 +75,22 @@ def content_from_params(ps)
...
   end
 
   def build_update_hash(update_params, id)
-    admin_allowed(:approved, :redirect)
+    admin_allowed(:beta_approved, :launch_approved, :redirect)
     content_update = content_from_params(update_params)
     unless content_update.blank?
       Project.find(id).primary_content.update!(content_update)
     end
+    if update_params[:live] == false
```
camallen: I think we'll need to add a little more state checking / modification here to ensure the projects switch between states correctly. A state machine like library helper class perhaps?
E.g. a project is 
``` JSON
{ "live": true, "beta_requested": true, "beta_approved": false, "launch_requested": false, "launch_approved": false }
```

and then an api request comes in to just set `{ launch_requested: true }`. The project states would then be 
``` JSON
{ "live": true, "beta_requested": true, "beta_approved": false, "launch_requested": true, "launch_approved": false }
```

edpaget: I'm not sure that's necessary. Project states that aren't 'live' are not used to change any behaviour other than what's shown where on the front-end. What would the problem be in having a project with both the requested flags set?

camallen: So i guess my only concern is having clearly defined states that we use in review to understand a projects progression from `initial` -> `beta` -> `launch` and how we filter these projects to zoo team users. 

Having the above scenario where both `beta_requested: true, launch_requested: true` and then an owner unlocks their project via `live: false` will result in both the requested fields set. This makes it harder to track what state the project is actually in.

camallen:
Also we'll need to add the check that only owner's can remove the lock state of a project from [here](https://github.com/zooniverse/Panoptes-Front-End/issues/240#issuecomment-98984793).
>Lock: If Lock = F, any collaborator can make changes. If Lock = T, only the project owner can make changes.

edpaget:
Yeah I disagree 100% with that. Owners and Collaborators should be treated the same since they are everywhere else. I'm not sure why that particular variable needs to be different.

camallen:
@aliburchard any reason in particular for only allowing the owner to make changes to a live / locked project?

app/controllers/api/v1/projects_controller.rb
edpaget: I'm not sure I understand what's unclear about the the project state in that case. The state is that the project isn't live, and the user has requested review for both launching and beta. All of that information can be presented in whatever interface we have for reviewing projects and the reviewer can determine what needs to happen next. 

If we need to keep track of the history of what's happened to project (ie that is has had a beta period) that seems like an entirely separate issue.

camallen:
Silence is deafening, merging this and we'll review later.

PR #950, opened 2015-06-06
copy set_member_subjects with subject_set

camallen:
Related to #942 copying a subject set didn't copy any subjects. This PR ensures that when copying a subject set from another project that the subjects are also copied.

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -96,8 +96,10 @@ class Api::V1::WorkflowsController < Api::ApiController
     orig_subject_set.dup.tap do |subject_set|
       subject_set.project_id = workflow_project_id
       subject_set.set_member_subjects_count = 0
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -79,15 +82,23 @@ def new_items(resource, relation, value)
...
         item
       end
     end
   end
+
+  def duplicate_subject_set_and_subjects(orig_subject_set)
+    orig_subject_set.dup.tap do |subject_set|
+      subject_set.project_id = workflow_project_id
+      subject_set.set_member_subjects_count = 0
+      subject_set.set_member_subjects = orig_subject_set.set_member_subjects.map do |sms|
```
edpaget: I'm not sure if ActiveRecord uses find_each internally for methods like map, but I don't think it does, so this will load all the SetMemberSubjects right away. It might be better to use find_each or find_in_batches to set this up.

camallen: Good catch - will fix.

PR #953, opened 2015-06-08
make zoo home user subscription migration idempotent

camallen:
reworked the import process to find all uniq user : project combinations and summate their know duplicates if they exist. This process is not repeatable for the same result but is non-optimal due to the use or serialized summary field that can't be queried / summated in the DB.


edpaget:
I'm confused what you mean by it being not repeatable

camallen:
now repeatable...my bad.

edpaget:
:+1:

PR #955, opened 2015-06-08
Add mixin to serialize nil values to defaults

### LINKED ISSUE #720: Types should be predictable.

edpaget:
Closes #720

### DIFF
--- a/app/serializers/access_control_list_serializer.rb
+++ b/app/serializers/access_control_list_serializer.rb
@@ -1,3 +1,4 @@
 class AccessControlListSerializer
   include ACLSerializer
+  include BlankTypeSerializer
 end
--- a/app/serializers/aggregation_serializer.rb
+++ b/app/serializers/aggregation_serializer.rb
@@ -1,5 +1,6 @@
 class AggregationSerializer
   include RestPack::Serializer
+  include BlankTypeSerializer
 
   attributes :id, :created_at, :updated_at, :aggregation
   can_include :workflow, :subject
--- a/app/serializers/classification_serializer.rb
+++ b/app/serializers/classification_serializer.rb
@@ -1,5 +1,6 @@
 class ClassificationSerializer
   include RestPack::Serializer
+  include BlankTypeSerializer
   attributes :id, :annotations, :created_at, :metadata
   can_include :project, :user, :user_group, :workflow
 
--- a/app/serializers/collection_role_serializer.rb
+++ b/app/serializers/collection_role_serializer.rb
@@ -1,5 +1,6 @@
 class CollectionRoleSerializer
   include ACLSerializer
+  include BlankTypeSerializer
 
   attributes :id, :roles
   can_include :user_group, :resource
--- a/app/serializers/collection_serializer.rb
+++ b/app/serializers/collection_serializer.rb
@@ -2,6 +2,7 @@ class CollectionSerializer
   include RestPack::Serializer
   include OwnerLinkSerializer
   include FilterHasMany
+  include BlankTypeSerializer
 
   attributes :id, :name, :display_name, :created_at, :updated_at,
     :slug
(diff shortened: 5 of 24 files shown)


parrish:
Sorry, this just caught my eye.  If you want defaults for attributes, why would you not want to set this on the database?  Serializing attribute defaults differently than their actual default seems pretty confusing.

For what it's worth, restpack properly serializes db defaults when they're set.

edpaget:
In #720 Brian mentioned he'd prefer attributes to never be null, there are a few places, namely the UserSerializer, where attributes won't be shown if a user doesn't have sufficient permission to see them. This just makes them an empty string, or 0, or false instead of null.

parrish:
Yeah, this is kind of all over the place (and maybe I should be commenting on #720).

Null means missing without a default value and we shouldn't try to change that.

If a user doesn't have permissions to view an attribute, it shouldn't be part of the response.

If an actual default exists, it should exist on the database column.

In the examples raised:

> A project's "introduction" value is null until it's manually defined. An introduction is a string. It should never be anything else. It should never be null

Should be a default empty string value.

> but what about numbers count: 0

Counters should always default to zero.

Honestly, setting defaults to fake a DTD of some sort, is going to cause more headaches in the long run.  I'm seriously confused about what's actually motivating this.

edpaget:
So I really have no opinion on whether null values should be returned from the API or not. According to @brian-c it was easier for him to not have them be returned, so we agreed to do that. Not returning an attributes is effectively the same as setting it to null, so those attributes are still returned in the response. 

What kind of issue do think this would cause?

parrish:
> Not returning an attributes is effectively the same as setting it to null

Test whether it's a missing value or a missing attribute.  e.g. `response.foo // => null` vs. `response.hasOwnProperty('foo') // => false`

A simple example of how this could end up biting you:
  - show a model with serialized attribute: 0, but actually nil
  - expect model to show up in a query with the filter ?attribute=0

Other than project contents, the schema seems to have most defaults defined.

It seems like a much easier approach to just add blank/empty defaults to that table.

edpaget:
Yeah I'm curious what @brian-c and @camallen think. I think I mostly agree with what you're saying, but I still think we should return the attributes with some sort of null or empty value when a user isn't permitted to see their actual value to keep responses consistent.

I don't have a problem adding the defaults to contents, so I'll do that as well.

brian-c:
No strong opinion here.

Users' `email` field could be a good example:

A user is allowed to see their own email address, so it's the string `"foo@bar.com"`.

A user is not allowed to see another user's email address, so I'd expect the other user to just not have an `email` field. Their email address isn't `null`, we just can't see it.

Let's pretend a user is allowed to say "I don't want to give you my email address." I'd expect _that_ value to be `null`, (assuming we didn't set it up with some kinda `has_email` boolean).

camallen:
>Null means missing without a default value and we shouldn't try to change that.

Yep

>If a user doesn't have permissions to view an attribute, it shouldn't be part of the response.

As Brian mentions the value isn't null in the DB so a representing that as an empty string is an arbitraty choice for us. I can't see any recommendations in the JSON-API spec about this so I think it's up to us to decide what we want to do. I think it makes the client simpler to just serialize empty / missing values here. 

>If an actual default exists, it should exist on the database column.

Agreed. Rails counters will be set to 0 so we should mimic this behavriour in any of our custom counter fields. 

I think the example of filtering on a value that has been serialized with a different value is a good example of what can go wrong with this. Any value that can be filtered on should have a sensisble default in the DB. If it's a non-DB value then it won't be filterable by restpack anyway. This should take care of most of the problems. 

For non-filterable values, I don't think there is any problem serializing missing values with sensible defaults based on their types.

edpaget:
I'll do this differently.

PR #959, opened 2015-06-09
When talk isn't configured log api attempts

edpaget:
There's no production talk at the moment, so just log that we tried to talk (heh) to it and do nothing.

### DIFF
--- a/app/workers/talk_admin_create_worker.rb
+++ b/app/workers/talk_admin_create_worker.rb
@@ -3,6 +3,8 @@ class TalkAdminCreateWorker
 
   def perform(project_id)
     Project.find(project_id).create_talk_admin(client)
+  rescue TalkApiClient::NoTalkHostError => e
+    Rails.logger.info e # TODO: remove when production talk is deployed
   end
 
   def client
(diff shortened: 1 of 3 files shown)


app/workers/talk_admin_create_worker.rb
```diff
@@ -3,6 +3,8 @@ class TalkAdminCreateWorker
 
   def perform(project_id)
     Project.find(project_id).create_talk_admin(client)
+  rescue TalkApiClient::NoTalkHostError => e
+    Rails.logger.info e
```
camallen: Can you log an error in honey badger as well just so we know this is happening outside of the logs? As per [here](https://github.com/zooniverse/Panoptes/blob/db3030d0c4276f0668fab93f61eb3e41ae6f8564/app/controllers/api/events_controller.rb#L60)

edpaget: I can, but that's just going to clutter up honeybadger with notifications. We can stop rescuing this error when all the Talk environments are deployed and it'll go through to Honeybadger normally.

camallen: Fair call, Let's just remember to rip it back out.

edpaget: I'll add a note

PR #960, opened 2015-06-09
User login

parrish:
This is based on the discussion that came up on zooniverse/Panoptes-Front-End#513.

The gist is that we shouldn't be conflating what a user displays their name as publicly with what they login with.

Moreover, we can restrict wacky characters in logins while supporting old zooniverse home logins via display_name.


app/models/user.rb
```diff
@@ -31,8 +31,9 @@ class User < ActiveRecord::Base
 
   belongs_to :signup_project, class_name: 'Project', foreign_key: "project_id"
 
-  validates :display_name, presence: true, uniqueness: { case_sensitive: false }
-  validates :display_name, format: { without: /\$|@|\s+/ }, unless: :migrated
+  validates :login, presence: true, uniqueness: { case_sensitive: false }
+  validates :login, format: { with: /\A[\w\-\.]{3,}\z/ }, if: :migrated
```
camallen: Always validate the login format.

app/models/user.rb
```diff
@@ -65,9 +68,9 @@ def self.from_omniauth(auth_hash)
     auth = Authorization.from_omniauth(auth_hash)
     auth.user ||= create do |u|
       u.email = auth_hash.info.email
+      u.display_name = auth_hash.info.name
+      u.login = u.display_name.gsub /\s+/, '_'
```
camallen: Should sub also this remove other non-permitted chars in the login that may creep in?

app/models/user.rb
```diff
@@ -82,8 +85,13 @@ def self.reflect_on_association(association_name)
     end
   end
 
-  def self.find_for_database_authentication(conditions = {})
-    where('"users"."display_name" ILIKE ?', conditions[:display_name]).first || super
+  def self.find_for_database_authentication(conditions = { })
+    conditions = conditions.dup
```
camallen: this just reads strangely to me, can you rename the method parameter to be warden_conditions? 

https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address#for-activerecord

db/fig_dev_seed_data/fig_dev_seed_data.rb
```diff
@@ -36,13 +36,13 @@ def yellow(text); colorize(text, 33); end
 
 #setup an admin user
 attrs = { password: password,
-          display_name: 'zooniverse_admin',
+          login: 'zooniverse_admin',
           email: 'no-reply@zooniverse.org' }
 admin = User.create(attrs) do |user|
   user.build_identity_group
```
camallen: Can you please add the `user.admin = true` in here as well? Something that was forgotten about.

db/schema.rb
```diff
@@ -398,6 +399,7 @@
   add_index "users", ["display_name"], name: "users_display_name_trgm_index", using: :gist, opclass: {"display_name"=>:gist_trgm_ops}
```
camallen: @edpaget what's the use case for fuzzy matching on the display_name now? All @ mentions just using login field now.

lib/zooniverse_user.rb
```diff
@@ -41,7 +41,7 @@ def password=(plain_password)
 
   def import
     #use the indexed field
-    user = User.find_or_initialize_by(display_name: login)
+    user = User.find_or_initialize_by(login: login)
```
camallen: If we're converting zoo home login fields to a restricted char set then these finds won't work by default. Suggest we add a sanitized login method on the User model to handle the conversion.

lib/zooniverse_user.rb
```diff
@@ -76,7 +76,8 @@ def setup_panoptes_user_account(pu)
       pu.build_avatar(external_link: true, src: avatar_to_url)
       pu.build_identity_group
     end
-    pu.display_name = login
+    pu.login = login
```
camallen: This will have to be a sanitized login too.

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -402,12 +403,12 @@ def update_request
 
     context "with a valid replace put operation" do
       before(:each) { update_request }
-      let(:new_display_name) { "Mr_Creosote" }
+      let(:new_login) { "Mr_Creosote" }
```
camallen: Are we going to allow people to change their login names? If this is going to be used as the URL slug as per [here](https://github.com/zooniverse/Panoptes-Front-End/issues/513#issuecomment-110486503) then I don't think we should let the login field be updated.

lib/user_info_scrubber.rb
```diff
@@ -21,7 +21,7 @@ def self.scrub_details
     @user.update_columns(email: nil)
   end
 
-  def self.scrubbed_display_name(tries)
+  def self.scrubbed_login(tries)
```
camallen: This isn't used anymore, can go.

db/schema.rb
edpaget: Searching for users. The collaborator editor searches by display name and I assume the @ mentions could too. Similar to the interface for mentioning people on  Github/Twitter/Facebook.

app/models/user.rb
parrish: Probably.  I'll generalize and make the sanitizing more complete.

spec/controllers/api/v1/users_controller_spec.rb
edpaget: I think we should but with a strong warning that it'll break stuff.

parrish:
@camallen Let me know if I missed anything you'd like addressed

camallen:
@parrish looks good :+1: 

Not sure this can be merged straight up since it'll break existing user accounts and logins. @edpaget thoughts on migrating the existing users to work with this? Also should we remove the slug field from users if we're using user logins for url's?

parrish:
Working on a migration now -- got a bit hung up on indexing

edpaget:
@camallen I think I already removed the slug column from users. I think we can eliminate it from UserGroups too and use the "name" field their similarly to "login".

app/serializers/user_serializer.rb
```diff
@@ -3,7 +3,7 @@ class UserSerializer
   include RecentLinkSerializer
   include MediaLinksSerializer
 
-  attributes :id, :display_name, :credited_name, :email, :created_at,
+  attributes :id, :login, :display_name, :credited_name, :email, :created_at,
     :updated_at, :type, :firebase_auth_token, :global_email_communication,
     :project_email_communication, :beta_email_communication,
     :slug, :max_subjects, :uploaded_subjects_count, :admin
```
edpaget: You can remove slug here. And the slug method in the body.

app/models/user.rb
```diff
@@ -126,7 +144,7 @@ def email_changed?
   def build_identity_group
     raise StandardError, "Identity Group Exists" if identity_group
     build_identity_membership
-    identity_membership.build_user_group(display_name: display_name)
+    identity_membership.build_user_group(display_name: login)
```
edpaget: This should still be display_name and the name field should be set to login.

edpaget:
Can we set display_name = login when a user is created with just a login? Just to make sure that field isn't empty for front-end use.

parrish:
I'm setting the display_name like this:

``` ruby
before_validation :default_display_name, on: [:create, :update]

def default_display_name
  self.display_name ||= login
end
```

app/models/user.rb
parrish: The user group has both display_name and name.

Is this right?

``` ruby
user_group.name == user.login
user_group.display_name == user.display_name
```

edpaget: Yes.

app/serializers/user_serializer.rb
parrish: There's a problem with the identity group build:

``` ruby
# User#build_identity_group
identity_membership.build_user_group(display_name: display_name, name: login)

# UserGroup
validates :display_name, format: { without: /\$|@|\s+/ }, unless: :identity?

# UserGroup#identity?
```

`identity?` will never be true on creation since that query can't run on the unpersisted records.
Since User#display_name can be unrestricted, shouldn't UserGroup#display_name be unrestricted?

edpaget: Yes it can be.

parrish: > Yes it can be. 

Is in response to?

> identity? will never be true on creation since that query can't run on the unpersisted records.

(I can't see how that's possible)

or

> Since User#display_name can be unrestricted, shouldn't UserGroup#display_name be unrestricted?

If that's the case, I'll remove the display_name validation, but validate the name attribute with the same format as User#login?

edpaget: Sorry the second one.

camallen:
@parrish - missed these updates, any more work to come on this? Also how long does the user migration take to run?

parrish:
I do need to resolve an index miss from devise not using the case insensitive email index.  Locally at least, I'd ballpark the data migration to run in about 10 hours.

parrish:
I also wanted to double check that you were aware that this method

``` ruby
  def email_changed?
    authorizations.blank?
  end
```
is overriding the change tracking method from ActiveModel::Dirty.

edpaget:
Ah that explains another issue I ran into. I think that was there to work around some sort of issue with Devise when someone signed up using an omniauth provider and refused to provide their email, but I did not leave detailed enough notes about it.

parrish:
I only mentioned it because I ended up tracking down the source of the case sensitive validation to: https://github.com/plataformatec/devise/blob/master/lib/devise/models/validatable.rb#L31

That method can prevent the email uniqueness from being validated.

edpaget:
Go ahead and remove it and see if any tests fail. It was long enough ago, I have no idea what the reason for doing that was, sorry.

parrish:
After removing `email_changed?` it failed a spec:

```
User::from_omniauth an invalid user should raise an exception
 Failure/Error: expect{ User.from_omniauth(auth_hash) }.to raise_error(ActiveRecord::RecordNotUnique)
   expected ActiveRecord::RecordNotUnique but nothing was raised
 # ./spec/models/user_spec.rb:73:in `block (4 levels) in <top (required)>'
```

So I just changed it to fail on the validation (via `create!`) instead of the unique constraint error.

edpaget:
Awesome thanks for tracking that down.

camallen:
This looks good. I've got some concerns about diverging from a consistent schema.rb and the migration ordering. I'm working on a PR to address my concerns and will hopefully have it ready soon.

camallen:
https://github.com/parrish/Panoptes/pull/1

camallen:
@parrish looks like the lowercase email spec on the DB is failing. Perhaps try the .valid? to access the devise email check.

parrish:
@camallen got it in 002a1686dc6a.

Apparently, I still had a lowercase index in my test db.

camallen:
@parrish can you check the Gemfile and Gemfile.lock agree? 

travis is wigging with `You are trying to install in deployment mode after changing`

parrish:
@camallen It was fixed in d675d828f.

Also, you'll want to specify (at the minimum) the major version of any gem you're using in production in the gemfile

edpaget:
@parrish doesn't the Gemfile.lock contain that information already? It won't change your gem version in there unless you run bundle update.

lib/unique_for_owner_validator.rb
```diff
@@ -13,8 +13,8 @@ def validate(record)
     end
   end
 
-  def unique_fields
-    %i(display_name)
+  def unique_fields_for(record)
```
marten: Did you consider putting this on the record class, instead of switching based on the type of record here? ie:

```ruby
class User
  def self.unique_fields; %i(login); end
end

record.class.unique_fields
```

parrish:
@edpaget until you run bundle update.  It's really easy for gems to bump without spec failures, yet introduce some weird behavior in production.  I sent #988 that locks production gems to a very liberal range.

lib/unique_for_owner_validator.rb
parrish: I'm indifferent really.

I feel like this is a bit of a hack anyway.  The custom validator probably ought to be replaced with a scoped uniqueness validation.

marten: Yeah, I had a better look, don't see anything that can't be done with the standard one I think.

camallen: @marten feel free to issue a PR for this one.

PR #985, opened 2015-06-17
Expand README for the first-use experience

marten:
Added a couple minor notes that might not be clear to someone who hasn’t done much with Docker yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ It's possible to run Panoptes only having to install the `fig_rake` gem. Alterna
   + If you ran `bundle install`: `rake configure:dev_docker`
   + If you did not: `cp dockerfiles/Dockerfile.dev Dockerfile`
 
+0. Install [Docker and Docker Compose](https://docs.docker.com/compose/install/).
+
 0. Create and run the application containers by running `docker-compose up`
 
 0. After step 5 finishes, open a new terminal and run `frake db:create db:migrate` to setup the database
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,19 +55,24 @@ It's possible to run Panoptes only having to install the `fig_rake` gem. Alterna
 0. `cd` into the cloned folder. Run either `bundle install` or `gem install fig_rake`
 
 0. Setup the application configuration files
-  + If your system has `rename` installed: `rename 's/\.yml.hudson$/\.yml/' config/*.yml.hudson`
```
edpaget: Please keep this line.

marten: Trouble was that it left a dirty working copy, since the template files are renamed, and thus count as "deleted" as compared to the last commit. Alternatively, perhaps adding `&& git checkout config` would also solve it?

edpaget: I think that's a matter of personal preference. To me a dirty working dir doesn't matter, since I'm not going to `git add .` 

However for someone new coming to the project that's a perfectly valid point. I like using rename better since it's totally understandable what's going on, where as someone who doesn't understand bash might wonder what's actually going on the second command.

marten: Agreed on that, that's why I installed rename instead of copy/pasting magic. Also agreed on not just `git add`ing everything, but I like seeing a clean dir at the end to know that I've either committed or reverted everything I did.

edpaget: Sure not a big deal.

PR #988, opened 2015-06-17
Add version specifications to Gemfile

parrish:
Adds the minimum amount of versioning to the gemfile to reduce breaking changes during gem updates.


edpaget:
Eh so the reason I don't like having explicit versions for anything that's not Rails is to try to keep us on the latest version of Gems. We should update a gem find out if anything breaks when we run our tests and fix our code to keep using the latest version.

Otherwise we end up with the Zoo-Home situation where no one could run it for a while because of stuck Gem dependency.

parrish:
Stuck meaning depending on a yanked version.

The downside to being too specific about versions is that
  - you have to manually change them to update them
  - if you have conflicting dependencies, you need to manually resolve them

The downside to being to liberal with your versions is that
  - you upgrade gems unknowingly unless you're very cautious
  - any gem that you depend on will eventually break your app

Locking versions to a major version number (or minor for 0.x gems) is loose enough that you're not going to be dealing with dependency conflicts in most cases.

You'll still end up with a broken app occasionally when gems aren't 100% semantically versioned. This just adds a bit of sanity to it.

camallen:
I'm ok with major version numbers. Bundler should sort / warn on dependency issues. 

Travis may not pick up on gem bump problems automatically (we cache the build) but our CI docker builds and subsequent AMI staging tests should warn on major failures.

@parrish do you wan to rebase this?

parrish:
@camallen Rebased

PR #991, opened 2015-06-17
Speed up destroy_links for subject sets

### LINKED ISSUE #974: destroy_links is slow
similar to https://github.com/zooniverse/Panoptes/issues/672

Same issue as update_links, a destroy on each as well as any subsequeunt count callbacks, etc. Instead we should bulk delete the `set_member_subjects`, the associated `subject_workflow_counts` and manually queue the QueueRemovalWorker. 

Once the `set_member_subjects` are taken care of we can then remove the `subject_set` and it's `subject_set_workflows` by a normal delete probably (taking care to reset the subject_set#set_member_subject relation).

### LINKED ISSUE #987: sending transcriptions/classifications/markings directly to Cassandra
@rogerhutchings and @VVH would like to run a beta of annoTate starting July 1st. I'd like to be able to read in the transcriptions directly from Cassandra. @camallen thinks this is doable

edpaget:
Delete links without callbacks and manually queue workers. Closes #974

### DIFF
--- a/spec/controllers/api/v1/subject_sets_controller_spec.rb
+++ b/spec/controllers/api/v1/subject_sets_controller_spec.rb
@@ -205,4 +205,38 @@ describe Api::V1::SubjectSetsController, type: :controller do
 
     it_behaves_like "is destructable"
   end
+
+  describe '#destroy_links' do
+    context "removing subjects" do
+      let(:sms) { subject_set.set_member_subjects }
+      it 'should remove the subjects from the set' do
+        delete :destroy_links, subject_set_id: subject_set.id, link_relation: :subjects,
+          link_ids: subject_set.subjects.pluck(:id).join(',')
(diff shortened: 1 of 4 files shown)


camallen:
@ed - can you update the issue that this PR closes? Should be #974

edpaget:
Sorry should be fixed.

spec/controllers/api/v1/subject_sets_controller_spec.rb
```diff
@@ -205,4 +205,38 @@
 
     it_behaves_like "is destructable"
   end
+
+  describe '#destroy_links', :focus do
```
camallen: Can you remove the focus here?

edpaget: I was wondering why travis ran so fast.

PR #994, opened 2015-06-18
Optimize user login migration

camallen:
allow a list of display_name and only migrate users that don't have a login field set.

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -6,8 +6,6 @@ namespace :migrate do
   task user_login_field: :environment do
 
     user_display_names = ENV['USERS'].try(:split, ",")
-
-    total = User.count
     validator = LoginUniquenessValidator.new
 
     null_login_users = User.where(login: nil)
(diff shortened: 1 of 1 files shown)


edpaget:
Users can't have a missing login since the login column has a NOT NULL constraint.

camallen:
I got rid of that here - https://github.com/zooniverse/Panoptes/pull/989/files

Needs to be added back in after the migration.

edpaget:
iNo it's still there in that commit the `change_column :users, :login, :string, null: false` at the end.

camallen:
sorry pasted the wrong link, https://github.com/zooniverse/Panoptes/pull/993/files

Looks like staging got this migration. Did you run it manually on your computer as the jenkins deploy logs just have a failure then a empty success? 

Production got the one I ran this morning to so that column now exists, no null constraint (validation is on the model) and owner and beta users have been migrated. The rest need to be run now.

edpaget:
Yeah I ran the development migration in pretty much the same way you're doing it now. Should I kick of the rest of the migration?

PR #998, opened 2015-06-18
case insensitive owner filter on the idenity group

camallen:
older identity groups may have a different case to the login field so we need to use a case insensitive query path here.

http://preview.zooniverse.org/panoptes/#/projects/VVH/prettiest-pugs
http://preview.zooniverse.org/panoptes/#/projects/OxfordGreg/star-trek-aliens

### DIFF
--- a/app/controllers/concerns/filter_by_owner.rb
+++ b/app/controllers/concerns/filter_by_owner.rb
@@ -8,7 +8,7 @@ module FilterByOwner
   def add_owner_ids_to_filter_param!
     owner_filter = params.delete(:owner).try(:split, ',')
     unless owner_filter.blank?
-      groups = UserGroup.where('lower(name) = ?', owner_filter.map(&:downcase))
+      groups = UserGroup.where(UserGroup.arel_table[:name].lower.in(owner_filter.map(&:downcase)))
       @controlled_resources = controlled_resources
         .joins(:access_control_lists)
         .where(access_control_lists: {user_group: groups})
--- a/app/controllers/concerns/filter_by_owner.rb
+++ b/app/controllers/concerns/filter_by_owner.rb
@@ -8,7 +8,7 @@ module FilterByOwner
   def add_owner_ids_to_filter_param!
     owner_filter = params.delete(:owner).try(:split, ',')
     unless owner_filter.blank?
-      groups = UserGroup.where(name: owner_filter)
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/filter_by_owner.rb
```diff
@@ -8,7 +8,7 @@ module FilterByOwner
   def add_owner_ids_to_filter_param!
     owner_filter = params.delete(:owner).try(:split, ',')
     unless owner_filter.blank?
-      groups = UserGroup.where(name: owner_filter)
+      groups = UserGroup.where('lower(name) = ?', owner_filter.map(&:downcase))
```
marten: This used to do a `WHERE name IN ('a', 'b', 'c')` instead of comparison, does equality comparison work for this too in Postgresql?

marten: Ah, right the specs pass because there's only a single filter. If you'd add a second owner name, you'd see the specs fail with a SQL syntax error:

```
[9] pry(main)> puts UserGroup.where("lower(name) = ?", ['a', 'b'])
  UserGroup Load (1.6ms)  SELECT "user_groups".* FROM "user_groups" WHERE (lower(name) = 'a','b')
PG::DatatypeMismatch: ERROR:  argument of WHERE must be type boolean, not type record
LINE 1: SELECT "user_groups".* FROM "user_groups" WHERE (lower(name)...
                                                        ^
: SELECT "user_groups".* FROM "user_groups" WHERE (lower(name) = 'a','b')
ActiveRecord::StatementInvalid: PG::DatatypeMismatch: ERROR:  argument of WHERE must be type boolean, not type record
LINE 1: SELECT "user_groups".* FROM "user_groups" WHERE (lower(name)...
                                                        ^
: SELECT "user_groups".* FROM "user_groups" WHERE (lower(name) = 'a','b')
from /Users/marten/.rvm/gems/ruby-2.2.1/gems/activerecord-4.2.2/lib/active_record/connection_adapters/postgresql_adapter.rb:592:in `async_exec'
```

camallen:
Good catch, my bad. Can you issue a PR against my branch? Otherwise I'll get to this later.

marten:
Will do.

marten:
See https://github.com/camallen/Panoptes/pull/3 for a fix.

PR #1004, opened 2015-06-22
ensure user login field cannot be null

camallen:
Seems the prod & staging db’s are slightly inconsistent due to manual migrations last week. Ensure the prod system has a not null constraint on the user’s login column.


marten:
LGTM

PR #1007, opened 2015-06-23
Unique per user unsubscribe end point and links in emails

### LINKED ISSUE #865: unsubscribe email end point with token auth
Replicate behaviour from here: https://www.zooniverse.org/unsubscribe for use in links in emails from the zooniverse. 

When a user clicks a link in an email we send them it will unsubscribe the user from all project email lists and set their user global & project email communication preferences to false.

We need to add a per-user token to embed into the link that will authenticate the user request.

camallen:
Closes #865 - add unique user unsubscribe end point `/unsubscribe` using a generated token based on the user's unique login names. 

Also adds links to default email layouts. I wasn't 100% on the multi addressed email (e.g. classification data mailer) view to include the instanct per user unsubscribe token. Instead i've added general links that users can follow (via login) to manage their subscriptions. If needed i'll send one email at a time and include the per user "instant unsubscribe" links in each email.

Also we may need to add a generic unsubscribe via email end point (and front-end form) if we don't have mail merge features in our mailers. @mrniaboc status on mail merge with jisc mail?


camallen:
Still need to queue a jisc mail removal when they unsubscribe.

app/views/layouts/jisc_mail.html.erb
```diff
@@ -0,0 +1,5 @@
+<html>
+  <body>
+    <%= yield %>
```
edpaget: I don't think this file is necessary since JSIC mail command have to be in plaintext emails anyway.

camallen: I'll rip it out.

PR #1008, opened 2015-06-23
Export labels of used tools when the task type is drawing

### LINKED ISSUE #979: Label tools in data export
Looking at the raw classifications, drawing tool markings are just associated with "tool 1", "tool 2" etc. I don't know how the tool number maps to the tool label in the workflow. Can we report the tool label instead of the tool number?

marten:
Fixes #979.

### DIFF
--- a/spec/lib/formatter/csv/annotation_for_csv_spec.rb
+++ b/spec/lib/formatter/csv/annotation_for_csv_spec.rb
@@ -1,0 +1,21 @@
+require 'spec_helper'
+
+RSpec.describe Formatter::Csv::AnnotationForCsv do
+  let(:classification) { create(:classification) }
+
+  let(:annotation) do
+    {
+      "task" => "interest",
+      "value" => [{"x"=>1, "y"=>2, "tool"=>1},
+                  {"x"=>3, "y"=>4, "tool"=>2},
+                  {"x"=>5, "y"=>6, "tool"=>1}]
+    }
+  end
+
+  it 'adds the tool labels for drawing tasks' do
+    formatted = described_class.new(classification, annotation).to_h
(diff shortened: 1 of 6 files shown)


camallen:
@marten let us know when this is ready for review.

spec/lib/formatter/csv/annotation_for_csv_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it 'adds the task label' do
+    formatted = described_class.new(classification, annotation).to_h
+    expect(formatted["task_label"]).to eq("Draw a circle")
+  end
+
+  it 'adds the tool labels for drawing tasks' do
+    formatted = described_class.new(classification, annotation).to_h
+    expect(formatted["value"][0]["tool_label"]).to eq("Green")
```
camallen: hey if we've got multiple related expectations can you use the new `aggregrate_failures` feature in [rspec](http://rspec.info/blog/2015/06/rspec-3-3-has-been-released/)?

marten: nice, hadn't looked at rspec 3.3 yet.

marten:
@camallen my proposal would be to go ahead with this branch initially (just adding labels inline to the json). I can continue to work on uploading zipfiles with a secondary file describing the project, but it'll be a little more involved as the dump worker logic currently always expects things to be text/csv. 

It occurs to me we could also do a third export download with the project definition, but then we also have to get the frontend involved, and I think it'd be nicer to have the classifications export be a single self-contained and self-describing download.

camallen:
@marten I think we should liase with @ggdhines and @aliburchard to see what they need before implementing something else.

Issue #1009, opened 2015-06-23
Include Subject Links in Collection


edpaget:
Forgot to do this.

PR #1010, opened 2015-06-24
Filter role controllers by user_id

edpaget:
Also serialize collection_roles link in response


edpaget:
This also closes #1009 now too.

PR #1012, opened 2015-06-24
Allow wildcard filtering of display_names fields

edpaget:
Also creates a trgm index on the fields to make them faster


camallen:
@edpaget some spec failures

PR #1014, opened 2015-06-24
Add translatable workflow_description field to project

edpaget:
Resolves zooniverse/Panoptes-Front-End#371

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150624135643) do
+ActiveRecord::Schema.define(version: 20150624155122) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -141,20 +141,21 @@
...
     t.json     "guide"
-    t.text     "faq",               default: ""
-    t.text     "result",            default: ""
-    t.text     "education_content", default: ""
-    t.jsonb    "url_labels",        default: {}
+    t.text     "faq",                  default: ""
+    t.text     "result",               default: ""
+    t.text     "education_content",    default: ""
+    t.jsonb    "url_labels",           default: {}
+    t.text     "workflow_description", default: "", null: false
```
camallen: why only this one null?

edpaget: I don't know.

camallen: No probs, we should eventually migrate these cols to default and set null: false.

PR #1016, opened 2015-06-25
Add ouroboros compatibility

edpaget:


edpaget:
Wanted to make sure all the fiddly bits like sanitized logins and identity_groups were happening in Panoptes instead of having that code in two places.

spec/models/user_spec.rb
```diff
@@ -534,6 +534,31 @@
     end
   end
 
+  describe "#update_ouroboros_created", :focus do
```
camallen: focus needs to be removed.

app/models/user.rb
```diff
@@ -51,6 +51,7 @@ class User < ActiveRecord::Base
 
   before_validation :default_display_name, on: [:create, :update]
   before_validation :setup_unsubscribe_token, on: [:create]
+  before_validation :update_ouroboros_created, on: [:create]
```
camallen: What's the plan to get this validation to run? I assume the first time they login and it updates the password? I am wondering about stargazing users and if they've already got a bcrypt password hash it may not run.

edpaget: When you login devise update some fields on the model. I assumed it still triggered normal validations then.

camallen: Yep - looks like it'll get called [here](https://github.com/plataformatec/devise/blob/181920886e4e60783f330b94072b0423345f714f/lib/devise/strategies/database_authenticatable.rb#L11)

edpaget: We don't include `:lockable` on our model.

camallen: yeah - it looks like all the devise `save` methods are `validate: false`. As long as the before validate is working i'm good.

edpaget: You're correct. The validation won't be called on login. Since they're being skipped anyway doing this as a before_save should work. I think I'm actually going to do it in both places so users created through this method can still be references in other places (like being assigned as collaborators).

edpaget: Actually I don't know about that either.

camallen:
LGTM

PR #1018, opened 2015-06-25
Make sure to still remove tempfile after spec runs

marten:
@camallen mentioned in #979 that running the specs leaves exports in the ./tmp. This is the reason that happens.


marten:
Yay random failures... :unamused:

camallen:
Happy days, we picked this up yesterday - @edpaget thinks it's todo with the number of users we create (20 odd which is too many) and the default page responses (20 i think) so the expected result isn't in the returned set. 

I've kicked off another run but agree we should get this fixed.

PR #1019, opened 2015-06-25
optimize the subscription migration

camallen:
cache as much as possible to reduce the load on the target db. Not really happy with the coupling between the instance and class methods but i didn't want to refactor to extract the state to a migration driver.


camallen:
@marten - can you review this, there are no specs attached to these migrations.

marten:
Tests are still running, but LGTM

PR #1023, opened 2015-06-25
Make sure to update the timestamp if nothing else changes

### LINKED ISSUE #1006: Touch exports models to make sure their updated_at field is changed
Make sure the updated_at field is set the time of the most recent export when a new one is requested so the Front-End accurately indicates when the last request happened.

marten:
Actually fixes #1006


edpaget:
LGTM

edpaget:
First thing after this craziness is to fix that failure.

PR #1024, opened 2015-06-25
Anyone could create roles that's not good

edpaget:


edpaget:
Thank you based @chrissnyder

camallen:
LGTM

PR #1025, opened 2015-06-25
Welcome emails with unsubscribe via email

### LINKED ISSUE #841: Welcome email for new users
Do we want to have a welcome email for new signups ready? 

If so this should be in a background worker for sending.

camallen:
closes #841 - welcome emails will be sent when a non-migrated user is created. Also added a json unsubscribe by email end point for a page like current zoo home unsubscribe form.


edpaget:
Closed via d9946e215d273440e39deea81d62b6312b7c6f49

PR #1026, opened 2015-06-25
Set User#zooniverse_id for panoptes-only users

parrish:


app/models/user.rb
```diff
@@ -53,6 +53,7 @@ class User < ActiveRecord::Base
   before_validation :setup_unsubscribe_token, on: [:create]
   before_validation :update_ouroboros_created
   before_save :update_ouroboros_created
+  after_create :set_zooniverse_id
```
camallen: I assume we don't want to add this to migrated users.

edpaget: It doesn't look like it will since it happens with the ||=

parrish: Yep.  Just panoptes users that don't have a zooniverse_id

camallen:
LGTM

PR #1032, opened 2015-06-26
Robustness against various unexpected data formats

marten:
With these changes, it gets through the entire staging dump, which is where the most strange data probably is.


camallen:
Solid - LGTM

PR #1035, opened 2015-06-28
Add RSpec binstub with Spring enabled

marten:
Not sure why bin/rspec wasn't committed in the repo, since the Guardfile does reference it. What this out of developer preferences or just because `bin/**` is ignored in .gitignore?


camallen:
The gaurdfile references bin/rspec to run via spring (it could use spring rspec). LGTM

PR #1046, opened 2015-06-30
Use Docker compose 1.3 alternate dockerfile

marten:
The only thing I'm not certain of is how Jenkins builds the production container. Does it use `rake configure:prod_docker` (which this branch removes) anywhere or was that just for local use?


camallen:
Excellent stuff - been waiting for a while to use the alternate docker file feature. All prod builds will use the default expected `Dockerfile` in the root.

marten:
Yeah, I just noticed it was released 18 june.

marten:
@camallen I assume that's a LGTM?

camallen:
Happy days!

PR #1049, opened 2015-06-30
Adds a unique index to User#zooniverse_id

parrish:
Fair warning, this is a unique index and there are currently non-unique values in the db.

Currently there are around 59 records that need fixing

```sql
select zooniverse_id, count(*)
from users
group by zooniverse_id
having count(*) > 1
order by count(*) asc
```


parrish:
Actually, this isn't ideal as we'll want to reference the id on the insert to create the zooniverse_id

PR #1055, opened 2015-07-01
Add rack-timeout using default settings

edpaget:
Adds rack timeout with a default of 15s to process a request. This way
we will at least timeout on the SSL SYSCALL Errors and be able to
process other requests


camallen:
LGTM

camallen:
I don't think we need this now.

edpaget:
I still think it'd be useful to add, since we're still in a situation where a blocked thread can block all the others unless it made to to a point where it could yield execution. I'll reopen when it's actually ready though.

PR #1061, opened 2015-07-02
allow delete and show on singluar media relations

### LINKED ISSUE #1060: User avatars and profile headers can't be deleted
GET /users/3/avatar returns with `{"href": "/users/3/avatar"}`.

DELETE /users/3/avatar results in a 422.

Same deal with `profile_header`. Projects' `attached_images` are fine though, so it's not a media-wide thing.

camallen:
Fixes #1060, this probably needs a refactor, we could look at adding single and collection methods for the routes in this controller and switch them in the routes setup based on relation types.

### DIFF
--- a/app/controllers/api/v1/media_controller.rb
+++ b/app/controllers/api/v1/media_controller.rb
@@ -2,6 +2,8 @@ class Api::V1::MediaController < Api::ApiController
   doorkeeper_for :update, :create, :destroy, scopes: [:medium]
   resource_actions :default
 
+  before_action :create_conditions, only: [ :create, :create_collection ]
+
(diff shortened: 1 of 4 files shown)


edpaget:
Is there anyway we just do `DELETE /users/32/avatar` and not add the show routes either? The avatar is already uniquely identified at `DELETE /users/32/avatar`, and the serializer is returning that as the sole URL for the avatar.

camallen:
Yeah, i'll refactor into collection and resource routes to clean up the controller actions.

camallen:
@edpaget - still working on this, needs a final refactor.

PR #1065, opened 2015-07-02
Allow unicode chars in the login field

camallen:
Putting this out there since it cropped up as an error on login. Adding basic support for unicode characters in login, however this may break @ - mentions in talk. 

I'm voting yes especially to target non-ascii char based languages over the world. @mrniaboc thoughts?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,7 +2,7 @@ class User < ActiveRecord::Base
   include Activatable
   include Linkable
 
-  ALLOWED_LOGIN_CHARACTERS = '[\w\-\.]'
+  ALLOWED_LOGIN_CHARACTERS = '[[:word:]\-\.]'
   USER_LOGIN_REGEX = /\A#{ ALLOWED_LOGIN_CHARACTERS }+\z/
 
   devise :database_authenticatable, :registerable,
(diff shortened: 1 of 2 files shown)


edpaget:
I 100% disagree with this since this will make any @-ing system totally unworkable. The ASCII character set is the lowest common denominator in terms of computer input. Anyone can type it on pretty much any device. 

As soon as we allow arbitrary unicode characters in login names, then we segregate who who can @ to people whose name characters are in a set you can easily type. 

We have display names so people can already choose to be represented in their profile and in the larger name that appears next to their talk posts in whatever form they want to be.

DarrenMcRoy:
I'm really with Ed on this one. We should do whatever Twitter already does here. They seem to have no issues with being global. :+1:

edpaget:
I addition, this would allow project URL spoofing to happen. If we allows any unicode character in URLs then users wouldn't know whether zooniverse.org/#/projects/zooniverse/Galaxy-Zoo takes them to our Galaxy Zoo page or another user's page who's using Cyrillic characters to look like zooniverse.

astopy:
I’m with Ed. This will have all kinds of unpredictable security implications.

camallen:
Ah - The latter point about mimicking other accounts is alarming, that will open up a can of worms. I going to close this and we'll deal with existing users on a case by case basis.

PR #1071, opened 2015-07-06
Ensure existing user activity events link to the correct projects

camallen:
Once this is active I'll fix the migrated data to link back to the correct projects.


marten:
LGTM

PR #1074, opened 2015-07-06
allow panotpes zoo id formats

camallen:
events controller is rejecting new user activities on old projects

### DIFF
--- a/app/schemas/event_create_schema.rb
+++ b/app/schemas/event_create_schema.rb
@@ -16,7 +16,7 @@ class EventCreateSchema < JsonSchema
 
     property "zooniverse_user_id" do
       type "string", "integer"
-      pattern "^[0-9]*$"
+      pattern "^.+[0-9]*$"
     end
 
     property "count" do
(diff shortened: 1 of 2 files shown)


app/schemas/event_create_schema.rb
```diff
@@ -16,7 +16,7 @@ class EventCreateSchema < JsonSchema
 
     property "zooniverse_user_id" do
       type "string", "integer"
-      pattern "^[0-9]*$"
+      pattern "^.+[0-9]*$"
```
edpaget: Do you want to make this pattern more explicity? as in "^(panoptes-|)[0-9]+$"

camallen: i could i just don't want to have to redo. It'll currently fail on finding the user and not just reject the message outright before even attempting to process.

edpaget: Sure. Not like anyone but us will be using this endpoint.

edpaget:
:shipit:

PR #1075, opened 2015-07-06
Better user search using pg_search gem

### LINKED ISSUE #1034: Need to better weight user searching and match inside name
From zooniverse/Panoptes-Front-End#859, I think this is a three part thing:

+ [ ] Need to search using '%fragment%' to match inside a displayname (
   + this was the whole reason for doing the trigram index don't know why I didn't do it.
   + Why meg can't find Darren. 
+ [ ] Exact matches need to be matched first 
   + When I type in 'zookeeper' the user 'zookeeper' should be matched first
+ [ ] It should maybe search both display_names and logins

edpaget:
Sets up user search ranked by relevance along both the display_name and
login fields.

Closes #1034


edpaget:
I still need to test this a bit locally to make sure it isn't super slow for real-world usage.

edpaget:
Took me a while to figure out the correct way to index this, but it is __fast__ now.

camallen:
Looks good.

PR #1081, opened 2015-07-07
Project pref serialized responses

### LINKED ISSUE #1080: add activity count to user project preference serialized response
For use in a replacement of the zoo home ribbon.

camallen:
closes #1080 use the counts on the UPP if they are set, otherwise sum from the user seens in the DB.


camallen:
updating this - just realised something.

PR #1082, opened 2015-07-07
Project pages

edpaget:
This separates project pages into a new resource speeding up load times for project listings between 3-5x. 

It will also allow us to support custom page types. 

I'm still working on the corresponding Front-End pull request.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -17,6 +17,8 @@ ActiveRecord::Schema.define(version: 20150706185722) do
   enable_extension "plpgsql"
   enable_extension "intarray"
   enable_extension "pg_trgm"
+  enable_extension "hstore"
+  enable_extension "postgres_fdw"
 
   create_table "access_control_lists", force: :cascade do |t|
     t.integer  "user_group_id", index: {name: "index_access_control_lists_on_user_group_id"}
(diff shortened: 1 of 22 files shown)


db/schema.rb
```diff
@@ -17,6 +17,8 @@
   enable_extension "plpgsql"
   enable_extension "intarray"
   enable_extension "pg_trgm"
+  enable_extension "hstore"
```
camallen: Seems these extensions have crept in.

spec/models/project_page_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+RSpec.describe ProjectPage, type: :model do
```
camallen: doy you want to remove this or leave it pending?

camallen:
LGTM - nice work!

db/schema.rb
edpaget: Will fixed. I wish I could pin down what caused this.

spec/models/project_page_spec.rb
edpaget: I should probably add some specs.

PR #1084, opened 2015-07-08
use user table functional indexes

camallen:
Update our code to ensure we use the functional indexes on the users table, happy to refactor to class methods on user if needed.


lib/zooniverse_user.rb
```diff
@@ -40,8 +40,8 @@ def password=(plain_password)
   end
 
   def import
-    #use the indexed field
-    user = User.find_or_initialize_by login: User.sanitize_login(login)
+    user = User.find_by("lower(login) = '#{User.sanitize_login(login).downcase}'")
```
edpaget: Let's use zooniverse_id here.

ouroboros/galaxy_zoo.rb
```diff
@@ -9,7 +9,7 @@
 mongo = Moped::Session.new ['127.0.0.1:27017']
```
edpaget: Can we just remove this file/dir?

PR #1088, opened 2015-07-09
Ensure retire workers fire for incoming complete classifications

### LINKED ISSUE #1087: 60 logged in volunteers since July 5 saw the same subject, but it is supposed to be retired after 20 classifications
We sent out a newsletter to the Planet Four volunteer list about Terrains and I noticed that pretty much Talk was empty. There were very few new posts.  I originally though this was a problem with Talk posting again. 

Then I looked at the subjects, just yesterday alone 61 people all except 1 who wasn't logged in, saw the same subject and are posting on it. It's supposed to retire at 20 and all the correct live settings as far as I can tell are active and @camallen  uploaded our dataset so I don't think it's a data upload issue. We did have the subject set being used while new subjects were being added but I don't presume that would cause this issue. 

I don't think I should publish the csv file here - check out subject 483978 and 483873

I'm also confused even if these subjects are not retiring and why if I have 8000 subjects in the subject set that  these show so frequently even if it wasn't retiring? Is this a bug on the backend?

camallen:
Should fix #1087 - we we're updating the seen subjects then testing if a user had seen the subject to fire the count workers. Seems we only ever managed to count towards retirement was when the ClassificationWorker fired off before the UserSeenSubject update seens transaction completed.

This PR reworks the order of events in the lifecycle transact method to make sure the count worker fires before we update the user seens. I don't think this will impact on duplicate subjects for users see [this commit](https://github.com/zooniverse/Panoptes/commit/2ec96166d2e619bf0cdc4a447edc1770a435a026) but it's something to keep in mind when reviewing.

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -13,16 +13,19 @@ class ClassificationLifecycle
       message = "Background process called before persisting the classification."
       raise ClassificationNotPersisted.new(message)
     end
-
     dequeue_subject
-    update_seen_subjects
     ClassificationWorker.perform_async(classification.id, action.to_s)
   end
 
   def transact!(&block)
     Classification.transaction do
-      mark_expert_classifier if subjects_are_unseen_by_user?
+      #NOTE: ensure the block is evaluated before updating the seen subjects
+      # as the count worker won't fire if the seens are set, see #should_count_towards_retirement 
       instance_eval(&block) if block_given?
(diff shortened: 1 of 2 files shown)


marten:
Seems fine to me. But I do wonder why we're splitting this up over multiple background jobs in the first place. Wouldn't one or two (lifecycle and kafka) be sufficient?

PR #1089, opened 2015-07-09
update subscribe worker params

camallen:
allow subscribe worker to accept email and name params but only forward the email onto the mailer.


edpaget:
We should make people login instead of just using their display_name. 

With this is someone wanted to be jerk they should just unsubscribe all of Zooniverse by putting in the publically available display_names.

PR #1107, opened 2015-07-14
Allow admin any mime type

### LINKED ISSUE #1078: Allow arbitrary mime types for subject locations for admins
For Mammoth Bats, I'll be creating subjects with a mime type of `video/mp4` and it's returning a validation error of `Error: Validation failed: Locations is invalid`.

For typical projects/media uploads this makes sense, but for admins when we upload subjects, could this be literally anything?

camallen:
closes #1078 - this must be used via the admin mode flag.


edpaget:
I think this is fine for now. In the near future maybe we should just remove the content-type restrictions entirely, and only enforce it through the Front-End.

PR #1108, opened 2015-07-14
Retire via API call

marten:
* Not sure if there's any way to **not** automatically retire. The project builder seems to insist on a numeric threshold. If there's no way, we need to add that too probably.

* We need a good way to authenticate this API without using an OAuth token belonging to the actual project owner.

* API documentation, how do I add that?

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -5,7 +5,7 @@ class Api::V1::WorkflowsController < Api::ApiController
   doorkeeper_for :update, :create, :destroy, :retire_subject, scopes: [:project]
   before_action  :reject_live_project_changes, only: [ :create ]
 
-  resource_actions :default
+  resource_actions :default, :retire_subject
   schema_type :json_schema
 
   def show
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -2,7 +2,7 @@ class Api::V1::WorkflowsController < Api::ApiController
   include Versioned
   include TranslatableResource
 
-  doorkeeper_for :update, :create, :destroy, scopes: [:project]
+  doorkeeper_for :update, :create, :destroy, :retire_subject, scopes: [:project]
   before_action  :reject_live_project_changes, only: [ :create ]
 
   resource_actions :default
@@ -27,6 +27,11 @@ class Api::V1::WorkflowsController < Api::ApiController
 
   def destroy_links
     super { |workflow| refresh_queue(workflow) }
+  end
+
+  def retire_subject
+    SubjectRetirementWorker.perform_async(params[:subject_id], params[:id])
+    head :ok
   end
 
   private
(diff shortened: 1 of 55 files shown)


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -29,6 +29,11 @@ def destroy_links
     super { |workflow| refresh_queue(workflow) }
   end
 
+  def retire_subject
+    SubjectRetirementWorker.perform_async(params[:subject_id], params[:id])
+    head :ok
```
marten: Not sure if there's something else we'd want to respond with here.

config/routes.rb
```diff
@@ -77,7 +77,11 @@
         json_api_resources :pages, controller: "project_pages"
       end
 
-      json_api_resources :workflows, links: [:subject_sets], versioned: true
+      json_api_resources :workflows, links: [:subject_sets], versioned: true do
+        member do
+          post :retire_subject
```
marten: I'm open to suggestions for a better location, it feels slightly wacky where it is now but I can't think of anything better.

app/controllers/api/v1/workflows_controller.rb
edpaget: How about :no_content?

config/routes.rb
edpaget: Why not make this is a JSON API relationship called `retired_subjects` and retire them by posting to `/workflows/1/links/retired_subjects` would also all for easy 'unretiring' by doing deleting `/workflows/1/links/retired_subjects/1,11,2`

camallen:
> Not sure if there's any way to not automatically retire. The project builder seems to insist on a numeric threshold. If there's no way, we need to add that too probably.

We could add a flag to workflow to use a retirement service instead of counts.

>We need a good way to authenticate this API without using an OAuth token belonging to the actual project owner.

Agreed

> API documentation, how do I add that?

test it here, http://docs.panoptes.apiary.io/# and edit the `apiary.apib`

config/routes.rb
marten: I like that. The inverse might also be an idea: `/subjects/1/links/retired_workflows`, since that's how it is stored in the db?

camallen: Either works but I think this fits better on the workflow, as to me subjects can be shared across projects so I think primarily on the specific worklow (and hence project details) that we're retiring for.

marten: I agree with that, that's also the way I'd reason about it. I assume the database approaches this differently because a workflow has many subjects and the queries might get too slow? Is that also the reason behind the current array-of-ids vs a join table for `SubjectWorkflowRetirements`?

camallen: Not sure i'm following 100%, I think your reasoning is questioning the situation where we have a `workflow_id` and a `subject_id` but we don't know the implementation details of the `set_member_subject_id` so we can't just append to the SMS `retired_workflow_ids` array.

This is due to the fact that the client applications deal in subject ids (raw media locations, metadata, etc) and the internal SubjectSet, SetMemberSubject models are really just how we implmenet the links between projects & subjects. 

So your retire end points would need to determine the internal link models to retire and update the SMS `retired_workflow_ids` array. Make sense? PM me if you want to chat more about this.

marten:
Closed in favour of #1277. Rebasing was harder than just looking at the diff and starting fresh.

PR #1109, opened 2015-07-14
test collection update links and provide better error message

camallen:
Adding better messaging and explicity update_links testing. I spec'd this out when testing a HBadger error.


edpaget:
:+1:

PR #1110, opened 2015-07-14
Update README.md

chrissnyder:


edpaget:
Mind adding 'or lower'?

edpaget:
Actually nevermind that's really all its tested with.

PR #1112, opened 2015-07-14
handle dup inserts that violate the db index constraints

camallen:
gracefully handle an index violation that passes model validation.


edpaget:
:+1:

PR #1113, opened 2015-07-14
Upload classification and subject dumps as gzip

### LINKED ISSUE #1077: Compressing the csv data dumps for subjects and classifications
Future request - I'm noticing it's already starting to take minutes for the classification database for P4T to download. Also one of my collaborators mentioned that he uses safari and it automatically tries to load the csv when he clicks the link from his email since it's uncompressed  - eventually it might be better if these were compressed

edpaget:
Closes #1077


edpaget:
Actually I'm going to do this slightly differently.

PR #1114, opened 2015-07-14
Upload classification and subject dumps as gzip

### LINKED ISSUE #1077: Compressing the csv data dumps for subjects and classifications

edpaget:
Closes #1077

### DIFF
--- a/app/workers/concerns/dump_worker.rb
+++ b/app/workers/concerns/dump_worker.rb
@@ -14,7 +14,15 @@ module DumpWorker
   end
 
   def temp_file_path
+    "#{Rails.root}/tmp/#{project_file_path.join("_")}"
+  end
+
+  def csv_file_path
     "#{Rails.root}/tmp/#{project_file_path.join("_")}.csv"
+  end
+
+  def gzip_file_path
+    "#{Rails.root}/tmp/#{project_file_path.join("_")}.gz"
   end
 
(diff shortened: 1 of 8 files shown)


edpaget:
This won't solve the automatically opening the csv in Safari issue. Maybe I should include a note to 'right-click select Save As' on the link in the email?

app/workers/concerns/dump_worker.rb
```diff
@@ -14,9 +14,21 @@ def mailer
...
+
+  def csv_file_path
     "#{Rails.root}/tmp/#{project_file_path.join("_")}.csv"
   end
 
+  def gzip_file_path
+    "#{Rails.root}/tmp/#{project_file_path.join("_")}.gz"
+  end
+
+  def s3_object_path
```
camallen: is this used?

camallen:
Looks good. just wondering about the empty method.

app/workers/concerns/dump_worker.rb
edpaget: No

PR #1131, opened 2015-07-16
Migrate retirement to subject workflow counts

marten:
After merging maybe someone can show me how to deploy and run rake tasks. Data migration should run in under a minute. After that I'll send a followup PR to clean up leftover code.


camallen:
I like the idea of this, as it'll allow us to find out when a subject was retired. Can you fix the failing specs?

marten:
@camallen I forgot that `joins` does an `INNER JOIN` which filtered out the unseen subjects (which wouldn't have a count yet).

edpaget:
Looks good to me. :shipit:

PR #1132, opened 2015-07-16
convert page size string values to ints

camallen:
allow page size selection params to cascade into the subject selector.

### DIFF
--- a/lib/subject_selector.rb
+++ b/lib/subject_selector.rb
@@ -58,7 +58,10 @@ class SubjectSelector
   end
 
   def default_page_size
-    params[:page_size] ||= 10
+    page_size = params[:page_size]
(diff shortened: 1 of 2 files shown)


lib/subject_selector.rb
```diff
@@ -58,7 +58,10 @@ def missing_subject_set_error
   end
 
   def default_page_size
```
marten: I think "default_page_size" is a bit of a misnomer given that it can be overridden using params.

marten:
Minor naming niggle, but LGTM

camallen:
this isn't right...fix coming

marten:
bnfge-iccaaqpyi jpg-large

camallen:
just in case i didn't get the message the first time.

PR #1133, opened 2015-07-16
Always return randomly selected subjects from the queue

### LINKED ISSUE #1033: nonrandom subject queing
On an ouroboros project, if I refresh the page without classifying, I am served a new image every time.

On Panoptes (specifically http://preview.zooniverse.org/panoptes/#/projects/aliburchard/wildebeest-watch), if I refresh the page without classifying, I am presented with the same image. Even if I close the tab and open a new tab and go to the wildebeest watch landing page, I am ultimately presented with the same image. This happens until I submit a complete classification for the subject in question.

marten:
PostgreSQL will always sort the integers when performing some array operations, including union and difference. That meant that `set_member_subject_ids` would become sorted upon both enqueue and dequeue of any subject (but the initial creation of the queue would be with a shuffled array).

By simply always returning a random selection from the list of enqueued ids, we ensure that we always return a random selection. This also allows for more complicated ordering on a per-workflow basis later on by doing something other than `array.sample()`.

It also means you can use F5 to get a different subject (which @aliburchard wanted I believe).

Fixes #1033

### DIFF
--- a/app/models/subject_queue.rb
+++ b/app/models/subject_queue.rb
@@ -101,10 +101,6 @@ class SubjectQueue < ActiveRecord::Base
   end
 
   def next_subjects(limit=10)
-    if user_id
-      set_member_subject_ids[0..limit-1]
-    else
-      set_member_subject_ids.sample(limit)
-    end
+    set_member_subject_ids.sample(limit)
   end
 end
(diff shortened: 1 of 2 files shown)


edpaget:
:+1:

PR #1151, opened 2015-07-20
Remove legacy style of retiring subjects

marten:


app/models/set_member_subject.rb
```diff
@@ -48,4 +58,11 @@ def remove_from_queues
   def set_random
     self.random = rand
   end
+
+  def timestamp_newly_retired_workflows
+    retired_subject_workflow_counts.each do |record|
+      record.retired_at ||= Time.now
+      record.save if record.persisted? and record.changed?
```
camallen: Loosk like this would be better as a method on `retired_subject_workflow_count`, e.g. `record.update_retired!`.

app/models/set_member_subject.rb
```diff
@@ -48,4 +54,10 @@ def remove_from_queues
   def set_random
     self.random = rand
   end
+
+  private
+
+  def timestamp_newly_retired_workflows
+    retired_subject_workflow_counts.each(&:retire!)
```
marten: Records just added to the collection (`retired_subject_workflow_counts << record` etc) will be there too.

app/models/set_member_subject.rb
```diff
@@ -15,6 +16,7 @@ class SetMemberSubject < ActiveRecord::Base
   can_through_parent :subject_set, :update, :show, :destroy, :index, :update_links,
     :destroy_links
 
+  before_save :timestamp_newly_retired_workflows
```
camallen: So you'd be like `retired_subject_workflow_counts << non_retired_record` and then leveraging the save functionatliy of the `sms` to retire it?

Just wondering if we should be setting this retired value on the record before adding to the association Or having an explicit method to do it instead of co-opting the save function.

marten: Yes. Personally, I'm in favour of an explicit method. I think I let myself damage the design in favour of not changing too much specs (laziness, basically). That's because it bites with factory_girl (can't do `create(:set_member_subject, retired_workflows: [workflow])` anymore), as well as the JSONAPI controller magic.

Do you know how to make [the SMS controller](https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/set_member_subjects_controller.rb#L6-L7) use that method instead of the association to manage the links?

camallen: Yeah, you can pass a block to `update_links` in the controller [here](https://github.com/zooniverse/Panoptes/blob/master/lib/json_api_controller/updatable_resource.rb#L20) to perform custom work on the resource before the save, ala [this](https://github.com/zooniverse/Panoptes/blob/31316cab492cf2d5775f13b7e43bc6f790b38d54/app/controllers/api/v1/subject_sets_controller.rb#L13). Is this what you are looking for?

app/controllers/api/v1/set_member_subjects_controller.rb
```diff
@@ -5,4 +5,16 @@ class Api::V1::SetMemberSubjectsController < Api::ApiController
 
   allowed_params :create, :priority, links: [:subject, :subject_set, retired_workflows: []]
   allowed_params :update, :priority, links: [retired_workflows: []]
+
+  def create
+    super { |set_member_subject| set_member_subject.retired_subject_workflow_counts.each(&:retire!) }
```
camallen: Do you want to extract the common method call to a specific method on the sms? 
``` ruby 
def retire_all_subject_workflow_counts
  retired_subject_workflow_counts.each(&:retire!)
end

#then the block here could be
super { |set_member_subject| set_member_subject.retire_all_subject_workflow_counts }
```

camallen:
LGTM - @marten you happy this is good to go?

PR #1174, opened 2015-07-22
User ouroboros created dup login clashes

camallen:
attempt to avoid clashes on login sanitation as some users are signing up via ouroboros with possibly clashing login names after login sanitation. this adds an extra loop to try and bump the login name using a counter to avoid the clashes.

### DIFF
--- a/lib/tasks/user_limit.rake
+++ b/lib/tasks/user_limit.rake
@@ -15,11 +15,15 @@ namespace :user_limit do
     end
   end
 
+  def find_user_by_login(login)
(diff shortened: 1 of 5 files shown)


lib/tasks/user_limit.rake
```diff
@@ -15,15 +15,11 @@ namespace :user_limit do
     end
   end
 
-  def find_user_by_login(login)
-    User.find_by!("lower(login) = '#{login.downcase}'")
```
camallen: This rake task should raise the error..i'll fix this.

PR #1181, opened 2015-07-23
avoid optimistic lock errors when deactivating workflows

camallen:
HBadger is reporting these errors from our retirement worker. It's most likely a counter being updated behind the scenes so just avoiding the error as we just want to deactivate the workflow when it’s finished.

Maybe not the best way to avoid these errors so happy to change if there are better ideas.

### DIFF
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -15,8 +15,15 @@ class RetirementWorker
 
   def deactivate_workflow!(workflow)
     if workflow.finished?
-      workflow.active = false
-      workflow.save!
+      ignore_optimistic_lock { workflow.update_attribute(:active, false) }
     end
+  end
+
+  private
+
+  def ignore_optimistic_lock
+    ActiveRecord::Base.lock_optimistically = false
+    yield
+    ActiveRecord::Base.lock_optimistically = true
   end
 end
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -15,15 +15,7 @@ class RetirementWorker
 
   def deactivate_workflow!(workflow)
     if workflow.finished?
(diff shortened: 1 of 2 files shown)


app/workers/retirement_worker.rb
```diff
@@ -15,8 +15,15 @@ def perform(count_id)
...
-      workflow.active = false
-      workflow.save!
+      ignore_optimistic_lock { workflow.update_attribute(:active, false) }
     end
   end
+
+  private
+
+  def ignore_optimistic_lock
+    ActiveRecord::Base.lock_optimistically = false
```
edpaget: So I'm not 100% this will be thread safe. Since sidekiq runs in one process and executes a bunch of threads won't this disable optimistic locking for anything running in a concurrent sidekiq thread?

camallen: Yeah i don't know..I could test in staging. Any other thoughts to handle these? (see prod sidekiq retires).

edpaget: Yeah I think if you use the class method `::update` it skips looking so you could do 

```ruby
Workflow.update(workflow.id, {activate: false})
```
but I can't find any docs to back up my memory.

parrish: I just did a quick proof of concept on this:  https://gist.github.com/parrish/41efa5613077f9dcf818

Basically if you want thread isolated variables, you'll need to set them on the thread.

In the case of an instance variable on ActiveRecord::Base, you might have luck redefining the attribute as a thread variable.  Seems pretty hacky though

edpaget: Alternatively `::update_all` definitely skips looking so 

```ruby
Workflow.where(id: workflow.id).update_all(active: false)
```

camallen:
Thanks @edpaget and @parrish - nice catch on thread saftety, should have thought about that.

PR #1182, opened 2015-07-23
dequeue set_member_subject ids after selection

### LINKED ISSUE #1176: Dequeue Subjects after requesting for selection
Try and avoid race conditions between a classification being submitted to a slow API node and the next request for subjects hitting a free one and returning the queue before any dequeue event fires. 

So move the dequeue from [here](https://github.com/zooniverse/Panoptes/blob/546f1d2acc8d965132bc9badd644bfdaddf215c2/lib/classification_lifecycle.rb#L16) to [here perhaps](https://github.com/zooniverse/Panoptes/blob/546f1d2acc8d965132bc9badd644bfdaddf215c2/lib/subject_selector.rb#L24).

camallen:
closes #1176 - get set member subject ids out of the queue asap and let them replenish if they are needed when the queue's replenish.

This should hopefully avoid any race condition between classification processing and next subjects calls by increasing the time window between.

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -13,14 +13,13 @@ class ClassificationLifecycle
       message = "Background process called before persisting the classification."
       raise ClassificationNotPersisted.new(message)
     end
-    dequeue_subject
     ClassificationWorker.perform_async(classification.id, action.to_s)
   end
 
   def transact!(&block)
     Classification.transaction do
       #NOTE: ensure the block is evaluated before updating the seen subjects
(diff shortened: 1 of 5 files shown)


camallen:
So I can't reproduce the failing specs in isolation, time to check out rspec bisect, https://relishapp.com/rspec/rspec-core/v/3-3/docs/command-line/bisect

parrish:
@camallen Holy crap bisect is useful.  Perfect timing.  Thanks!

camallen:
@parrish - wow, that bisect feature is awesome!

PR #1191, opened 2015-07-27
Return to mainline Rails release

edpaget:
My Rails turned out to be unnecessary since the problem was due to
changes in docker networking code


camallen:
LGTM

PR #1197, opened 2015-07-27
Handle CRSF errors with a null session

### LINKED ISSUE #1117: 404 Errors in the API need to be handled by the API Controller
...or be handled in such a way as they can skip the ApplicationControllers' CSRF protection. Otherwise the proper error won't be reported.

edpaget:
We only use session to get OAuth tokens, which will fail 422 with a null
session anyway, so instead of throwing and error on non-api routes and
skipping CSRF protection on api routes, we'll just use the null_session
handling instead.

Closes #1117


camallen:
LGTM - out of interest how is the GET on a missing api route testing the CSRF protection via null session instead of exception?

edpaget:
Ah yes that should be a POST.

PR #1198, opened 2015-07-27
Allow subjects to be linked to external objects

### LINKED ISSUE #1038: Allow externally linked media sources for subjects
Allow externally hosted images via the subjects create / update end point.

edpaget:
Closes #1038

### DIFF
--- a/spec/controllers/api/v1/subjects_controller_spec.rb
+++ b/spec/controllers/api/v1/subjects_controller_spec.rb
@@ -326,6 +326,20 @@ describe Api::V1::SubjectsController, type: :controller do
     end
 
     it_behaves_like "is updatable"
+
+
+    it 'should create externally linked media resources' do
+      default_request user_id: authorized_user.id, scopes: scopes
+      external_locs = [{"image/jpeg" => "http://example.com/1.jpg"}, {"image/jpeg" => "http://example.com/2.jpg"}]
+      external_src_urls = external_locs.map { |loc| loc.to_a.flatten[1] }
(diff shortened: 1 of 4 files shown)


spec/controllers/api/v1/subjects_controller_spec.rb
```diff
@@ -309,23 +309,33 @@
...
     it_behaves_like "is updatable"
+
+
+    it 'should create externally linked media resources' do
+      default_request user_id: authorized_user.id, scopes: scopes
+      update_params[:subjects].merge!(locations: [{"image/jpeg" => "http://example.com/1.jpg"},
+        {"image/jpeg" => "http://example.com/2.jpg"}])
+      put :update, update_params.merge(id: resource.id)
+      locations = Subject.find(created_instance_id("subjects")).locations
+      expect(locations.map(&:external_link)).to all( be true )
```
camallen: for coverage i think we should we test the explicit addition of the supplied URL to the `locations.src` value?

``` ruby
external_locs = [{"image/jpeg" => "http://example.com/1.jpg"}, {"image/jpeg" => "http://example.com/2.jpg"}]
external_src_urls = external_locs.map { |loc| loc.to_a.flatten[1] }
#...
aggregate_failures "external srcs" do
  expect(locations.map(&:external_link)).to all( be true )
  expect(locations.map(&:src)).to match_array(external_src_urls)
end
```

camallen:
Nice work on this, it'll be super useful :+1:

PR #1206, opened 2015-07-28
Add rake task to create user project prefences

### LINKED ISSUE #1122: Setup project preferences for all users that classified on snaphot supernova
add these in after #1103 is done.

edpaget:
Also make sure each user only has one upp object per project

Closes #1122

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -78,5 +78,25 @@ namespace :migrate do
       end
       puts "Updated #{ missing_token_count } users have unsubscribe tokens."
     end
+
+    desc "Create project preferences for projects classified on"
+    task create_project_preferences: :environment do
+      project = Project.find(ENV["PROJECT_ID"])
+
+      if user = User.find_by(id: ENV["USER_ID"])
+        p "Updating: #{user.login}"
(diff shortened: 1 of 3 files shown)


lib/tasks/migrate.rake
```diff
@@ -78,5 +78,25 @@ namespace :migrate do
...
       puts "Updated #{ missing_token_count } users have unsubscribe tokens."
     end
+
+    desc "Create project preferences for projects classified on"
+    task create_project_preferences: :environment do
+      project = Project.find(ENV["PROJECT_ID"])
+
+      if user = User.find_by(id: ENV["USER_ID"])
+        p "Updating: #{user.login}"
+        UserProjectPreference.create!(user: user, project: project)
```
camallen: this will throw an error if the user / project preference already exists. do you want to handle it?

edpaget: Nope it's a Rake task, so the person running it will be able to inspect the environment if that output is unexpected.

camallen:
Code looks fine, is this for stargazing 2015 UPP's?

edpaget:
Yup. Once this is merged/deployed I can kick off a Rake task to create them.

Issue #1139, opened 2015-07-16
Change Project slug to include owner name?


edpaget:
Any thoughts on making the project slug include the owner-name by default when the response is serialized? eg

```json
{
  "projects": {
    ...,
    "slug": "edwardothegreat/edwardos-sad-project"
  }
}
```

And similarly changing the slug filter param to reflect this `GET /api/projects?slug=edwardothegreat%2Fedwardos-sad-project`.

I think this would definitely benefit the front-end implementation in terms of saving extra requests/includes, but it would also make the `slug` make more sense since it will fully uniquely identify a project by itself.

camallen:
Sounds useful. I can't think of any drawbacks at the moment. Are there any you can outline?

aweiksnar:
:+1: I'm in favor of this

edpaget:
There's the database hit, but that could be covered up by eagerloading, and we incur it anyway since almost all the time people are doing `GET /api/projects?include=owner`.

parrish:
> I think this would definitely benefit the front-end implementation in terms of saving extra requests/includes

`api.type('projects').get({ slug: 'edwardos-sad-project', owner: 'edwardothegreat', include: ['background', 'pages', 'workflows'] })` seems to work just fine, so I'm not sure why that would reduce the number of calls.  (Apparently `owner` isn't a valid include)

>  it would also make the slug make more sense since it will fully uniquely identify a project by itself

Generally agreed

parrish:
Yep, just double checked that: `{"errors":[{"message":":owner is not a valid include for Project"}]}`

edpaget:
Sorry I mistyped it 'owners'.

parrish:
Yep, just caught that myself

edpaget:
It would reduce the number of calls if you want to build a link to the Project's page, since to do the `api.type('projects').get({ slug: 'edwardos-sad-project', owner: 'edwardothegreat', include: ['background', 'pages', 'workflows'] }` you have to know the owner name which is an additional request at the moment.

parrish:
Not in general,  it's a URL param.  #/projects/edwardothegreat/edwardos-sad-project

edpaget:
Only one case though. I ran into this building links from Collection subjects to their Talk pages. Also the /projects route.

chrissnyder:
+1, if for no other reason than to simplify link generation on the FE. Tons of cases in Talk where this would make life a bit easier.

parrish:
[...]

PR #1208, opened 2015-07-28
Keep logins and display_names in sync

edpaget:
...between users and their identity groups


edpaget:
This is moving us towards #1139.

Really these fields probably belong in a separate `identity` table that is shared by users and groups.

camallen:
Do you want to add an issue in future plans to extract the fields to a seperate table?

PR #1210, opened 2015-07-28
Pluralize media links if they need to be

### LINKED ISSUE #1209: Workflow.links.attached_images has "id": [...] instead of "ids": [...]
Singular instead of plural. Otherwise it's gold, thanks @edpaget!

edpaget:
Closes #1209

### DIFF
--- a/app/serializers/concerns/media_links_serializer.rb
+++ b/app/serializers/concerns/media_links_serializer.rb
@@ -61,7 +61,12 @@ module MediaLinksSerializer
                             href: media_href(model, link),
                             type: link.to_s.pluralize
                            }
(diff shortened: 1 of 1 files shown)


app/serializers/concerns/media_links_serializer.rb
```diff
@@ -57,11 +57,17 @@ def add_links(model, data)
     data = super
     self.class.media_links.each do |link|
       id = data[:links].delete(link)
+      p id
```
camallen: debug statement.

camallen:
LGTM after removing the debug statement.

PR #1214, opened 2015-07-29
Create recents

camallen:
with a migration rake task to catch up on the missing ones.


edpaget:
Have you tried running the rake task against any test data?

camallen:
yeah - just ran it on my dev db which is a copy of staging. 

Checked the missing recents query count matched `Classification.count - Recent.count` (we only do 1 subject classifcations currenlty) and then checked some of the created Recent data looks good. Also looks like the `Recent.create_from_classification(classification)` is good and the recents were getting created. Any concerns?

edpaget:
Nope just wanted to check that it had happened otherwise I would have run against my staging backup.

edpaget:
:+1:

PR #1217, opened 2015-07-29
use the json accept header for talk api

camallen:
get the json rep of the resources, avoid the html redirect on the main repo

### DIFF
--- a/lib/talk_api_client.rb
+++ b/lib/talk_api_client.rb
@@ -78,14 +78,14 @@ class TalkApiClient
   end
 
   def initial_request
-    request('get', '').body.each do |_, resource|
+    request('get', '', accept_header: "application/json").body.each do |_, resource|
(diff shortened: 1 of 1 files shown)


lib/talk_api_client.rb
```diff
@@ -78,14 +78,14 @@ def create_connection(adapter)
   end
 
   def initial_request
-    request('get', '').body.each do |_, resource|
+    request('get', '', accept_header: "application/json").body.each do |_, resource|
```
edpaget: Why is this one different?

parrish: This won't need to be different after zooniverse/Talk-Api@c46b359160e435aa5 is deployed

camallen: Well that makes it easier!

edpaget:
We still need the other change though right @camallen?

camallen:
@edpaget which other change?

edpaget:
Nevermind. I'm having trouble reading today.

PR #1219, opened 2015-07-29
Form project slugs to include owner name

parrish:
As discussed in #1139

I'm not sure if/when we want this to go, but it's a pretty straightforward change.  Of course, it'll require updating the existing records as well as the front end.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -12,7 +12,7 @@ class Project < ActiveRecord::Base
 
   EXPERT_ROLES = [:expert, :owner]
 
-  acts_as_url :display_name, sync_url: true, url_attribute: :slug, allow_duplicates: true
+  acts_as_url :slugged_name, allow_slash: true, sync_url: true, url_attribute: :slug, allow_duplicates: true
 
   has_many :workflows
   has_many :subject_sets, dependent: :destroy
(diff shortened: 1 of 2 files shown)


edpaget:
I was planning to leave them as two separate fields on the resource (`slug` and `owner_name`), so we wouldn't have to regenerate all the existing slugged fields. 

I created: https://github.com/zooniverse/Panoptes/pull/1208, so project resources could eager load the owning user group and just use the name field instead of having to handle both owner types and make an additional database call.

parrish:
That's less useful than a static field -- talk would still be calling the API on every comment render if it needs to be dynamically generated.

edpaget:
Why couldn't talk just include the owner_name field in the project response?

edpaget:
Actually nevermind, I like your solution better. I'm going to send you a PR to you PR to take care of the user_group stuff.

parrish:
Basically talk needs to be able to point users to front-end endpoints -- that's not possible at the moment without having to either add more foreign tables or consume the API -- both of which aren't ideal.  If projects stored either a fully formed slug (or slug + owner_name) then this would work

edpaget:
Cool that make sense. I send you a super trivial patch to just handle the user group case https://github.com/parrish/Panoptes/pull/2

edpaget:
:+1: Will merge as soon as travis finishes.

PR #1220, opened 2015-07-29
Update collection slugs similarly to project slugs

edpaget:
Add a rake task to update the slugs in Jenkins

### DIFF
--- a/spec/support/slugged_name.rb
+++ b/spec/support/slugged_name.rb
@@ -1,0 +1,15 @@
+RSpec.shared_examples "has slugged name" do
+  describe '#slugged_name' do
+    let(:resource){ create described_class.name.underscore, display_name: 'Some Awesome Project / Other Stuff', owner: owner }
+    subject{ resource.slug }
+    context "when owner is a user" do
+      let(:owner){ create :user, login: 'somebody' }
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -17,6 +17,8 @@
   enable_extension "plpgsql"
   enable_extension "intarray"
   enable_extension "pg_trgm"
+  enable_extension "hstore"
```
camallen: these extensions crept in again..

spec/support/slugged_name.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.shared_examples "has slugged name" do
+  describe '#slugged_name' do
+    let(:owner){ create :user, login: 'somebody' }
```
camallen: do you want to spec the case where an owner is a user group?

db/schema.rb
edpaget: Everytime I migrate without dropping first...

camallen: yeah i normally do a `rake db:migrate:reset RAILS_ENV=true` but it'd be nice to not have to worry about this. I think it's cause the extensions exist on the staging db (perhaps an old migration) so the schema dumper reads it and adds it to the the dump.

PR #1223, opened 2015-07-30
Deactivated controlled resources

### LINKED ISSUE #1222: Only include active resources in controlled_resources scopes
Deactivated ("Deleted") projects / users, etc are still included in our API scopes for controller actions, these should not be included.

camallen:
Closes #1222 - only use active resources in controlled scopes when the controlled resource responds to the activatable enum scope.


edpaget:
LGTM :shipit:

edpaget:
Ah the failing spec bring up a good point that there should be a way to see memberships in the inactive/invited states for group administrators.

camallen:
Will look into it.

PR #1224, opened 2015-07-30
only send notifications for persisted projects

camallen:
skip this callback on create actions, only run on updates.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -43,7 +43,7 @@ class Project < ActiveRecord::Base
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
 
-  before_save :send_notifications
+  after_update :send_notifications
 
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -43,7 +43,7 @@ class Project < ActiveRecord::Base
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
 
-  before_save :send_notifications
+  before_save :send_notifications, if: Proc.new { |project| project.persisted? }
```
edpaget: my bad on this I meant to only run on update can we do that instead of this proc?

camallen: yeah for sure. i didn't see after_update in the api docs list but it does exist.

edpaget:
LGTM :ship:

PR #1228, opened 2015-07-31
Public aggregation access by workflow config

### LINKED ISSUE #1213: Add workflow flag to enable public access to aggregations endpoint
zooniverse/annoTate#59 - I need public access to the aggregations endpoint for my workflow, so I can show people what's already been done.

cc @camallen @VVH

camallen:
Closes #1213 - Add per workflow aggregation config object and open up the index route to allow access if the supplied workflow id is public.

### DIFF
--- a/app/controllers/api/v1/aggregations_controller.rb
+++ b/app/controllers/api/v1/aggregations_controller.rb
@@ -1,15 +1,9 @@
 class Api::V1::AggregationsController < Api::ApiController
 
-  doorkeeper_for :index, :show, scopes: [:project], unless: :public_access?
   doorkeeper_for :create, :update, scopes: [:project]
   resource_actions :create, :update, :show, :index
   schema_type :json_schema
-
-  skip_before_action :check_controller_resources, only: :show,  if: :public_resource_workflow?
-
   before_action :filter_by_subject_set, only: :index
-  before_action :filter_by_public_workflows, only: :index, if: :public_workflows?
-  before_action :filter_by_public_resource_workflow, only: :show, if: :public_resource_workflow?
 
   private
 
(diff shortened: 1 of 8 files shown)


app/models/aggregation.rb
```diff
@@ -14,8 +14,15 @@ class Aggregation < ActiveRecord::Base
   validate :aggregation, :workflow_version_present
 
   def self.scope_for(action, user, opts={})
-    return super unless (action == :show || action == :index) && !user.is_admin?
-    joins(:workflow).merge(Workflow.scope_for(:update, user, opts))
+    case
+    when action == :index && opts.has_key?(:public_workflow_ids)
```
edpaget: Why not just always return public aggregations through the index or show routes?

app/controllers/api/v1/aggregations_controller.rb
```diff
@@ -1,16 +1,42 @@
...
+    @workflow_ids ||= params.slice(:workflow_id, :workflow_ids)
+      .values.join(",").split(',')
+  end
+
+  def public_workflows?
+    return @public_workflows if @public_workflows
+    @public_workflows = false
+    unless workflow_ids.blank?
+      public_count = Workflow.where(id: workflow_ids)
+        .where("aggregation ->> 'public' = 'true'").count
```
edpaget: Will we need an index on this object?

camallen: I was thinking the workflow ids would be restrictive enough.

app/models/aggregation.rb
camallen: Trying to tap into controlled resources in RoledController for the before_action [here](https://github.com/zooniverse/Panoptes/blob/1e8b9cadfbf67c60354158fddd4d4a1980eb9696/lib/role_control/roled_controller.rb#L8)

camallen: Just realised that effectively skips without valid`resource_ids`..i'll just override the index / show actions controlled_resources then. keep them out of the authorisation calls.

app/controllers/api/v1/aggregations_controller.rb
edpaget: Makes sense to me.

camallen:
reworked - controller is a bit heavier now to allow access to the public show route as well. Tried refactoring to a module but was really just hiding code in another file.

app/models/aggregation.rb
```diff
@@ -14,8 +14,11 @@ class Aggregation < ActiveRecord::Base
   validate :aggregation, :workflow_version_present
 
   def self.scope_for(action, user, opts={})
-    return super unless (action == :show || action == :index) && !user.is_admin?
-    joins(:workflow).merge(Workflow.scope_for(:update, user, opts))
+    if (action == :show || action == :index) && !user.is_admin?
+      joins(:workflow).merge(Workflow.scope_for(:update, user, opts))
```
edpaget: I think you can make your controller much lighter by doing

```ruby
joins(:workflow).merge(Workflow.scope_for(:update, user, opts)).or('"workflow"."aggregation"->>'public' = 'true')
```

But you'd have to index the aggregation json blob.

edpaget: And you also wouldn't be restricted to explicitly requested public workflows, but that doesn't seem like a thing we do elsewhere.

camallen: Pretty sure this would still fail the doorkeeper access for non-logged in users unless i just turn it off completely and return empty responses instead of 401s

edpaget: You could just remove https://github.com/camallen/Panoptes/blob/public_aggregations_rebase/app/controllers/api/v1/aggregations_controller.rb#L3. People will get 404s on #show routes, and empty sets if there are no public aggregations for them to se.

camallen: cool - i did like the 401 on those routes but i really don't like the amount of code in the controller to do this.

edpaget: Yeah if you want to enforce a logged in user, there's also the 'require_login' before_action.

camallen: Had a look at using an `or` query but can't keep getting malformed queries. Any chance you can take a look? latest in https://github.com/camallen/Panoptes/commits/public_aggregations

camallen: Also looks like rails AR gets or in rails 5 release, https://github.com/rails/rails/commit/9e42cf019f2417473e7dcbfcb885709fa2709f89#commitcomment-9535536

edpaget: Yeah I backported the that patch: https://github.com/zooniverse/Panoptes/blob/master/lib/gem_ext/rails/query_methods.rb what kind of malformed query are you seeing? I've been using it in a few places so far.

camallen: Yeah, it was the backport or. I think it was with null relation (`.none`) and bind param values but wasn't 100%. Will look into it further on monday.

edpaget:
@camallen this good to go?

camallen:
Yeah should be if you're happy with it.

PR #1229, opened 2015-07-31
Add private_profile flag to users

### LINKED ISSUE #1066: User Visibility Controls
Users should be able to set their profiles to private, such that other users won't be able to see their feed or stats pages.

edpaget:
Allows users to mark their classification counts (and possibly future
parts of their profile) as private

Closes #1066

### DIFF
--- a/db/migrate/20150727212724_add_private_profile_to_users.rb
+++ b/db/migrate/20150727212724_add_private_profile_to_users.rb
@@ -1,0 +1,5 @@
+class AddPrivateProfileToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :private_profile, :boolean, index: { where: "(private_profile = true)" }, default: false
+  end
+end
--- a/db/migrate/20150727212724_add_private_profile_to_users.rb
+++ b/db/migrate/20150727212724_add_private_profile_to_users.rb
@@ -1,5 +1,5 @@
 class AddPrivateProfileToUsers < ActiveRecord::Migration
   def change
-    add_column :users, :private_profile, :boolean, index: { where: "(private_profile = true)" }, default: false
+    add_column :users, :private_profile, :boolean, index: { where: "(private_profile = false)" }, default: true
   end
 end
--- a/app/models/concerns/preferences.rb
+++ b/app/models/concerns/preferences.rb
@@ -16,7 +16,14 @@ module Preferences
 
     def scope_for(action, user, opts={})
       return all if user.is_admin?
-      user.send("#{@preferences_for}_preferences")
+      case action
+      when "show", "index"
+        self.joins(:user, @preferences_for).where(:users => {private_profile: false},
+          @preferences_for.to_s.pluralize => {private: false})
+          .or(user.send("#{@preferences_for}_preferences"))
+      else
+        user.send("#{@preferences_for}_preferences")
+      end
     end
   end
 end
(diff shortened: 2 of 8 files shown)


db/migrate/20150727212724_add_private_profile_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddPrivateProfileToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :private_profile, :boolean, index: { where: "(private_profile = true)" }, default: false
```
edpaget: Not sure if this should be true or false by default.

camallen: me either but i'm leaning to false by default with the option to go private.

camallen:
LGTM

edpaget:
Thanks just want to get some input from @DarrenMcRoy @chrissnyder @aliburchard about whether this should be public or private by default.

DarrenMcRoy:
I agree with @camallen and vote public by default. I'll almost always vote for "opt-out" of features like this; the point of the feature is that it's good for the community, so it should be the default; the point of having it opt-out is that not everyone will want it, so they can turn it off.

edpaget:
Do we know is good for the community? We have in the past been strong resistent to making classificaitons counts public.

edpaget:
For instance, when science teams want 'top classifier' lists.

chrissnyder:
I agree with Ed in that I don't think the answer here is as clear as it first appears. This is really the first time the Zoo has had any sort of public profile (right?), so it's uncharted territory for us.

Pros for public profile:

- Easier to manage (no decisions about what is public or not... everything is just public).
- More things to share, discuss.

Pros for private profile:

- Gives users a bit of a sense of control over their Zooniverse experience.
- Caters to those who want to keep their work more private.
- Potentially avoid problems like Ed mentions above.

Looking at the above, I'm leaning towards saying private is actually a better default mode. If we are in doubt that something should be public or not, that feels like an indication that it should indeed be private.

camallen:
@mrniaboc thoughts on private profiles by default?

mrniaboc:
Time to reference some actual data! When asked what things would make users increase/decrease their participation in the Zooniverse, the only one they reacted strongly against was competition with other users. This makes me lean heavily towards private as default.

participation-survey

aliburchard:
Im all about stars being private by default. I like that @mrniaboc had data to auggest that people would prefer it, because I don't have anything other than a general sense of betrayal and annoyance when I've discovered my information has been made public without realizing it.

eatyourgreens:
@aliburchard Makayla has a 500 page PhD thesis full of user research into that sense of betrayal and annoyance.

+1 to keeping stats private by default. If we switch to public by default, we should test the idea with our volunteers first.

aliburchard:
@eatyourgreens that is a *lot* of pages...

edpaget:
So it sounds like this should be bring by default. I'll update to make sure that that's the case.

vrooje:
I know I'm a little late to the party, but I agree with @mrniaboc. Classification counts private by default... though I'd actually prefer to make the ribbons public by default. The ribbons aren't really a competition, just a description.

DarrenMcRoy:
I can admit when I've been overruled. :-)

edpaget:
@vrooje I'm not sure its possible to have one be public and one be private. And I think that should also be an optional thing, just since it is an change in established behaviour.

vrooje:
Hmm, okay. I guess I'm thinking ahead to when we might eventually let people put mini-ribbons beneath their avatars on Talk. I've talked to some people who would be okay with that because it's based on percentages, but who would never want their classification *numbers* published.

edpaget:
@camallen I think this should be good to go pending travis.

edpaget:
@camallen don't know what's up with the failures. They're from the priority selection?

camallen:
is this rebased of master?

PR #1237, opened 2015-08-05
config cors middleware from yaml file

### LINKED ISSUE #1233: Extract cors domain whitelist to config file
make it read from yaml config file instead of hard coded.

camallen:
closes #1233 - Replace env specific hard coded configs for a derived envs from the yaml file. the yaml file requires requires some care to conform to the expected format


camallen:
config file for deployment is in place.

PR #1238, opened 2015-08-05
set csrf headers on sign_in json request

### LINKED ISSUE #1236: Add CSRF route
For zooniverse/Panoptes-Front-End#1312

Currently the front end has to scrape the HTML on arbitrary pages to get a CSRF token. Ideally it should be able to just fetch one from, say, /csrf as plain text.

camallen:
closes #1236 - spec out the behaviour required for the sign in json route


edpaget:
LGTM :+1:

PR #1242, opened 2015-08-06
preserve queue id ordering

### LINKED ISSUE #1241: Subject queue operations should maintain the existing order
`Subject.enqueue_update` and `dequeue_update` [here](https://github.com/zooniverse/Panoptes/blob/520453d62b926e9e6fff11330520d42a4e744d3f/app/models/subject_queue.rb#L92) should maintain order of the queue as it was.

Postgresql int array union and diff sorts the queue after manipulation (http://www.postgresql.org/docs/9.1/static/intarray.html)
screen_shot_2015-07-16_at_14 22 02_360

image

camallen:
closes #1241 - turns out it's hard to maintain order of arrays in postgres. All the operators i found / tried modified returned sorted lists (asc).

This will just iterate over the scoped models calling update with ruby land array ordering. These methods should not be called outside of background workers unless the scope is restrictive enough e.g. [here](https://github.com/zooniverse/Panoptes/blob/0bce1f69f7bdd192b284bfd8be31250a596fad4f/lib/subject_selector.rb#L88).

I looked into something funky in the db to keep the `update_all` command but failed to get it working properly. Any ideas for AREL / SQL optimisations here are welcome.

image
```
select * FROM 
unnest(array[8,7,6,1,2,5]) WITH ORDINALITY AS set1(ids, num)
full join
(select * from unnest(array[5, 55]) as ids) as set2
on set1.ids = set2.ids
order by set1.num
```

### DIFF
--- a/spec/models/subject_queue_spec.rb
+++ b/spec/models/subject_queue_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.describe SubjectQueue, :type => :model do
+RSpec.describe SubjectQueue, type: :model do
   let(:locked_factory) { :subject_queue }
   let(:locked_update) { {set_member_subject_ids: [1, 2, 3, 4]} }
 
@@ -280,6 +280,11 @@ RSpec.describe SubjectQueue, :type => :model do
           expect(sms_ids.size).to eq(sms_ids.uniq.size)
         end
 
+        it 'should maintain the order of the set' do
+          ordered_list = ues.set_member_subject_ids | [ sms.id ]
+          SubjectQueue.enqueue(workflow, sms.id, user: user)
+          expect(ues.reload.set_member_subject_ids).to eq(ordered_list)
+        end
 
         context "when a queue's existing SMSes are deleted", sidekiq: :inline do
           before(:each) do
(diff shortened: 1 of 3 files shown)


spec/models/subject_queue_spec.rb
```diff
@@ -280,6 +280,13 @@
           expect(sms_ids.size).to eq(sms_ids.uniq.size)
         end
 
+        it 'should maintain the order of the set' do
+          ordered_list = ues.set_member_subject_ids | [ sms.id ]
+          SubjectQueue.enqueue(workflow, sms.id, user: user)
+          ues.reload.set_member_subject_ids.each_with_index do |id, index|
```
marten: Doesn't normal array comparison work here? Surely `[1,2,3] != [3,2,1]`?

camallen: yeah i'll update it.

marten:
LGTM.

If we ever *really* need more speed and less queries, we could construct bulk updates by querying a batch and then updating multiple rows [a la this](http://stackoverflow.com/questions/18797608/update-multiple-rows-in-same-query-using-postgresql). Don't think there's any Arel magic to build such a query, so not worth the effort and uglier code until we know we need the speed (IMHO).

PR #1246, opened 2015-08-07
make priority sort order static

camallen:
sort by ascending order for priority to allow open ended lists and prepend via negative numbers.

### DIFF
--- a/lib/postgresql_selection.rb
+++ b/lib/postgresql_selection.rb
@@ -25,7 +25,7 @@ class PostgresqlSelection
       query = query.where(subject_set_id: opts[:subject_set_id])
     end
     if workflow.prioritized
-      query = query.order(priority: opts.fetch(:order, :desc))
+      query = query.order(priority: :asc)
     end
     @available = query
   end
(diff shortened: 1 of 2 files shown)


edpaget:
Can you explain a little more why this needs to work this way? I'm a little confused.

camallen:
I assume you mean why default sort to asc?

edpaget:
Yeah. I'm not sure why it accomplishes what you stated in the commit message.

camallen:
allow addition to subject sets of new data that has higher and lower priority, i.e. lower just take last priority and inc. higher then prepend so if you start at 1 then 0,-1,-2, etc. Most priority sets will be static but may want new data.

marten:
Not entirely sure why that wouldn't just be reversed for descending order. Also not sure why a higher number for priority should not equate a higher priority.

camallen:
@marten I implemented a mech to allow the sort order to be parameterized but it had to rebuild the queue on changes. in the past all our priority sets have been single order but as you mentioned highest priority value is highest priority. Reversing the order allows the user to define the highest order in their set and then they can data append at will. If they need to prepend they can use negative values. So the use case currently is transcription via page numbers, where 0/1 comes first and we can add pages if needed.

Issue #1069, opened 2015-07-04
multiple users reporting duplicate images while signed in


vrooje:
Our 2 moderators on GZ Bars have both reported seeing duplicate images. I'm going to check the subject list for duplicates and will update but just wanted to flag this in case it's related to zooniverse/Panoptes#787 or something else that isn't duplicate subjects.

vrooje:
Update: I've just checked the metadata in the subject download from the project and I can't find any duplicates in the image name, so I don't think this is a duplicate-subject issue.

camallen:
Any talk links that indicate they have seen duplicate images?

vrooje:
I don't think so, yet... the mods posted on Talk but didn't link to specific images. Will ask them to do so in the future.

camallen:
@vrooje closing for now, please re-open if you have more info.

vrooje:
https://www.zooniverse.org/#/projects/vrooje/galaxy-zoo-bar-lengths/talk/21/280?page=1&comment=1515

From @Capella05:
 > Just got this image to classify for the second time - 465197

vrooje:
Also found the [subject comment thread](https://www.zooniverse.org/#/projects/vrooje/galaxy-zoo-bar-lengths/talk/32/565), which I suspect is redundant but am including it just in case.

vrooje:
I've now got more information about this:

Strangely I *don't* have a duplicate in the classification export for Capella05 on subject id 465197, despite the fact that my export was requested more than 24 hours after the reported duplicate.

I do have a different duplicate for that user, however:

`username   subject_id   metadata`
`Capella05  {"464985":  {"started_at":"2015-07-01T19:23:02.756Z","user...`
`Capella05  {"464985":  {"started_at":"2015-07-07T19:16:23.117Z","user...`

And I have a total of 159 duplicates from various users. That's about 1% of the classifications. Examples (with dummy usernames):

`thisuser  {"458602":  {"started_at":"2015-06-01T16:57:55.437Z","user...`
`thisuser  {"458591":  {"started_at":"2015-06-01T16:58:00.380Z","user...`
`thisuser  {"458584":  {"started_at":"2015-06-01T16:58:05.496Z","user...`
`thisuser  {"458550":  {"started_at":"2015-06-01T16:58:10.840Z","user...`
[11 more classifications of non-duplicate subjects]
`thisuser  {"458602":  {"started_at":"2015-06-01T16:59:23.785Z","user...`
[...]

PR #1251, opened 2015-08-10
add notifications around duplicate being queued

camallen:
Linked to #1069, try and get some visibility about queueing errors. This will check if
 1. a duplicate id is being enqueued to the current queue
 2. the queue is growing beyond 2 * the set default
 3. a seen before id is being enqueued

This is temporary and should be deployed after #1249 is in and the queues have been emptied. Manual inspection of the queues showed some have grown very large and may not be refreshing at all.


marten:
Tests fail but codewise lgtm

PR #1254, opened 2015-08-11
Access queue in order they were built for a logged in user

camallen:
Revert "Always return randomly selected subjects from the queue, since the queue is always sorted by id"

This reverts commit 0f9e41a6383590be807ec87cf3b66428b14aa6e3.

This will be needed for priority queues.

### DIFF
--- a/spec/models/subject_queue_spec.rb
+++ b/spec/models/subject_queue_spec.rb
@@ -371,7 +371,7 @@ RSpec.describe SubjectQueue, type: :model do
       end
 
       it 'should return the first subjects in the queue' do
-        expect(ues.next_subjects).to_not match_array(ues.set_member_subject_ids[0..9])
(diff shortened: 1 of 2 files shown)


spec/models/subject_queue_spec.rb
```diff
@@ -371,7 +371,7 @@
       end
 
       it 'should return the first subjects in the queue' do
-        expect(ues.next_subjects).to_not match_array(ues.set_member_subject_ids[0..9])
+        expect(ues.next_subjects).to match_array(ues.set_member_subject_ids[0..9])
```
edpaget: I don't think `match_array` cares about order. If you use `eq` it will.

camallen: thanks will update.

PR #1257, opened 2015-08-11
strip any seen before ids before enqueue

camallen:
do not enqueue subjects the user has already seen. workaround until the underlying cause of the seen_before dups are found. Note: we still get notified of the error event in honey badger but this will not queue the seen before dup for future classification.

### DIFF
--- a/app/models/subject_queue.rb
+++ b/app/models/subject_queue.rb
@@ -105,8 +105,8 @@ class SubjectQueue < ActiveRecord::Base
         curr_size: sq.set_member_subject_ids.size,
         append_size: sms_ids.length
       )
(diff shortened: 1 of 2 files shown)


edpaget:
Do you think the underlying issue is a gap in time between when the Subjects are selected to be enqueues and the UserSeenSubjects are updated?

camallen:
yeah could be - should we refill only after new data comes in? this is a workaround in the meantime. Also i'm refactoring this...will update this PR.

edpaget:
Yeah I think it might be worth exploring doing a one classificaiton comes in, one subject gets dequeued and one subject gets queued model.

camallen:
i'm refactoring this to get more info into why the seen befores and straight dups of the queue are attempting to queue up. Will issue a new PR with the changes.

PR #1260, opened 2015-08-12
refactor to dup removal class for existing queue

camallen:
more robust dup removal and cleaner service class. 

Not sure how long we want this to hang around so specc'd its behaviour through subject queue, can spec the class itself if needed but kind of want to get this into prod to get better messaging (more info on dups)  and not queue any kind of dup for end users.


camallen:
@edpaget or @marten i'd like to get this out this evening, any chance of a review.

PR #1264, opened 2015-08-13
Media put

edpaget:
+ Allows media resources to updated via PUT. 
+ Adds a client and configuration for connecting to the aggregation API
+ Adds aggregations_exports 
+ Add metadata state to dump resources, so we can reflect progress in the front-end.


camallen:
LGTM. quick question..how is the media scope update authorization setup here?

edpaget:
It should be the parent resources's update scope.

PR #1265, opened 2015-08-13
Add rack-timeout using default settings

camallen:
Supersedes #1055    

Adds react timeout with a default of 15s to process a request.  Also bump gem minor updates lock mysql2 gem minor updates that won't build. Hopefully this will help with the stuck puma thread while we figure out a way to fix it.


edpaget:
What kind of error does rack::timeout throw without any `rescue_from`? I think the error it raises will be uncaught leading to a 50x error type. If we make sure the error is 408, the frontend has code to catch and retry.

camallen:
500's by the looks, we can catch it and return a different error but need to take care that we don't get stuck there...https://github.com/heroku/rack-timeout#errors

camallen:
what does the front end do on 500's now?

edpaget:
Shows `Error: 0`. I don't think its necessarily reasonable to retry on 50x requests since they indicate an internal server error, whereas certain 40x errors are recoverable.

edpaget:
I had a local branch where I was working on rescuing and returning. I think I got hung up figuring out tests for it.

edpaget:
Nevermind looks like I don't have it anymore. I don't mind merging this in and figuring out the proper handling later.

camallen:
I was just going to say let's get this in and see what happens. I think these stuck threads that would raise the timeout error never return anyway. they get stuck looking for a FUTEX that is never set so the associated client request ends up timing out i assume.

camallen:
I see this as a stop gap, if it's causing too many problems we can rip it out and look at another solution.

PR #1267, opened 2015-08-13
Update aggregation api

edpaget:


app/controllers/api/v1/projects_controller.rb
```diff
@@ -69,6 +69,7 @@ def create_subjects_export
   def create_aggregations_export
     media_params[:metadata] ||= { recipients: [api_user.id] }
     medium = create_or_update_medium(:aggregations_export, media_params)
+    p medium
```
camallen: debug statement

camallen:
LGTM

app/controllers/api/v1/projects_controller.rb
edpaget: Thanks, will remove

PR #1271, opened 2015-08-14
remove rack-timeout

camallen:
merge this when we're happy that the stuck thread issue is gone. maybe mon morning...


camallen:
@edpaget - any reason not to remove this? There was one search error timeout on /users/ index but nothing since.

edpaget:
Why not just keep it? At least we get notified if there are horrendous delays.

marten:
Turns out it causes problems leaving it in: http://stackoverflow.com/a/31167112 is probably what caused https://app.honeybadger.io/projects/40595/faults/14858040. Rebase and merge?

Issue #982, opened 2015-06-16
Ability to pause and upause subjects


mschwamb:
I got told to move this issue over here. 

What happens if a science team puts in way more data than the classification rated can actually handle on then site and decide to switch tactics wanting to pause most of the live subjects except a small subset - ie Planet Four going for classifying full seasons during stargazing to targeting one region and classifying all the data for four seasons of it - switching to the new scheme meant pausing most of the season 1 data we uploaded

currently you'd have to make a new subject set and merge the previous classifications already in hand in processing. Is there any way/menu to pause and unpause subjects from the workflow editor or the backend?

camallen:
Our plan in panoptes was to hand over control of subject data management to project collaborators and thus to have all subjects active the whole time. This will fit the majority of use cases but you raise a good point. 

Perhaps this is really about best practices to organise large amounts of subject data into sets and then being able to choose to activate/deactivate those sets for selection by linking to workflows. This setup would give you the control you desire. Failing that we could add an active state on the subject that would enable subjects to be served to volunteers. This state could then be modified through the front end like the subject uploader - this will add more data management complexity to the project owner though, linked to #886.

In my mind i think getting the subject data setup correctly in the first instance, noting that normal user will be limited to 10K subjects and will liaise with us to go beyond this amount so we can advise on best practice then.

@aliburchard thoughts re best practice for data management?

mschwamb:
Well I'm not sure telling people best practices will help the situation. Especially if it's a group that can't do data analysis it would mean re-uploading everything and losing the previous classifications because I presume the aggregation software woudn't be able to handle that and then having duplicate Talk pages. 

Planet Four was a Zooniverse made project with core team helping with data management and discussions (both Arfon, Stuart, Chris were directly involved in communicating with the science team (I joined after the build was nearly complete),  the Zooniverse devs were also making the subject data (parsing from large images). It was well known what was going in but, we still needed to pause subjects after stargazing to get a science output. Other projects like Cyclone Center have done that before as well focusing on just a single storm for looking at all storm data for a year. 

So I think really having the future ability to pause even if it's through the API to have states like Ouroboros does to make some live or pause a subject will be a very handy capacity.

mschwamb:
Since @chrissnyder has done the bulk of pausing and unpausing on Planet Four for us, he might be able to comment about how frequently this is done on other custom projects and about the handiness of such a feature

camallen:
[...]

PR #1277, opened 2015-08-17
Make retiring subject possible via update_links on a workflow

### LINKED ISSUE #1098: Retiring subjects via API
I'm figuring out how to retire subjects by API call. The logical place for that seems to be using the `SetMemberSubjectController` (as per @edpaget's comments in #519). 

But to use that API (and for things like `SubjectWorkflowCount#retire!`), the API caller needs the id of the join model (why does a join table even have it's own ids?), which has been hidden since #519.

How do we see this fitting in the API?

/cc @camallen @edpaget

marten:
Fixes #1098


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -69,12 +69,28 @@ def extract_strings(tasks)
...
+    end
+  end
+
   def new_items(resource, relation, value)
-    items = construct_new_items(super(resource, relation, value), resource.project_id)
-    if items.any? { |item| item.is_a?(SubjectSet) }
-      items
+    case relation
+    when :retired_subjects, 'retired_subjects'
+      # TODO: scope_for
```
marten: Any ideas? Might not be needed? Any user allowed to update a workflow can now mark any of that workflow's subjects as retired, which seems fine by me?

camallen: This will be nice to have as researchers may want to remove subjects from circulation, see https://github.com/zooniverse/Panoptes/issues/982#issuecomment-112829720

edpaget: @camallen do you mean we should allow people with the 'researcher' role to retire subjects?

camallen: no - i was just using that term...should have said project owners / collabs.

edpaget: Ah cool that what marten's code should do.

app/models/subject_workflow_count.rb
```diff
@@ -1,4 +1,6 @@
 class SubjectWorkflowCount < ActiveRecord::Base
+  include Linkable
```
camallen: Does linkable need to be included?

spec/models/workflow_spec.rb
```diff
@@ -145,6 +145,45 @@
     end
   end
 
+  describe '#retire_subject' do
+    it 'marks a counted subject as retired' do
+      workflow = create(:workflow_with_subject_sets)
```
camallen: Do you want to extract some of these setup lines to let blocks?

```ruby
describe '#retire_subject' do
  let(:workflow) { create(:workflow_with_subject_sets) }
  let(:subject) { create(:subject, subject_sets: workflow.subject_sets) }

#and override in the let's in different context blocks?
end
```

spec/models/workflow_spec.rb
```diff
@@ -145,6 +145,45 @@
...
+      expect(subject.retired_for_workflow?(workflow)).to be_truthy
+      expect(SubjectWorkflowCount.retired.count).to eq(2)
+    end
+
+    it 'leaves a retired subject retired' do
+      workflow = create(:workflow_with_subject_sets)
+      subject = create(:subject, subject_sets: workflow.subject_sets)
+
+      workflow.retire_subject(subject.id)
+      workflow.retire_subject(subject.id)
```
camallen: will this update the `retire_at` timestamp?

marten: No, `#retire!` has a guard that early-returns when it's already retired.

camallen: cool want to add an `aggregate_failures` double expectation that the retire at stamp doesn't change?

app/models/subject_workflow_count.rb
marten: Apparently not anymore with the approach I ended up with. Good catch.

camallen:
Awesome work, merging.

PR #1278, opened 2015-08-17
do not attempt to find unseens for logged out user

camallen:


lib/set_member_subject_selector.rb
```diff
@@ -11,7 +11,7 @@ def initialize(workflow, user)
 
   def set_member_subjects
     to_classify = select_set_member_subjects_to_classify
-    unless to_classify.exists?
+    if !to_classify.exists? && user
```
edpaget: Do you want to do this check the other way around? `user && !to_classify.exists?` it should save a db query to check to_classify in cases where no user is supplied.

camallen: I only want to run this if there main selection is empty as a fallback to queue data.

edpaget: Sorry `.exists?` is doing a db check right? I was just suggesting you can test the user condition so you don't do the query if it's not met.

camallen: yeah - sorry. that makes sense. now i look at it i think this is wrong anyway. i should not be queueing possibly retired data otherwise we'll overclassify. but instead just returning an empty scope that will trigger a db select anyway.

Issue #1247, opened 2015-08-07
Project newsletters for old projects -- do people have to re-opt-in?


DarrenMcRoy:
I'm slightly concerned because I've noticed that it doesn't appear that I am still automatically subscribed to the newsletters for any of the old projects that I've classified on.

For instance, if we send a newsletter to Chicago Wildlife Watch users, is it only going to go to those who have manually resubscribed to it... ? Because that cannot be too many people?

I also checked the "Get email updates from the Projects you classify on" button, and it didn't check off the list below.

And I tried doing a classification on Snapshot Serengeti to make it appear in my list of checkboxes, with the "Get email updates from the Projects you classify on" box checked, and when Snapshot appeared, it was unchecked.

![screen shot 2015-08-07 at 3 43 11 pm](https://cloud.githubusercontent.com/assets/8987864/9145695/38d42e84-3d1c-11e5-8fad-b31fbcf57dfc.png)

Just a bit concerned over whether we've lost track of who is subscribed to what, and if all our users were imported from old projects with empty newsletter settings. What's up there?

camallen:
This is related to the stats migration. @mrniaboc will be able to construct lists in the meantime but we need to get this fixed.

> I also checked the "Get email updates from the Projects you classify on" button, and it didn't check off the list below.

This is intended behaviour, this field is used when a user classifies for the first time on a project. if ticked and a user classifies (for the first time) then any subsequent project email preference will be set to true.

>And I tried doing a classification on Snapshot Serengeti to make it appear in my list of checkboxes, with the "Get email updates from the Projects you classify on" box checked, and when Snapshot appeared, it was unchecked.

Yeah this is a gap between the planned migration being run around the cut over period. The current code updates the counts an existing preference for legacy projects. It's not responsible for communication prefs since the canonical version of this should have been migrated.

So we haven't lost track we just haven't consolidated the data. In theory any new modifications to project prefs will have true / false values in panoptes DB. Old communication pref data is in zoo home so we need to combine the two when migrating.

edpaget:
Closed by 1279

PR #1279, opened 2015-08-17
Import and store legacy project counts by workflow

### LINKED ISSUE #1103: Legacy project activity counts should respect individual workflows
Linked to https://github.com/zooniverse/Panoptes-Front-End/issues/1028#issuecomment-121093777

Live event counts are per worklow so legacy workflow sums can be lost by classifying on legacy projects that only have the live workflow count. Update the activity count to be a json field and re-migrate the per workflow data. In turn update the events controller to update the per workflow counts.

edpaget:
Closes #1103 #1247

### DIFF
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -2,8 +2,6 @@ module Api
 
   class EventsController < ApplicationController
 
-    skip_before_action :verify_authenticity_token, if: :json_request?
-
     KNOWN_EVENTS = %w( activity workflow_activity )
 
     def self.resource_name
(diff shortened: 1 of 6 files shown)


app/controllers/api/events_controller.rb
```diff
@@ -2,8 +2,6 @@ module Api
 
   class EventsController < ApplicationController
 
-    skip_before_action :verify_authenticity_token, if: :json_request?
```
camallen: i assume that since we don't raise auth token error anymore this won't be a problem.

edpaget: Yup it should just have a null session.

camallen:
pretty sure this is good. Do you want to mark the [migrated ones](https://github.com/camallen/Panoptes/blob/4c5cba47fa25b307112cb7074f946eca15b4d701/lib/zooniverse_user_subscription.rb#L83) in case you have to restart the migrations?

edpaget:
Not really I'm going to run on Jenkins so if something interrupts the run I assume it's something that'll need to be fix globally.

camallen:
Cool - merging.

PR #1285, opened 2015-08-18
[WIP] V2 api

edpaget:
No need to merge yet. I just wanted to get this opened so people could take a look and offer ideas also critique the format.

When we move to a JSON API compliant spec, we should cut down on some extraneous API routes, mostly for join tables like `set_member_subjects` and instead make more of the links meaningful. So instead of just linking from a project to workflows, link from a project to active_workflows and inactive_workflows. 

I decided not use Apiary for this, since it's a huge PITA. Since we use a regularized format for our JSON we can just explain the format once, then list and describe what the attributes and links mean, which seems more useful than just showing example API responses.

### DIFF
--- a/doc/api_v2.md
+++ b/doc/api_v2.md
@@ -259,7 +259,13 @@ JSON API responses contain four top-level sections
 
 ### Clients
 
-Clients for JSON-API written in a variety of languages can be found [here]().
+Clients for JSON-API written in a variety of languages can be found [here](http://jsonapi.org/implementations/).
+
+#### Examples
+
++ [Using the Zooniverse Classification API with Python](python_client.md)
++ [Using the Zooniverse Classification API with Javascript](javascript_client.md)
++ [Using the Zooniverse Classification API with Ruby](ruby_client.md)
 
 ### Resources
 
--- a/doc/api_v2.md
+++ b/doc/api_v2.md
@@ -1,0 +1,256 @@
+# Zooniverse Classification API V2
+
+The Zooniverse Classification API (version 2) is a [JSON-API]() compliant API.
+
+<!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-generate-toc again -->
+**Table of Contents**
+
+- [Global](#global)
+    - [Content-Types](#content-types)
+- [Authorization](#authorization)
+    - [Authorization Code](#authorization-code)
+    - [Client-Credentials](#client-credentials)
+    - [Resource Owner Credentials](#resource-owner-credentials)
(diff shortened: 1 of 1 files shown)


chrissnyder:
Would this be an appropriate place to discuss broader directions we'd like to see the API take, or should it be scoped to how we organize and structure the responses?

edpaget:
Fine with me

edpaget:
I'd like to basically have the next verison documented before starting implementing it anyway.

camallen:
Looks good. Lots of detail in there, probably best to go over this in person when you are here next week.

brian-c:
Looks like this'll be awesome. I really dig the idea of more specific links.

chrissnyder:
- Huge +1 on the idea of more specific links. Should really make working with the data structures more enjoyable.
- For errors, I'd really like it if we focused on making useful and informative `code` values, rather than trying to make the `details` user-readable. User meaning the volunteer. While they might be useful to a developer, they really aren't appropriate to directly output within an interface, and I don't see a problem asking a dev to have the lookup table from code to explanation handy.
- This might not be the right place to bring it up, but are we still looking to manage the roles ACL internally within Panoptes? Really scared that this will get out of control from a client's perspective as keep adding new services if we don't figure it out soon.
- Anyone have any thoughts on making task workflows follow a strict schema? I find it a bit scary that an innocent change to the project builder could materially affect how tasks are saved, and not having that be catchable seems bad. Especially as we build out custom projects that use workflows created within the project builder.

edpaget:
I don't think it matters whether Panoptes or another service manages roles and access permissions, but yeah they do need to be centralized somehow.

parrish:
In terms of meaningful link keys, I'm totally on board with that.  That definitely clarifies the intent.

As far as the spec changes go, I'm just okay with it.

I'm not 100% sold on how the spec is becoming more verbose over time.  Being more explicit with `relationships` could possibly be more flexible in cases of wacky associations.  Conversely I find a loss of flexibility in `included` simply being a mixed type array -- we're going to have to map that by type every time we make multiple includes on requests.

Honestly, most of the spec changes have been changes without real benefit or even an obvious motivation.  I'd probably be less annoyed if I picked up the spec after 1.0.

> we should cut down on some extraneous API routes, mostly for join tables like set_member_subjects

It's up in the air as to whether or not that route (in specific) is unneeded.  It certainly would be necessary if we have to do per-project subject pages.

As far as permissions being moved outside of the app/database, I'd say we're a long way off from needing that.  Currently that'd just be adding complexity.

PR #1286, opened 2015-08-19
Add fallback to compile assets on start

astopy:
To temporarily work around a bug in Docker which can stop the assets
container from being created during build. It should be rare that this
happens, and worst case this will just add a couple of minutes to the
deploy time.

### DIFF
--- a/scripts/docker/start.sh
+++ b/scripts/docker/start.sh
@@ -20,6 +20,11 @@ else
     git pull
   fi
 
+  if [ ! -d public/assets ]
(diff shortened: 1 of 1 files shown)


camallen:
merging...it's be nice not to have to rely on this.

marten:
I assume a load balancer upstream will correctly wait for this to be processed and becoming responsive to web requests before stopping old containers?

astopy:
Yup.

PR #1289, opened 2015-08-19
compile assets if missing for staging

camallen:
used in staging env related to #1286


camallen:
my bad thought staging was running under foreman not supervisor.

PR #1293, opened 2015-08-20
Kafka classification stream using AMS instead of RestPack

### LINKED ISSUE #1072: Include subjects in the Kafka classification JSONs
For the SWAP integration with Kafka the worker needs to have access to the metadata of the subjects of the classification. It needs that to check if a subject is a training subject. To avoid dealing with caching  and making API calls to fetch the metadata, it's easiest if we post the subject JSON along with the classification straight into Kafka.

marten:
The V1Adapter I've added is compatible (for this case, in general there might be a few minor differences with RestPack, haven't exhaustively checked) with output from RestPack. The V1Adapter is basically a copy of an older AMS adapter from before they started working towards JSON-API 1.0, with a few naming changes to be like RestPack.

AMS has a built-in JsonApi adapter (which I've made default in initializer) which follows the 1.0 spec for JSON-API.

Apart from that, AMS "just works" with things like belongs_to_many, has support for virtual associations as well, just define a method on the serializer and state at the top that it has_many or belongs_to that method name.

Fixes #1072

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,8 +9,8 @@ gem 'doorkeeper', '~> 1.4.1'
 gem 'devise', '~> 3.0'
 gem 'versionist', '~> 1.0'
 gem 'rack-cors', '~> 0.3', require: 'rack/cors'
-gem 'restpack_serializer', github: "edpaget/restpack_serializer", branch: "dev" # API
-gem 'active_model_serializers', '0.10.0.rc2'
+gem 'restpack_serializer', github: "edpaget/restpack_serializer", branch: "dev" # REST API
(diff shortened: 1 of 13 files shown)


edpaget:
This looks great.

camallen:
yeah - i like it

marten:
Awesome, I'll expand the KafkaSubjectSerializer and poke you for a final review.

marten:
Added metadata to subject serialization, should be good enough for Nero now. LGTY?

edpaget:
:shipit: once travis finishes.

PR #1298, opened 2015-08-20
Switch Puma to listen on port 81

astopy:
So that we can use net=host in Docker when we deploy, without it conflicting with Nginx.


astopy:
Note that no more deployments can happen until this is merged :)

camallen:
LGTM

PR #1300, opened 2015-08-21
Prepend queue ids

### LINKED ISSUE #1245: Add ids to front of user queue
Adding subjects to `subject_queue` should allow the option of adding to the front of the queue for use in projects like spacewarps retirment scheme.

camallen:
closes #1245 - does what it says on the tin.

I'd rather have a `has_many :subjects, through: :set_member_subjects` relation on SubjectQueue working but the has_many_through scope didn't work with the belongs_to_many link.

### DIFF
--- a/spec/controllers/api/v1/subject_queues_controller_spec.rb
+++ b/spec/controllers/api/v1/subject_queues_controller_spec.rb
@@ -12,6 +12,7 @@ describe Api::V1::SubjectQueuesController, type: :controller do
   let!(:set_member_subjects) do
     create_list(:set_member_subject, 3, subject_set: workflow.subject_sets.first)
   end
+  let(:set_member_subject_ids) { set_member_subjects.map(&:id) }
   let(:subjects) { set_member_subjects.map(&:subject) }
   let(:subject_ids) { subjects.map(&:id) }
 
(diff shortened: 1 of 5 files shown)


edpaget:
LGtM

camallen:
i've got to run rspec bisect on this to get a reproducable failing case.

spec/controllers/api/v1/subject_queues_controller_spec.rb
```diff
@@ -31,18 +34,50 @@
 
   describe "#update" do
     let(:test_attr) { :set_member_subject_ids }
-    let(:test_attr_value) { subjects.map(&:id) }
+    let(:test_attr_value) { subject_ids }
```
edpaget: These should be sms_ids

spec/controllers/api/v1/subject_queues_controller_spec.rb
```diff
@@ -31,18 +34,50 @@
...
+    let(:test_relation_ids) { subject_ids }
```
edpaget: So should this.

camallen: the api requires subject ids to be passed in which are then converted to sms ids on the fly, so i need different expectations. I'm running bisect to get a reproducible test case on this.

PR #1302, opened 2015-08-24
Save classifications to Cassandra

### LINKED ISSUE #987: sending transcriptions/classifications/markings directly to Cassandra

edpaget:
Duplicate classification records into Cassandra for aggregation engine

Closes #987

### DIFF
--- a/app/models/cassandra/classification.rb
+++ b/app/models/cassandra/classification.rb
@@ -7,6 +7,7 @@ class Cassandra::Classification
   key :workflow_version, :int
   key :user_ip, :inet
   key :user_id, :int
+  key :classification_id, :int
   column :created_at, :timestamp
   column :updated_at, :timestamp
   column :user_group_id, :int
@@ -20,7 +21,7 @@ class Cassandra::Classification
 
   def self.from_ar_model(classification)
     attrs = classification.attributes.dup
-    attrs.delete("id")
+    attrs["classification_id"] = attrs.delete("id")
     attrs.delete("subject_ids").each do |subject_id|
       create!(attrs.merge(subject_id: subject_id))
     end
--- a/app/models/cassandra/classification.rb
+++ b/app/models/cassandra/classification.rb
@@ -1,0 +1,59 @@
+class Cassandra::Classification
+  include Cequel::Record
+
+  key :project_id, :int
+  key :workflow_id, :int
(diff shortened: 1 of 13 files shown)


edpaget:
I put in a PR for my Cequel changes https://github.com/cequel/cequel/pull/249

app/models/cassandra/classification.rb
```diff
@@ -0,0 +1,59 @@
...
+  column :gold_standard, :boolean
+  column :expert_classifier, :boolean
+  column :metadata, :text
+  column :annotations, :text
+
+  after_create :save_subject
+
+  def self.from_ar_model(classification)
+    attrs = classification.attributes.dup
+    attrs.delete("id")
```
camallen: could the panoptes id be used to link a classification with multiple subjects for say a pairwise comparison?

edpaget: Greg's annotation code wasn't using it, and I'm not :100: % sure of the affect of including it as part of the cassandra key. We could add it though.

camallen: ok, @ggdhines any use case for including the panoptes-id key to group multi subject classifications together?

ggdhines: if you mean multi frame classifications like WW, then no (seems to work fine with just multiple frames). if you mean multiple subjects as with Solar Stormwatch - then maybe. Haven't though about that. Maybe include it to be on the safe side

camallen: yeah this is for projects like solar stormwatch - will it break your current schema?

ggdhines: yeah - stormwatch will probably break a bunch of things. But will have to deal with that sort of project separately

camallen: @ggdhines will including the panoptes id key break things or will the aggregation engine just not run for projects like stormwatch? We're only interested in the former, the later can be punted till we run a project like this.

ggdhines: @cam - the id key shouldn't break anything

edpaget:
Whoo!

PR #1310, opened 2015-08-26
The expires_in option is only in newer versions of the AWS SDK gem

### LINKED ISSUE #1177: One hour is too short for holding classification dumps
I requested classifications from the Project Builder, but they didn't arrive within ten minutes, and I had a lot of other things to do. I came back to my email an hour and a half later -- 61 minutes after the email had arrived -- and was unable to get my classifications. So now I'm re-requesting them, which is a waste of processing on the Panoptes side and a pain in the neck for me. Can we up that holding time to, say, 6 hours? Or 24?

marten:
In the version we use it's called `expires`.

Fixes #1177 for real this time.

### DIFF
--- a/lib/media_storage/aws_adapter.rb
+++ b/lib/media_storage/aws_adapter.rb
@@ -44,7 +44,7 @@ module MediaStorage
       if opts[:private]
         object(path).url_for(:read,
                              secure: true,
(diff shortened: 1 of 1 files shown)


astopy:
Is there a reason we can't use the latest version of the gem?

Issue #1311, opened 2015-08-27
Remove authorization requirement for 1st Party applications. 


edpaget:
Cross Ref https://github.com/zooniverse/wildcam-gorongosa/issues/94

PR #1312, opened 2015-08-27
Skip authorization step if the application is ours

edpaget:
Closes zooniverse/wildcam-gorongosa#94
Closes zooniverse/Panoptes#1311

### DIFF
--- a/spec/controllers/authorizations_controller_spec.rb
+++ b/spec/controllers/authorizations_controller_spec.rb
@@ -41,6 +41,15 @@ describe AuthorizationsController, type: :controller do
     sign_in owner
   end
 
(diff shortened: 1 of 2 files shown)


spec/controllers/authorizations_controller_spec.rb
```diff
@@ -41,6 +41,15 @@
     sign_in owner
   end
 
+  context "a grant from a first party application" do
+    let!(:app) { create(:first_party_pap, owner: owner) }
```
marten: First party pap?

edpaget: whoops

PR #1318, opened 2015-09-02
Allow subject sets to be filtered by metadata

edpaget:
Introduced metadata filters for subject sets via a param in the form
metadata.field

Adds a gin index to subject sets


spec/controllers/api/v1/subject_sets_controller_spec.rb
```diff
@@ -55,6 +55,12 @@
 
     it_behaves_like 'is indexable'
     it_behaves_like 'has many filterable', :workflows
+
+    it "is filterable by metadata", :focus do
```
camallen: remove focus.

camallen:
LGTM after a rebase

PR #1319, opened 2015-09-02
Scope enqueues/dequeues to subject set

edpaget:
This was making the subject and classification routes unbearably slow.

For grouped workflows we can assume that subjects are only queued in
their respective group queue and make sure we're only dequeuing from
that install of all the queues


edpaget:
Once I have this passing I'm going to merge it in. Sorry all, this should resolve the slowness from today.

edpaget:
Also this will only fix the controllers not the worker. We need to figure out something else for them still.

PR #1322, opened 2015-09-02
handle empty database subject selection

camallen:
don't try and idx an empty array and raise an error to indicate no data to select from for the client


edpaget:
LGTM

PR #1323, opened 2015-09-02
dequeue subjects only once

camallen:
when the classification comes in and not when serving out as well.

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -41,9 +41,9 @@ class ClassificationLifecycle
   end
 
   def dequeue_subjects
-    sms = SetMemberSubject.by_subject_workflow(subject_ids, workflow.id).pluck(:id, :subject_set_id)
-    sms_ids = sms.map(&:first)
-    set_id = workflow.grouped ? sms.map { |s| s[1] }.first : nil
+    sms_scope = SetMemberSubject.by_subject_workflow(subject_ids, workflow.id)
+    sms_ids = sms_scope.pluck(:id)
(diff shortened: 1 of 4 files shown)


edpaget:
Can we do it the other way around? Otherwise we have the possiblity of duplicates from when the front-end requests more subjects before classifications have come in.

camallen:
yep can do note that it too may have a race condition too where the queue is below min threshold and it kicks off an enqueue task that may requeue subjects that have just been selected

camallen:
kinda sounds like we need both ends dequeing...

chrissnyder:
PFE does track subjects seen in a given session. The only thing it affects is wether we display the badge "you already classified this" or not. We could look through the subjects returned from the API and prioritize ones that haven't been seen.

camallen:
@chrissnyder how does that relate to https://github.com/zooniverse/Panoptes-Front-End/issues/1427?

chrissnyder:
I think they are one in the same. We could apply that prioritization to both logged in and not logged in users. Basically just making https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/classify.cjsx#L146 a bit smarter.

camallen:
@edpaget what do you think about this? I'd like to not rely on clients having to track this data to ensure we don't get dups.

camallen:
@chrissnyder - any ETA on the smarter subject session tracking?

edpaget:
Yeah I don't think we should rely on clients at all for not showing dups.

camallen:
closing in favour of #1327

PR #1324, opened 2015-09-02
Expose ports for easier debugging

marten:
I'm happy to change the ports if it collides with anyone's docker containers, but this just makes it so much easier to open a postgresql shell etc. I keep making this change and then not committing it, so I figured it'd poll opinions on a PR.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ postgres:
   environment:
     - "POSTGRES_USER=panoptes"
     - "POSTGRES_PASSWORD=panoptes"
+  ports:
+    - "5432:5432"
 
 zookeeper:
   image: zooniverse/zookeeper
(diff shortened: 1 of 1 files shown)


camallen:
happy for this to go in...i think default ports make sense

marten:
Cool. It might collide with linux-folk since you can use docker straight on your machine, and you might have PG running locally too?

astopy:
Maybe don't use the default ports to avoid it colliding?

marten:
@astopy do you know what IP exposed ports bind to by default? If not 127.0.0.1, can we make that so, would help accidentally exposing a database dump?

astopy:
You can add the IP in the port specification (as IP:host_port:container_port). It’ll bind to 0.0.0.0 by default I think, but if you change it to 127.0.0.1 that’ll probably break for docker-machine.

marten:
Just merging this then. Feel free to change the ports if it bugs anyone.

PR #1325, opened 2015-09-02
explicit queue build for the subject set

camallen:
don't assume building on all subject sets for a workflow, allow them to be explicitly specified.


edpaget:
LGTM

PR #1327, opened 2015-09-02
spec dequeue on both selection and classification receipt

camallen:
spec dequeue on both ends to ensure we don't allow dups


edpaget:
LGTM

PR #1331, opened 2015-09-03
use order by random for selection

camallen:
~~looks like the prev select between random was doing a seq scan, this change should use an index scan instead.~~ 

It's still a seq scan with an index scan for the order by.

Update: ran some benchmarks and the techniques are similar in terms of response times but the select between two random points doesn't handle larger limits well and relies on loop breaking to not spin lock. Also in my testing the distribution coming out of between random points was sequential, order by random looked more random.

```
#ran @ 3 x sequentially

## larger limits

"random sample for 1000"
"limit was -> 100"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.002905)

"random sample for 1000"
"limit was -> 999"
"Using order by random()"
  0.010000   0.000000   0.010000 (  0.004057)

"random sample for 1000"
"limit was -> 999"
"Using order by select between two random points random()"
"breaking loop with length -> 983"
  0.160000   0.000000   0.160000 (  0.314260)

"random sample for 1000"
"limit was -> 999"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.004051)

#smaller limits

"random sample for 2500"
"limit was -> 20"
"Using order by select between two random points random()"
  0.010000   0.000000   0.010000 (  0.003676)

"random sample for 2500"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.002653)

"random sample for 2500"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.003201)

"random sample for 5000"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.004333)

"random sample for 5000"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.003080)

"random sample for 5000"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.002887)


-----------------------------------------------------------

"random sample for 2500"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.003406)

"random sample for 2500"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.003782)

"random sample for 2500"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.004394)

"random sample for 5000"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.003915)

"random sample for 5000"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.004824)

"random sample for 5000"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.003908)
```

```
[15] pry(#<PostgresqlSelection>)> query.order("RANDOM() #{direction}").explain
=> EXPLAIN for: SELECT "set_member_subjects"."id","set_member_subjects"."random","set_member_subjects"."priority" FROM "set_member_subjects" INNER JOIN "subject_sets" ON "set_member_subjects"."subject_set_id" = "subject_sets"."id" INNER JOIN "subject_sets_workflows" ON "subject_sets"."id" = "subject_sets_workflows"."subject_set_id" WHERE "subject_sets_workflows"."workflow_id" = $1  ORDER BY RANDOM() asc [["workflow_id", 1]]
                                                                      QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=16.91..16.92 rows=1 width=47)
   Sort Key: (random())
   ->  Nested Loop  (cost=15.04..16.90 rows=1 width=47)
         ->  Hash Join  (cost=14.91..16.26 rows=1 width=55)
               Hash Cond: (set_member_subjects.subject_set_id = subject_sets_workflows.subject_set_id)
               ->  Seq Scan on set_member_subjects  (cost=0.00..1.25 rows=25 width=51)
               ->  Hash  (cost=14.79..14.79 rows=10 width=4)
                     ->  Bitmap Heap Scan on subject_sets_workflows  (cost=4.23..14.79 rows=10 width=4)
                           Recheck Cond: (workflow_id = 1)
                           ->  Bitmap Index Scan on index_subject_sets_workflows_on_workflow_id_and_subject_set_id  (cost=0.00..4.23 rows=10 width=0)
                                 Index Cond: (workflow_id = 1)
         ->  Index Only Scan using subject_sets_pkey on subject_sets  (cost=0.13..0.63 rows=1 width=4)
               Index Cond: (id = set_member_subjects.subject_set_id)
(13 rows)

[16] pry(#<PostgresqlSelection>)> query.where('"set_member_subjects"."random" BETWEEN random() AND random()').explain
=> EXPLAIN for: SELECT "set_member_subjects"."id","set_member_subjects"."random","set_member_subjects"."priority" FROM "set_member_subjects" INNER JOIN "subject_sets" ON "set_member_subjects"."subject_set_id" = "subject_sets"."id" INNER JOIN "subject_sets_workflows" ON "subject_sets"."id" = "subject_sets_workflows"."subject_set_id" WHERE "subject_sets_workflows"."workflow_id" = $1 AND ("set_member_subjects"."random" BETWEEN random() AND random()) [["workflow_id", 1]]
                                                                QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=5.28..17.54 rows=1 width=47)
   Join Filter: (subject_sets.id = set_member_subjects.subject_set_id)
   ->  Hash Join  (cost=5.28..15.88 rows=1 width=8)
         Hash Cond: (subject_sets_workflows.subject_set_id = subject_sets.id)
         ->  Bitmap Heap Scan on subject_sets_workflows  (cost=4.23..14.79 rows=10 width=4)
               Recheck Cond: (workflow_id = 1)
               ->  Bitmap Index Scan on index_subject_sets_workflows_on_workflow_id_and_subject_set_id  (cost=0.00..4.23 rows=10 width=0)
                     Index Cond: (workflow_id = 1)
         ->  Hash  (cost=1.02..1.02 rows=2 width=4)
               ->  Seq Scan on subject_sets  (cost=0.00..1.02 rows=2 width=4)
   ->  Seq Scan on set_member_subjects  (cost=0.00..1.62 rows=3 width=51)
         Filter: (((random)::double precision >= random()) AND ((random)::double precision <= random()))
(12 rows)
```

### DIFF
--- a/lib/postgresql_selection.rb
+++ b/lib/postgresql_selection.rb
@@ -37,11 +37,12 @@ class PostgresqlSelection
   end
 
   def sample(query=available)
-    query.where('"set_member_subjects"."random" BETWEEN random() AND random()')
+    direction = [ :asc, :desc ].sample
+    query.order("RANDOM() #{direction}")
   end
 
   def limit
-    opts.fetch(:limit, 20).to_i
+    @limit ||= opts.fetch(:limit, 20).to_i
   end
 
   def selection_strategy
@@ -55,7 +56,7 @@ class PostgresqlSelection
   def select_results_randomly
     enough_available = limit < available_count
     if enough_available
-      construct_random_sample_to_limit
+      sample.limit(limit).pluck(:id)
     else
       available.pluck(:id).shuffle
     end
@@ -63,16 +64,5 @@ class PostgresqlSelection
 
   def select_results_in_order
     available.limit(limit).pluck(:id)
-  end
-
-  def construct_random_sample_to_limit
-    results = []
-    attempt = 0
-    until results.length >= limit do
-      results = results | sample.limit(limit).pluck(:id)
-      attempt += 1
-      break if attempt >= MAX_RANDOM_SAMPLE_ATTEMPTS
-    end
-    results
   end
 end


camallen:
hold off on this, just running some benchmarks

edpaget:
I'm really curious how this doesn't result in a seq scan. Presumably the database needs uses index scans to create a set of results, but then it need to iterate through each row and call the `random()` function then sort by the result of `random()` which will result in a seq scan.

camallen:
@edpaget see above, your right it's still a seq scan.

edpaget:
I'm curious if there's a difference on really large sets 50,000k or more?

camallen:
I can run a test on this in a bit

marten:
I wonder if we might be better off doing it in two queries:

```
ids = available.pluck(:id).shuffle[0..limit]
available.where(id: ids)
```

edpaget:
Cool I totally get preformance being the same at this order of magnitude, but I think we're going to find order by slow, the more queries we run.

camallen:
I think the limit will keep this to a reasonable scope size and it's cleaner code. More data to come.

edpaget:
The limit doesn't affect the number of rows it needs to sort. It still has to call random for every row that matches the where clause, then sort every row to decide which 20 to return.

camallen:
Right - I've tested at large (100K) and small (21) subjects and tested small and large limits to see how this behaves. Looks like order by random is the optimal solution and differences in query times at larger set sizes are pretty negligible, so for most of our use cases it'll be fast enough. Note the order between randoms exhibits looping behaviour (stuck puma issue a while back) when the limit is near the available size and blows out for large sets ~14s for the looping query whereas the order by random increased inline with expectations ~0.5s

See the bottom as well for an example of the actual resulting random distribution results (the behaviour we actually want) which sees a more random distribution doing to order by rand on primary key.

```
## smaller page sizes

"random sample for 100000"
"limit was -> 20"
"Using order by random()"
  0.010000   0.000000   0.010000 (  0.043217)

"random sample for 100000"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.039316)

"random sample for 100000"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.019653)

"random sample for 100000"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.010259)


## larger page sizes

"random sample for 100000"
"limit was -> 1000"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.042889)

"random sample for 100000"
"limit was -> 1000"
"Using order by random()"
  0.010000   0.000000   0.010000 (  0.060722)

"random sample for 100000"
"limit was -> 1000"
"Using order by select between two random points random()"
  0.010000   0.000000   0.010000 (  0.023967)

"random sample for 100000"
"limit was -> 1000"
"Using order by select between two random points random()"
  0.010000   0.000000   0.010000 (  0.015656)

## limits approaching small set sizes
 
"random sample for 21"
"limit was -> 20"
"Breaking loop with 19 selected"
  0.060000   0.000000   0.060000 (  0.069741)

"random sample for 21"
"limit was -> 20"
"Using order by select between two random points random()"
  0.010000   0.000000   0.010000 (  0.021218)

## limits approaching large set size

"random sample for 100000"
"limit was -> 99999"
"Using order by random()"
  0.400000   0.000000   0.400000 (  0.516620)

"random sample for 100000"
"limit was -> 99999"
"Using order by random()"
  0.380000   0.000000   0.380000 (  0.496844)

"random sample for 100000"
"limit was -> 99999"
"Using order by select between two random points random()"
"Breaking loop with 98062 selected"
  7.910000   0.030000   7.940000 ( 14.656803)

"random sample for 100000"
"limit was -> 99999"
"Using order by select between two random points random()"
"Breaking loop with 98005 selected"
  8.120000   0.020000   8.140000 ( 14.800690)


## Result id orders and randomness

"random sample for 100000"
"limit was -> 20"
"Using order by random()"
=> [47907, 24490, 1031, 90372, 18025, 70665, 699, 53319, 67091, 33043, 33595, 14971, 54231, 27814, 71114, 60083, 83977, 45973, 50553, 22964]

"random sample for 100000"
"limit was -> 20"
"Using order by select between two random points random()"
=> [99999, 99998, 99986, 99981, 99979, 99977, 99972, 99961, 99953, 99950, 99948, 99942, 99929, 99919, 99916, 99909, 99900, 99888, 99877, 99875]
```

edpaget:
That seems a lot slower...

camallen:
which one was a lot slower? Also did you see the resulting random distribution from the queries? Looks like the current implementation doesn't provide a random distribution like we actually want.

edpaget:
Using order by random seems a great deal slower 

```
"random sample for 100000"
"limit was -> 20"
"Using order by random()"
  0.000000   0.000000   0.000000 (  0.039316)

"random sample for 100000"
"limit was -> 20"
"Using order by select between two random points random()"
  0.000000   0.000000   0.000000 (  0.019653)
```

Almost 200ms slower, and 100k isn't that big of a subject set.

edpaget:
Never mind I can't read.

edpaget:
Sure let's try it this way.

camallen:
cool, agree that the random sort on large sets may be problematic but the current implementation has problems as well.

PR #1332, opened 2015-09-03
remove cache counters

camallen:
seems the project / worfklow counter cache from classifications are causing lots of deadlocked updates and slowing the api nodes down. Removing to see if this is the problem.


edpaget:
@camallen can you add `:disabled` to the two failing specs.

camallen:
on it

Issue #1193, opened 2015-07-27
Should we allow subjects to be deleted via the API


camallen:
Seems any collaborator / project owner can delete subjects via the API. Since subjects can be re-used across the zooniverse should we still allow this?

mkosmala:
Yes, especially for testing. To get my projects up and running, I create the project and then add three images to each workflow. I don't care about these images; they're just for testing. When I'm ready to do a beta or launch, I delete these dummy images via the API and load my real images in.

mkosmala:
(But can be tricky, as you note. So I would vote for a solution that makes deleting subjects relatively easy -- but perhaps subject to some constraints.)

brian-c:
Yes, but not if they've been classified in a live project?

Maybe that check could happen automatically when they're removed from a subject set. I don't think the user should have to know the difference between deleting a resource and deleting the only link to it.

camallen:
So a classification in any project could be a proxy to being commented in talk / collected. Can anyone think of a way to comment / collect a subject without classifying it somehow first?

So removal from a subject set just removes the link between `subject_set` and `subject`, it doesn't actually delete any subject (these just get orphaned from the site now for reference in any collections / talk, etc). 

This functionality is really why i wanted to start this discussion as we don't provide an interface to delete subjects except via the API which may cause problems if more advanced users decide to issue `DELETE /subjects/id` without understanding the consequences. @mkosmala did you actually delete the subject resources via the delete described above or just the `subject_set`?

mkosmala:
I can't think of a way to comment or collect a subject via the front-end without first classifying it. 

And, all my "deletes" were through the front-end. I've both clicked the little trash can icon to delete images one at a time and deleted entire subject sets full of subjects. I haven't looked to see what this did on the back-end -- if I understand you, @camallen, I really just removed links between the subjects and subject_sets when I did these things. 

I have not issued any ```DELETE```s via the back-end. (Seems a bit scary to do that... I would probably consult with you guys to see what the consequences would be if I were to try it that way.)

But a follow-up question: if my subjects are just getting orphaned when I trash-can them (or delete entire subject sets) and they aren't really being deleted, do they still count toward my subject quota?

camallen:
> I have not issued any DELETEs via the back-end. (Seems a bit scary to do that... I would probably consult with you guys to see what the consequences would be if I were to try it that way.)

Yep - that's why i'm wondering if we should allow this end point at all.

[...]

PR #1337, opened 2015-09-07
Count subjects directly, not individual set members

### LINKED ISSUE #1162: Why is SubjectWorkflowCount associated with a SetMemberSubject?
```
         ┌──────────────────┐      ┌───────────────┐       ┌──────────────────┐   
     ┌──▶│SetMemberSubject 1│◀─────│    Subject    │──────▶│SetMemberSubject 2│◀─┐
     │   └──────────────────┘      └───────────────┘       └──────────────────┘  │
     │             │                                                 │           │
     │             ▼                                                 ▼           │
     │   ┌──────────────────┐                              ┌──────────────────┐  │
     │   │   SubjectSet 1   │                              │   SubjectSet 2   │  │
     │   └──────────────────┘                              └──────────────────┘  │
     │             ▲                                                 ▲           │
     │             └───────────────────────┬─────────────────────────┘           │
     │                                     │                                     │
     │  ┌──────────────────────┐   ┌───────────────┐   ┌──────────────────────┐  │
     └──│ SubjectWorkflowCount │──▶│   Workflow    │◀──│ SubjectWorkflowCount │──┘
        └──────────────────────┘   └───────────────┘   └──────────────────────┘   
```

I did some digging, and it seems like:

* A subject can theoretically be in two sets, both of which are assigned to the same workflow. I couldn't find an actual usage of this.
* If this is the case, it would get selected for queueing with a double chance, since the query for that doesn't do a `SELECT DISTINCT`.
* We always increment the counter for all SMSes for a given subject+workflow, irrespective of which subject_set it was selected from to put in the queue (I don't think we even keep track of this).
* Therefore, counts are usually in sync. If you add a set to a live workflow, and introduce duplication between subjects that way, then they might be out of sync.
* When any of the counters goes over the threshold, we retire the all SetMemberSubjects for that workflow (not per set)

Proposal:

```
┌──────────────────┐      ┌───────────────┐       ┌──────────────────┐
│SetMemberSubject 1│◀─────│    Subject    │──────▶│SetMemberSubject 2│
└──────────────────┘      └───────────────┘       └──────────────────┘
          │                       ▲                         │         
          ▼                       │                         ▼         
┌──────────────────┐  ┌──────────────────────┐    ┌──────────────────┐
│   SubjectSet 1   │  │ SubjectWorkflowCount │    │   SubjectSet 2   │
└──────────────────┘  └──────────────────────┘    └──────────────────┘
          ▲                       │                         ▲         
          │                       ▼                         │         
          │               ┌───────────────┐                 │         
          └───────────────│   Workflow    │─────────────────┘         
                          └───────────────┘                           
```

* Classifications deal with subjects. Why should the counter for (basically) number of classifications deal with SMSes?
* This would eliminate a whole bunch of join statements.
* There is something which grouped workflows that I haven't looked into how that ties in with this. I think that just means we want subjects out of all the sets in the queue. Which would be really strange to use when you have overlap between sets.

I'm happy to pick this discussion up in a week and a half, this ticket is mostly so I don't forget about it.

(Note to future self: [chat link](https://zooniverse.slack.com/archives/D06EY3EE9/p1437475092000023))

marten:
This isn't actually a difference in counting styles, because we used to increment counters and mark retirement on all SMSes for a subject-workflow pair. This just removes that so that we just have a single SWC for each pair.

Fixes #1162

### DIFF
--- a/app/controllers/api/v1/subject_sets_controller.rb
+++ b/app/controllers/api/v1/subject_sets_controller.rb
@@ -86,7 +86,6 @@ class Api::V1::SubjectSetsController < Api::ApiController
   private
 
   def remove_linked_set_member_subjects(resource, set_member_subjects)
-    SubjectWorkflowCount.where(set_member_subject: set_member_subjects).delete_all
     QueueRemovalWorker.perform_async(set_member_subjects.pluck(:id), resource.workflows.pluck(:id))
     CountResetWorker.perform_async(resource.id)
     set_member_subjects.delete_all
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -42,9 +42,17 @@ class SetMemberSubject < ActiveRecord::Base
   end
 
   def self.non_retired_for_workflow(workflow)
-    by_workflow(workflow)
-    .joins("LEFT OUTER JOIN subject_workflow_counts ON subject_workflow_counts.subject_id = set_member_subjects.subject_id")
-    .where('subject_workflow_counts.id IS NULL OR subject_workflow_counts.retired_at IS NULL')
+    if SubjectWorkflowCount::BACKWARDS_COMPAT
+      by_workflow(workflow)
+        .joins("LEFT OUTER JOIN subject_workflow_counts swc1 ON swc1.set_member_subject_id = set_member_subjects.id")
+        .joins("LEFT OUTER JOIN subject_workflow_counts swc2 ON swc2.subject_id = set_member_subjects.subject_id")
+        .where('swc1.id IS NULL OR swc1.retired_at IS NULL')
+        .where('swc2.id IS NULL OR swc2.retired_at IS NULL')
+    else
+      by_workflow(workflow)
+      .joins("LEFT OUTER JOIN subject_workflow_counts ON subject_workflow_counts.subject_id = set_member_subjects.subject_id")
+      .where('subject_workflow_counts.id IS NULL OR subject_workflow_counts.retired_at IS NULL')
+    end
   end
 
   def self.unseen_for_user_by_workflow(user, workflow)
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -65,6 +65,15 @@ class SetMemberSubject < ActiveRecord::Base
     retired_workflows.pluck(:id)
   end
 
+  def retired_workflows
+    if SubjectWorkflowCount::BACKWARDS_COMPAT
+      workflow_ids = SubjectWorkflowCount.retired.by_subject(subject_id).pluck(:workflow_id)
+      Workflow.where(id: workflow_ids)
+    else
+      super
+    end
+  end
+
   def retired_workflows=(workflows_to_retire)
     workflows_to_retire.each do |workflow|
       workflow.retire_subject(subject)
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -4,9 +4,10 @@ class SetMemberSubject < ActiveRecord::Base
 
   belongs_to :subject_set, counter_cache: true, touch: true
   belongs_to :subject
-  has_many :subject_workflow_counts, dependent: :destroy
   has_many :workflows, through: :subject_set
-  has_many :retired_subject_workflow_counts, -> { retired }, class_name: 'SubjectWorkflowCount'
+
+  has_many :subject_workflow_counts, through: :subject
+  has_many :retired_subject_workflow_counts, -> { retired }, through: :subject, class_name: 'SubjectWorkflowCount', source: 'subject_workflow_counts'
   has_many :retired_workflows, through: :retired_subject_workflow_counts, source: :workflow
 
   validates_presence_of :subject_set, :subject
@@ -42,7 +43,7 @@ class SetMemberSubject < ActiveRecord::Base
 
   def self.non_retired_for_workflow(workflow)
     by_workflow(workflow)
-    .joins("LEFT OUTER JOIN subject_workflow_counts ON subject_workflow_counts.set_member_subject_id = set_member_subjects.id")
+    .joins("LEFT OUTER JOIN subject_workflow_counts ON subject_workflow_counts.subject_id = set_member_subjects.subject_id")
     .where('subject_workflow_counts.id IS NULL OR subject_workflow_counts.retired_at IS NULL')
   end
 
@@ -53,7 +54,7 @@ class SetMemberSubject < ActiveRecord::Base
   end
 
   def retire_workflow(workflow)
-    count = subject_workflow_counts.find_or_create_by!(workflow_id: workflow.id)
+    count = SubjectWorkflowCount.find_or_create_by!(workflow_id: workflow.id, subject_id: subject_id)
     count.retire!
   end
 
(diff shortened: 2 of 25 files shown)


app/models/set_member_subject.rb
```diff
@@ -53,7 +54,7 @@ def self.unseen_for_user_by_workflow(user, workflow)
   end
 
   def retire_workflow(workflow)
-    count = subject_workflow_counts.find_or_create_by!(workflow_id: workflow.id)
+    count = SubjectWorkflowCount.find_or_create_by!(workflow_id: workflow.id, subject_id: subject_id)
     count.retire!
```
camallen: Should we kick off the dequeue for retired subjects here similar to what i've proposed in #1339?

db/migrate/20150827124834_alter_subject_workflow_counts_add_subject_id.rb
```diff
@@ -0,0 +1,33 @@
+class AlterSubjectWorkflowCountsAddSubjectId < ActiveRecord::Migration
+  def up
+    add_column :subject_workflow_counts, :subject_id, :integer, index: true
```
camallen: not null?

db/migrate/20150827124834_alter_subject_workflow_counts_add_subject_id.rb
```diff
@@ -0,0 +1,33 @@
...
+    add_foreign_key :subject_workflow_counts, :subjects
+
+    # Create new aggregated (per subject) SWC records
+    execute <<-END
+      INSERT INTO subject_workflow_counts (subject_id, workflow_id, classifications_count, created_at, updated_at, retired_at)
+      SELECT sms.subject_id,
+             workflow_id,
+             SUM(classifications_count) AS classifications_count,
```
camallen: Do we want to run a sum here? I thought we updated each sms for a subject / workflow combo so in theory they should have the same classification count.

db/migrate/20150827124834_alter_subject_workflow_counts_add_subject_id.rb
```diff
@@ -0,0 +1,33 @@
...
+      ORDER BY classifications_count DESC
+    END
+
+    # Remove old SWCs
+    execute <<-END
+      DELETE FROM subject_workflow_counts WHERE subject_id IS NULL AND set_member_subject_id IS NOT NULL;
+    END
+
+    add_index :subject_workflow_counts, [:subject_id, :workflow_id], unique: true
+    change_column_null :subject_workflow_counts, :subject_id, false
```
camallen: ah - good one

app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -86,7 +86,6 @@ def destroy_relation(resource, relation, value)
   private
 
   def remove_linked_set_member_subjects(resource, set_member_subjects)
-    SubjectWorkflowCount.where(set_member_subject: set_member_subjects).delete_all
```
camallen: are we keeping these around for reference?

camallen:
Looks good - i'm wondering about the best way to get this deployed with the migration timing. Maybe extract the migration code to a rake task and have both counters being updated for the meantime.

marten:
Ill double check, but I recall trying that statement on a local copy of
production (wondering how slow it would be) and it only taking a few
seconds, which is why I didn't rake it.

app/controllers/api/v1/subject_sets_controller.rb
marten: Could but then we can't set subject_id  to not null

camallen:
if it's super quick then great.

app/controllers/api/v1/subject_sets_controller.rb
camallen: So looks like you've removed the cleanup for linked SWCounts and are thus keeping them around. Is this correct or are they getting removed another way?

db/migrate/20150827124834_alter_subject_workflow_counts_add_subject_id.rb
marten: You're right. Brainfart on my end. I'll swap it for a MAX().

app/controllers/api/v1/subject_sets_controller.rb
marten: I guess we should clean them up if/when a subject or workflow is destroyed (does that ever happen?). Is a `ON DELETE CASCASE` on the foreign key relation in pg good enough for you?

camallen: workflows yes - subjects sort of(https://github.com/zooniverse/Panoptes/issues/1193), no need for pg cascades we can just hook into the API lifecycle events like this method used to. If we go in and hack on the console and break things then that's our fault.

camallen: also if there is a reason to keep them around (like tracking retirement) then maybe we should.

marten:
Just ran it again, takes 21 seconds to migrate.

camallen:
that sounds fine if we do it during a low time.

camallen:
@marten it'd be nice to get this into production. did you want to rebase and fixup the sum counters?

marten:
Sure. I was waiting for things to stabilise a bit, but I guess that's now.

app/controllers/api/v1/subject_sets_controller.rb
marten: What's the reason for not liking a pg cascade? Too magical or is there a performance reason I'm unaware of? I don't think there's any need to keep a counter for a subject/workflow that no longer exists. We should generally either do soft-deletions or just remove everything associated (imho).

camallen: no performance consideration. principle of least surprise and the functionality being visible in code.

app/models/set_member_subject.rb
marten: I'm opting not to touch queuing in this PR given how many fires we've been fighting around it, let me know if you think we should create an issue and follow up in another PR.

camallen: yeah - put an issue in so we can have a look at it.

app/controllers/api/v1/subject_sets_controller.rb
marten: I just put in a `dependent: :destroy` on the has_many relations, and set the foreign keys to restrict so we can't accidentally make a mess on the console. Probably the best of both explicitness and referential integrity.

app/models/set_member_subject.rb
marten: Actually, this method is only used in a spec file. Getting rid of it.

camallen:
LGTM - will need a rebase and we should schedule this in a down period (early UK time perhaps?)

camallen:
@marten are you still going to modify this to be backwards compatible and extract the migration to a runnable rake task? If so we will hold off on this.

marten:
I started on it yeah. Halfway through I think

marten:
@camallen okay, this is now compatible regardless of whether the data has been converted yet.

1. Deploy this code
2. Run rake task (locks the swc table for about a minute, so run outside peak hours)
3. Make branch that removes BACKWARDS_COMPAT by assuming it's set to false instead of true
4. Deploy that.

I will prepare step 3 now.

marten:
Cleanup commit after rake has been run: https://github.com/marten/Panoptes/commit/7cb16cbb4478682bda8859f2e487d2015e6db132

PR #1343, opened 2015-09-08
update queues using optimistic locking with retires

camallen:
update queues respecting stale versions with a retry component. hopefully we won’t see too many re-raised stale object errors in sidekiq but with the combination of retires in code and sidekiq we should get consistent data in the queues.

we can always look at serializable transaction isolation level as well but thought i'd try this method first.

### DIFF
--- a/app/models/concerns/with_optimistic_retry.rb
+++ b/app/models/concerns/with_optimistic_retry.rb
@@ -9,12 +9,9 @@ module WithOptimisticRetry
       yield
     rescue ActiveRecord::StaleObjectError => e
       begin
-        # Reload lock_version in particular.
         reload
         retries -= 1
-      rescue ActiveRecord::RecordNotFound
-        # If the record is gone there is nothing to do.
-      else      
+      else
         if retries > 0
           retry
         else
--- a/app/models/concerns/with_optimistic_retry.rb
+++ b/app/models/concerns/with_optimistic_retry.rb
@@ -1,0 +1,26 @@
+# http://apidock.com/rails/ActiveRecord/Persistence/reload
+module WithOptimisticRetry
+
+  private
+
+  def with_optimistic_retry(retries=nil)
+    retries ||= 10
+    begin
+      yield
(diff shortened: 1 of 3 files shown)


app/models/concerns/with_optimistic_retry.rb
```diff
@@ -0,0 +1,26 @@
...
+    retries ||= 10
+    begin
+      yield
+    rescue ActiveRecord::StaleObjectError => e
+      begin
+        # Reload lock_version in particular.
+        reload
+        retries -= 1
+      rescue ActiveRecord::RecordNotFound
+        # If the record is gone there is nothing to do.
```
edpaget: Should this raise?

edpaget:
This looks nice. I'm still super annoyed we can do atomic sql updates. I'm worried at some point we're going to accidentally overwrite another change, but this should overall prevent that from happening.

app/models/concerns/with_optimistic_retry.rb
camallen: in our case yes.

edpaget: Sorry I mean should it re-raise the error after catching it. It shouldn't happen since don't ever delete queues at the moment right?

app/models/concerns/with_optimistic_retry.rb
```diff
@@ -0,0 +1,25 @@
...
+  def with_optimistic_retry(retries=nil)
+    retries ||= 10
+    begin
+      yield
+    rescue ActiveRecord::StaleObjectError => stale
+      begin
+        reload
+        retries -= 1
+      rescue ActiveRecord::RecordNotFound => not_found
+        raise not_found
```
edpaget: You could just not catch it now ;)

camallen:
yeah - I was fondly thinking of atomic updates in a database that will remain nameless.

app/models/concerns/with_optimistic_retry.rb
camallen: haha!

PR #1347, opened 2015-09-09
User counter-cache gem for most counters

### LINKED ISSUE #1341: Reinstate removed classification counters
get these back with background count workers or redis based key counts

https://github.com/zooniverse/Panoptes/pull/1332/files

edpaget:
Closes #1341


edpaget:
Please double/triple check what I'm doing here with https://github.com/wanelo/counter-cache it was quite late as I wrote this.

edpaget:
When this is deployed we'll need to add an extra `counter:` section to the redis configuration.

camallen:
Just for reference, i think we're all ok to shelve this one for the moment and have a look at fixing this with alternate strategies like background workers or a different data store.

PR #1352, opened 2015-09-09
Filter out retired subjects when pulling from the queue

marten:
This means we don't need to care about dequeueing retired subjects, and
can pay that price slowly over time when people actually start using
their queues.

### DIFF
--- a/app/models/subject_queue.rb
+++ b/app/models/subject_queue.rb
@@ -108,10 +108,14 @@ class SubjectQueue < ActiveRecord::Base
 
   def next_subjects(limit=10)
     if user_id
-      set_member_subject_ids[0..limit-1]
+      non_retired_set_member_subject_ids[0..limit-1]
     else
-      set_member_subject_ids.sample(limit)
+      non_retired_set_member_subject_ids.sample(limit)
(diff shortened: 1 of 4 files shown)


camallen:
@marten rebase?

marten:
Rebased.

marten:
I guess I should've also added a spec to check we don't return retired subjects. On it now.

camallen:
mind reader

camallen:
LGTM

edpaget:
Do we want to do with #1353 ?

camallen:
I think we can use them together, dequeue on select, enqueue on classification receipt if below min and then selection don't show retired subjects. Thoughts?

edpaget:
It should be much cheaper to retire subjects from the worker now, so we could still do that and save the hit on subject selection.

marten:
I'm fine with deploying one first, and waiting a bit to see how that changes load. This PR should be easy to rebase in any case.

camallen:
sounds like a plan

camallen:
closing this, let's revisit if we end up needing this.

PR #1353, opened 2015-09-09
Use int array functions in subject queue

edpaget:


app/models/subject_queue.rb
```diff
@@ -84,18 +84,16 @@ def self.create_for_user(workflow, user, set_id: nil)
...
+    query.update_all([dequeue_sql, sms_ids])
   end
 
-  # NOTE: if the query scope returns a large set then
-  # don't use this query outside of a background Worker
-  # SQL optimisations welcome here...no postgres ops preseved order
   def self.enqueue_update(query, sms_ids)
-    query.find_each { |sq| sq.enqueue(sms_ids) }
+    return if sms_ids.blank?
+    enqueue_sql = "set_member_subject_ids = subarray(uniq(sort(set_member_subject_ids | array[?])), 0, #{DEFAULT_LENGTH})"
```
camallen: can this just be `set_member_subject_ids = array[?]` since we've already remove dups and seen_before?

spec/models/subject_queue_spec.rb
```diff
@@ -479,7 +455,6 @@
       end
 
       context "optimisitic updates" do
```
camallen: this can go i think

app/models/subject_queue.rb
edpaget: That's not happening anymore. Since we're not iterating through all the queues. We're only removing the seen before.

camallen: got it now

lib/non_duplicate_sms_ids.rb
```diff
@@ -1,52 +1,23 @@
...
-    queue.user
-  end
-
-  def workflow
-    queue.workflow
-  end
-
-  def curr_queue_size
-    curr_queue_ids.size
+    @seen_before = seen_before_set.map(&:id)
```
camallen: can this become a pluck?

lib/non_duplicate_sms_ids.rb
```diff
@@ -1,52 +1,23 @@
 class NonDuplicateSmsIds
```
camallen: what do you think about renaming this now? the original name was a pretty average anyway.

camallen:
LGTM - we should just notify on the open dups ticket that this may changing behaviour.

lib/non_duplicate_sms_ids.rb
edpaget: Sure

marten: I'd do that in the `seen_before_set` itself. We're not running multiple queries (`exists?` and `pluck`) against it, better to just load the plucked array once and perform presence check against that array (technically).

edpaget:
It really shouldn't introduce dups. The original problem with `uniq()` in postgres was that the array had to be sorted first and we weren't doing that.

lib/non_duplicate_sms_ids.rb
edpaget: Doesn't matter to me. I'll just it SeenSubjectRemover.

lib/non_duplicate_sms_ids.rb
edpaget: Cool will change to do that.

lib/seen_subject_remover.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def append_ids_size
+    append_ids.length
+  end
+
+  def seen_before_set
+    @seen_before ||= if user_seen_subject
+      SetMemberSubject.where(id: append_ids)
+        .joins(:subject)
+        .where(subjects: { id: user_seen_subject.subject_ids })
```
camallen: i think @marten was talking about plucking here to create the array and then testing in the error_params if seen_before_set.size > 0

marten:
LGTM

PR #1357, opened 2015-09-09
Subject uploader counter cache

camallen:
avoid trying to update this counter during subject creates. only update this value when checking for subject limits currently but it could be called at other times.

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -15,36 +15,12 @@ class UserSerializer
 
   media_include :avatar, :profile_header
 
-  def credited_name
-    @model.credited_name
-  end
-
-  def email
-    @model.email
-  end
-
-  def global_email_communication
-    @model.global_email_communication
-  end
-
-  def project_email_communication
-    @model.project_email_communication
-  end
-
-  def beta_email_communication
-    @model.beta_email_communication
-  end
-
   def firebase_auth_token
     FirebaseUserToken.generate(@model)
   end
 
   def max_subjects
     @model.subject_limit
-  end
-
-  def uploaded_subjects_count
-    @model.uploaded_subjects_count
   end
 
   def admin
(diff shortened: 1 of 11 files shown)


app/serializers/user_serializer.rb
```diff
@@ -43,10 +23,6 @@ def max_subjects
     @model.subject_limit
   end
 
-  def uploaded_subjects_count
```
marten: If you're removing it from the API, and recounting every time you upload a subject, why have the counter in the DB at all?

marten: Also needs to be removed from the "include_#{me_only_attribute}?" thing in the private section of this serializer I presume.

marten: Nevermind, you're just removing useless wrappers around an attribute that Restpack can find on it's own.

camallen: thanks - good point. i thought it was once per set of uploads but you're right, it's per subject.

camallen: i'll move to a query based count on each create, i'll index that field and make sure it's unscoped.

marten: You need to remove `uploaded_subjects_count` from the [`attributes` list](https://github.com/zooniverse/Panoptes/pull/1357/files#diff-270dcd01883f952843b67f5bb9c5c93dL9) and the [section making sure it's only serialized when it's your own user](https://github.com/zooniverse/Panoptes/pull/1357/files#diff-270dcd01883f952843b67f5bb9c5c93dL69)

camallen: I thought we'd probably want that in there still, maybe as an optional attribute?

marten: Yeah I missed that it's now a method on user, thought the column disappeared and so it would raise if requested. Noticed it later when it popped up in the spec. LGTM and Merged.

PR #1358, opened 2015-09-10
focus set with random sampling for selection

camallen:
try head and tail window sampling of the subjects table with a random sample of the results. @edpaget @parrish @marten thoughts?


edpaget:
We're going to need to add a a worker right away to regenerate the random column, right? Otherwise this is going to server ~2000 subjects at the start.

parrish:
Still needs the OFFSET and the random value reassignment.

Probably would improve things as is though

camallen:
yeah - we could have a daily job to regenerate the random column values..

camallen:
random value assignment on selection will swamp the db index update, right?

edpaget:
It would need to be more often that than unless we could with Michael's OFFSET suggestion. ~2000 subjects active in a day might be way too low for some projects.

parrish:
Somewhat.  Depends on the index.  It looks like it's a non-composite btree which isn't terrible.

parrish:
Hourly is probably fine

marten:
Or we could just only reset random column when `rand < 0.05` or something

marten:
Upside: if there's lots of activity, we reset the column more often so we distribute over the entire set better instead of focussing on 2000 subjects for an hour.

Downside: if there's lots of activity, the DB has more load, and then on top of that has to update this index more often.

camallen:
@edpaget @parrish added a worker to reassign the random attribute of the "focussed" set thanks to @marten . thoughts?

edpaget:
I'd be concerned that it could queue a bunch of these workers and update all the random values multiple times in short succession.

marten:
Re earlier:

> It would need to be more often that than unless we could with Michael's OFFSET suggestion. ~2000 subjects active in a day might be way too low for some projects.

Well if a subject is retired we'll automatically pick the next, so it's a sliding limit if I'm not mistaken. But the 2000 limit is up for debate, as is the 0.05 (together they *accidentally* average out so that each subject will have seen one classification by the time we reshuffle).

edpaget:
Cool, well I'm good to try this out.

parrish:
If you're worried about triggering the random reassignment worker too frequently, you could always rate limit it: https://github.com/parrish/sidekiq-congestion

You'd probably want to set track_rejected to false and reject_with to reschedule.

parrish:
You need to change something to get anything approaching uniformly random.

You could choose a random starting point
``` ruby
# no guarantee you'll get the number you're fetching
direction = rand < 0.5 ? '>' : '<'
where("random #{ direction } ?", rand) # assuming a 0-1 range
```

or you could limit/offset it

``` ruby
# at least attempts to get the number you're fetching
amount = 2_000 # or however many you're fetching
ceiling = [(rand * 50_000).ceil, count - amount].min
offset(ceiling).limit(amount)
```

chrissnyder:
Playing devils advocate, is there any pressing concern that this would solve that bumping the db instance type wouldn't for the few weeks? There isn't a project launching in the near future, so we aren't going to get any massive spike either.

Feels like subject selection is such a core part of the API that we really need to take the time to get it right, rather than applying additional band aids.

camallen:
@chrissnyder not really, it'll be cheaper and it'd be nice not to have the db being worked if / when we scale back. I do think we have the tate re-email to go out soon but that won't use random selection.

camallen:
I've tested this strategy using 2 million set member subjects (2 x workflows, 2 x subject sets with ~50% assignment of the data into the selecting workflow) as well. However some note:
1. index updates are executed in line
2. i'm only updating the random value of selected set_member_subjects not the whole focus window so this PR needs an update if that's acceptable.

This won't reflect heavy concurrent usage of the index update so we need a uniq jobs / rate limiter in sidekiq to ensure we don't stress the index updates with lots of selections. This strategy shows fairly constant execution times but will impact performance on index updates (as you can see), i'd say that since we background this job most selection queries will be pretty quick. Code [here](https://github.com/camallen/Panoptes/commit/46c8461453a7576870717e31543d21b8bddddc7f)

image

edpaget:
It might be worth running your benchmarks on postgres running in a docker container with memory limits set to mirror a m3.medium instance [see here](https://docs.docker.com/reference/run/#runtime-constraints-on-cpu-and-memory)

```
docker run -m 3.75G -c 1 postgresql
```

camallen:
Update, I used a memory / cpu limited pg docker container `-m 3750M --cpu-shares 2` (it wouldn't start with 1 cpu limit).

response times were pretty similar to previous readings for both selection strategies, the main difference was the resulting random distribution of selected sms ids. Not surprisingly the random offset had better distribution by default but still suffers from poor performance approaching 1M rows (where we are now). The focus set selection maintained pretty consistent performance, the only variation here was due to the window select size we pluck from. It doesn't spread selection as much but rather creates a focus set of data that is a function of the activity on the project and the selection space size. For instance a 5K selection window over 250K subjects is 12.5K subjects being in the focus set. Retirement will modify this by reducing the selection space as will more people requesting data will result in greater spread via more frequent index updates and the smaller the selection space the greater the spread we get.

In summary i think we should go with this strategy over the offset selection. I suggest running at the 5K selection window for fast queries (~40ms in my benchmarking but not sure what we'll see in prod db / network latency, etc). 

I'm going to update this PR with a YAML configuration directive for this window size and only index update the selected sms ids. Also we'll need to rate limit the index updates so i guess this should come in too, though i'd like to get this out there sooner than later (next week now). 

Any thoughts / objects appreciated.

image

edpaget:
This sounds fine to me.

camallen:
cool - i'll finish this off over the weekend / first thing next week.

camallen:
I've built the prod / staging config file and will push it once this has been approved.

edpaget:
This looks good to :ship: to me.

PR #1360, opened 2015-09-11
use offset strategy to select random data

camallen:
Related to https://github.com/zooniverse/Panoptes/pull/1358#issuecomment-139301488 - this technique should randomly offset into the available data, select a window over the data and randomly sample into that window.

### DIFF
--- a/lib/postgresql_selection.rb
+++ b/lib/postgresql_selection.rb
@@ -31,12 +31,14 @@ class PostgresqlSelection
   end
 
   def available_count
-    available.except(:select).count
+    @available_count ||= available.except(:select).count
   end
 
   def sample(query=available)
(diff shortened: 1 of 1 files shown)


camallen:
so this strategy seems to scale linearly but i'd say becomes too slow at larger offsets with larger table spaces. I tested synthetic data up to 2 million set member subjects (2xworkflows, 2x subject sets with ~50% assignment of the data into the selecting workflow - so probably not super realistic of the way we select data).

image

camallen:
see, https://github.com/zooniverse/Panoptes/pull/1358#issuecomment-139619412 for details on the selection distribution vs timing benchmarks.

camallen:
closing in favour of #1358

PR #1362, opened 2015-09-14
Dequeue non logged in on classification receipt not subject selection

camallen:
I was looking into rate limiting the non-logged in queue rebuilds and thought instead that it should not dequeue as often. As we don't guarantee that  non-logged in won't see duplicates then we should only dequeue from the non-logged in on classifications receipt.

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -17,20 +17,19 @@ class ClassificationLifecycle
   end
 
   def transact!(&block)
+    dequeue_for_non_logged_in
     Classification.transaction do
       update_classification_data
-
       #NOTE: ensure the block is evaluated before updating the seen subjects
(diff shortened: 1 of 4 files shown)


lib/classification_lifecycle.rb
```diff
@@ -17,20 +17,19 @@ def queue(action)
...
-
       create_recent
       update_seen_subjects
-      #to avoid duplicates in queue, do not refresh the queue before updating seen subjects
-      refresh_queue
       publish_to_kafka
       save_to_cassandra unless Rails.env == 'production'
     end
+    #to avoid duplicates in queue, do not refresh the queue before updating seen subjects
+    refresh_queue
```
edpaget: This should still be within the transaction right? Otherwise it'll update the queue without the seen subjects?

camallen: I figured this wouldn't be called if there was an error in execution above.

PR #1369, opened 2015-09-16
serialize workflows retirement defaults

### LINKED ISSUE #1361: get default number of classifications for retirement
I want to be able to get the default number of classifications for a subject to be retired.  Currently, if the default is used, I get u'retirement': {}.

camallen:
closes #1361


marten:
Not sure what's up with the spec failure, but don't see how that would be caused by this.

PR #1370, opened 2015-09-16
Rate limit enqueue subjects worker

camallen:
specifically the non-logged in queue, allow only 1 every 10 seconds by default. i have built and setup the config file for this.

### DIFF
--- a/config/initializers/subject_enqueue_config.rb
+++ b/config/initializers/subject_enqueue_config.rb
@@ -1,0 +1,18 @@
+module Panoptes
+  module SubjectEnqueue
+    def self.config
+      @config ||= begin
+                    file = Rails.root.join('config/subject_enqueue_config.yml')
+                    YAML.load(File.read(file))[Rails.env].symbolize_keys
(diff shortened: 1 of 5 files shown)


camallen:
@marten can you take a quick look?

config/initializers/subject_enqueue_config.rb
```diff
@@ -0,0 +1,18 @@
...
+      @config ||= begin
+                    file = Rails.root.join('config/subject_enqueue_config.yml')
+                    YAML.load(File.read(file))[Rails.env].symbolize_keys
+                  rescue Errno::ENOENT, NoMethodError
+                    {  }
+                  end
+    end
+
+    def self.congestion_opts
+      config[:congestion_opts].symbolize_keys
```
marten: Rescuing and defaulting to an empty hash in `config` won't help much if you end up calling symbolize_keys on nil here.

camallen: good point, will fix

marten:
LGTM otherwise

PR #1376, opened 2015-09-17
rate limit data dump workers

camallen:
set to default limit of 1 per day configurable via the yaml file and customizable via a direct overide in the worker class


marten:
Should we have any kind of UI for this to let people know they can't request a new one?

camallen:
probably some text changes in the front end.

PR #1382, opened 2015-09-23
[WIP] config to disable classification count retirmenent

### LINKED ISSUE #1368: disable counter retirement for live projects with custom retirement
Linked to https://github.com/zooniverse/Ancient-Lives/issues/18#issuecomment-140395139

Projects like annotate and the ancient lives rebuild will use custom retirement instead of the counter based ones. We should support counters on these live projects but provide a mechanism to disable the retirement from happening via these raw counters.

@ggdhines - re annotate is there any reason to have both custom retirement via aggregation engine and a counter retirement set to a higher number? I.e. retire from aggregation if conditions X otherwise retire from counts if they reach 15?

camallen:
closes #1368 - allow workflow's with custom retirement (aggregation engine) to disable the count retirement worker.

### DIFF
--- a/lib/retirement_schemes/classification_count.rb
+++ b/lib/retirement_schemes/classification_count.rb
@@ -4,8 +4,16 @@ module RetirementSchemes
       @count = count
     end
 
-    def retire?(count)
-      count.classifications_count >= @count
+    def retire?(sw_count)
+      if disabled?
+        false
+      else
+        sw_count.classifications_count >= @count
+      end
+    end
+
+    def disabled?
+      !!@count.to_s.match(/\Adisabled\z/i)
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/retirement_schemes/classification_count.rb
```diff
@@ -4,8 +4,16 @@ def initialize(count)
       @count = count
     end
 
-    def retire?(count)
-      count.classifications_count >= @count
+    def retire?(sw_count)
+      if disabled?
```
marten: If you invert this to `enabled?`, this method body could just be `enabled? && sw_count.classifications_count >= @count` which I think reads better.

marten:
It might be cleaner to just have a different retirement scheme that does absolutely nothing. But I think that will require adapting PFE to support selecting between multiple schemes. Mental note to refactor this into a Never retirement scheme whenever we pick that up?

camallen:
i'll have a look at a refactor to use a disabled retirement scheme.

marten:
LGTM if you want to merge this in the meantime btw.

marten:
Going to steal this WIP

camallen:
thanks :+1:

camallen:
closing in favour of #1528

PR #1388, opened 2015-09-28
send email when aggregation update is finished

camallen:
notify the requester / project owner when the aggregation export has finished uploading

### DIFF
--- a/app/controllers/api/v1/media_controller.rb
+++ b/app/controllers/api/v1/media_controller.rb
@@ -34,8 +34,8 @@ class Api::V1::MediaController < Api::ApiController
   def update
     error_unless_exists
     set_controlled_resources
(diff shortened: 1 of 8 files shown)


edpaget:
LGTM

app/controllers/api/v1/media_controller.rb
```diff
@@ -162,4 +162,11 @@ def create_conditions
     .scope
     check_controller_resources
   end
+
+  def send_aggregation_ready_email
+    finished = params[:media][:metadata].try(:[], "state") == "finished"
```
edpaget: Hey know you merged this, but this need to be changed to "ready"

camallen: arghh

PR #1392, opened 2015-09-29
do not rate limit the mailer worker

camallen:
seperate the rate limiting for dump workers into a composable module.

### DIFF
--- a/app/workers/aggregation_data_mailer_worker.rb
+++ b/app/workers/aggregation_data_mailer_worker.rb
@@ -1,5 +1,6 @@
 class AggregationDataMailerWorker
-  include DumpWorker
+  include Sidekiq::Worker
+  include DumpMailerWorker
 
   attr_reader :medium
 
--- a/app/workers/aggregation_data_mailer_worker.rb
+++ b/app/workers/aggregation_data_mailer_worker.rb
@@ -1,5 +1,4 @@
 class AggregationDataMailerWorker
-  include Sidekiq::Worker
   include DumpWorker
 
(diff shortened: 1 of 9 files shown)


app/workers/aggregation_data_mailer_worker.rb
```diff
@@ -1,5 +1,4 @@
 class AggregationDataMailerWorker
-  include Sidekiq::Worker
   include DumpWorker
```
edpaget: Why is the mailer including the DumpWorker? Couldn't we have a DumpMailerWorker module instead?

camallen: Yep - i did it this was as the `DumpWorker` is a dumping ground of responsibilities. I'll refactor and cleanup.

app/workers/classifications_dump_worker.rb
```diff
@@ -1,7 +1,10 @@
 require 'csv'
 
 class ClassificationsDumpWorker
+  include Sidekiq::Worker
   include DumpWorker
+  include DumpMailerWorker
```
edpaget: Sorry I still don't get why this would be included in the non-Mailer Workers. Shouldn't the `DumpWorker` module contain all the necessary functionality for a data dump worker? And the `DumpMailerWorker` module should only have common functionality for the data dump mailer workers?

camallen: normal dump workers (e.g. classification) sends emails, so ive extracted the dump email functionality to a separate module for use in the dump mailer worker.

PR #1393, opened 2015-09-29
Delay the subject_set counter action

### LINKED ISSUE #1383: Subject Create route is super Slow
Ping @saschaishikawa for more information, but he was trying to create subjects for the whaling project and brought down to the API each time.

edpaget:
Closes #1383 

@camallen let me know if I'm using the congestion control gem correctly here


app/workers/subject_set_subject_counter_worker.rb
```diff
@@ -0,0 +1,19 @@
+class SubjectSetSubjectCounterWorker
+  include Sidekiq::Worker
+
+  sidekiq_options congestion: {
+    interval: 1.hour,
+    max_in_interval: 10,
+    min_delay: 3.minutes,
+    reject_with: :cancel,
```
marten: We might want :reschedule? Not entirely sure, which is why a test for expected behaviour might be nice even though it's the gem's job to make it so. I think what we want is for each enqueue to push the job ahead? At the very least, we want to make sure that a job always occurs after the last model was created.

edpaget: This job is called with a `perform_in(3.minutes, ...)` from the SetMemberSubject's after create callback, so it should only be invoked after the last SMS is created.

marten: I meant a test just to see that the reschedule/cancel works like we'd want in case of congestion.

edpaget:
@camallen, @marten I'm totally lost on the last two test failures

```
1) Api::V1::ProjectsController#destroy behaves like is destructable an authorized user with proper scopes should return 204

     Failure/Error: expect(response).to have_http_status(:no_content)

       expected the response to have status code :no_content (204) but it was :conflict (409)

     Shared Example Group: "is destructable" called from ./spec/controllers/api/v1/projects_controller_spec.rb:856

     # ./spec/support/destructable.rb:15:in `block (4 levels) in <top (required)>'

  2) Api::V1::ProjectsController#destroy behaves like is destructable an authorized user with proper scopes should delete the resource

     Failure/Error: expect{resource_class.find(resource.id)}.to raise_error(ActiveRecord::RecordNotFound)

       expected ActiveRecord::RecordNotFound but nothing was raised

     Shared Example Group: "is destructable" called from ./spec/controllers/api/v1/projects_controller_spec.rb:856

     # ./spec/support/destructable.rb:19:in `block (4 levels) in <top (required)>'
```

This is caused by the set member subjects being deleted which is triggering an update of the `set_member_subjects_count` to subtract 2 from the count and increment the lock_version column on on of the SubjectSets. Trying to delete the just updated SubjectSet causes an ActiveRecord::StaleObjectError, which is causing the above errors. 

I have no idea where that update statement is being issued from, do you have any clue what could be causing it?

camallen:
Turns out rails is using doing counter caching even though the setting is turned off on the relation. Change the column name to something different (`set_member_subjects_total_count`) and it doesn't run the update.

marten:
Did you try setting it to explicit `false` (not just removing it) on *both* sides of the relation?

camallen:
I got the same conflict with `belongs_to :subject_set, counter_cache: false`.

marten:
Rails-- :disappointed:

edpaget:
This is super stupid! I tried the `counter_cache: false`, earlier. I think this must be a Rails bug, no way this is intended behaviour.

edpaget:
Turns out this is fixed in Rails 4.2.4!

camallen:
sweet

camallen:
@edpaget - congestion setup looks ok. I've been removing these to config files so we don't have to modify code to change the settings, just rebuild / deploy, happy for this to go as is though.

config/congestion_control_config.yml.hudson
```diff
@@ -0,0 +1,25 @@
...
+  congestion_opts:
+    interval: 86400
+    max_in_interval: 1
+    min_delay: 43200
+    reject_with: !ruby/symbol cancel
+  
+counter_default: &counter_default
+  congestion_opts:
+    interval: 360
+    max_in_interval: 1o
```
camallen: 10?

spec/models/subject_set_spec.rb
```diff
@@ -70,7 +70,7 @@
       end
     end
 
-    describe "#set_member_subjects_count" do
+    describe "#set_member_subjects_count", :disabled do
```
camallen: does this still need to be disabled with the `after(:create)` block in the SMS factory?

edpaget: Yes. The counter won't run right away (or at all since its in a test) so this won't be consistent immediately.

camallen:
@edpaget rebase?

edpaget:
Can we get rid of this merge protection? There's not really a reason to do it since travis-ci merges master into your PR when it's tested if that's possible. 

See here:
screen shot 2015-10-05 at 8 58 47 am

marten:
@edpaget I'm all for it.

edpaget:
Yeah we could always just restart the CI if that's what we were concerned about.

edpaget:
I mostly annoys me that `Update branch` doesn't use a rebase.

config/initializers/congestion_control_config.rb
```diff
@@ -0,0 +1,38 @@
...
+            CongestionConfig.new(config)
+          end
+        end
+      end
+    end
+
+    def self.load_config
+      @config ||= begin
+                    file = Rails.root.join('config/congestion_control_config.yml')
+                    p YAML.load(File.read(file))
```
camallen: debug statement

config/initializers/congestion_control_config.rb
```diff
@@ -0,0 +1,38 @@
...
+          define_method worker do
+            CongestionConfig.new(config)
+          end
+        end
+      end
+    end
+
+    def self.load_config
+      @config ||= begin
+                    file = Rails.root.join('config/congestion_control_config.yml')
```
camallen: does this exist for deploy yet?

edpaget: Nope I'll get it up though.

config/congestion_control_config.yml.hudson
camallen: bump^

config/initializers/congestion_control_config.rb
edpaget: It's online now.

camallen:
sweet LGTM

PR #1400, opened 2015-10-01
Fix migrate task for SWC

marten:


camallen:
LGTM - what's up with travis?

marten:
The classic quick-fix screwup where I changed to 1.9-style hashes but didn't remove the initial `:` in front the task symbol.

PR #1404, opened 2015-10-03
Remove SWC backwards compatibility

marten:


camallen:
looks good, just looks like this `SubjectWorkflowCount::BACKWARDS_COMPAT` is still used but not defined.

marten:
Ah, more appeared when I rebased onto master. But that's why I used a constant instead of a `TODO` comment :)

PR #1406, opened 2015-10-05
workflow display_name must be present

camallen:
closes https://github.com/zooniverse/Panoptes-Front-End/issues/1696


edpaget:
:shipit:

PR #1407, opened 2015-10-05
use json schema for user project prefs

### LINKED ISSUE #1401: project_preferences tutorial_completed_at field
Could we add a `tutorial_completed_at` field?

I'd like to store it as a date so we can tell the user if a tutorial has changed since they last went finished it.

Defaults to `null`.

camallen:
closes #1401, allow front ends to store free form data in the preferences object.


edpaget:
:shipit:

PR #1410, opened 2015-10-06
classification heartbeat check

camallen:
scheduled job to check classifications are being received within a window period and alert via email / honeybadger if not


camallen:
i've added a config file for this. i plan to test on staging once this PR goes through.

PR #1412, opened 2015-10-06
use 15 min schedules to check for missing data

camallen:
avoid heavy cpu load via ice cube, https://github.com/tobiassvn/sidetiq/wiki/Known-Issues


edpaget:
I'm fine to merge this whenever it passes.

edpaget:
Also ruby whoo! :dancer:

PR #1414, opened 2015-10-07
Add tutorial resource

### LINKED ISSUE #1399: Create a tutorial_steps resource
Projects are getting simple Chimp&See-style popup tutorials.

We'll need a tutorial_steps resource, each with its own `media` and a `content`.

Media will live in the project's attached media, so maybe that should be a link?

Content'll just be Markdown.

They should be reorderable. Do linked resources maintain their order when it changes? Otherwise we could have an explicit `index` field or something.

edpaget:
Closes #1399

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -13,6 +13,7 @@ class Project < ActiveRecord::Base
 
   EXPERT_ROLES = [:expert, :owner]
 
+  belongs_to :default_tutorial, class_name: "Tutorial"
   has_many :workflows
   has_many :subject_sets, dependent: :destroy
   has_many :classifications
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -174,6 +174,7 @@ class Api::V1::ProjectsController < Api::ApiController
   end
 
   def new_items(resource, relation, value)
+    return super if relation == :default_tutorial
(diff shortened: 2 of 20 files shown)


app/models/project.rb
```diff
@@ -13,6 +13,7 @@ class Project < ActiveRecord::Base
 
   EXPERT_ROLES = [:expert, :owner]
 
+  belongs_to :default_tutorial, class_name: "Tutorial"
```
camallen: should this association be the other way around? `Project has_one :default_tutorial` and a `Tutorial belongs_to :project`?

edpaget: Why? The key is on the project table.

camallen: i thought it would clean up the inverse relation FK specification.

edpaget: Sure if you'd prefer it the other way I'll switch it around. It made more sense to me that the project had the default defined on it, but I can see the other way too.

camallen: i really don't mind. just asked the question. you decide

edpaget: Up to you man. You're going to be maintaining it. I thought I answered the question in my post above, that I think it makes sense that the project holds the reference to its default tutorial, but I could see it either way.

camallen: leave it

camallen:
This LGTM, you happy @edpaget?

edpaget:
Yup

PR #1423, opened 2015-10-12
Optimize exports

marten:
Most importantly I've added a cache for the reified old versions of workflows and contents. This seemed to be amazingly slow even if dealing with the latest version. I'm also preloading subjects and handling the check for retiredness less stupidly.


camallen:
LGTM - rebase?

PR #1427, opened 2015-10-12
Use tsvector instead of trgrm

### LINKED ISSUE #1230: Search has difficulty with short user names
As @edpaget discovered, short user names like mine (DZM) don't work with the search algorithm. Opening an issue to work on that.

### LINKED ISSUE #1296: Searching for users with just @ rack timeout
Getting rack timeout errors on https://panoptes.zooniverse.org/api/users?search=@&page_size=10

We should probably only look for users with a character after the @ symbol.

edpaget:
Closes #1296 
Closes #1230


edpaget:
Note the migration that creates the tsvector columns took about 30 minutes to run on my machine against a copy of the production database. I'm also not sure if the change to a sql dump will require the postgresql-client-9.4 package to be installed on the production Docker container as well.

config/database.yml.hudson
```diff
@@ -2,6 +2,7 @@ default: &default
   adapter: postgresql
   encoding: unicode
   user: panoptes
+  username: panoptes
```
edpaget: This took an infuriatingly long time to find.

config/application.rb
```diff
@@ -21,6 +21,8 @@ class Application < Rails::Application
     config.action_controller.action_on_unpermitted_parameters = :raise
     config.middleware.insert_before ActionDispatch::ParamsParser, "CatchApiJsonParseErrors"
 
+    config.active_record.schema_format = :sql
```
camallen: do we have to use sql dump format?

config/database.yml.hudson
camallen: do we need both user and username?

db/migrate/20151012162248_change_user_search_index.rb
```diff
@@ -0,0 +1,29 @@
+class ChangeUserSearchIndex < ActiveRecord::Migration
+  def up
+#    remove_index :users, name: "users_idx_trgm_login_display_name"
```
camallen: comments?

db/migrate/20151012162248_change_user_search_index.rb
```diff
@@ -0,0 +1,29 @@
...
+        tsv, 'pg_catalog.english', login, display_name
+      );
+    SQL
+
+    now = Time.current.to_s(:db)
+    update("UPDATE users SET updated_at = '#{now}'")
+  end
+
+  def down
+  #  execute <<-SQL
```
camallen: comments? we should probably have a reversal here.

config/application.rb
edpaget: Yes. we have to setup triggers to create the tsvector columns. Also I'm not sure why we were avoiding it other than aesthetic reasons? Our code isn't portable to any other databases.

db/migrate/20151012162248_change_user_search_index.rb
edpaget: Thanks had to comment this out to fix a local broken migration. It's removed now.

config/database.yml.hudson
edpaget: Yes. This is a stupid thing in rails (though from what I can tell it's fixed in 5.0). It dumps the sql format using the postgres command line tools, and sets variables for the database connection through the PG* environment vars. For some reason it sets `PGUSER` from a `username` field instead of the `user` field which it uses for connections otherwise.

camallen:
@marten any thoughts on this?

config/application.rb
camallen: I do use the schema doc a bit but I'm fine with using the sql dump format.

camallen:
LGTM , @edpaget can you rebase? we should try and get this migrated in some low period.

edpaget:
Yup rebased. I'm still not sure if we need to add the postgresql-client stuff to the prod dockerfile, so I guess we can see if the staging deploy breaks and add it if so?

camallen:
that's what i was thinking.

PR #1428, opened 2015-10-13
correctly handle aggregation export media updates

camallen:
aggregation engine uses update url's like `https://panoptes.zooniverse.org/api/projects/1154/aggregations_export` that don't have a media resource id. this PR correctly handles the media resource lookup for sending the aggregation export success email.

### DIFF
--- a/app/controllers/api/v1/media_controller.rb
+++ b/app/controllers/api/v1/media_controller.rb
@@ -34,8 +34,8 @@ class Api::V1::MediaController < Api::ApiController
   def update
     error_unless_exists
     set_controlled_resources
-    send_aggregation_ready_email
     super
+    send_aggregation_ready_email
   end
 
   def destroy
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/media_controller.rb
```diff
@@ -34,8 +34,8 @@ def show
   def update
     error_unless_exists
     set_controlled_resources
-    send_aggregation_ready_email
     super
+    send_aggregation_ready_email
```
edpaget: Won't this not get fired due the `render` in the super method?

marten: Render doesn't stop further execution (except in before filters). Just sets the value to be rendered when it's Rails' turn again.

camallen: should be getting called, the spec tests that the mailer gets called with the correct media resource id. I wanted to add it after to ensure the .save! is called on the update action so failed updates won't trigger this.

PR #1440, opened 2015-10-15
Subject importer

marten:
This is based on the script used to load up Wildcam, but with tests and with ids as parameters. This way we can just kick off an import from Jenkins.

### DIFF
--- a/app/workers/subject_import_worker.rb
+++ b/app/workers/subject_import_worker.rb
@@ -1,0 +1,84 @@
+require 'csv'
+
+class SubjectImportWorker
+  include Sidekiq::Worker
+
+  def perform(project_id, user_id, subject_set_id, csv_url)
+    project = Project.find(project_id)
+    zoo_user = User.find(user_id)
+    subject_set = project.subject_sets.find(subject_set_id)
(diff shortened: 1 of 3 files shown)


edpaget:
LGTM

app/workers/subject_import_worker.rb
```diff
@@ -0,0 +1,84 @@
+require 'csv'
+
+class SubjectImportWorker
+  include Sidekiq::Worker
```
camallen: should we push this to a data queue, `sidekiq_options queue: :data_high`

camallen:
LGTM too

app/workers/subject_import_worker.rb
marten: Are the queues live yet? Or will the normal worker pick them up for now?

camallen: not yet - adam's setting them up and no normal worker won't pick them up. so up to you about specifying the queue.

marten: Merging it now, and telling slackbot to remind me to move the queue next week

camallen: cool

PR #1446, opened 2015-10-20
explicit pref belongs_to options

camallen:
don't dependant destroy a project when deleting a user project preference, allow the relation to have the correct options passed via the preferences_for method.


camallen:
@marten should be good to go now..can you please review?

PR #1451, opened 2015-10-23
ensure collections default to private

camallen:
linked to https://github.com/zooniverse/Panoptes-Front-End/pull/1822 ensure collections in panoptes default to explicit private setting.

This migration took 2 secs to run on my machine so should be good to go out as is.

### DIFF
--- a/db/migrate/20151023103228_add_default_to_collection_private.rb
+++ b/db/migrate/20151023103228_add_default_to_collection_private.rb
@@ -1,0 +1,8 @@
+class AddDefaultToCollectionPrivate < ActiveRecord::Migration
+  def change
+    private_setting = true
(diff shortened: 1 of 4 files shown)


db/migrate/20151023103228_add_default_to_collection_private.rb
```diff
@@ -0,0 +1,7 @@
+class AddDefaultToCollectionPrivate < ActiveRecord::Migration
+  def change
+    change_column :collections, :private, :boolean, default: false
```
marten: Shouldn't this be true if you want them to default to private?

camallen: whoops, yep.

PR #1456, opened 2015-10-26
[WIP] Project tracking via links

camallen:
**DON'T MERGE THIS YET** this will break talk collection searching so we need to get a plan for before rolling this out as well.

`#destroy_links` in `updatable_resources` now also saves the updated resource to allow the collection `project_ids` to be updated.

### DIFF
--- a/app/controllers/api/v1/collections_controller.rb
+++ b/app/controllers/api/v1/collections_controller.rb
@@ -28,8 +28,7 @@ class Api::V1::CollectionsController < Api::ApiController
 
   def add_relation(resource, relation, value)
     if relation == :subjects && value.is_a?(Array)
-      added_project_ids = Subject.unscoped.uniq.where(id: value).pluck(:project_id)
-      resource.project_ids = resource.project_ids | added_project_ids
+      resource.project_ids = update_project_ids(resource, value)
     end
     super
   end
@@ -45,9 +44,12 @@ class Api::V1::CollectionsController < Api::ApiController
 
   private
 
+  def params_project_ids
+    @params_project_ids ||= (params.delete(:project_ids) || params.delete(:project_id)).try(:split, ',')
+  end
+
   def filter_by_project_ids
-    if ids_string = (params.delete(:project_ids) || params.delete(:project_id)).try(:split, ',')
-      project_ids = ids_string.split(",")
+    if project_ids = params_project_ids
       @controlled_resources = controlled_resources.where.overlap(project_ids: project_ids)
     end
   end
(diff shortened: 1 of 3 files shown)


db/migrate/20151024080849_collections_belongs_to_many_projects.rb
```diff
@@ -0,0 +1,8 @@
+class CollectionsBelongsToManyProjects < ActiveRecord::Migration
+  def change
+    add_column :collections, :project_ids, :integer, array: true, default: []
+    Collection.where.not(project_id: nil).update_all("project_ids = ARRAY[project_id]")
+    add_index  :collections, :project_ids, using: 'gin'
```
marten: Weren't there issues with this? I thought we didn't have this index on `classifications.subject_ids` for some specific reason?

camallen: yeah db index updates under heavy load (i.e. stargazing) would hit the db cpu really hard. This is on collections so will not have as many concurrent updates to deal with. If it's a problem we can drop it and figure out queries later.

app/controllers/api/v1/collections_controller.rb
```diff
@@ -22,5 +22,21 @@ def build_resource_for_create(create_params)
     add_user_as_linked_owner(create_params)
     super(create_params)
   end
-end
 
+  def add_relation(resource, relation, value)
```
camallen: @marten, @parrish - I'm not 100% sold on the way i'm doing this as it implies we track the originating project link and not necessarily the project context the subject was collected in, e.g. subject uploaded to project A but reused in project B. 

It could instead default to linked project tracking if no project_id param is supplied? thus allowing the context to be passed in via link params.

camallen:
I'm not 100% sold on the way i'm doing this as it implies we track the originating project link and not necessarily the project context the subject was collected in, e.g. subject uploaded to project A but reused in project B.

It could instead default to linked project tracking if no project_id param is supplied? thus allowing the context to be passed in via link params.

db/migrate/20151024080849_collections_belongs_to_many_projects.rb
marten: deal.

parrish:
I'm on the fence here.  It almost makes sense for this behavior to be a fallback if the client doesn't specify a project id in the params when adding the subject.

I'll see about setting Talk up for multiple project ids as well.

camallen:
I'm going to rework this to add a project_id param and use the current as a fallback default.

marten:
What's the status of this PR?

camallen:
super stale, I'll reopen it later when i get back to this.

PR #1459, opened 2015-10-29
expose csrf token headers

camallen:
ui app needs this header exposed for a dev server


marten:
I'm not entirely sure what this does, and how it affects security. @zooniverse/developers Anyone else with a clue who can review this?

camallen:
used when copying config files for sensible defaults when smoke testing the application. Basically the sign out route uses the CSRF token from the response headers, if they aren't exposed then the response object won't be able to access them.

PR #1466, opened 2015-11-02
update aws adapater s3 signed urls

camallen:
use correct default expiry times and convert to use the expires param in seconds to expire.

### DIFF
--- a/lib/media_storage/aws_adapter.rb
+++ b/lib/media_storage/aws_adapter.rb
@@ -40,7 +40,7 @@ module MediaStorage
     end
 
     def get_path(path, opts={})
-      expires = (opts[:get_expires] || @get_expiration).minutes.from_now
+      expires = expires_in(opts[:get_expires] || @get_expiration)
       if opts[:private]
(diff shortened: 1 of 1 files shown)


lib/media_storage/aws_adapter.rb
```diff
@@ -9,8 +9,8 @@ class AwsAdapter < AbstractAdapter
     def initialize(opts={})
       @prefix = opts[:prefix] || Rails.env
       @bucket = opts[:bucket]
-      @get_expiration = opts.fetch(:expiration, {})[:get] || DEFAULT_PUT_EXPIRATION
-      @put_expiration = opts.fetch(:expiration, {})[:put] || DEFAULT_GET_EXPIRATION
+      @get_expiration = opts.fetch(:expiration, {})[:get] || DEFAULT_GET_EXPIRATION
+      @put_expiration = opts.fetch(:expiration, {})[:put] || DEFAULT_PUT_EXPIRATION
```
marten: Ha!

PR #1467, opened 2015-11-02
Associate classification and subjects through join

marten:


db/migrate/20151027134345_convert_classification_subject_ids_to_join_table.rb
```diff
@@ -0,0 +1,12 @@
+class ConvertClassificationSubjectIdsToJoinTable < ActiveRecord::Migration
+  def change
+    create_table :classification_subjects, id: false do |t|
+      t.integer :classification_id, null: false
+      t.integer :subject_id, null: false
+    end
+
+    add_index :classification_subjects, [:classification_id, :subject_id], unique: true, name: 'classification_subjects_pk'
```
camallen: we will want to search just on `subject_id` in this table?

marten: I guess we should have indexes on both columns seperately, for lookups of all subjects for a classification and vice versa. Good catch.

spec/factories/classifications.rb
```diff
@@ -21,13 +17,7 @@
...
     workflow
     subject_ids do
-      if build_real_subjects
-        #TODO: clean this up -> this cascading create_list always builds:
-        # 7 Projects, 3 Workflows, 2 SubjectSets, 2 Subjects
-        create_list(:set_member_subject, 2).map(&:subject).map(&:id)
-      else
-        (1..10).to_a.sample(2)
-      end
+      create_list(:set_member_subject, 2).map(&:subject).map(&:id)
```
camallen: will this run every time we build the factory.. it creates a very large number of models that aren't always needed, hence i added a transient feature.

camallen:
Looks good, how long does the data migration run?

spec/factories/classifications.rb
marten: There were only a few that actually used that transient, and I changed those into `build_stubbed` so they wouldn't make things slower. We still want to clean those associations up though, for one the subjects aren't associated with the same project as the classification, which is weird.

camallen: cool

marten:
About 19 seconds with the temporary index I just added.

camallen:
looks like that cassandra merge conflicts..rebase?

marten:
Done.

camallen:
pinging @ggdhines, we're modifying the database schema for classification -> subjects relations. Once this is deployed (stag first then prod) you'll have to lookup the join table to find the links. This PR has details.

marten:
and @ggdhines feel free to pull me aside if you need help with the sql

marten:
Shall we wait on merging this until the deploy for collections, so we don't do too many things at once?

camallen:
Up to you, should we plan for data migrations tomorrow?

camallen:
getting this into staging for testing with the aggregation engine

ggdhines:
@marten 's already helped me with one query. The following is to get the set of all subjects which have been retired for a given workflow id and since a certain date:
SELECT * FROM "subjects" INNER JOIN "subject_workflow_counts" ON "subject_workflow_counts"."subject_id" = "subjects"."id"
WHERE "subject_workflow_counts"."workflow_id" = 9 AND "subject_workflow_counts"."retired_at" >= '2000-01-01 00:00:00'

@camallen  @marten anything I need to change for this one?

marten:
No, that's unchanged.

PR #1469, opened 2015-11-03
Remove cassandra

### LINKED ISSUE #1458: Remove cassandra

marten:
Fixes #1458


camallen:
looks like the db create step in travis failed...https://github.com/zooniverse/Panoptes/blob/1a5a5bd12117773f89ca0c91cdc4ad827c28af8d/.travis.yml#L8

Should we use the sql schema for creating the db now?

marten:
Nah, the silver searcher just didn't look into yml files so I missed the cequel reference there.

camallen:
Want to remove the cassandra service addon for travis too?

Also i think we should just use `RACK_ENV=test bundle exec rake configure:travis db:create db:structure:load` instead of running the migrations now we have a sql dump that should work.

marten:
Good idea, will make sure the structure file keeps working too.

PR #1471, opened 2015-11-05
Take new rows without subject_ids into account

marten:
Otherwise we would be trying to insert null values and failing the
constraints.


camallen:
LGTM

Issue #1398, opened 2015-09-30
Way to have classification counts and user participation counts to start when the project launches and not during beta and building


mschwamb:
I'm building a new project and we're adding in images to check the background and play with the subject presentation (since it's a combination of several images we're uploading as one). Currently when we click we can see the classification counts going up on the project front page, but this isn't going to be reflective of the actual project stats when we launch. On custom projects I've been involved in the classification database was typically zeroed out right before launch. Is there a way to mark prelaunch classifications and not count them towards the project status counters ?

camallen:
I think https://github.com/zooniverse/Panoptes-Front-End/pull/1673 will solve this issue in the future. We currently count all classifications, pretty sure these counts will disappear as noise after launching.

mschwamb:
It won't fix the whole issue because it won't record the annotations. We're also testing our design tree and subject design with the team and looking at the data afterwards. We'll be iterating a bunch of times arguing over image contrast and scaling between the different components. I agree eventually it will be noise, but when you add in the beta test it will take some time to become noise. 

Feel free to label as won't fix and close if you disagree.  thought it would be a back end issue and didn't look at the front end.  Looks like it something that others want the same feature  #273

camallen:
Is it really just that you are thinking about old projects launching from scratch where we hid the beta classification work / counts and "started" with a clean slate (after lots of internal iteration)?

Happy to leave this open as a discussion. In my opinion Panoptes doesn't use this paradigm as a project evolves over time and hopefully reaches launch status. Where is the distinction to what classifications we count, perhaps only live state classifications?

mschwamb:
well, yes, but it's not really an old way. I'm finding that bit stayed the same with the project builder, it's just that there's no telecon with the developer it's happening in the meeting with the science team and one of us is opening the project builder website. 

 At least for my two project builder projects Terrains and the one  I'm currently building that has been my experience.  The research team for my new project is debtating over whether we should have one image bigger with the reference stars smaller  whether we need four color contrast levels or just the grey scale image and then a inverted black --> white image. So the only difference I'm finding is the iteration is the same or a bit more, because we can do it quickly and make the entire team test it and check the result.

Also I still find the best way to get help examples is to run through 100s of images in the interface to find the examples. The ticket you pointed to does solve those clicks racking up since it won't record those.

[...]

PR #1472, opened 2015-11-06
Add project launch date

marten:
TODO:

* [x] increment counters manually where counter cache is removed
* [x] update/recount workflow.classifications_count

### DIFF
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -34,7 +34,10 @@ module Api
       if event_params_check
         if upp = user_project_preference
           upp.legacy_count[create_params[:workflow]] = create_params[:count]
-          response_status = :ok if upp.save
+          if upp.save
+            Project.increment_counter :classifiers_count, upp.project_id
+            response_status = :ok
+          end
         end
       end
       render status: response_status, nothing: true
--- a/app/models/classification.rb
+++ b/app/models/classification.rb
@@ -23,6 +23,16 @@ class Classification < ActiveRecord::Base
   scope :created_by, -> (user) { where(user: user) }
   scope :complete, -> { where(completed: true) }
 
+  after_create do
+    Project.update_counters project.id, classifications_count: 1
+  end
+
+  after_destroy do
+    if project.launch_date.nil? || created_at >= project.launch_date
+      Project.update_counters project.id, classifications_count: -1
+    end
+  end
+
   def self.scope_for(action, user, opts={})
     return all if user.is_admin?
     case action
--- a/app/models/classification.rb
+++ b/app/models/classification.rb
@@ -23,18 +23,6 @@ class Classification < ActiveRecord::Base
   scope :created_by, -> (user) { where(user: user) }
   scope :complete, -> { where(completed: true) }
 
-  after_create do
-    Project.update_counters project.id, classifications_count: 1
-    Workflow.update_counters workflow.id, classifications_count: 1
-  end
-
-  after_destroy do
-    if project.launch_date.nil? || created_at >= project.launch_date
-      Project.update_counters project.id, classifications_count: -1
-      Workflow.update_counters workflow.id, classifications_count: -1
-    end
-  end
-
   def self.scope_for(action, user, opts={})
     return all if user.is_admin?
     case action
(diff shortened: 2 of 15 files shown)


app/workers/reset_project_counters_worker.rb
```diff
@@ -0,0 +1,37 @@
+class ResetProjectCountersWorker
+  include Sidekiq::Worker
```
marten: maybe put this on the high queue?

marten:
Fixes #1398 btw

app/workers/reset_project_counters_worker.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  sidekiq_options queue: :data_high
+
+  def perform(project_id)
+    project = Project.find(project_id)
+
+    return unless project.launch_date
```
camallen: what about beta projects, etc that haven't got a launch date? Should we have a default on this field?

app/workers/reset_project_counters_worker.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    project.classifications_count = project.classifications.where("created_at >= ?", project.launch_date).count
+    project.classifiers_count = UserProjectPreference.connection.execute(<<-SQL)[0]["count"]
+      SELECT COUNT(DISTINCT(upp.user_id)) count FROM user_project_preferences upp
+      INNER JOIN classifications c ON c.user_id = upp.user_id
+      INNER JOIN projects p ON p.id = upp.project_id
+      WHERE upp.project_id = #{project.id} AND c.project_id = #{project.id} AND c.created_at >= p.launch_date;
+    SQL
+    project.save!
```
camallen: wondering if we should just update the columns here and avoid potential model validation errors like lock_version, etc?

app/models/classification.rb
```diff
@@ -23,6 +23,18 @@ class Classification < ActiveRecord::Base
   scope :created_by, -> (user) { where(user: user) }
   scope :complete, -> { where(completed: true) }
 
+  after_create do
```
camallen: i'm not a huge fan of callbacks that modify state in associated models, i can see the reasoning here but any reason why we can't extract these to service objects and then just link them via the controller / call them when testing?

app/workers/reset_project_counters_worker.rb
marten: I was thinking they probably only want to reset when they go live and thus have a launch date. If we have some other date in this field that's not the launch date, I think we'd want to find a different name for the column while we still easily can.

app/models/classification.rb
marten: Mostly because it's an enormous pain to call a service object instead of creating AR records with our current JSONApi stuff. Maybe I should refactor one or two controllers and show how I'd do it without all that (in a fresh branch).

marten:
I'm voting against my previous thought wrt unretire in this branch. Smaller branches are better.

app/workers/reset_project_counters_worker.rb
marten: Fine by me. That's why I did the save-bang method, so it at least wouldn't go unnoticed, but a surgical replacement of just the two values and ignoring validations is fine too.

app/models/classification.rb
camallen: we've already got a service object to do this, so it'd just be a hook on there.

marten:
@camallen moved callbacks to service objects.

camallen:
Looks good, specs failing though.

app/controllers/api/events_controller.rb
```diff
@@ -34,7 +34,10 @@ def process_incoming_event
       if event_params_check
         if upp = user_project_preference
           upp.legacy_count[create_params[:workflow]] = create_params[:count]
-          response_status = :ok if upp.save
+          if upp.save
+            Project.increment_counter :classifiers_count, upp.project_id
```
camallen: This events controller is for legacy projects (ouroboros and older) only, i assume this is to keep counts in sync for those projects?

camallen: I got this now, ignore above.

camallen:
LGTM - if you are happy, lets get it live!

PR #1473, opened 2015-11-06
Do not allow users to login with display_name

camallen:
Removed ability to login with display_name, removed unique constraint on user#display_name index, switched to normal model validations for user#login and removed old code.

I still need to check that this won't break the front end but i'm almost 100% sure we don't use unique display_name filtering for anything anymore.


db/migrate/20151106172531_remove_uniqueness_display_name_index_users.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveUniquenessDisplayNameIndexUsers < ActiveRecord::Migration
+  def change
+    remove_index :users, column: :display_name, unique: true
```
marten: Dont we still want display names to be unique? What happens if I start impersonating moderators by setting my display name the same?

camallen: I'm not sure that it's a real problem as we report the @user  handle and the display name together in talk.

image

marten: Guess not then.

camallen: Can you think of another case where we'd need this? It'd be nice not to have to fix any conflicts if we remove this validation.

marten: I'm fine with multiple Jim's having the same display name. But we will always have to include the username too. Don't know of any other areas where we might not be doing that currently.

camallen: @brian-c @aweiksnar @parrish does the front end require unique display_names anything (filtering?) these days? Pretty sure the API uniquely filters on slugs and owners only.

camallen:
I think we are good to go on this.

marten:
LGTM

PR #1478, opened 2015-11-11
rework commit_id.txt build

camallen:
build the latest commit tag during the docker build and then copy across to public on start.

This is needed since docker hub now respects the .dockerignore file and the .git data isn't available at runtime in the resulting images.


marten:
LGTM

PR #1481, opened 2015-11-12
add script to hot restart puma

camallen:
most likley to be used after a migration to ensure the AR schema is consistent with the DB.

Looks like there will be some changes to AR that will fix some of these errors but probably not good enough for all migration schema changes, https://github.com/rails/rails/issues/12330

This will have to be combined with the jenkins prod db migration job once it's live.

### DIFF
--- a/scripts/docker/restart_puma.sh
+++ b/scripts/docker/restart_puma.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# simple script to hot restart puma, https://github.com/puma/puma#restart
+# e.g. docker exec -it 'panoptes_container_name' bash -c "/rails_app/scripts/docker/restart_puma.sh"
+pkill -F tmp/pids/server.pid -USR2


scripts/docker/restart_puma.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+# simple script to hot restart puma, https://github.com/puma/puma#restart
+# e.g. docker exec -it 'panoptes_container_name' bash -c "/rails_app/restart_puma.sh"
+kill -s SIGUSR2 `ps aux | grep [p]uma | awk '{print $2}'`
```
marten: The containers should have pkill, which would make it as simple as `pkill -USR2 puma`.

marten:
LGTM

camallen:
this won't work as a hot restart with the way we're running puma under rails. we'd need to run puma as as standalone server (which i think impacts on log format / collecting). I'm looking into running puma this way and what changes we'll need. in the meantime this can get deployed just not used.

PR #1483, opened 2015-11-12
Puma cluster mode

camallen:
~~**Don't merge this yet**~~ This is good to go now.

~~We'll need to get the logging format sorted as it's not using the rails log format any more.~~ This setup is req'd in order to use the restart_script in #1481.

Bumps puma gem, switches to running puma on it's own in clustered mode with 2 workers that use code preloading leveraging MRI COW mem savings. It should respect AR connections on boot / app restart (see config/puma.rb). 

All up this should be more resource efficient for the production nodes some simple dev benchmarking showed a speedup of the projects page load of 1-1.5s (4.5 over 5.5). Time will tell what we see in prod. I also tested some hot restarts running the whole stack in docker and both AR and redis connections didn't grow but reset and stabilised to prev values.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,7 +198,7 @@ GEM
       slop (~> 3.4)
     pry-rails (0.3.4)
       pry (>= 0.9.10)
-    puma (2.13.4)
+    puma (2.15.3)
     rack (1.6.4)
     rack-cors (0.4.0)
     rack-protection (1.5.3)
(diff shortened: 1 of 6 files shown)


marten:
Are we sure Redis connections didn't just remain stable because that connection is now being reused across processes after forking?

marten:
LGTM if you think it's fine. Probably want to make sure we've had some requests on staging around a restart before deploying production.

camallen:
Nope but i took https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#sidekiq-tries-to-use-a-connection-from-a-child-process-without-reconnecting to mean it should be ok due to the was sidekiq uses connections. 

Plan is to test on staging and run some restarts and see how it goes.

PR #1491, opened 2015-11-16
serialize default retirement criteria in workflow dumps

### LINKED ISSUE #1417: Default retirement rules show as {} in workflow export
In one of my projects, I've set workflows to retire at a count of 5. In the "retirement" column of the workflows export, the value is ```{"options":{"count":5},"criteria":"classification_count"}```

In my other project, I left the workflows at their default value (retire at a count of 15). The "retirement" column for these workflows is just ```{}```. 

Earlier versions of workflows in my first project also have ```{}``` listed for the "retirement" column, so my guess is that when the retirement rules haven't been changed from their default, nothing ends up in the workflows export.

camallen:
closes #1417

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -76,6 +76,18 @@ class Workflow < ActiveRecord::Base
     end
   end
 
+  def retirement_with_defaults
+    retire_criteria = self.retirement
+    if retire_criteria.blank?
+      {
+        criteria: Workflow::DEFAULT_CRITERIA,
+        options: Workflow::DEFAULT_OPTS
+      }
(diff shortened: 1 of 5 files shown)


marten:
What would happen if we just change `Workflow#retirement` to return the default? Wouldn't that make the frontend happy too, by handing it the defaults instead of (probably?) having to duplicate them over there?

camallen:
Yeah - we do this in the serializer for the UI. I felt like it was more a presentation issue and not model layer responsibility but didn't like that i was duplicating code...happy to rework it to default in the model layer, will update.

PR #1493, opened 2015-11-17
complement search scope with a login match

camallen:
some logins are hard to find for tsvector, this will add the user login scope to the search whenever the search result set drops below a threshold. The only drawback is an  extra count function that i have limited but it's a db cost that should be pretty fast.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2467,6 +2467,13 @@ CREATE INDEX user_groups_display_name_trgm_index ON user_groups USING gist (disp
 
 
 --
+-- Name: users_idx_trgm_login_display_name; Type: INDEX; Schema: public; Owner: -; Tablespace: 
+--
+
+CREATE INDEX users_idx_trgm_login_display_name ON users USING gin ((((COALESCE((login)::text, ''::text) || ' '::text) || COALESCE((display_name)::text, ''::text))) gin_trgm_ops);
+
+
+--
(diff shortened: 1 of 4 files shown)


db/structure.sql
```diff
@@ -2460,11 +2460,20 @@ CREATE INDEX user_groups_display_name_trgm_index ON user_groups USING gist (disp
...
+CREATE INDEX users_idx_trgm_login_display_name ON users USING gin ((((COALESCE((login)::text, ''::text) || ' '::text) || COALESCE((display_name)::text, ''::text))) gin_trgm_ops);
+
+
+--
 -- Name: tsvectorupdate; Type: TRIGGER; Schema: public; Owner: -
 --
 
 CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE ON users FOR EACH ROW EXECUTE PROCEDURE tsvector_update_trigger('tsv', 'pg_catalog.english', 'login', 'display_name');
 
+ALTER TABLE users DISABLE TRIGGER tsvectorupdate;
```
camallen: @marten any idea why the update trigger for tsvector is getting disabled here? I haven't figured it out.

camallen: had to manually add / remove the trigger again in the migration...

marten:
Merge conflict (probably structure.sql) but LGTM so feel free to merge it after fixing.

PR #1504, opened 2015-11-20
add field guide project resource

### LINKED ISSUE #1502: Projects need a FieldGuide resource
Very very similar to the Tutorial resource.

It'll contain an list of `items` (probably a JSON blob?), each with a `title` string, a `content` string, and an `icon` image source string that'll refer to a Media resource, so it needs `attached_images` too.

camallen:
closes #1502 

@brian-c can you check i've got the schema right (app/schemas/field_guide_create_schema.rb) and that the `icon` attribute will contain the linked id for the media resource?

### DIFF
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -32,15 +32,22 @@ class Medium < ActiveRecord::Base
     self.src ||= MediaStorage.stored_path(content_type, type, *path_opts)
   end
 
+  def linked_resource_details
+    details = /\A(?<resource>field_guide|[a-z]+)_(?<media_type>\w+)/i.match(type)
+    [ details[:resource], details[:media_type] ]
+  end
+
+  def pluralize_media_type?(media_type)
+    ["attached_image"].include?(media_type)
+  end
+
   # TODO: This method is a good argument for converting this into a STI model
   def location
-    case type
-    when "project_attached_image", "workflow_attached_image", "tutorial_attached_image"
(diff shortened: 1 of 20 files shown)


app/models/medium.rb
```diff
@@ -32,15 +32,31 @@ def create_path
...
 
+  def plural_types
+    @plural_types ||= %w(project workflow tutorial field_guide)
+  end
+
+  def plural_types_regex
+    @plural_types_regex ||= /\A(#{plural_types.join("|")})_attached_image\z/
+  end
+
+  def split_type
```
marten: My suggestion would be to do it like this:

```ruby
  # TODO: This method is a good argument for converting this into a STI model
  def location
    details = linked_resource_details

    if pluralize_media_type?(details[:media_type])
      "/#{resource.pluralize}/#{linked_id}/#{media_type.pluralize}/#{id}"
    else
      "/#{resource.pluralize}/#{linked_id}/#{media_type}"
    end
  end

  def linked_resource_details
    /\A(?<resource>field_guide|\w+)_(?<media_type>\w+)/i.match(type)
  end

  def pluralize_media_type?(media_type)
    ["attached_image"].include?(media_type)
  end
```

A couple of specs might help us remember that/why we're doing this too.

camallen:
@marten updated, thanks for the regex example.

camallen:
@brian-c this should be on staging soon.

PR #1507, opened 2015-11-23
Cellect for subject selection

### LINKED ISSUE #1373: Use cellect as a selection service to build queues from
Instead of running the Postgres selection by default, i was thinking we can use Cellect as a selection end point to select data to enqueue for a user's subject queue. 

This would be different to the way we used it previously where we just passed all concurrent subject requests so we would require less Cellect servers and would fit into the extraction of task allocation services we're thinking of building. 

Thanks to @marten's selection diagram for inspiration.

camallen:
closes #1373 

Adds Cellect as a subject selection service back to panoptes. There is a global config `Panoptes.cellect_on` that turns cellect selection on and each worfklow has the option of opting into cellect via a configuration directive or failing that the subjects set size. All the cellect work has been backgrounded so we shouldn't see any timeouts in user facing requests if cellect nodes are unreachable.

For reference, the PR was Reworked from this [original commit](https://github.com/edpaget/Panoptes/commit/939be03227e3a71bbdc5bf0bf4f970503d0d79c1)


marten:
I thought I taught Hound not to complain about a lot of these... Apparently not, sorry.

marten:
Ahh, now I get it. @camallen you should rebase onto master so you get the `.ruby-style` file which will disable all these annoying checks. Since you have none, Hound just checks against their defaults, which are based on some community-maintained Ruby style guide.

spec/workers/enqueue_subject_queue_worker_spec.rb
```diff
@@ -53,23 +53,135 @@
...
+        it 'should attempt to queue the selected set' do
+          allow(Subjects::CellectClient).to receive(:get_subjects).and_return(result_ids)
+          expect(SubjectQueue).to receive(:enqueue)
+          run_selection
+        end
+
+        context "when the cellect client can't reach a server" do
+          it "should fall back to postgres strategy" do
+            allow(Subjects::CellectClient).to receive(:get_subjects)
+              .and_raise(Subjects::CellectClient::ConnectionError)
```
camallen: I actually prefer to see the . on the trailing line so i know it's related to the line above with/without indentation.

camallen:
@marten @parrish if/when you get time i'd appreciate your thoughts on this, no rush. Merry Christmas to you both.

bin/rails
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env ruby
-begin
```
marten: Something went wrong with a rebase here. Or did you mean to get rid of spring?

camallen: Yeah i did remove it, i couldn't get it playing nicely with cellect and it's connection to zk on app boot when using docker-compose for testing. See [this commit](https://github.com/camallen/Panoptes/commit/47f9d82009fff8bf1e3ad365a930d304b92dcb2b), happy to put it back in but the Cellect NodeSet (client in this case) but had dead actors due the app creating a ZK connection on boot.

parrish:
:+1: 

Just reading through and poking around in a local setup, this seems to work.  I'd like to see this deployed to staging for a bit before a production deploy though.

camallen:
@parrish i think this is good to go, i've disabled the cellect client on boot and re-added spring with an explicit require before using the client. This has avoided errors on the connection / node_set form the client in my testing.

parrish:
You may also want an explicit `require 'cellect/client'` in lib/subjects/cellect_session.rb (depends on load order), but this looks good.

:shipit:

PR #1508, opened 2015-11-23
Publish events to Kafka

marten:
Two events are implemented:

* a classification is made: publish an event to indicate this. the event-id is derived from the record id, so that duplicates aren't possible. this allows us to build histograms of how many classifications we see per timeslot

* a subject is retired: publish the current values of various counters, so that we can chart these over time later.

### DIFF
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -9,6 +9,7 @@ class RetirementWorker
       count.retire! do
         SubjectQueue.dequeue_for_all(count.workflow, count.set_member_subject_ids)
         deactivate_workflow!(count.workflow)
+        publish_to_event_stream(count.workflow)
       end
     end
   end
@@ -17,5 +18,14 @@ class RetirementWorker
     if workflow.finished?
       Workflow.where(id: workflow.id).update_all(active: false)
     end
+  end
+
+  def publish_to_event_stream(workflow)
+    EventStream.push('workflow_counters', Time.now,
+      project_id: workflow.project_id,
+      workflow_id: workflow.id,
+      subjects_count: workflow.subjects_count,
+      retired_subjects_count: workflow.retired_subjects_count,
+      classifications_count: workflow.classifications_count)
   end
 end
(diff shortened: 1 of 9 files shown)


app/workers/retirement_worker.rb
```diff
@@ -18,4 +19,13 @@ def deactivate_workflow!(workflow)
       Workflow.where(id: workflow.id).update_all(active: false)
     end
   end
+
+  def push_counters_to_event_stream(workflow)
```
camallen: no event_id here, you happy to use the defaults from the es writer [here](https://github.com/zooniverse/zoo-event-stats/blob/dcdc0ec411d5924b32965e43a7b5f48e445c6334/lib/output/elasticsearch_writer.rb#L48)?

marten: The `EventStream` class defaults to a UUID so that Kafka consumers always see the event_* fields, but leaves the option of overriding it (eg. deduping classifications).

lib/multi_kafka_producer.rb
```diff
@@ -27,6 +27,7 @@ def self.publish(topic, *msgs_and_keys)
   KAFKAS = { kafka: 'jruby-kafka', poseidon: 'poseidon' }
 
   def self.default_adapter
+    return :null if Rails.env.test?
```
camallen: was this actually attempting to publish data in the test harness?

spec/lib/multi_kafka_producer_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 shared_examples "default adapter" do
   it 'should set the default adapter for the platform' do
+    allow(Rails.env).to receive(:test?).and_return(false)
     MultiKafkaProducer.adapter =  nil
```
camallen: not sure i'm following this, if it's set to nil before the let block runs won't it just be nil in the expectation?

lib/multi_kafka_producer.rb
marten: Yes, all tests just stub out `MultiKafkaProducer.publish` but if you don't you'll end up actually hitting Kafka (which we probably almost never need in the test suite). It's no longer needed for this branch though, so I've taken it out. I might add a Test adapter later which remembers stuff sent to it, which means tests can use that instead of setting mock expectations whenever they check stuff sent into Kafka.

PR #1511, opened 2015-11-25
Upgrade doorkeeper

marten:
I've converted `doorkeeper_for` into a custom `require_authentication` wrapper instead of the before_filter directly, because that will tie in nicely with my next PR.

As far as my testing could tell we're still secure.


camallen:
specs failing.

config/initializers/doorkeeper.rb
```diff
@@ -18,6 +18,8 @@ def self.optional
 
   realm "Panoptes"
 
+  grant_flows ["authorization_code", "client_credentials", "implicit", "password"]
```
camallen: Do we have to explicitly declare the flows we are using now? Also are we using all these?

camallen:
Looks fine to me, rebase needed.

config/initializers/doorkeeper.rb
marten: They removed `implicit` and `password` from the defaults in 2.0.

* `implicit` is probably what we're using in PFE
* `password` is used by some frontends (get access token using username/password without ever redirecting to Panoptes). They shouldn't use this, but I'm not sure if they've migrated away yet

They added the other two as new defaults:

* `client_credentials` is for that Faraday middleware I wrote (get access token using application key/secret)
* `authorization_code` is the full redirect and temporary token -> access token exchange that all external apps that have a backend should be using.

camallen: pretty sure we don't have any password flow apps and see who shouts.

camallen: we're using implicit for non zooniverse.org domains and a custom implementation of devise session -> token for PFE.

camallen:
Looks like i was wrong and PFE is using the resource owner passwords grant type...https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Resource-Owner-Password-Credentials-flow#configuration

camallen:
this all looks fine, as long as we don't break login for PFE this is good.

PR #1512, opened 2015-11-25
[WIP] Allow login via http basic

marten:


app/controllers/api/api_controller.rb
```diff
@@ -40,17 +41,51 @@ class ApiController < ApplicationController
...
+        check_doorkeeper_authentication!
+        @current_resource_owner = User.find_by_id(doorkeeper_token.resource_owner_id)
+      else
+        check_doorkeeper_authentication!
       end
     end
 
+    def check_doorkeeper_authentication!
+      (self.class.doorkeeper_scopes_per_action || {}).each do |actions, scopes|
+        if actions.empty? || actions.include?("all") || actions.include?(params[:action])
```
marten: It was already set up this way, but I always get nervous when stuff isn't private-by-default. Should I switch our controllers over so that they must specify whenever an action is supposed to be public?

camallen: what sort of dsl are you talking about? 
`require_authentication :create, :update, except: [:index, :show], scopes: [:project]`

marten: Talked this through IRL, we're going to make it so that you specify both public and private actions. Unspecified actions raise an exception and are thereby completely inaccessible. Maybe even decorator-style:

```ruby
auth :public \
def index; end

auth :private, scopes: [:project] \
def show; end
```

marten:
I might do the public/private thing in another PR, but with `faraday-panoptes` I'm closing this PR since it's now just about as easy to just use OAuth's client_credentials flow.

PR #1515, opened 2015-11-27
Attempt to speed up user searches

### LINKED ISSUE #1509: check pg_search scopes for user search are using indexes
seeing large cpu spikes when searching for small strings, e.g. `d` & `dzm`. Check the query is using indexes and the resulting sql is optimised.

camallen:
closes #1509 and attempts to address [this](https://github.com/zooniverse/Panoptes/issues/1230#issuecomment-159040608)

removes some sparse indexes that weren't used as the query was a full table scan for both and adds a layer strategy to finding users with narrow and broad scopes. Seems its a trade off between high accuracy and speed.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -16,7 +16,13 @@ class Api::V1::UsersController < Api::ApiController
   alias_method :user, :controlled_resource
 
   search_by do |name, query|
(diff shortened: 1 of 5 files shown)


DarrenMcRoy:
Cool! :dart:

camallen:
@DarrenMcRoy feel free to try it out to see if it's faster.

PR #1518, opened 2015-12-01
Set finished_at instead of deactivating workflows

marten:
Instead of deactivating, set a timestamp column. This allows us to see
which workflows are supposed to be visible but have just run out of
data. In turn, that will allow us to automatically unfinish a workflow
when more data becomes available (either due to recounts or simply
because more data is added to a subject set).


camallen:
Lgtm

PR #1524, opened 2015-12-01
update workflow serialized response

### LINKED ISSUE #1520: Expose workflows' public_gold_standard field
I can put a value into `public_gold_standard` just fine, and it works as expected (thanks @camallen) but it doesn't seem to be included in the workflow serialization.

camallen:
closes #1520 added public_gold_standard and added default for null finished_at

### DIFF
--- a/app/serializers/workflow_serializer.rb
+++ b/app/serializers/workflow_serializer.rb
@@ -9,7 +9,7 @@ class WorkflowSerializer
              :created_at, :updated_at, :finished_at, :first_task, :primary_language,
              :version, :content_language, :prioritized, :grouped, :pairwise,
              :retirement, :retired_set_member_subjects_count, :href, :active,
-             :aggregation, :configuration
+             :aggregation, :configuration, :public_gold_standard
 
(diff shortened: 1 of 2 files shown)


app/serializers/workflow_serializer.rb
```diff
@@ -49,4 +49,8 @@ def content
   def retirement
     @model.retirement_with_defaults
   end
+
+  def finished_at
+    @model.finished_at || ""
```
marten: Why is an empty string cleaner than a null? Is that just because the parser in JS-land is stupid?

camallen: something like that. we agreed a while back to always serialize a default value (mostly via db defaults) so i just added it in for consistent behaviour.

brian-c: I'd actually expect this to be null, since it's an unset date, not text.

camallen: cool, it'll serialize as a string by default. will remove the empty string.

PR #1527, opened 2015-12-02
Save resource before retiring

marten:
Works around optimistic locking issues (retiring updates a counter on
the workflow and increments the lock version, so afterwards the
existing workflow object cannot be saved until doing a reload.

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -72,7 +72,9 @@ class Api::V1::WorkflowsController < Api::ApiController
 
   def add_relation(resource, relation, value)
     if relation == :retired_subjects && value.is_a?(Array)
+      resource.save!
       value.each {|id| resource.retire_subject(id) }
+      resource.reload
     else
       super
     end
@@ -81,7 +83,9 @@ class Api::V1::WorkflowsController < Api::ApiController
   def new_items(resource, relation, value)
     case relation
     when :retired_subjects, 'retired_subjects'
+      resource.save!
       value.flat_map {|id| resource.retire_subject(id) }
+      resource.reload
     when :subject_sets, 'subject_sets'
       items = construct_new_items(super(resource, relation, value), resource.project_id)
       if items.any? { |item| item.is_a?(SubjectSet) }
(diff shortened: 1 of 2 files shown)


camallen:
is this just for retiring by the api?

marten:
Yes. It's one of the UpdateableResource things that does a save afterwards (which tries to set a default value for the ranked_model). Automatic retirement was fine.

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -72,7 +72,9 @@ def extract_strings(tasks)
 
   def add_relation(resource, relation, value)
     if relation == :retired_subjects && value.is_a?(Array)
+      resource.save!
```
camallen: won't the reload on it's own be sufficient here?

camallen:
If i'm getting this right, we could also use a `retire_without_count!`method on the subject_workflow_count and manually bump the counter manually on the update_links save event.

app/controllers/api/v1/workflows_controller.rb
marten: Yes, but I don't feel confident enough in the whole JsonApiController stuff that I dare say whether there's any way there aren't any unsaved changes. If you do, then cool I'll remove this.

camallen: the `add_relation` method here should only be called via the `update_links` so the only state change here is retiring the subject. Pretty sure this would be ok to reload as we've got no other attributes to change here.

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -81,7 +83,9 @@ def add_relation(resource, relation, value)
   def new_items(resource, relation, value)
     case relation
     when :retired_subjects, 'retired_subjects'
+      resource.save!
       value.flat_map {|id| resource.retire_subject(id) }
+      resource.reload
```
camallen: does this code even get called when using the `update_links` route? If not we can just enforce `retired_subjects` on a workflow via the update_links route and not the update resource links to avoid workflow instance updates combined with retired counter bumps that cause the lock errors.

camallen: also we can modify the yield block in `updatable_resource#update_links` to allow a forced reload of the workflow but we currently use it for post persistence operations so we'd have to deal with that.

marten:
Okay, so merge this now and I'll extract to service object later in the context of the trailblazer-gem branch? LGTY?

camallen:
Yeah - lets do punt this and just pay the extra db trip.

PR #1528, opened 2015-12-02
Custom retirement config

### LINKED ISSUE #1368: disable counter retirement for live projects with custom retirement

marten:
Closes #1382
Closes #1368

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -71,15 +71,11 @@ class Workflow < ActiveRecord::Base
   end
 
   def retirement_with_defaults
-    retire_criteria = self.retirement
-    if retire_criteria.blank?
-      {
-        criteria: Workflow::DEFAULT_CRITERIA,
-        options: Workflow::DEFAULT_OPTS
-      }
-    else
-      retire_criteria
-    end
+    self.retirement.presence || default_retirement_options
+  end
+
+  def default_retirement_options
+    {criteria: Workflow::DEFAULT_CRITERIA, options: Workflow::DEFAULT_OPTS}
   end
 
   def retired_subjects_count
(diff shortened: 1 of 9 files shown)


lib/retirement_schemes/classification_count.rb
```diff
@@ -1,11 +1,21 @@
...
+      if disabled?
+        false
+      else
+        sw_count.classifications_count >= @count
+      end
     end
 
-    def retire?(count)
-      count.classifications_count >= @count
+    def disabled?
```
camallen: is this used anymore or are we just going to overwrite the criteria scheme to NeverRetire?

marten: No, forgot to remove it after adding NeverRetire.

app/models/workflow.rb
```diff
@@ -67,25 +67,13 @@ def retire_subject(subject_id)
...
-      RetirementSchemes::ClassificationCount.new(*params)
-    else
-      raise StandardError, 'invalid retirement scheme'
-    end
+    criteria = retirement_with_defaults.fetch('criteria')
+    options = retirement_with_defaults.fetch('options')
+    scheme_class = RetirementSchemes.for(criteria).new(options)
   end
 
   def retirement_with_defaults
```
camallen: I am using this for default rep in the serialzer, https://github.com/zooniverse/Panoptes/blob/f2a92ce3d3e48b3874d29801e3c10e3be9283912/app/serializers/workflow_serializer.rb#L50

camallen:
LGTM

PR #1534, opened 2015-12-03
Fix application form path and method

### LINKED ISSUE #1530: oauth application edit form submit is broken
submits to `/applicatoins.:id` not `/applications/:id`

marten:
Fixes #1530


app/views/doorkeeper/applications/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_for :application, url: oauth_applications_path(application), html: {class: 'form-horizontal', role: 'form'} do |f| %>
+<%= form_for :application, url: application.new_record? ? oauth_applications_path(application) : oauth_application_path(application), method: application.new_record? ? :post : :put, html: {class: 'form-horizontal', role: 'form'} do |f| %>
```
camallen: can this be like the doorkeeper form instead? https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/views/doorkeeper/applications/_form.html.erb#L1

marten: For the URL yes, but our routes don't like the normal POST that happens, so seems like the method thing needs to stay.

camallen:
cool, LGTM.

PR #1536, opened 2015-12-07
Allow sorting by number of people that have classified

marten:
Sadly Restpack doesn't support customizing the parameters used for each sorting, so I can't really call this "popularity" outward-facing, unless anyone has any ideas?


marten:
Also, do we want an index on this, or would we rather it do a seq scan? It's not too many projects, and at the same time there's probably a lot of updates to the counts, so we'd probably rather take the scan?

camallen:
not sure about the index, seems we'd be updating it all the time.

PR #1537, opened 2015-12-07
Project activity counts

camallen:
Daily project activity counts based on 24 window of classification counts.

### DIFF
--- a/spec/workers/calculate_project_activity_worker_spec.rb
+++ b/spec/workers/calculate_project_activity_worker_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+describe CalculateProjectActivityWorker do
+  let(:worker) { described_class.new }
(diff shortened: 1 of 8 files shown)


spec/workers/calculate_project_activity_worker_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  let(:user) { project.user }
+
+  let(:classifications) do
+    create_list(:classification, 2, project: project, workflow: workflow)
+    c = create(:classification, project: project, workflow: workflow)
+    c.update_column(:created_at, 25.hours.ago)
+  end
+
+  describe '#perform' do
+
```
marten: What's up with the blank lines after a `describe` or `context`? Do we have a style guide about this?

camallen: i find them more readable, what do you prefer?

PR #1541, opened 2015-12-08
Resolves #1514 should not increment classifiers count every time a user performs an action

### LINKED ISSUE #1514: Incorrect project classifiers count
The `events_controller` api is incorrectly updating the project classifier count on all activity messages. It should only count them when the `user_project_preference` is created for the first time and thus indicating a new user for a project.

For reference, the `events_controller` is used to link legacy project activity counts to the new zooniverse api, this is the way we keep counters in sync between legacy projects and older api's.

amyrebecca:
Spec changes were already provided by Cam Allen

Because calling upp.save automatically unsets upp.new_record? I capture the state of upp.new_record? in was_new and then decide whether or not to increment the counter based on that value.

### DIFF
--- a/spec/controllers/api/events_controller_spec.rb
+++ b/spec/controllers/api/events_controller_spec.rb
@@ -171,6 +171,11 @@ describe Api::EventsController, type: :controller do
           end.to change { UserProjectPreference.count }.from(0).to(1)
         end
 
+        it "should increment the project's classifiers count", :focus do
+          expect(Project).to receive(:increment_counter).with(:classifiers_count, project.id)
+          post :create, first_visit_event_params
+        end
+
         context "with an unknown user zooniverse_id" do
 
           it "should not create the user project preference model (upp)" do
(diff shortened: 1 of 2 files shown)


spec/controllers/api/events_controller_spec.rb
```diff
@@ -171,6 +171,11 @@ def overridden_params(new_params)
           end.to change { UserProjectPreference.count }.from(0).to(1)
         end
 
+        it "should increment the project's classifiers count", :focus do
```
camallen: still got a `:focus` set on here so the CI only tested those specs, just remove it.

spec/controllers/api/events_controller_spec.rb
```diff
@@ -240,6 +245,11 @@ def overridden_params(new_params)
             end.to_not change { UserProjectPreference.count }.from(1)
           end
 
+          it "should not increment the project's classifiers count", :focus do
```
camallen: same focus here.

camallen:
LGTM once the full CI runs. Thanks @amyrebecca!

PR #1542, opened 2015-12-08
Remove zoo_home connection and models

marten:


camallen:
this looked ok to me. can we get it out next week?

marten:
Yes, not going to deploy it right before the weekend ;)

camallen:
it'll need a rebase too.

Issue #1546, opened 2015-12-09
Issues with project search


alexbfree:
To see these issues, please visit http://preview.zooniverse.org/panoptes-front-end/miniztm/?env=production#/projects , which uses the project search service for the search box on the right. This reveals a couple of issues, you will see both if you start typing `"Snapshot Serengeti"`.

### First issue: Non-public projects(?) being included
If you type `"snapsh"` you can see a project called "Snapshots at Sea" listed. I do not have access to this project so it should not be listed in search.

### Second issue: No results with first few characters
We get no results back for 1,2,3 and 4 length search strings. This produces strange behaviour in the autocomplete, where somebody might be confused by the apparent absence of results - it's not obvious that you have to keep typing the first 5 characters.

astopy:
First issue only seems to happen if you have admin mode turned on.

CKrawczyk:
I am seeing "Snapshots at Sea" and I don't have admin access at all.

eatyourgreens:
Same here, and I'm not logged in at all.

camallen:
snapshots at sea is a public project, https://www.zooniverse.org/projects/tedcheese/snapshots-at-sea, we can add more restrictive search filters to only find launch_approved, etc flagged projects.

search is something we can work on, see for some details https://github.com/zooniverse/Panoptes/issues/1230

camallen:
closed by #1547

PR #1547, opened 2015-12-10
Added tsvector to projects model

srallen:
Adds tsvector to projects mode for #1546

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -74,8 +74,13 @@ class Project < ActiveRecord::Base
 
   pg_search_scope :search_display_name,
     against: :display_name,
-    using: :trigram,
-    ranked_by: ":trigram"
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -74,8 +74,13 @@ class Project < ActiveRecord::Base
 
   pg_search_scope :search_display_name,
     against: :display_name,
-    using: :trigram,
-    ranked_by: ":trigram"
+    using: { tsearch: {
+      prefix: true,
```
marten: Could you indent this one more level so it's clearer that it all belongs to tsearch?

marten:
LGTM

app/models/project.rb
srallen: Fixed in latest commit.

PR #1549, opened 2015-12-11
always setup project email prefs on first classification receipt

camallen:
take care of the situtation where projects like shakespeares world setup the upp before the first classification is submitted and the email prefs / counter incs were missed.

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -36,14 +36,14 @@ class ClassificationLifecycle
 
   def process_project_preference
     if should_create_project_preference?
-      existing_record = true
-      upp = UserProjectPreference.where(user: user, project: project).first_or_create do |up|
-        Project.increment_counter :classifiers_count, project.id
-        up.email_communication = user.project_email_communication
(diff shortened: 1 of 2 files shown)


lib/classification_lifecycle.rb
```diff
@@ -36,14 +36,17 @@ def transact!(&block)
...
-        existing_record = false
+        save_record = true
+      end
+      if first_classifcation = upp.email_communication.nil?
+        save_record = true
+        Project.increment_counter :classifiers_count, project.id
+        upp.email_communication = user.project_email_communication
       end
-      upp.touch if existing_record
+      save_record ? upp.save! : upp.touch
```
marten: I wonder if you could just replace the whole `save_record` variable with `record.changed?`:

```ruby
      upp = UserProjectPreference.where(user: user, project: project).first_or_initialize do |up|
        up.preferences = {}
      end

      if first_classifcation = upp.email_communication.nil?
        Project.increment_counter :classifiers_count, project.id
      end

      upp.changed? ? upp.save! : upp.touch
```

camallen: yep - thanks.

Issue #1372, opened 2015-09-16
Missing foreign key relations


marten:
We're missing a FK between projects and workflows, so:

```sql
SELECT * FROM workflows 
LEFT OUTER JOIN projects on workflows.project_id = projects.id
WHERE projects.id IS NULL;
```

returns 22 rows.

There's only 7 FK relations in the schema, so we're probably missing others too?

PR #1551, opened 2015-12-15
Deactivate instead of destroy projects

camallen:
Seems the api allowed full project delete and some background workers were failing with missing record.  I've reworked the destroy route to just deactivate instead, note that users with rights can hard delete workflows, subject_sets and set_member_subjects so any deactivate could possibly be a shell.

I'm thinking we should keep projects around for posterity and possibly restores. Should we just allow the project to be hard deleted instead?


marten:
LGTM

I noticed a few of those missing projects when I was looking at data for nightly exports. Should we add foreign key constraints on more stuff?

camallen:
can we just take notes on this issue, https://github.com/zooniverse/Panoptes/issues/1372?

marten:
Yeah I didn't mean in this PR. You'd have to re-add deleted project records before you could set up that FK anyways, and that's definitely out of scope for this PR.

PR #1552, opened 2015-12-15
Provide a fast action for listing projects

parrish:
This is a bit hacky, but it cuts the high throughput, 1+ second action from zooniverse.org/projects down to ~25 millis and also doesn't interfere with any of the other requests against the endpoint.

Insofar as I can tell, the best (only?) way to approach this is to just bypass the performance bloat from restpack/json-api, and controller/resource abstractions.  Definitely open to suggestions though.

If we're cool with this proposal
  - I'll add specs
  - I've got a corresponding PR (zooniverse/Panoptes-Front-End#2029) for the front end that
    - adds the `homepage=true` param
    - eliminates the unnecessary api calls to fetch the owner
    - eliminates the unnecessary api calls to fetch the avatars


camallen:
Thanks, I like this and its fast. I was working on a branch to add context to the index route (i.e. the homepage param) that would cut down the associated model cruft in the RP serializer. @marten thoughts?

lib/fast_project_serializer.rb
```diff
@@ -0,0 +1,72 @@
+class FastProjectSerializer
```
camallen: we could probably leverage the v1 json api serialiser @marten built, e.g. https://github.com/zooniverse/Panoptes/blob/9826fe1f9f82b3ba2efac914319529a9429ee628/lib/kafka_classification_serializer.rb

marten:
Seems okay to me, but I wonder if we shouldn't just make a `/api/homepage` route at this point.

parrish:
> we could probably leverage the v1 json api serialiser @marten built

Ah, I didn't realize that AMS was included too.  Not sure if I could get it to produce an identical format though.

> Seems okay to me, but I wonder if we shouldn't just make a /api/homepage route at this point.

I mentioned this earlier, but the front end passes through query params to the api request:  e.g. https://www.zooniverse.org/projects?homepage=true requests https://www.zooniverse.org/api/projects?include=avatar&launch_approved=true&homepage=true

That lets us deploy this and see how it does in production just by adding a feature flag.

Beyond that, I'm pretty indifferent.

I'll go ahead and spec out the controller response today.

marten:
Yeah the v1adapter I added was most of the way there (as in i might have missed some edge cases, for instance I didnt need pagination for my goals so that would need a test).

parrish:
This should be good to go now I think.  I also added in the ability to handle translations should it be needed

camallen:
LGTM - rebase?

parrish:
Rebased!

PR #1553, opened 2015-12-16
[WIP] Nightly exports

marten:


app/workers/warehouse_dump_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class WarehouseDumpWorker
+  include Sidekiq::Worker
+  include Sidetiq::Schedulable
+
+  recurrence(backfill: true) { daily.hour_of_day(1) }
+
+  def perform(last_occurence, current_occurence)
+    from = Time.at(last_occurence)
+    till = Time.at(current_occurence)
```
marten: Fair point here by Hound, don't let me merge until I add specs to check that this setup doesn't miss/duplicate rows around DST changes.

marten:
Moving this to a separate agent in https://github.com/zooniverse/warehouse/

PR #1555, opened 2015-12-16
Projects index speed up

camallen:
similar to #1552 but using the default v1 serializers to provide a slimmed down resource representation. I've also tried to optimise data sideloading eager loads and language queries from loaded / non-loaded associations. 

This PR uses the `?cards=true` params on the projects index route to get a cut down resource rep. Also benchmarking on my machine shows this route come in at ~85ms.

@marten @parrish thoughts? Also we could try out both this and the other PR and get the front end calling on the different override params for testing.

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -26,7 +26,7 @@ class Api::V1::ProjectsController < Api::ApiController
                     :introduction,
                     :url_labels]
 
-  before_action :eager_load_includes, only: :index
+  before_action :eager_load_relations, only: :index
   before_action :filter_by_tags, only: :index
   before_action :downcase_slug, only: :index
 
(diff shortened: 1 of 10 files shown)


parrish:
It works as far as I can tell, although I can't see why that spec is failing.

Pros vs #1552 
  - It's not completely bypassing what the rest of the app is doing.

Cons vs #1552
  - Response times are ~85 ms vs ~25 ms, but both are way better than what's in master
  - There's still an N+1 on project_contents, but it's fast, so probably not a big deal
  - This feels really complex (to me) for a one-off case, but that could also just be the established pattern of abstractions in Panoptes

Either one will work IMO, but it might be overkill.  If the changes positively impact other actions (or other cases for the index action), then this is probably worth it.

camallen:
The main difference with this implementation is that this PR will still support translated project contents by default which we'll need long term, hence the code changes to `translatable.rb`. For current needs #1552 works fine in my opinion and i think we should look to merge both and switch the client to consume which one it wants. 

I'll try and track down the N+1 on project_contents.

marten:
It's not the easiest to read code in the world, but that's mostly the fault of the existing architecture I think.

camallen:
anything you want me to modify to make it more readable? I really didn't like the eager load override code but i couldn't figure a way of ensuring we always eager load the associations.

PR #1557, opened 2015-12-18
Pass in the permissions scope to the fast project serializer

parrish:
Updates #1552 to use the permissions scope from the query.

Performance hit is around 20-25 ms for this (basically double).


camallen:
LGTM

parrish:
Nope.  This is dumb.  We're going to use something simpler.

parrish:
Just double checked against production, all `launch_approved: true` projects are public so this extra step is completely unnecessary.  I'll create a separate issue for building out a state machine around transitioning project state.

PR #1565, opened 2015-12-30
Faster user search on login field

### LINKED ISSUE #1554: Can't find existing user when adding collaborators
Think this is a backend issue. Trying to add user jenmunroe as an expert and researcher to Shakespeare's World. There is a jmunroe but apparently that's not the right person. @camallen I know you've sorted out a few of these before and can maybe help again?

camallen:
closes #1554 - limit user search scopes to login field only, removed the prefix matching and updated trigram to use indexes. This will now attempt to find a user by login field and if missing try and match on a full search using tsvectors and trigram (fuzzy). 

Note, tsvectors can return large, similar rank results sets for similar strings so the fuzzy matching is used in the rank to tie break, in my testing it worked well so limiting the page size (like the front-end does) should return the most relevant hits first.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -17,7 +17,7 @@ class Api::V1::UsersController < Api::ApiController
 
   search_by do |name, query|
     search_names = name.join(" ")
-    login_search = query.where("lower(login) = '#{search_names.downcase}'")
+    login_search = query.where("lower(login) = ?", search_names.downcase)
     if login_search.exists?
       login_search
     else
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -16,7 +16,13 @@ class Api::V1::UsersController < Api::ApiController
   alias_method :user, :controlled_resource
 
   search_by do |name, query|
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -17,11 +17,11 @@ class Api::V1::UsersController < Api::ApiController
 
   search_by do |name, query|
     search_names = name.join(" ")
-    fast_search = query.search_name_fast(search_names)
-    if fast_search.exists?
-      fast_search
+    login_search = query.where("lower(login) = '#{search_names.downcase}'")
```
parrish: Avoid a SQL injection here with
```ruby
query.where "lower(login) = ?", search_names.downcase
```

parrish:
Looks good otherwise.  Needs a rebase from master first though.

app/controllers/api/v1/users_controller.rb
camallen: good catch, bad campbell

camallen:
fixed and rebased, specs running now.

camallen:
fixed and rebased, specs running now.

parrish:
:+1:

PR #1566, opened 2016-01-04
pg search indexes

camallen:
waiting on #1565 - i reviewed the use of search indexes and made some changes. 

1. notably never use a trigram threshold as it currently will [not use an index](https://github.com/Casecommons/pg_search/issues/278).
2. don't use prefixes for tsvectors just use the trigram properly for fuzzy matches.
3. user_groups only search the display name for trigrams to keep the index size down.
4. use gin trigram op indexes, i couldn't get the derived pg_search queries to use the gist indexes.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -28,6 +28,6 @@ class Collection < ActiveRecord::Base
 
   pg_search_scope :search_display_name,
     against: :display_name,
-    using: { trigram: { threshold: 0.1 } },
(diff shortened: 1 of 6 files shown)


parrish:
Looks like this needs a rebase too.

camallen:
rebased - @parrish or @marten it'd be cool to get this out today.

parrish:
:shipit:

PR #1567, opened 2016-01-04
Remove unused columns

### LINKED ISSUE #1560: remove unused columns
Noting that SubjectWorkflowCount set_member_subject_id attribute this isn't being used anymore and confused my when i looked at the association *_ids vs attribute *_id, so best to remove this from the schema.

Also the subject_ids on classifications too.

marten:
To do this zero-downtime-style, **deploy first**, then migrate later.

Fixes #1560

### DIFF
--- a/db/migrate/20160104131622_remove_deprecated_columns.rb
+++ b/db/migrate/20160104131622_remove_deprecated_columns.rb
@@ -1,0 +1,6 @@
+class RemoveDeprecatedColumns < ActiveRecord::Migration
+  def change
+    remove_column :subject_workflow_counts, :set_member_subject_id
+    remove_column :classifications, :subject_ids
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -187,7 +187,6 @@ CREATE TABLE classifications (
     gold_standard boolean,
     expert_classifier integer,
     metadata jsonb DEFAULT '{}'::jsonb NOT NULL,
-    subject_ids integer[] DEFAULT '{}'::integer[],
     workflow_version text
 );
 
@@ -810,7 +809,6 @@ ALTER SEQUENCE subject_sets_workflows_id_seq OWNED BY subject_sets_workflows.id;
 
 CREATE TABLE subject_workflow_counts (
     id integer NOT NULL,
-    set_member_subject_id integer,
     workflow_id integer,
     classifications_count integer DEFAULT 0,
     created_at timestamp without time zone NOT NULL,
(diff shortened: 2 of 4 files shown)


db/migrate/20160104131622_remove_deprecated_columns.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveDeprecatedColumns < ActiveRecord::Migration
+  def change
+    remove_column :subject_workflow_counts, :set_member_subject_id
```
camallen: these migrations will run before the prod code has updated to not use the columns right?

db/structure.sql
```diff
@@ -2228,20 +2226,6 @@ CREATE UNIQUE INDEX index_subject_sets_workflows_on_workflow_id_and_subject_set_
...
---
-
-CREATE INDEX index_subject_workflow_counts_on_set_member_subject_id ON subject_workflow_counts USING btree (set_member_subject_id);
-
-
---
--- Name: index_subject_workflow_counts_on_sms_id_and_workflow_id; Type: INDEX; Schema: public; Owner: -; Tablespace: 
---
-
-CREATE UNIQUE INDEX index_subject_workflow_counts_on_sms_id_and_workflow_id ON subject_workflow_counts USING btree (set_member_subject_id, workflow_id);
```
camallen: are we losing the worfklow index as well or just the compound one here?

db/migrate/20160104131622_remove_deprecated_columns.rb
marten: only if we run the deploy after migrating, i thought we had them as two tasks in jenkins so we could deploy before migrations?

camallen: ha - i just read the note. we can choose when to migrate so this will work, sorry should have read it properly.

camallen:
LGTM

db/structure.sql
marten: Seems like there's no index on just `workflow_id`. Probably hasn't been an issue because no subject is used in many workflows (after using the `subject_id` index, we're left with probably up to 2 or 3 rows, and usually only one.

db/migrate/20160104131622_remove_deprecated_columns.rb
```diff
@@ -0,0 +1,6 @@
...
+    remove_column :classifications, :subject_ids
```
camallen: one last question...removing this column will still allow the classification -> subjects association to build correctly right?

marten: The association is already a `has-many through` so yes, reading and even assignment works fine.

PR #1569, opened 2016-01-05
fallback selection should return some data

camallen:
non-logged in queues may not replenish via the subject selector for grouped selection. So this will now instead just return some data from the workflow taking into account the group if set.

### DIFF
--- a/lib/subjects/postgresql_selection.rb
+++ b/lib/subjects/postgresql_selection.rb
@@ -17,6 +17,15 @@ module Subjects
       results.take(limit)
     end
 
+    def any_workflow_data(options={})
+      @opts = options
+      any_workflow_data_scope
+      .order(random: [:asc, :desc].sample)
+      .limit(limit)
+      .pluck("set_member_subjects.id")
+      .shuffle
+    end
+
     private
 
     def available
@@ -79,6 +88,19 @@ module Subjects
 
     def reassign_random?
       rand < Panoptes::SubjectSelection.index_rebuild_rate
+    end
+
+    def any_workflow_data_scope
+      scope = workflow.set_member_subjects
+      if workflow.grouped
+        if subject_set_id = opts[:subject_set_id]
+          scope = scope.where(subject_set_id: subject_set_id)
+        else
+          msg = "subject_set_id parameter missing for grouped workflow"
+          raise Subjects::Selector::MissingParameter.new(msg)
+        end
+      end
+      scope
     end
   end
 end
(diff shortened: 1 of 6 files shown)


lib/subjects/selector.rb
```diff
@@ -111,5 +107,16 @@ def dequeue_for_logged_in_user(sms_ids)
         DequeueSubjectQueueWorker.perform_async(workflow.id, sms_ids, queue_user.try(:id), subject_set_id)
       end
     end
+
+    def select_some_workflow_sms_ids(select_limit)
```
camallen: I'm going to refactor this, it's not where it should be.

lib/subjects/postgresql_selection.rb
```diff
@@ -17,6 +17,15 @@ def select(options={})
       results.take(limit)
     end
 
+    def any_workflow_data(options={})
+      @opts = options
```
marten: Why is opts not just a constructor param? Query methods that change state (apart from caching) are icky, and it seems like all the code ever does is `PsqlSelection.new(workflow, user).select(opts)`, so why not just do `PsqlSelection.new(workflow, user, opts).select`?

camallen: fair call - will do.

marten:
LGTM

lib/subjects/postgresql_selection.rb
camallen: going to rework as explicit method params and remove the state...reading the specs their use case is to modify the behaviour of the method at runtime.

marten: Seems like it's more to make the specs easier to write, but you could just as well do `def subject(opts); described_class.new(workflow, user, opts); end`.

For instance, consider that `@available` would need to be invalidated if `opts[:subject_set_id]` changes. But it isn't actually invalidated. So there's bugs with actually using the same object multiple times with different opts anyways.

camallen:
@marten refactored to move the setup state into the constructor but left a limit override on the method to change selection set size on the fly.

marten:
I'm going to call YAGNI on that.

camallen:
your probably right, i did need it for testing the sliding window spec [here](https://github.com/zooniverse/Panoptes/blob/4db3fca4e7fb9ab644826437c7ffc057fe36d241/spec/lib/subjects/postgresql_selection_spec.rb#L48) and i didn't want to rewrite that spec (though it does need it). 

This fixes a bug so we should get it out and then fix the spec / limit override later.

marten:
That's fine. It's not worse than the original. LGTM if you get the specs green.

camallen:
should be good to go now @marten

PR #1576, opened 2016-01-05
Only use cellect when global config and the workflow settings match

camallen:
check that cellect is on and the workflow is configured to use cellect.


marten:
LGTM

PR #1577, opened 2016-01-06
add lifecycled at migration for classifications

### LINKED ISSUE #1517: add lifecycled_at field to classification
On the event where classification saves but can't be queued for background work. We'd set the `lifecycled_at` field on successful classification lifecycle processing and add a sidetiq scheduled job to reprocess any that haven't been set.

camallen:
closes #1517 adds a lifecycled at timestamp, that is updated in the lifecycle. Also have a rake task to backfill, i wasn't sure if this will break the app adding in this col, anyway to pre-load the cols on the model to not break?

### DIFF
--- a/db/migrate/20160106120927_add_lifecycled_at_to_classifications.rb
+++ b/db/migrate/20160106120927_add_lifecycled_at_to_classifications.rb
@@ -1,0 +1,5 @@
+class AddLifecycledAtToClassifications < ActiveRecord::Migration
+  def change
+    add_column :classifications, :lifecycled_at, :timestamp, index: true
+  end
+end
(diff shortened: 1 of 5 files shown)


marten:
To close #1517 I think we should also add the sidetiq worker that retries stuff. We can do that in another branch though.

LGTM

camallen:
i should probably make that worker

camallen:
@marten is there anything we can do to stop the col add busting the query cache here?

marten:
Don't think so, apart from trying to restart puma & sidekiq as quickly as possible.

camallen:
yeah - that's what I thought too..merging.

marten:
Here's hoping https://github.com/rails/rails/pull/22170 gets merged for Rails 5.

PR #1580, opened 2016-01-06
Non lifecycled classification worker

camallen:
merge this only after #1577 and the data migration to mark all as processed is finished.

Introduces a scheduled worker to automatically reprocess classifications that are saved but not lifecycled (redis down / api failure but db persistence, etc).


camallen:
@marten the data is migrated and this should be good to go. The only problem i can see is the `Classification.where(lifecycled_at: nil)` is not hitting that partial index via rails (in prod only is on dev). Thoughts?

camallen:
this is good to go now, index is being used after db reboot.

PR #1582, opened 2016-01-06
Sanitize User login search criteria.

parrish:
Also prevents non-exact matches for short (< 3 character) queries.

At the moment, the specs have one failure, but I'd like some feedback here.

### DIFF
--- a/spec/controllers/api/v1/users_controller_spec.rb
+++ b/spec/controllers/api/v1/users_controller_spec.rb
@@ -186,76 +186,104 @@ describe Api::V1::UsersController, type: :controller do
       end
 
       describe "search" do
-        context "by display_name" do
-          context "fuzzy matching against the login field " do
-            let(:display_name) { user.display_name }
-            let(:index_options) { { search: display_name } }
+        let(:index_options) { { search: user.login } }
 
-            it "should respond with both login matches" do
-              expect(json_response[api_resource_name].length).to eq(2)
+        it "should respond with the correct user", :aggregate_failures do
+          expect(json_response[api_resource_name].length).to eq(1)
+          expect(json_response[api_resource_name][0]['login']).to eq(user.login)
+        end
+
+        context "fuzzy matching against the login field" do
+          let(:similar_user){ create :user, login: 'bill_murray', display_name: 'Bill Murray' }
+          let(:index_options) { { search: similar_user.display_name } }
+
+          it 'should respond with the user' do
+            result_id = created_instance_id(api_resource_name)
+            expect(result_id).to eq(similar_user.id.to_s)
+          end
+        end
+
+        context "non-matching display_name or login" do
+          let(:index_options) { { search: "bill murray" } }
+
+          it "should not return any data" do
+            expect(json_response[api_resource_name].length).to eq(0)
+          end
+        end
+
+        context "with a limited page size" do
+          let(:index_options) { { search: user.login, page_size: 1 } }
+
+          it "should respond with 1 item" do
+            expect(json_response[api_resource_name].length).to eq(1)
+          end
+
+          it "should respond with the correct item" do
+            result = json_response[api_resource_name][0]['login']
+            expect(result).to eq(user.login)
+          end
+        end
+
+        context "with partial strings" do
+          let(:index_options) { { search: partial } }
+
+          context "partials that don't hit any trigrams" do
+            let(:partial) { user.login[0..1] }
+
+            it "should not find any users" do
+              expect(json_response[api_resource_name].length).to eq(0)
             end
           end
 
-          context "non-matching display_name or login" do
-            let(:index_options) { { search: "bill murray" } }
+          context "partials that match the trigrams" do
+            let(:partial) { user.login[0..2] }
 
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -258,6 +239,40 @@
...
+            it 'should abort the query', :aggregate_failures do
+              expect_any_instance_of(User::ActiveRecord_Relation)
+                .to_not receive(:full_search_login)
+              expect(User).to receive(:none).and_call_original
+              get :index, index_options
+            end
+
+            # Why doesn't this pass?
+            #   login_search = query.where("lower(login) = ?", search_names.downcase)
+            # That expression doesn't modify the query criteria???
```
parrish: Super stumped here.

Maybe some wackiness coming upstream from the 'controlled_resources' or something.

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -186,25 +186,6 @@
       end
 
       describe "search" do
-        context "by display_name" do
```
parrish: Nix'ed this since we don't actually search by display_name anymore.

Curious that it wasn't failing though

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -249,7 +230,7 @@
           context "with hard to find tsvector" do
             let(:hard_name) { "S_Powell" }
             let(:hard_user) do
-              create(:user, login: hard_name, display_name: hard_name)
+              create(:user, login: hard_name)
```
parrish: Testing for login in this context.  Updated

spec/controllers/api/v1/users_controller_spec.rb
marten: I wonder if it's a result of UserSearchString inheriting from String. That usually doesn't work out too well.

lib/user_search_string.rb
```diff
@@ -0,0 +1,6 @@
+class UserSearchString < String
```
marten: I know that it should work in this particular case with only a constructor, but I'd still prefer just a class method like `UserSearchString.from(search_names)` instead of inheritance of a core type defined in C. Just to prevent someone else coming along adding a method on this class, and then being thoroughly confused at UserSearchStrings returning Strings when calling eg gsub.

spec/controllers/api/v1/users_controller_spec.rb
parrish: It's not.

``` ruby
search_by do |name, query|
  search_names = name.join(" ")

  # This is the query used for exact login matches:
  login_search = query.where("lower(login) = ?", search_names.downcase)

  if login_search.exists?
    login_search
  else
    # If there's a match, then the sanitized string is never used
    names = UserSearchString.new search_names
    names.present? ? query.full_search_login(names) : User.none
  end
end
```

lib/user_search_string.rb
parrish: > inheritance of a core type defined in C

There's nothing onerous about this code.  Very large portions of Ruby are defined (at least partially) in C.  It's also a very common practice in Ruby to inherit from primitives to create a specialized behavior.

> being thoroughly confused at UserSearchStrings returning Strings

`UserSearchString.new` returns a `UserSearchString`, which `.is_a?(String)`

marten: True, but it's not as if this class prevents you from having search strings that aren't "valid": `UserSearchString.new("@foo")` will return something equal to `"foo"`, but `UserSearchString.new("foo").prepend("@")` will get you one that is unfiltered. Inheriting from String got you a load of methods that allow you to change the string "unsafely". Not just the ones that modify one in-place either. Just plain `UserSearchString.new("@foo").gsub(/^/, "@")` will return a *new* UserSearchString that still has the at-symbol in front, just because it doesn't call the initializer again.

I know nobody will probably try to do this. Until one morning when I haven't had coffee and I do. 

All you're really trying to do is converting a user-input to something for internal use. That's just a function that takes a string, and returns a string. Why complicate things with something that works only if you know all the pitfalls?

camallen: I prefer the class method approach on this as well, `UserSearchString.from(search_names)`.

spec/controllers/api/v1/users_controller_spec.rb
camallen: it wasn't failing since the display name matches the login trigram threshold based on factory data. I left it in to show that searching with different attribute inputs can still match the login trigram data based on how similar the login / display_name data actually is, e.g camallen / cam allen

spec/controllers/api/v1/users_controller_spec.rb
camallen: It's just the build order for the before all index get request, try:
```
let(:short_user){ create :user, login: 'me' }
let(:index_options){ { search: short_user.login } }
```

parrish:
Updated to fix the failing spec and excluded the controversial OOP

camallen:
LGTM

PR #1583, opened 2016-01-07
Optimize unseen query

camallen:
convert the query to arel and use a not exists subselect with unnest instead of a not in any array comparator.


marten:
LGTM although I think if/when I ever have to debug this, I'll probably have to do a .to_sql first. :P But Arel allows us to keep working on top of by_workflow, which I do like.

camallen:
yeah - i was .to_sql'n a lot on this one. We need to return scopes on this so we can use it with grouping, priority / random ordering.

PR #1585, opened 2016-01-07
Add partial index concurrently to lifecycled_at

camallen:
merge after #1584 - this should build a partial index on lifecycled_at concurrently outside a transaction block


marten:
LGTM

PR #1588, opened 2016-01-08
Reload cellect vs retire cellect

camallen:
notify the cellect client of retire events and don't force cellect to reload the whole workflow.

### DIFF
--- a/app/workers/retire_cellect_worker.rb
+++ b/app/workers/retire_cellect_worker.rb
@@ -1,0 +1,15 @@
+class RetireCellectWorker
+  include Sidekiq::Worker
+
+  def perform(subject_id, workflow_id)
+    workflow = Workflow.find(workflow_id)
+    if Panoptes.use_cellect?(workflow)
+      smses = workflow.set_member_subjects.where(subject_id: subject_id)
+      smses.each do |sms|
+          Subjects::CellectClient
+          .remove_subject(subject_id, workflow_id, sms.subject_set_id)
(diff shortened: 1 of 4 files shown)


camallen:
@marten be good to get this out asap to stop the load on cellect / db

app/workers/retire_cellect_worker.rb
```diff
@@ -0,0 +1,15 @@
...
+  def perform(subject_id, workflow_id)
+    workflow = Workflow.find(workflow_id)
+    if Panoptes.use_cellect?(workflow)
+      smses = workflow.set_member_subjects.where(subject_id: subject_id)
+      smses.each do |sms|
+          Subjects::CellectClient
+          .remove_subject(subject_id, workflow_id, sms.subject_set_id)
+      end
+    end
+  rescue Subjects::CellectClient::ConnectionError, ActiveRecord::RecordNotFound
```
parrish: Always use explicit requires in Sidekiq workers for non-app constant paths or you'll raise autoload exceptions the first time it's accessed.

parrish:
I was trying to figure out why Cellect was running under a heavy CPU load.  This would certainly explain it.

LGTM other than the missing `require 'subjects/cellect_client'` in the worker.

parrish:
@camallen Since this needs to go, I'll just merge and deal with the autoload errors.

camallen:
Sweet thanks.

PR #1589, opened 2016-01-08
ignore collect connection errors on classification lifecycle

camallen:
just set and forget cellect ignore any connection errors


marten:
Perhaps this should be moved out to it's own worker so that it can retry independently, so that we don't end up showing duplicates. Alternatively, Cellect could listen to Kafka to get this info instead of Panoptes pushing it explicitly. I'll make a ticket about this.

I'm still merging this first though.

PR #1595, opened 2016-01-09
use the gold standard indexes

camallen:
make sure the queries use the gold standard indexes on classifications and workflow


marten:
Interesting, weird little edge-case with PG? Anyways, LGTM.

camallen:
yeah the sql resulted in `= t`which didn't match the index, i was a little surprised.

PR #1596, opened 2016-01-09
Lifecycle at window

### LINKED ISSUE #1593: Lifecycled at worker should have a window period
They should not pickup currently queued but not lifecycled, so a delay of like the last 15 mins or so should fix it. We could also skip re life cycling if the field is present.

camallen:
closes #1593 - only back process classifications that were skipped due to redis / api errors by ignoring the live stream via a window period (configurable via the ENV["LIVE_WINDOW"] var).

### DIFF
--- a/app/workers/classification_worker.rb
+++ b/app/workers/classification_worker.rb
@@ -9,7 +9,6 @@ class ClassificationWorker
 
   def perform(id, action)
     @classification = Classification.find(id)
-    return if classification.lifecycled_at
     @action = action
     lifecycle_classification
   end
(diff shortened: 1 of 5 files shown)


config/initializers/panoptes.rb
```diff
@@ -0,0 +1,5 @@
+module Panoptes
+  def self.lifecycled_live_window
+    @lifecycled_live_window ||= (ENV["LIVE_WINDOW"] || 15)
```
marten: I think you'll need to `to_i` that ENV var?

marten: Actually, probably `Integer(ENV["LIVE_WINDOW"] || 15)` might be better than `to_i`, since it'll raise if it's not parseable.

app/workers/classification_worker.rb
```diff
@@ -5,16 +5,28 @@ class ClassificationWorker
 
   sidekiq_options queue: :high
 
+  attr_reader :classification, :action
+
   def perform(id, action)
-    classification_lifecycle = ClassificationLifecycle.new(Classification.find(id))
+    @classification = Classification.find(id)
+    return if classification.lifecycled_at
```
marten: Won't this also return in case of `action == :update` (in which case it's been lifecycled before).

config/initializers/panoptes.rb
camallen: good one

app/workers/classification_worker.rb
camallen: yeah i think it will - nice catch again.

PR #1600, opened 2016-01-11
Audit indexes

camallen:
remove unused / duplicates AND update index defs / code to hit indexes

### DIFF
--- a/app/controllers/concerns/filter_by_metadata.rb
+++ b/app/controllers/concerns/filter_by_metadata.rb
@@ -11,7 +11,8 @@ module FilterByMetadata
     meta_params.each do |meta_key|
       _, key = meta_key.to_s.split(".")
       value = params.delete(meta_key)
-      @controlled_resources = controlled_resources.where("\"#{resource_sym}\".\"metadata\"->>:key = :value", key: key, value: value)
+      @controlled_resources = controlled_resources
+      .where("\"#{resource_sym}\".\"metadata\" @> ? ", { key => value }.to_json)
     end
   end
 end
--- a/db/migrate/20160111112417_remove_unused_indexes.rb
+++ b/db/migrate/20160111112417_remove_unused_indexes.rb
@@ -1,0 +1,53 @@
+class RemoveUnusedIndexes < ActiveRecord::Migration
+  def up
+    #was put in place for the aggregation engine, that never used it!
+    remove_index :classifications, column: :workflow_version
+    #this will be used soon for HHMI project group classificaton comparison
+    remove_index :classifications, column: :user_group_id
+    #this compound index is never used but the user_id one is
+    remove_index :memberships, column: [ :user_id, :identity ]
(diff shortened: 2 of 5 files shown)


db/migrate/20160111112417_remove_unused_indexes.rb
```diff
@@ -0,0 +1,53 @@
...
+    remove_index :memberships, column: [ :user_id, :identity ]
+    #not currently used and should only be used with a more restrictive filter
+    # e.g. user_group_id: X and overlap or contains roles..
+    remove_index :access_control_lists, column: :roles
+    #aggregations only make sense in the context of a subejct, workflow
+    # use the compound index instead
+    remove_index :aggregations, column: :subject_id
+    remove_index :subjects, column: :zooniverse_id
+    #index is not used as it's only for sorting with more restrictive filter
+    remove_index :user_project_preferences, column: :updated_at
```
parrish: Won't this be used now that this is merged? https://github.com/zooniverse/Panoptes-Front-End/pull/2094/files#diff-4e943d3edc8de512994b6a0031c1ed97R82

app/controllers/concerns/filter_by_metadata.rb
```diff
@@ -11,7 +11,8 @@ def filter_by_metadata
     meta_params.each do |meta_key|
       _, key = meta_key.to_s.split(".")
       value = params.delete(meta_key)
-      @controlled_resources = controlled_resources.where("\"#{resource_sym}\".\"metadata\"->>:key = :value", key: key, value: value)
+      @controlled_resources = controlled_resources
+      .where("\"#{resource_sym}\".\"metadata\" @> ? ", { key => value }.to_json)
```
parrish: Is this for subject_sets.metadata?

spec/controllers/api/v1/tags_controller_spec.rb
```diff
@@ -21,6 +21,18 @@
     let(:n_visible) { 2 }
 
     it_behaves_like "is indexable", false
+
+    describe "search by name" do
+      let(:resource) { create(:tag, name: "bowie") }
```
parrish: ;(

db/migrate/20160111112417_remove_unused_indexes.rb
camallen: Only if the api requests all project_preferences without the user_id filter. My testing was showing that index was more restrictive and the sort clause was using quick sort and missing the index. I removed it since it wasn't being used (186MB) but could add it back if it's needed. Don't think we ever request without the user_id but it could...

camallen: Maybe just add back after SGL.

app/controllers/concerns/filter_by_metadata.rb
camallen: yeah - that was the use case i was testing.

parrish:
Cool.  Looks good to me.

PR #1601, opened 2016-01-11
raise error when no data available for selection

### LINKED ISSUE #1586: raise no data error when none to select from
subject selector should raise no data error when there is no workflow data to select.

camallen:
closes #1586 - error gracefully with a useful message.


parrish:
:+1:

PR #1603, opened 2016-01-12
Add data publish worker

### LINKED ISSUE #1599: publishing to down kafka / event streams
@marten Should we move them to background workers that can fail at will? instead of possibly delaying the lifecycle events due to busted zookeeper  / network connections?

camallen:
closes #1599  - don't allow kafka problems to block the classification lifecycle. Not 100% on whether we should split these different jobs into separate workers to avoid a point of failure to create duplicate messages in kafka.


app/workers/publish_data_worker.rb
```diff
@@ -0,0 +1,16 @@
+class PublishDataWorker
```
marten: This seems a bit excessive. Without benchmarks, I don't really believe looking up the classification 4 times (lifecycle, this worker, two workers queued by this one) will actually be faster than inserting two items in the queue from the lifecycle worker and losing one classification lookup.

camallen: good call, it was. I've updated to just call the workers from the lifecycle now.

PR #1604, opened 2016-01-12
move processing off to the data worker

camallen:
move as much as we can off the api nodes and onto the data worker


marten:
LGTM

PR #1608, opened 2016-01-12
meet the guard criteria

camallen:
dont publish data unless it's really retired


marten:
LGMT

camallen:
getting this out before the show.

PR #1610, opened 2016-01-13
Deduplicate UPP and add unique index

marten:


db/migrate/20160113120732_dedupe_upps.rb
```diff
@@ -0,0 +1,19 @@
...
+          SELECT id
+          FROM (
+            SELECT id, ROW_NUMBER() OVER (PARTITION BY user_id, project_id ORDER BY id ASC) AS row
+            FROM user_project_preferences
+          ) dupes
+          WHERE dupes.row > 1
+        );
+      SQL
+
+      add_index :user_project_preferences, [:user_id, :project_id], unique: true
```
camallen: can we use this for queries instead of the 2 x single attribute indexes?

marten: I think we can get rid of the project one, but the user one we might want to keep because of https://github.com/zooniverse/Panoptes/blob/master/app/controllers/emails_controller.rb#L50

camallen: be good to get rid of any if we can, that project_id one is 187MB, user_is is 33MB

marten: do you want to take this one in your massive index removal?

camallen: yeah good call.

camallen:
how long does this take to run on your machine?

camallen:
LGTM - if it'll deploy quickly lets get it out.

marten:
A few seconds.

marten:
Done, index takes < 2 secs to build on my machine. The delete statement is about the same.

PR #1615, opened 2016-01-13
Add a cache for model version

marten:
Instantiating old Papertrail versions is slow, and we need this for
every workflow we return. Caching this makes it faster.

### DIFF
--- a/app/serializers/workflow_serializer.rb
+++ b/app/serializers/workflow_serializer.rb
@@ -25,11 +25,15 @@ class WorkflowSerializer
   end
 
   def version
-    "#{ModelVersion.version_number(@model)}.#{ModelVersion.version_number(content)}"
+    "#{@model.current_version_number}.#{content_version}"
   end
 
   def content_language
     content.language if content
+  end
+
+  def content_version
+    content.try(:current_version_number) || ModelVersion.default_version_num
   end
 
   def tasks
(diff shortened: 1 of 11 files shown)


app/serializers/workflow_serializer.rb
```diff
@@ -25,13 +25,17 @@ def self.links
...
-    "#{ModelVersion.version_number(@model)}.#{ModelVersion.version_number(content)}"
+    "#{@model.current_version_number}.#{content_version}"
   end
 
   def content_language
     content.language if content
   end
 
+  def content_version
+    content.try(:current_version_number) || ModelVersion.default_version_num
```
marten: This is the main bit I'd like to see refactored/cleaner, but didn't have a good idea how. If nobody else does, let's just leave it, but have a think.

camallen:
rebase?

app/serializers/workflow_serializer.rb
camallen: is the default_version_num method defined?

marten:
rebased.

app/serializers/workflow_serializer.rb
marten: Yeah that was already used by `ModelVersion.version_number`

camallen:
LGTM - lets check there are no locks on the workflows table before running this.

PR #1616, opened 2016-01-13
Remove indexes

camallen:
get rid of indexes where we can, some classification ones will have to be rebuilt (possibly with partial wheres).


marten:
LGTM

PR #1617, opened 2016-01-13
avoid the association assignment via link_params

camallen:
if the user has a valid token they'll have the correct it


marten:
LGTM

PR #1619, opened 2016-01-13
Adds explicit requires for non-eager-autoloaded constants.

parrish:
Resolves application boot-time errors like "RuntimeError: Circular dependency detected while autoloading constant <Whatever>"

Probably not high priority, but crashing on boot is never good.

### DIFF
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -1,3 +1,5 @@
+require 'subjects/set_member_subject_selector'
+
 class SetMemberSubject < ActiveRecord::Base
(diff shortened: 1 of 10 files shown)


marten:
I've been tempted to just throw `lib/` out of the eager/autoloading path and set up a single file that requires everything in an order that works.

parrish:
It's also worth mentioning that constant autoloading will work better with absolute scope: `::ConstantName` from a Sidekiq worker.

marten:
Anyways, LGTM

PR #1620, opened 2016-01-13
disable increment_counters for SGL

camallen:
turn these off for now...we can look at getting a better counter working later.


camallen:
Should i just put an if guard for the project on this instead?

parrish:
Looks good.  I'll merge when the test passes

parrish:
You mean an `if ENV['WHATEVER']` statement?  I think this is fine.

camallen:
just skipping if the project_id == SGL project id so we don't have to recount everything later..

parrish:
Ah.  Yeah, that'd be preferable

camallen:
closed in favour of #1621

PR #1625, opened 2016-01-13
don't dequque all for SGL

camallen:
avoid locking up the workers

### DIFF
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -7,7 +7,9 @@ class RetirementWorker
     count = SubjectWorkflowCount.find(count_id)
     if count.retire?
       count.retire! do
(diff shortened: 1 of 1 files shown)


parrish:
Should we decrease the queue size in this case?

parrish:
Part of the problem is that there's a race condition on trying to dequeue the same subject multiple times.

Maybe do a check that it's active before trying to dequeue it?

camallen:
I'm not 100% sure about the flow on effects of shrinking the queues, to say 10/20, halving could work. 

A couple of page reloads (10 selected per time) and they'll be empty but the only refill on classification submission. So an empty queue will fallback to PG selection. It may be all fine while we've got a bigger db but when we scale down there may be ripple effects. Thoughts?

parrish:
True.  I'll merge this and keep an eye on how it behaves.

PR #1626, opened 2016-01-14
Move project/workflow classifications counter updates to ratelimited

marten:
@parrish should we do `track_rejected` on this?

### DIFF
--- a/app/workers/project_classifications_count_worker.rb
+++ b/app/workers/project_classifications_count_worker.rb
@@ -1,0 +1,25 @@
+class ProjectClassificationsCountWorker
+  include Sidekiq::Worker
+
+  sidekiq_options congestion: {
+    interval: 60,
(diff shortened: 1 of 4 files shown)


app/workers/project_classifications_count_worker.rb
```diff
@@ -0,0 +1,25 @@
+class ProjectClassificationsCountWorker
+  include Sidekiq::Worker
+
+  sidekiq_options congestion: {
+    interval: 60,
```
camallen: Do we want to make these configurable via env variables with defaults? E.g. [this](https://github.com/zooniverse/Panoptes/blob/233d839b0c1f3b4d2759beede4f4aab658e0f159/config/initializers/panoptes.rb#L9)

camallen:
LGTM - we can get this out and look at changes in another PR.

PR #1630, opened 2016-01-14
Disable dequeue workers and reduce queue sizes

camallen:
avoid locking updates on the subject queue table with large dequeue scopes. Also changed the default queue size to be 2 x the default limit selections to reduce the chance of retired data being in them. This may push more queries into the db when selection happens on an empty queue. Something to keep in mind when running on a smaller db instance.


marten:
LGTM

PR #1631, opened 2016-01-14
Don't return retired subjects from queue

marten:
This should still fall back when all are retired.

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -33,12 +33,25 @@ module Subjects
 
     def sms_ids_from_queue(queue)
       sms_ids = queue.next_subjects(subjects_page_size)
-      if sms_ids.blank?
+      non_retired_ids = filter_non_retired(sms_ids)
+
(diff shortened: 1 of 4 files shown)


lib/subjects/selector.rb
```diff
@@ -33,14 +33,27 @@ def selected_subjects(queue)
...
     end
 
+    def filter_non_retired(sms_ids)
+      retired_ids = SetMemberSubject
+        .joins("INNER JOIN subject_workflow_counts ON subject_workflow_counts.subject_id = set_member_subjects.subject_id")
+        .where("subject_workflow_counts.retired_at IS NOT NULL")
+        .where(set_member_subjects: {id: sms_ids})
+        .pluck(:id)
+
+      retired_ids = Set.new(retired_ids)
```
camallen: a thought, could you do a distinct on the query to get a uniq array and then just do array diff operation?

camallen:
LGTM

PR #1632, opened 2016-01-14
Guard retirement

parrish:
Scenario is:
  - 3 people classify the same subject simultaneously.
  - The subject will retire after the next classification
  - The subject will attempt to retire and dequeue 3 times

This just adds a check to ensure it's not already retired.

I'm sure it's possible for a race condition to allow multiple retirements in cases of very high concurrency, but this should take care of the majority of them.

### DIFF
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -5,7 +5,7 @@ class RetirementWorker
 
   def perform(count_id)
     count = SubjectWorkflowCount.find(count_id)
-    if count.retire?
+    if count.retire? && !count.retired?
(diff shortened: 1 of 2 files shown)


parrish:
Also, in regards to #1630, everything still holds other than the dequeuing.

camallen:
LGTM. @parrish Can you elaborate on the "everything still holds"?

parrish:
In terms of preventing multiple attempts to retire the same subject.  E.g. it won't send multiple remove requests to cellect, it won't publish it multiple times, etc.

marten:
LGTM

PR #1633, opened 2016-01-14
Always pass in serialization context

marten:


camallen:
LGTM - i don't know why we weren't doing this by default via the selector.

marten:
@parrish has until the tests finish to come up with a reason :P

parrish:
Seems alright to me, it'll add some extra overhead to run those queries though.  Probably worth it in this case.

PR #1638, opened 2016-01-22
Upgrade to Ruby 2.3.0

parrish:
Primarily to fix [this bug](https://app.honeybadger.io/projects/40595/faults/23124244) and the performance hit it causes.

2.3 also includes [the lonely operator](https://bugs.ruby-lang.org/issues/11537) and other performance improvements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
   fast_finish: true
   allow_failures:
   include:
-    - rvm: 2.2.1
+    - rvm: 2.3.0
 
 services:
   - postgresql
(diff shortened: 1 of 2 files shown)


camallen:
LGTM, how does it fix that bug? 

Lonely operator looks interesting, i wonder what Sandi Metz would say about that.

marten:
I've previously used https://github.com/raganwald/andand for that lonely operator. It's a blessing and a curse really. You have to be really careful with it, or you'll just end up pushing nils everywhere and you can't debug shit anymore. It's awesome on the debug console though.

parrish:
This is the bug in question: https://bugs.ruby-lang.org/issues/10871

parrish:
This is the bug in question: https://bugs.ruby-lang.org/issues/10871

PR #1642, opened 2016-01-26
split the travis build up on different concerns

camallen:
Split the spec tasks up to the concurrent build limit on travis (currently 5) to decrease the total run time (prev 23m, now ~7m).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,25 +2,41 @@ language: ruby
 cache: bundler
 sudo: false
 
+bundler_args: --without development production
+
+before_install:
+  - 'echo ''gem: --no-ri --no-rdoc'' > ~/.gemrc'
+
 before_script: &before_script
   - psql -c 'create database travis_ci_test;' -U postgres
   - mkdir -p tmp
-  - RACK_ENV=test bundle exec rake configure:travis db:setup
+  - bundle exec rake configure:travis db:setup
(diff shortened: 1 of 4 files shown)


marten:
The main thought I had is that we'll need to be pretty vigilant about adding directories (for instance I'm adding `spec/operations` in a PR, which wouldn't get picked up on CI. Any thoughts on how to fail the build in such case?

camallen:
Yeah, a good point. I can't see an easy way to match and split up the build. Vigilance? We could try gems that do this for us too.

marten:
Maybe a step that just checks the directories against a known list, and fails if it doesn't match? That'd be good enough for me.

marten:
Hey, what happened to the Travis build status on this branch? All I see now if Hound's status.

camallen:
Not sure, I just rebased this to start adding the checks.

camallen:
Ok - this should be good to go. 

I've added a list of know spec directories and compared this to a current spec dir listing before running the specs, see [this run](https://travis-ci.org/zooniverse/Panoptes/builds/109367464) for a failure.

The known file entries and the actual spec cmds are not enforced. So when adding new spec directories the updating user will have to include new spec dirs in the travis commands **AND** add them to the `known_dirs.txt` file.

PR #1645, opened 2016-02-01
Ruby docker base update

camallen:
linked to https://github.com/zooniverse/docker-ruby/pull/3 - this should be fine to merge before the base PR is updated.

Install pg gem dep locally. Remove jruby gems and the pgclient install on the docker dev image (happy to put this one back in if peeps use it).


marten:
LGTM

PR #1648, opened 2016-02-02
Add libpg-dev to Dockerfile

astopy:
Was removed by zooniverse/docker-ruby#3

Don't merge yet... just waiting for the Docker build to finish to make sure that's everything we need.


camallen:
I already fixed this in #1645

PR #1650, opened 2016-02-04
increase security of ip hashing

### LINKED ISSUE #1649: Hashing in dumps in theory not strong enough
[This line](https://github.com/zooniverse/Panoptes/blob/master/lib/formatter/csv/classification.rb#L18) means that the search space for a salt is pretty limited. And since it's pretty easy to know both the unhashed and hashed version of your own IP, you can feasibly search for the salt.

camallen:
closes #1649 - use a random number lookup on the ip string instead of a hash function of the ip string

### DIFF
--- a/app/workers/classifications_dump_worker.rb
+++ b/app/workers/classifications_dump_worker.rb
@@ -8,10 +8,12 @@ class ClassificationsDumpWorker
 
   sidekiq_options queue: :data_high
 
-  def perform_dump(obfuscate_private_details=true)
+  def perform_dump(private=true)
     CSV.open(csv_file_path, 'wb') do |csv|
       cache = ClassificationDumpCache.new
-      formatter = Formatter::Csv::Classification.new(project, cache, obfuscate_private_details: obfuscate_private_details)
+      formatter = Formatter::Csv::Classification.new(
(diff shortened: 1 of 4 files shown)


app/workers/classifications_dump_worker.rb
```diff
@@ -8,10 +8,12 @@ class ClassificationsDumpWorker
 
   sidekiq_options queue: :data_high
 
-  def perform_dump(obfuscate_private_details=true)
+  def perform_dump(private=true)
```
marten: `private=true` sounds more like "this is a private dump" or "yes, include private stuff". Maybe `anonymize=true` if you're searching for something shorter?

lib/classification_dump_cache.rb
```diff
@@ -40,4 +40,12 @@ def workflow_content_at_version(workflow_content_id, version)
       old_version || workflow_content
     end
   end
+
+  def secure_user_ip(ip_string)
+    if rand_hex = @secure_ip_lookup[ip_string]
```
marten: Hmm, why not just `@secure_ip_lookup[ip_string] ||= SecureRandom.hex(10)`?

marten:
Minor comments, but LGTM

PR #1652, opened 2016-02-05
Allow joining groups with token

marten:


db/migrate/20160202155708_add_join_token_to_user_groups.rb
```diff
@@ -0,0 +1,5 @@
+class AddJoinTokenToUserGroups < ActiveRecord::Migration
+  def change
+    add_column :user_groups, :join_token, :string
```
marten: Should these tokens be unique across the table? For now I figured they don't need to, saves an index and is fine as long as we always check a token against a specific given user_group_id.

app/controllers/api/v1/memberships_controller.rb
```diff
@@ -4,9 +4,18 @@ class Api::V1::MembershipsController < Api::ApiController
   resource_actions :index, :show, :create, :update, :deactivate
   schema_type :strong_params
 
-  allowed_params :create, links: [:user, :user_group]
   allowed_params :update, :state
 
+  def create
```
marten: Obviously we'd extract this when we get more actions like this. For now I'm not wanting to over-engineer.

app/controllers/api/v1/memberships_controller.rb
```diff
@@ -4,9 +4,18 @@ class Api::V1::MembershipsController < Api::ApiController
   resource_actions :index, :show, :create, :update, :deactivate
   schema_type :strong_params
 
-  allowed_params :create, links: [:user, :user_group]
```
marten: Note: this is safe because ActiveInteraction whitelists arguments to only the filters (eg `string :join_token`) specified by the operation.

app/models/user_group.rb
```diff
@@ -82,9 +83,25 @@ def identity?
     !!memberships.where(identity: true).pluck(:identity).first
   end
 
+  def verify_join_token(token_to_verify)
+    join_token.present? && join_token == token_to_verify
+  end
+
+  def public?
```
marten: This and `#member?` are no longer needed after having removed the old behaviour. Remove them.

app/operations/memberships/create.rb
```diff
@@ -0,0 +1,26 @@
...
+  class Create < Operation
+    string :join_token, default: nil
+
+    hash :links do
+      integer :user
+      integer :user_group
+    end
+
+    def execute
+      raise Unauthenticated unless api_user.logged_in?
```
camallen: I `api_user` comes from the api_controller?

marten: Yeah, same as used by relation manager and json controller modules.

app/operations/memberships/create.rb
```diff
@@ -0,0 +1,26 @@
...
+      raise Unauthorized unless user_group.verify_join_token(join_token)
+      raise Unauthorized unless user.id == api_user.id
+
+      Membership.create! user: api_user.user, user_group: user_group, state: :active
+    end
+
+    def user
+      User.find(links["user"])
```
camallen: These will raise if not found. Just checking we don't want to catch this and propagate up to the api controller.

camallen: ignore this, should have kept reading.

app/operations/memberships/create.rb
camallen: ignore again - should have kept reading.

db/migrate/20160202155708_add_join_token_to_user_groups.rb
camallen: Chances of clashing are pretty slim i assume. If we ever searching the db for this token then we'll need an index and may as well make it unique.

camallen:
LGTM - sorry for the noise. Really like the service objects pattern - nice, clean, uncoupled and readable code.

simoneduca:
woot!

PR #1655, opened 2016-02-05
Make debugging with pry more awesome

marten:
* byebug adds commands like `step` (step-into), `next` (step-over), `continue`, `finish`. Also adds ability to set more breakpoints once you're in a pry session.

* stack_explorer adds `show-stack`, `up` and `down` to inspect state higher up in the call stack. byebug adds some of this, but that only works with a stack that starts at the point the pry started, so you can't look at state higher up the call chain, hence this additional gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,8 @@ end
 group :development, :test do
   gem 'foreman'
   gem 'pry-rails'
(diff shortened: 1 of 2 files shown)


camallen:
Nice, LGTM

PR #1656, opened 2016-02-05
Fetch and use latest AWS proxy ips on container build

marten:
We could do fancier and subscribe to the SNS topic that publishes changes, but this probably works acceptably for our purposes. Just have to keep deploying ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 
 # Ignore all config files
 /config/*.yml
+/config/aws_ips.json
 
 .vagrant/
 *.swp
--- a/config/initializers/loadbalancer_ips.rb
+++ b/config/initializers/loadbalancer_ips.rb
@@ -1,0 +1,13 @@
+aws_ips_file_path = Rails.root.join("config/aws_ips.json")
+
+# File should always exist in prod/stag but allow testing this in dev mode if file exists
+if Rails.env.production? or Rails.env.staging? or File.exists?(aws_ips_file_path)
+  json = File.read(aws_ips_file_path)
+  data = JSON.load(json)
+
+  proxy_ips = data.fetch("prefixes")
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@
 
 # Ignore all config files
 /config/*.yml
+/config/aws_ips.json
```
camallen: should we have a `.dockerignore` entry for this as well?

config/initializers/loadbalancer_ips.rb
```diff
@@ -0,0 +1,13 @@
+aws_ips_file_path = Rails.root.join("config/aws_ips.json")
+
+# File should always exist in prod/stag but allow testing this in dev mode if file exists
+if Rails.env.production? or Rails.env.staging? or File.exists?(aws_ips_file_path)
+  json = File.read(aws_ips_file_path)
```
camallen: if the file doesn't exist (though it should i see), this will stop the app booting right?

.gitignore
marten: No, I figured we'd compile it into the docker container image so that it's at least consistent for all instances (and so that we can't have AWS's site be down when we're starting. Either it's there, or we docker's build fails, was my philosophy.

config/initializers/loadbalancer_ips.rb
marten: Yes, but worst-case that should just prevent a deploy, which is probably better than this breaking and us storing ELB IPs again without getting notified.

marten: At least, I assume our deploys will fail and roll-back if new instances don't come up.

camallen: Yeah good point.

PR #1660, opened 2016-02-09
Add user group ids to classifications

marten:
This is so we can associate classifications with user groups (in post-processing done by education api)

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -86,6 +86,7 @@ class ClassificationLifecycle
     mark_expert_classifier
     add_seen_before_for_user
     add_project_live_state
+    add_user_groups
     add_lifecycled_at
     classification.save!
   end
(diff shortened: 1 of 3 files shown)


lib/classification_lifecycle.rb
```diff
@@ -118,6 +119,10 @@ def add_project_live_state
     update_classification_metadata(:live_project, project.live)
   end
 
+  def add_user_groups
+    update_classification_metadata(:user_group_ids, user.non_identity_user_group_ids)
```
camallen: I thought the client would supply this info to capture the correct context of the classification. I.e. i was classifying subject X from assignment Y in user_group P..

marten: Assignments aren't coming until later this year. But because Amy is (going to be) working on a PFE interface for creating / joining groups, I imagine we'll have groups of friends becoming groups, so having the groups for all projects might give neat insights with respect to clusters of users? Doing it in the backend also means we get it for custom frontends for free.

camallen:
LGTM - just need a spec change to match_array.

PR #1663, opened 2016-02-10
consistent event time format

camallen:
ensure a consistent event time format before serializing to kafka

### DIFF
--- a/app/workers/publish_classification_event_worker.rb
+++ b/app/workers/publish_classification_event_worker.rb
@@ -12,7 +12,7 @@ class PublishClassificationEventWorker
     if classification.complete?
       EventStream.push("classification",
         event_id: "classification-#{classification.id}",
-        event_time: classification.updated_at.iso8601,
+        event_time: classification.updated_at,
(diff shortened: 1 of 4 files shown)


app/workers/publish_classification_event_worker.rb
```diff
@@ -12,7 +12,7 @@ def perform(classification_id)
     if classification.complete?
       EventStream.push("classification",
         event_id: "classification-#{classification.id}",
-        event_time: "#{classification.updated_at}",
+        event_time: classification.updated_at.iso8601,
```
marten: Now you end up converting to a string, parsing it back, and then converting it to a string again. Can we not just require this to just be a Time instance, and always let EventStream call `.iso8601` on it?

camallen: Yeah - i didn't like it either but i wanted to use `a_hash_including` spec which fails when passing a datetime object, I'll rework it.

camallen:
@marten should be ready to go now

PR #1669, opened 2016-02-15
Serialize join_token to group admins

marten:


app/models/user_group.rb
```diff
@@ -79,6 +79,14 @@ def owns?(resource)
                             resource_type: resource.class.to_s)
   end
 
+  def has_admin?(user)
+    membership_for_user(user).try { |m| m.roles.include?("group_admin") }
+  end
+
+  def membership_for_user(user)
+    memberships.find_by(user_id: user.id)
```
camallen: i assume this will find the right one from memberships?

app/models/user_group.rb
```diff
@@ -79,6 +79,14 @@ def owns?(resource)
                             resource_type: resource.class.to_s)
   end
 
+  def has_admin?(user)
+    membership_for_user(user).try { |m| m.roles.include?("group_admin") }
```
camallen: should this method belong in the `memberships` model?

app/models/user_group.rb
```diff
@@ -79,6 +79,14 @@ def owns?(resource)
...
+  end
+
+  def membership_for_user(user)
+    memberships.find_by(user_id: user.id)
+  end
+
   def identity?
     !!memberships.where(identity: true).pluck(:identity).first
```
camallen: Just looking at this and thinking would exists work here?

camallen:
Looks good, just a couple of comments.

app/models/user_group.rb
marten: Membership has a unique index on (user_id, user_group_id) so can only ever be one membership.

PR #1671, opened 2016-02-15
Kinesis evaluation

marten:
**Please note that this will also start pushing `user_ip` into Kafka's `classifications` topic** (This should be fine, Nero is the only one reading from that topic, since ZooEventStats is using the `events` topic.

Both Kinesis and Kafka use the same serializer. I'm doing this so that we can stop having a separate `PublishClassificationEventWorker` (if/when we migrate), instead moving the onus on the stream consumer to pluck items out of the global stream that it cares about (by looking at `type` field) and choose how to publish that. I'll make that change when I swap ZooEventStats over from Kafka to Kinesis.

### DIFF
--- a/app/workers/publish_classification_worker.rb
+++ b/app/workers/publish_classification_worker.rb
@@ -11,16 +11,31 @@ class PublishClassificationWorker
   def perform(classification_id)
     @classification = Classification.find(classification_id)
     if classification.complete?
-      payload = [classification.project.id, classification_json]
-      MultiKafkaProducer.publish("classifications", payload)
+      publish_to_kafka!
+      publish_to_kinesis!
     end
   rescue ActiveRecord::RecordNotFound
   end
 
   private
 
-  def classification_json
-    KafkaClassificationSerializer
-    .serialize(classification, include: ['subjects']).to_json
+  def serialized_classification
+    @serialized_classification ||= KafkaClassificationSerializer.serialize(classification, include: ['subjects']).as_json
+  end
+
+  def publish_to_kafka!
+    MultiKafkaProducer.publish("classifications", kafka_payload)
+  end
+
+  def kafka_payload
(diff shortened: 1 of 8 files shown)


app/workers/publish_classification_worker.rb
```diff
@@ -11,16 +11,31 @@ class PublishClassificationWorker
...
+  def kafka_payload
+    [classification.project.id, serialized_classification.to_json]
+  end
+
+  def publish_to_kinesis!
+    KinesisPublisher.publish("classification", classification.workflow_id, kinesis_payload)
+  end
+
+  def kinesis_payload
+    serialized_classification["classifications"][0]
```
marten: I figured it'd be much better to push multiple classifications as multiple events in the stream, rather than always having a payload be an array.

lib/kinesis_publisher.rb
```diff
@@ -0,0 +1,35 @@
...
+  def self.client
+    @client
+  end
+
+  def self.client=(client)
+    @client = client
+  end
+
+  def self.stream_name
+    ENV.fetch('KINESIS_STREAM')
```
camallen: Should we have a default on this if the ENV var isn't set?

spec/workers/publish_retirement_event_worker_spec.rb
```diff
@@ -25,5 +25,11 @@
         .with("workflow_counters", payload_expectation)
       worker.perform(workflow.id)
     end
+
+    it "should publish via EventStream" do
```
camallen: nitpicky - maybe change the spec name to match the expectation, i.e. publish via kinesis.

camallen:
LGTM (pending specs) - i liked the format_data with version tracking nice idea. 

I think Nero / events are the only consumers of the panoptes kafka stream and events strips private fields (ip) so this should be fine.

PR #1672, opened 2016-02-15
fixes for getting started

camallen:
Audit the getting started section in the readme and make sure it works. explicit 9.4 postgres and make sure we have a pg_dump tool for db setup / migrations


marten:
LGTM

PR #1675, opened 2016-02-16
Add linked models to Kinesis

marten:
I started working towards having JSON-API linked models be in the stream too (Nero uses that feature of the Kafka stream). Needed a place to store some thoughts (since I won't be working all the way through them yet), so making the PR way ahead of it being ready. This'll be dependent on us wanting to complete the migration from Kafka -> Kinesis.

* Remove old `EventStream` (possibly rename `KinesisPublisher` to be `EventStream`)
* `KinesisPublisher` should just take JSON-API-conforming hashes. Since those may contain an array of models (remember, format is `{classifications: [...], linked: {...}}`), the publisher should take care of outputting one event per model in the main collection ("classifications"), passing along all (or just appropriate) linked data. I really feel like one line in the stream should always only contain one model plus linked data.
* There's a Kinesis `put_records` method for publishing multiple lines in bulk.

### DIFF
--- a/lib/kinesis_publisher.rb
+++ b/lib/kinesis_publisher.rb
@@ -17,6 +17,7 @@ class KinesisPublisher
       source: 'panoptes',
       type: event_type,
       version: '1.0.0',
+      timestamp: Time.now.utc.iso8601,
       data: data,
       linked: linked
     }.to_json
--- a/lib/kinesis_publisher.rb
+++ b/lib/kinesis_publisher.rb
@@ -1,23 +1,24 @@
 class KinesisPublisher
-  def self.publish(event_type, partition_key, data)
+  def self.publish(event_type, partition_key, data, linked = {})
     return unless client
 
     client.put_record(
       stream_name: stream_name,
       partition_key: partition_key.to_s,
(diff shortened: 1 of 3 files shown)


marten:
@camallen I started this with a broader scope, but actually I need those linked models before I can move Nero over. Could you give this a review?

lib/kinesis_publisher.rb
```diff
@@ -1,23 +1,25 @@
...
   end
 
-  def self.format_data(event_type, data)
+  def self.format_data(event_type, data, linked)
     {
       source: 'panoptes',
       type: event_type,
       version: '1.0.0',
-      data: data
+      timestamp: Time.now.utc.iso8601,
```
camallen: I assume we're tracking times here and we use the payload timestamp for actual event tracking?

camallen:
LGTM

lib/kinesis_publisher.rb
marten: Yeah normally we'd look at eg `data[:updated_at]`, but in cases where the payload isn't an actual record this is useful (eg the `workflow_counters` events)

PR #1676, opened 2016-02-18
rework queue updates

### LINKED ISSUE #1386: priority workflow ordering not being respected
Subject queue updates are sorting the sms id array in asc order and losing any priority ordering from the input. 

We should respect this workflow's priority ordering when when retrieving from the queue (ignoring any internal queue order from PG array ops) [here](https://github.com/zooniverse/Panoptes/blob/f4c3e7eb4d4cea32f2b0397198b5d940ed5cdb96/lib/subject_selector.rb#L27) and when random we should suffle the ids to mitigate the pg array sort order op.

camallen:
closes #1386 - rework queue updates (enqueue / dequeue) to be driven by the calling context and extracted to ruby land. Instead of calling updates via sql using `update_all` update the state on the specific known queues when we are using them.


lib/subjects/selector.rb
```diff
@@ -115,14 +116,16 @@ def user_subject_queue
...
       SubjectQueue.by_set(subject_set_id)
-        .find_by(user: user, workflow: workflow)
+        .find_by(user: queue_user, workflow: workflow)
     end
 
-    def dequeue_for_logged_in_user(sms_ids)
+    def dequeue_ids(queue, sms_ids)
       if queue_user
-        DequeueSubjectQueueWorker.perform_async(workflow.id, sms_ids, queue_user.try(:id), subject_set_id)
+        queue.dequeue_update(sms_ids)
```
camallen: Dequeue for logged in users is not in the request cycle. if this impacts too much we can move it to the background again.

lib/seen_subject_remover.rb
```diff
@@ -20,44 +19,17 @@ def append_ids_size
...
-                       .joins(:subject)
-                       .where(subjects: { id: user_seen_subject.subject_ids })
-                     else
-                       SetMemberSubject.none
-                     end.pluck(:id)
+    @seen_before ||=
+      if user_seen_subject
+        SetMemberSubject.where(id: append_ids)
+          .joins(:subject)
+          .where(subjects: { id: user_seen_subject.subject_ids })
```
marten: Why not just `.where(subject_id: user_seen_subject.subject_ids)`?

camallen: Apart from ensuring we haven't deleted the subject independent of the sms, no. I can't see a compelling reason for this though, i'll update.

camallen:
@marten is good to review again, turns out i hadn't re applied the seens remover, thanks!

app/workers/enqueue_subject_queue_worker.rb
```diff
@@ -1,36 +1,32 @@
...
   })
 
-  attr_reader :workflow, :user, :subject_set_id, :limit, :selection_strategy
+  attr_reader :queue, :workflow, :user, :subject_set_id, :limit, :selection_strategy
 
-  def perform(workflow_id, user_id=nil, subject_set_id=nil, limit=SubjectQueue::DEFAULT_LENGTH, strategy_override=nil)
-    @workflow = Workflow.find(workflow_id)
-    @user = User.find(user_id) if user_id
-    @subject_set_id = subject_set_id
+  def perform(queue_id, limit=SubjectQueue::DEFAULT_LENGTH, strategy_override=nil)
```
marten: This'll throw a bunch of errors in Sidekiq when we deploy (and build the AMI, for that matter) for existing jobs. Do we care?

app/workers/enqueue_subject_queue_worker.rb
```diff
@@ -70,4 +66,10 @@ def default_strategy_sms_ids
   def workflow_strategy
     @workflow_strategy ||= workflow.selection_strategy
   end
+
+  def strategy_sms_ids
+    selected_sms_ids.compact
+    rescue Subjects::CellectClient::ConnectionError
```
marten: Agree, common style when using def as a begin is to indent the rescue with the def..end

app/workers/enqueue_subject_queue_worker.rb
camallen: yeah i think so, how about i just disable this job `return nil` for this PR and then get a new one out after merge to revert?

app/workers/dequeue_subject_queue_worker.rb
```diff
@@ -2,19 +2,17 @@ class DequeueSubjectQueueWorker
...
-    }
-  })
-
-  def perform(workflow_id, sms_ids, user_id=nil, subject_set_id=nil)
-    workflow = Workflow.find(workflow_id)
-    user = User.find(user_id) if user_id
-    unless sms_ids.blank?
-      SubjectQueue.dequeue(workflow, sms_ids, user: user, set_id: subject_set_id)
-    end
+  def perform(queue_id, sms_ids)
```
marten: Same as enqueue worker, but in this case `workflow_id`s in the queue will get interpreted as `queue_id`s (of which we have more than workflows, so high chance there's a queue for that workflow_id), and Sidekiq might go about removing those sms_ids from that queue.

app/workers/enqueue_subject_queue_worker.rb
camallen: yeah my bad

app/workers/enqueue_subject_queue_worker.rb
marten: Sounds good. See the dequeue worker too.

spec/lib/subjects/seen_remover_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      user: nil,
+      workflow: nil,
+      subject_ids: subject_ids
+    )
+  end
+  subject { described_class.new(user, workflow, append_sms_ids) }
+
+  describe "#unseen_ids" do
+    it "should return the whole set if there are no seen subjects" do
+      allow(subject)
```
marten: I've never really liked stubbing of private methods on the object under test. Any way we could instead only stub interactions between objects, e.g. perhaps most of the work done by `seen_before_sms_ids` could be extracted to a class method/scope on SetMemberSubject? That way you can also easily test the behaviour of that (and we could decouple this service class from the DB entirely).

lib/subjects/seen_remover.rb
```diff
@@ -0,0 +1,36 @@
...
+        )
+      else
+        SetMemberSubject.none
+      end
+      seen_before.pluck(:id)
+    end
+
+    def user_seen_subject
+      @user_seen_subject ||= if user
+        UserSeenSubject.find_by(user: user, workflow: workflow)
```
marten: The only place the `user` is used seems to be here. Have you thought about passing the user_seen_subject in directly? Might make the tests easier, since you can just pass in the double you need for the test instead of having to stub out method calls on UserSeenSubject.

spec/lib/subjects/seen_remover_spec.rb
camallen: Me either...bad effort on my behalf. Will rework to use dep injection.

lib/subjects/seen_remover.rb
```diff
@@ -0,0 +1,36 @@
+module Subjects
+  class SeenRemover
+
+    attr_reader :user, :workflow, :append_sms_ids
+
+    def initialize(user, workflow, append_ids)
+      @user = user
+      @workflow = workflow
+      @append_sms_ids = append_ids
```
marten: This one is a very subtle point, so don't hesitate to disagree: to me it feels more like `user` and `workflow` make up the "context" for this operation, whereas `append_ids` is the main piece of data. Because of that, I would have personally made `append_ids` a parameter to `unseen_ids` (and passing it along as a param to `seen_before_sms_ids`). The downside is having to pass that along, the upside is that it makes it easier to "reuse". Not sure if we'll do that in production code, but for instance in the tests you could use the same object but call `unseen_ids` with different arguments: an empty array vs. a normal one comes to mind.

camallen: I refactored to inject the user seens dependency but I felt it was cleaner to use the append ids in the initializer. If the need to reuse the object with new append ids arises we should refactor this out.

spec/lib/subjects/seen_remover_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  let(:workflow) { instance_double("Workflow")}
+  let(:user) { instance_double("User") }
+  let(:uss) { instance_double("UserSeenSubject") }
+  let(:sms) { create(:set_member_subject) }
+  let(:subject_ids) { [ sms.subject.id ] }
+  let(:user_seens) { nil }
+
+  subject { described_class.new(user_seens, append_sms_ids) }
+
+  describe "#unseen_ids", :focus do
```
camallen: bad campbel..

marten:
LGTM

PR #1678, opened 2016-02-19
Remove Kafka-related code

marten:
No longer publishes to Kafka. Kinesis is enabled only in staging/production and only when a stream name to push to is configured via environment variables (makes it easy to stop this under load too, just remove the env var)


marten:
That last commit is completely optional, just my attempt at keeping the lib/ folder slightly more compact, noticed a couple of related things which are all basically wrapper modules around services we use/run (S3, Kinesis, talk, aggregation engine). Let me know how you feel about that, happy to revert it.

app/services/event_stream.rb
```diff
@@ -0,0 +1,47 @@
...
+  def self.client
+    @client
+  end
+
+  def self.client=(client)
+    @client = client
+  end
+
+  def self.stream_name
+    ENV['KINESIS_STREAM']
```
camallen: didn't we want a default on this?

camallen:
LGTM

app/services/event_stream.rb
marten: Yes, I removed the default so that it'll only turn the stream on when configured, in case someone other than us wants to run their own panoptes

PR #1681, opened 2016-02-23
re-enable disabled job after the worker api change

camallen:
Merge after the latest build / deploy - revert a temp workaround for production sidekiq jobs


marten:
LGTM. I'll leave it to you to merge when the deploy is done.

PR #1687, opened 2016-02-24
Use tempfiles instead of figuring out a file name ourselves

marten:


app/workers/concerns/dump_worker.rb
```diff
@@ -40,16 +46,14 @@ def dump_type
...
+    @csv_tempfile.path
   end
 
   def gzip_file_path
-    "#{Rails.root}/tmp/#{project_file_path.join("_")}.gz"
+    @gzip_tempfile ||= Tempfile.new(['export', '.gz'])
+    @gzip_tempfile.path
   end
 
   def project_file_path
```
astopy: This should probably replace all non-alphanum characters rather than just whitespace (i.e. change \s to \W).

camallen:
LGTM after @astopy's comment.

app/workers/concerns/dump_worker.rb
marten: I'm actually contemplating getting rid of all of that, and just using the project id instead of logins and display names. Changing the replacement like you say means you could create "collisions" (not really collisions since the file name itself is a UUID, but you could get your file written in the same directory as another project which has a similar name but just difference in non-alphanum chars)

astopy: Yeah, using project ID sounds like a good idea.

marten:
@camallen Will changing the project_file_path break existing Mediums? It'll get the new path_opts assigned, not sure if that will influence anything...

camallen:
I assume you mean in terms of creating new s3 paths? Doesn't look like it will, we don't supply path opts when [put'ing the file](https://github.com/zooniverse/Panoptes/blob/56d05cbdd930c3da4058398d85a4e58f4a4d448b/app/workers/concerns/dump_worker.rb#L86) which creates a s3 bucket_path [here](https://github.com/zooniverse/Panoptes/blob/56d05cbdd930c3da4058398d85a4e58f4a4d448b/app/services/media_storage/aws_adapter.rb#L37).

marten:
But `load_medium` does update the `path_opts`...

camallen:
yeah we should look at reaping those old files someday, pinging @astopy for ref.

PR #1688, opened 2016-02-25
rework SMS selector

### LINKED ISSUE #1661: Noticed II'm getting finished images on Fossil Finders but there's no message saying data is complete
If I'm logged in and seeing finished images, shouldn't the site be putting up the gone through all the data message for the logged in user? 

screen shot 2016-02-10 at 8 38 57 pm

camallen:
closes #1661 - new users should not see retired data and a fix for the workflow.finished? check.

Also turns out i hate my future self and wrote some hard to understand specs a while back...past_cam--

### DIFF
--- a/app/models/concerns/subject_counts.rb
+++ b/app/models/concerns/subject_counts.rb
@@ -10,7 +10,7 @@ module SubjectCounts
   end
 
   def finished?
-    @finished ||= if subject_sets.empty?
+    @finished ||= if subject_sets.empty? || subjects_count == 0
       false
     else
       retired_subjects_count >= subjects_count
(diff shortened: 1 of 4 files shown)


app/models/concerns/subject_counts.rb
```diff
@@ -10,7 +10,7 @@ def retired_subjects_count
   end
 
   def finished?
-    @finished ||= if subject_sets.empty?
+    @finished ||= if subject_sets.empty? || subjects_count == 0
```
camallen: this is to cover the case when 0 == 0 and when owners delete subjects after a data run.

marten:
LGTM

PR #1682, opened 2016-02-23
Stale queue selection

camallen:
closes #1679, #1661, #1640, #1677 - request fresh selection data when queues are stale (older than 30 mins). 

Refactored the selection code which is a chunk of this PR to it's own strategy selector.

PR #1689, opened 2016-02-26
pass the user not the api_user to the subject selector

camallen:
this should fix the no method errors introduced from #1682 & #1676


marten:
Belated LGTM :p

PR #1693, opened 2016-02-26
handle stale queue update error in selection

camallen:
may hit stale queue error when dequeuing in the request cycle for logged in users. this will catch the error and push it to the background.

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -128,17 +128,13 @@ module Subjects
     end
 
     def dequeue_ids(queue, sms_ids)
-      background_update = !queue_user
       if queue_user
-        begin
-          queue.dequeue_update(sms_ids)
-        rescue ActiveRecord::StaleObjectError
-          background_update = true
-        end
-      end
-      if background_update
+        queue.dequeue_update(sms_ids)
+      else
         DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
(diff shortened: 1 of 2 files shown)


lib/subjects/selector.rb
```diff
@@ -128,9 +128,15 @@ def find_subject_queue
...
 
     def dequeue_ids(queue, sms_ids)
+      background_update = !queue_user
       if queue_user
-        queue.dequeue_update(sms_ids)
-      else
+        begin
+          queue.dequeue_update(sms_ids)
+        rescue ActiveRecord::StaleObjectError
+          background_update = true
```
marten: I wonder if having more duplication would be more readable in this particular case, by not having to keep track of state in your head (the idea being following if-else branches is easier than remembering variable values).

```ruby
if queue_user
  begin
    queue.dequeue_update(sms_ids)
  rescue ActiveRecord::StaleObjectError
    DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
  end
else
  DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
end
```

marten:
LGTM with some considerations

lib/subjects/selector.rb
camallen: can do for sure, i normally try to extract duplication but agree about readability.

PR #1695, opened 2016-02-29
[WIP] Build docker image from ruby:2.3-alpine

astopy:
This changes the Dockerfile to build `FROM ruby:2.3-alpine` -- there's really no reason for us to build our own ruby image when there's an official one, and the -alpine version is tiny. This reduces the final panoptes image size to 130 MB (down from the current 248 MB).

This will need a thorough test when it's deployed to staging. Alpine uses musl libc rather than glibc, so we'll need to make sure this doesn't have any unintended effects.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,22 @@
-FROM zooniverse/ruby:2.3
+FROM ruby:2.3-alpine
 
 WORKDIR /rails_app
 
-RUN apt-get update && \
-    apt-get install --no-install-recommends -y git curl supervisor libpq-dev && \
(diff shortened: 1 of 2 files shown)


camallen:
can we move this to a staging docker file that we can use for the staging infrastructure via the `dockerfile: Dockerfile.staging` cmd? then we can get this out and running to see how it performs.

astopy:
After testing it looks like this will depend on cowboyd/therubyracer#378 unless we can find a workaround.

camallen:
closing this for now - that linked issue is pretty stale. Let's re-open when we look at this again.

PR #1696, opened 2016-02-29
Add project, workflow, workflow_content and user into stream

marten:
I still intend to look at compressing this data down, but I'll handle that with an update to the ZooStream gem (and need to make Nero/ZooStats forwards-compatible first).

### DIFF
--- a/app/serializers/event_stream_serializers/classification_serializer.rb
+++ b/app/serializers/event_stream_serializers/classification_serializer.rb
@@ -7,12 +7,30 @@ module EventStreamSerializers
       Serialization::V1Adapter.new(serializer, options)
     end
 
-    attributes :id, :created_at, :updated_at, :user_ip, :annotations, :metadata
+    attributes :id, :created_at, :updated_at, :user_ip, :workflow_version, :gold_standard,
+               :expert_classifier, :annotations, :metadata
 
-    belongs_to :project,  serializer: EventStreamSerializers::ProjectSerializer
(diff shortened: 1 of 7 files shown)


app/serializers/event_stream_serializers/classification_serializer.rb
```diff
@@ -7,12 +7,30 @@ def self.serialize(classification, options = {})
...
-    belongs_to :workflow, serializer: EventStreamSerializers::WorkflowSerializer
-    has_many   :subjects, serializer: EventStreamSerializers::SubjectSerializer
+    belongs_to :project,          serializer: EventStreamSerializers::ProjectSerializer
+    belongs_to :user,             serializer: EventStreamSerializers::UserSerializer
+    belongs_to :workflow,         serializer: EventStreamSerializers::WorkflowSerializer
+    belongs_to :workflow_content, serializer: EventStreamSerializers::WorkflowContentSerializer
+    has_many   :subjects,         serializer: EventStreamSerializers::SubjectSerializer
+
+    def workflow
+      version = workflow_version.split(".")[0].to_i
```
camallen: extract to common helper?

camallen:
LGTM

app/serializers/event_stream_serializers/classification_serializer.rb
marten: Couldn't really come up with a good abstraction. I'd want to call this `workflow_version` but that's what the combined workflow+workflow_content version is already called. :(

PR #1697, opened 2016-02-29
Update to attention based cellect

camallen:
Note: cellect is turned of in staging so this can get merged / deployed. We can look at updating the configs, etc and redeploying cellect_panoptes after.

This removes all dependency on Zookeeper, instead we use redis for cellect node discover via Attention gem.


app/services/media_storage/aws_adapter.rb
```diff
@@ -1,16 +1,13 @@
...
-    DEFAULT_GET_EXPIRATION = 60
-
     attr_accessor :prefix, :bucket
 
     def initialize(opts={})
       @prefix = opts[:prefix] || Rails.env
       @bucket = opts[:bucket]
-      @get_expiration = opts.fetch(:expiration, {})[:get] || DEFAULT_GET_EXPIRATION
-      @put_expiration = opts.fetch(:expiration, {})[:put] || DEFAULT_PUT_EXPIRATION
+      @get_expiration = opts.fetch(:expiration, {})[:get] || 60
```
marten: Might be a good use for [`Hash#dig`](https://blog.blockscore.com/new-features-in-ruby-2-3/):

`@get_expiration = opts.dig(:expiration, :get) || 60`

marten:
LGTM but @parrish might be better at reviewing this.

app/services/media_storage/aws_adapter.rb
camallen: nice

camallen:
@parrish mind having a look at this, it's mostly just bumping the client gem and removing some zk / celluloid cruft.

camallen:
I'm going to bump the cellect client gem and put a timeout in too, sensible default 0.75?

parrish:
750ms is probably fine, but there may be edge cases when the user seen set takes longer to load.  Happy to tune that later though.

parrish:
@camallen Is [this build error](https://travis-ci.org/zooniverse/Panoptes/jobs/113172536) related?  I'm guessing not.

camallen:
npoe that's a brittle intermittent spec that i have to re-write.

PR #1699, opened 2016-03-03
handle cellect client connection failures

camallen:
don't let a busted redis connection stop us from serving subjects


camallen:
I've removed the cellect client var in staging configs to test this.

marten:
LGTM

PR #1700, opened 2016-03-03
Allow multiple subject ids to retire

### LINKED ISSUE #1628: retire subject route params
allow subject id string / integer and arrays. Return a better response code for invalid retire params.

marten:
Fixes #1628


camallen:
Nice one.

PR #1703, opened 2016-03-04
ignore redis failure on classification create

camallen:
don't 500 that may cause a duplicate classification submission instead just carry on as the lifecycle event will eventually fire but still go ahead and notify that redis is down.


marten:
LGTM

PR #1704, opened 2016-03-04
Fix URL pointing towards user groups in the apiary docs

### LINKED ISSUE #1460: Incorrect URL on apiary documentation
Was using the documentation to create user groups via the API, kept getting 404 errors. Turns out the URL in the doc is 'api/groups' instead of 'api/user_groups', at least I assume because I got a real response for the latter.

marten:
Fixes #1460


camallen:
LGTM

PR #1705, opened 2016-03-04
update s3 headers for data dump exports

### LINKED ISSUE #1694: mime type/content-disposition for data exports
Safari (and presumably other browsers) displays data export files inline by default rather than saving them. We should set the appropriate metadata on S3 to stop it doing that. Also may as well set a more useful filename while we're doing that.

Since we're already using signed URLs we can just set the headers in the GET request. See the request parameters section here: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html We'll just want to set content-disposition to something like `Content-Disposition: attachment; filename=projectname-classifications.csv`

camallen:
closes #1694 put the content disposition to the s3 resource via the medium attributes, currently only set via the dump workers but could be expanded out to other media resources.


marten:
LGTM

PR #1706, opened 2016-03-04
Clean up PostgresqlSelection

marten:
Started with removal of those `before(:all)` statements, then to keep stuff at least somewhat speedy I'm splitting up some stuff so tests can be more focussed. Not perfect yet, but time to merge this before it gets out of hand :P

- [x] Figure out why this spec keeps randomly failing --> Doesn't want to fail locally :(

```
  1) Subjects::PostgresqlSelection selection #select grouped selection behaves like select for incomplete_project when a user has only seen a few subjects with a limit of 1 should return an unseen subject
     Failure/Error: expect(uss.subject_ids).to_not include(subject.select.first)
       expected [417, 430, 419, 427, 410] not to include 427
     Shared Example Group: "select for incomplete_project" called from ./spec/lib/subjects/postgresql_selection_spec.rb:52
     # ./spec/support/shared_examples_for_postgresql_selection.rb:36:in `block (4 levels) in <top (required)>'
```

### DIFF
--- a/spec/support/shared_examples_for_postgresql_selection.rb
+++ b/spec/support/shared_examples_for_postgresql_selection.rb
@@ -26,6 +26,8 @@ shared_examples "select for incomplete_project" do
     let(:args) { opts.merge(limit: limit) }
     let!(:uss) do
       subject_ids = sms_scope.sample(seen_count).map(&:subject_id)
+      puts "All: #{sms_scope.map(&:subject_id).inspect}"
+      puts "Seen: #{subject_ids.inspect}"
       create(:user_seen_subject, user: user, subject_ids: subject_ids, workflow: workflow)
     end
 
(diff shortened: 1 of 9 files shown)


spec/support/shared_examples_for_postgresql_selection.rb
```diff
@@ -0,0 +1,75 @@
...
+        result = selector.select
+        expect(result).to match_array(result.to_a.uniq)
+      end
+    end
+
+    # Account for the loop cut off limit constructing the random sample
+    it 'should always return an approximate sample of subjects up to the unseen limit' do
+      unseen_count.times do |n|
+        limit = n+1
+        results_size = Subjects::PostgresqlSelection.new(workflow, user, (args || {}).merge(limit: limit)).select.length
```
camallen: can't see why these are using select.length i think it might have been to workaround count errors which should be fixed by something like .count(:all), i think we should change it.

camallen:
LGTM pending the count syntax concern.

spec/support/shared_examples_for_postgresql_selection.rb
marten: .select is just what the method on PostgresqlSelection is called. It's not Ruby's `Enumerable#select`

camallen: That makes a lot more sense!

PR #1708, opened 2016-03-07
Secure cookies

camallen:


config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
-
-Rails.application.config.session_store :cookie_store, key: '_Panoptes_session'
+secure_option = (Rails.env.development? || Rails.env.test?) ? false : true
```
astopy: Maybe `secure_option = not (Rails.env.development? || Rails.env.test?)`

marten:
wouldnt setting config.force_ssl in the environments have the same effect?

camallen:
yeah that setting would but don't we want to terminate ssl layer at the CDN?

camallen:
going to use the force_ssl config as it provides a fall back mechanism should the site be deployed without an ssl termination in front of it.

marten:
Cloudfront will set the correct header which Rails looks for: http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html#x-forwarded-proto

camallen:
Thanks - just found the relevant rack source, https://github.com/josh/rack-ssl/blob/master/lib/rack/ssl.rb#L41

marten:
LGTM

PR #1710, opened 2016-03-07
raise connection error when no cellect hosts available

camallen:
handle the case where the cellect node_set is empty and thus the returned host will be nil.

### DIFF
--- a/lib/subjects/cellect_client.rb
+++ b/lib/subjects/cellect_client.rb
@@ -81,7 +81,7 @@ module Subjects
         super(action, params) do
           params[:host] = @session.reset_host
         end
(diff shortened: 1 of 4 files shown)


marten:
LGTM, but I feel like all this should really be done by the cellect client object (and thus be extracted to the cellect gem)

camallen:
I think it's valid behavior for a client to return nil if there are no hosts...it's up to the user to figure out what todo in that scenario, ignore / raise.

camallen:
@marten Ah - do you mean move the client code from here to the gem? We could look at that.

marten:
Yes. But again, fine to merge until then

PR #1711, opened 2016-03-07
gem upgrades and puma version bump

camallen:
mostly straight upgrades but puma got a major version bump to 3.0 https://github.com/puma/puma/blob/master/History.txt


marten:
LGTM

PR #1712, opened 2016-03-08
add seperate route for health check

camallen:
This route will be used for specialized nginx directive to allow force_ssl to work with ELB health checks over http.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -112,6 +112,7 @@ Rails.application.routes.draw do
     end
   end
 
+  get "health_check", to: "home#index"
   root to: "home#index"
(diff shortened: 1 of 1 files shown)


marten:
One idea: should it be a special controller action that makes sure to touch the DB and Redis, but do both in a lightweight query?

camallen:
I was thinking that as well - could do. Can we get this into deployment and see if it works and then modify the health_check route to test the required services?

marten:
Yes, didn't mean that it should block this one. It's LGTM, but you assigned to astopy so I'll leave verdict to him.

PR #1715, opened 2016-03-09
Do not send emails when recipients list is empty

marten:


camallen:
should we also stop the mailer worker starting if emails is empty, [here](https://github.com/zooniverse/Panoptes/blob/589dc42cfadaec4bfe96c7bf60e6bd7c73f486d7/app/workers/concerns/dump_mailer_worker.rb#L9)?

PR #1718, opened 2016-03-10
Initial work on PR template checklist

marten:
Github now supports a template for the description of each pull request.

**Let me know what else you can come up with**

### Review checklist


camallen:
Are the changes covered by tests?

PR #1725, opened 2016-03-14
double quote the content-disposition filename

### LINKED ISSUE #1723: Adjust the name for the csv dumps so they open on macs (and likely other computers)
I think somehow the csv dumps when they get more intelligible fillenames now have quotes. When I try and click on it for example to open a few classifications from a test project I get this 

screen shot 2016-03-13 at 8 08 49 pm

if I do an ls, I get this:

/Users/mschwamb/Downloads/'europa-zoo-test-classifications.csv'

You can see the quotes visible. Can they please be removed.

camallen:
closes #1723 - otherwise safari gets weird

### DIFF
--- a/app/workers/concerns/dump_worker.rb
+++ b/app/workers/concerns/dump_worker.rb
@@ -113,6 +113,6 @@ module DumpWorker
 
   def content_disposition
     name = project.slug.split("/")[1]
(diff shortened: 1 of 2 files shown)


marten:
LGTM

PR #1732, opened 2016-03-18
extract scope concerns to seperate routes

### LINKED ISSUE #1713: Allow all users to fetch their classifications for a subject.
This issue is in reference to the following API call:
http://docs.panoptes.apiary.io/#reference/classification/classification-collection/list-all-classifications

The use-case for this issue is driven by a key feature in Ancient Lives, which is temporarily available here: https://preview.zooniverse.org/ancientlives/. The exact use-case is a user returning to an image of a papyrus fragment to continue / update their transcription.

In the current implementation, only users in administrative roles for a project (e.g., collaborator) can fetch classifications via the API. If a user is logged in, the user should have the privilege of fetching the data that they have contributed for the subject.

The API's response is empty for a user that has no administrative affiliation with the project. The response should, however, be populated with their past classifications.

A secondary issue is returning *only* the most recent classification, but that is another matter. I have front-end code that identifies the most recent classification currently, but future projects may benefit from the ease of retrieving the most recent.

Spoke with @camallen about this earlier today.

camallen:
closes #1713 

extracting the overloading the show/index query scopes to route concerns. 

**NOTE:** this may break some existing functionality since by default show / index routes will now return just completed classification for the requesting user and not all classifications the user has access to from a project / that are incomplete.  I don't think it does but I'd like to make sure this api change doesn't break any existing functionality that PFE / custom projects rely on.

new routes:
GET      /api/classifications                         -> default, completed classfications for this user
GET      /api/classifications/incomplete       -> only incomplete for the user 
GET      /api/classifications/project             -> all classifications the user has project access to
GET      /api/classifications/gold_standard -> only gold standard for marked workflows

TODO:
- [x] Update the api docs...remove the cruft on special for params and outline the collection routes

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -524,20 +524,20 @@ key.
             }
 
 ### List all classifications [GET]
-Only lists a classifications the active user has made,
-classifications from other members of a user_group, or projects the
-user has edit permissions for.
+Only lists classifications the active user has made or projects the user has edit permissions for.
 
-Classifications have the special *for* parameter that indicates the
-scope of classifications you'd like to retrieve. Possible options are:
+Classifications have special collection routes that indicate the scope you would like to retrieve.
 
-+ 'user' will fetch the current user's past classifications.
-+ 'user_groups' will fetch classifications from groups a user is a member of
-+ 'projects' will fetch classifications from projects a user has 'edit' permissions for
+Possible options are:
 
-Any of the for scopes may be further filtered using the *project_id*
++ '/' default, will fetch the current user's past complete classifications.
++ 'incomplete' will fetch the current user's past incomplete classifications.
++ 'project' will fetch classifications from projects a user has 'edit' permissions for
++ 'gold_standard' will fetch gold standard classifications for all marked workflows
+
+
+Any of the scopes may be further filtered using the *project_id*, *workflow_id*
 and *user_group_id* parameters.
-
 + Parameters
     + page (optional, integer) ... index of the collection page, 1 is default
     + page_size (optional, integer) ... number of items on a page. 20 is default
--- a/spec/models/classification_spec.rb
+++ b/spec/models/classification_spec.rb
@@ -130,45 +130,70 @@ describe Classification, :type => :model do
     let(:user) { ApiUser.new(create(:user)) }
     let(:project) { create(:project, owner: user.owner) }
     let(:user_group) { create(:user_group) }
-    let!(:classifications) do
+    let(:other_user_classification) { create(:classification) }
+    let(:other_user) { ApiUser.new(other_user_classification.user) }
+    let!(:membership) do
       create(:membership, roles: ['group_admin'], user: user.owner,
-             user_group: user_group, state: :active)
-      [create(:classification, user: user.owner),
-       create(:classification, user: user.owner, completed: false),
-       create(:classification, project: project),
-       create(:classification, project: project, completed: false),
-       create(:classification, user_group: user_group),
-       create(:classification)]
+        user_group: user_group, state: :active)
     end
 
-    it 'should return an ActiveRecord::Relation' do
-      expect(Classification.scope_for(:show, user)).to be_a(ActiveRecord::Relation)
+    describe "#show/index" do
+      it 'should return an ActiveRecord::Relation' do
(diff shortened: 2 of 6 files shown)


camallen:
@brian-c mind having a look at these route changes for classifications? Pretty sure we don't actually use them currently (i.e. look at peoples history / project dumps via the api) but be good to get your thoughts on it.

apiary.apib
```diff
@@ -524,20 +524,20 @@ key.
             }
 
 ### List all classifications [GET]
-Only lists a classifications the active user has made,
-classifications from other members of a user_group, or projects the
-user has edit permissions for.
+Only lists a classifications the active user has made or projects the user has edit permissions for.
```
parrish: > Only lists a classifications

Typo

apiary.apib
```diff
@@ -524,20 +524,20 @@ key.
...
 
-Classifications have the special *for* parameter that indicates the
-scope of classifications you'd like to retrieve. Possible options are:
+Classifications have special collection routes that indicate the scope of you'd like to retrieve.
```
parrish: > the scope of you'd

Typo

spec/models/classification_spec.rb
```diff
@@ -130,45 +130,70 @@
...
+
+      it 'should return all complete classifications the requesting user has made' do
+        expected = other_user_classification
+        expect(Classification.scope_for(:index, other_user)).to match_array(expected)
+      end
 
-    it 'should return all incomplete classifications a user has made' do
-      expected = classifications[1]
-      expect(Classification.scope_for(:show, user)).to include(expected)
+      it 'should not return incomplete classifications for a project' do
```
parrish: This seems somewhat counter-intuitive.  Can a user not request their own incomplete classifications?

spec/models/classification_spec.rb
```diff
@@ -130,45 +130,70 @@
...
+        create(:classification, user: user.owner, completed: false)
+        expect(Classification.scope_for(:show, user)).to be_empty
+      end
     end
 
-    it 'should not return completed classifications for a user' do
-      expected = classifications[0]
-      expect(Classification.scope_for(:show, user)).to_not include(expected)
+    describe "#project" do
+      it 'should return all project classifications if the user can update it' do
```
parrish: Can the user update the classification because they own the project here?

parrish:
This looks good other than a possible confusing implied intent [here](https://github.com/zooniverse/Panoptes/pull/1732#discussion-diff-56696043R166).  If I'm right about the update being allowed because `user` is the owner, maybe renaming the `user` helper to `owner` would clarify it.

I'll wait for @brian-c to chime in before merging though.

brian-c:
Looks like all PFE master currently uses is a POST to `/api/classifications` and a GET from `/api/classifications/gold_standard`, so this shouldn't break anything.

Any reason you're implementing these as scopes versus allowing explicit queries from the client? I'm thinking something like `/api/classifications?workflow=123&subjects=234&complete=false` would be more consistent and predictable on the client's end.

parrish:
> more consistent and predictable on the client's end

With the amount of filtering, scoping, including, etc. that can be done on this endpoint, it gets incredibly nasty for the API.  The alternative to this would be to logically split the endpoints by params.  That would require a pretty significant rewrite though and could still end up somewhat sloppy.  This seems like a good middle ground.

That said, it does break convention with a lot of the other resources.  Kind of on the fence here I guess.

brian-c:
That's fair. I figured most URL queries would translate pretty cleanly into SQL queries, so all the server would have to worry about would be read permissions. Probably naive on my part.

parrish:
A lot of gets funky when you're dealing with complicated associations -- trying to preload/join/filter multiple tables.  I think this is a fair example of how JSON API can be really counterproductive on the backend.

camallen:
@parrish, poor choice of words, if they have update access on the project is what it should have said. The user can update the classification if it is incomplete and they own it.

camallen:
@brian-c agree that its not standard but overloading the index route just got too funky. Basically decided to extract them out, if its causing problems we can revisit. 

The main issue is making sure a normal user can access their own data. I think the original implementation was a bit ambitious on 1 route.

Issue #1727, opened 2016-03-16
subject_sets endpoint doesn't seem to be respecting sort parameter


alexbfree:
After implementing https://github.com/zooniverse/Panoptes-Front-End/pull/2310 (fix for https://github.com/zooniverse/Panoptes-Front-End/issues/2291) which has now been deployed to PFE, I am not seeing the subject sets in the correct order. 

From the network tab, it looks like the response from the API has not respected the sort parameter...

![screenshot 2016-03-16 17 08 22](https://cloud.githubusercontent.com/assets/1473244/13821494/d848b404-eb99-11e5-9165-1f2721de0394.png)
![screenshot 2016-03-16 17 08 51](https://cloud.githubusercontent.com/assets/1473244/13821498/da69938e-eb99-11e5-840a-f41caa2490df.png)

This did seem to sort correctly in my local environment, so there may be something else going on; but the frontend fix to apply the sort parameter is definitely in place, the fieldname matches, and I can't explain the response seen here...

Can someone else explain this or is it indeed that the API is not respecting the sort order?

camallen:
@alexbfree we'll need a sort by declaration on the serializer, e.g. [like this](https://github.com/zooniverse/Panoptes/blob/cde5f4cb48d1f33be656f0eb403294419f895d35/app/serializers/recent_serializer.rb#L6) but [here](https://github.com/zooniverse/Panoptes/blob/03b096db6c33be6535a1c87d4552642ec1bf3839/app/serializers/subject_set_serializer.rb#L7). 

Fancy having a crack at it?

alexbfree:
ok that sounds straightforward enough with those pointers. I will have a go and get you a PR to review. I may ask for help setting up a local Panoptes API!

camallen:
cool - happy to help, the readme has instructions on getting setup for dev

alexbfree:
this has been fixed in the API by #1733

PR #1733, opened 2016-03-18
Add sorting of subject sets

alexbfree:
Fixes issue #1727 

Allow display_name sorting for subject sets


camallen:
Thanks @alexbfree nice work!

alexbfree:
@camallen How can i know if this has been deployed to production? it should have autodeployed right?
A quick test of the PFE testcase in #1727 is still showing unsorted values in the sorted by display_name subject_sets Panoptes request in the network monitor tab... not sure if this is because this fix hasn't yet been deployed or if something else odd is going on...

camallen:
It's not deployed yet, lets do it first thing monday.

PR #1736, opened 2016-03-21
add retired sms counts to subject set serialized response

### LINKED ISSUE #986: Get retired subject counts by subject set
As discussed with @camallen, I'd like to show progress counts by subject set (in my case, e.g. pages of a book)

camallen:
closes #986 - happy to rework this implementation if needed but it makes sense to me.

### DIFF
--- a/app/serializers/subject_set_serializer.rb
+++ b/app/serializers/subject_set_serializer.rb
@@ -2,9 +2,14 @@ class SubjectSetSerializer
   include RestPack::Serializer
   include FilterHasMany
 
-  attributes :id, :display_name, :set_member_subjects_count, :metadata,
-    :created_at, :updated_at, :href
+  attributes :id, :display_name, :set_member_subjects_count,
+    :retired_set_member_subjects_count, :metadata, :created_at, :updated_at,
+    :href
   can_include :project, :workflows
   can_sort_by :display_name
   can_filter_by :display_name
+
+  def retired_set_member_subjects_count
+    SubjectWorkflowCount.by_set(@model.id).retired.count
+  end
 end
--- a/spec/controllers/api/v1/subject_sets_controller_spec.rb
+++ b/spec/controllers/api/v1/subject_sets_controller_spec.rb
@@ -22,7 +22,8 @@ describe Api::V1::SubjectSetsController, type: :controller do
   let(:api_resource_name) { 'subject_sets' }
 
   let(:api_resource_attributes) do
-    %w(id display_name set_member_subjects_count created_at updated_at metadata)
+    %w( id display_name set_member_subjects_count created_at updated_at
+        metadata retired_set_member_subjects_count)
   end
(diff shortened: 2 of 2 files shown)


app/serializers/subject_set_serializer.rb
```diff
@@ -2,9 +2,14 @@ class SubjectSetSerializer
...
-    :created_at, :updated_at, :href
+  attributes :id, :display_name, :set_member_subjects_count,
+    :retired_set_member_subjects_count, :metadata, :created_at, :updated_at,
+    :href
   can_include :project, :workflows
   can_sort_by :display_name
   can_filter_by :display_name
+
+  def retired_set_member_subjects_count
+    SubjectWorkflowCount.by_set(@model.id).retired.count
```
marten: Is that going to be an N+1 query effectively?

camallen: yeah, it will be - any way to avoid paying the cost without a cache attribute?

marten: Thinking more about this: we retire per workflow, so this number might be higher than the number of subjects in the set.

camallen: yeah - could be in the case of workflow subjects in more than 1 set. I can't see an easy way around that edge case..?

marten: Don't think it's an edge-case, if you have two workflows that share the same subject sets you'll notice this immediately. I stand by my previous suggestion in #986 that this should probably be a hash from either `{workflow_id: retired_count}` in a serialized SubjectSet, or `{subject_set_id: retired_count}` in a serialized Workflow.

/cc @rogerhutchings

marten:
Just had a chat with @rogerhutchings, who's not going to use this immediately. It's now more of a "would be nice to have for future projects" kind of feature, and maybe we shouldn't burden all subject set API calls with this additional N+1 query, for something that nobody is going to use right now.

I'm thinking maybe we should have a stats resource under `/projects/123/statistics` to contain this information in a cached manner. I'm not too worried about caching this sorts of counter as long as we store it with a timestamp and update it only by recomputing the value. 

My thoughts would be it would be something like:

```sql
CREATE TABLE statistics (
  id integer NOT NULL,
  key character varying NOT NULL,
  data jsonb DEFAULT '{}'::jsonb NOT NULL,
  project_id integer NOT NULL,

  -- optional foreign keys if this stat pertains only to a specific combination of these
  workflow_id integer,      
  subject_set_id integer,
  subject_id integer,       
  user_id integer,          

  created_at timestamp,
  updated_at timestamp,
  expires_at timestamp
)


CREATE UNIQUE INDEX foo ON statistics USING btree (key, project_id, workflow_id, subject_set_id, subject_id, user_id)
```

camallen:
Yeah - that statistics idea is much better.

PR #1738, opened 2016-03-22
Allow empty workflow tasks on create / update

### LINKED ISSUE #1728: Allow for empty workflow tasks
Currently, tasks cannot be set to `{}`. Desired behavior is to be able to start with an empty tasks list (zooniverse/Panoptes-Front-End#934).

camallen:
closes #1728 allow empty workflow tasks

When creating / updateing workflow tasks, allow the workflow_content strings to be empty but ensure they are a hash (by default in db and controller task strings extractor) to avoid the null db constraint. This should be fine client side as the task set and the contents don't get out of sync, i.e both are `{}` so the serialized tasks object (i.e. the translated strings re-inserted) is just an empty hash.

### DIFF
--- a/app/controllers/api/v1/workflow_contents_controller.rb
+++ b/app/controllers/api/v1/workflow_contents_controller.rb
@@ -3,8 +3,5 @@ class Api::V1::WorkflowContentsController < Api::ApiController
 
   require_authentication :all, scopes: [:project]
   resource_actions :default
-  schema_type :strong_params
-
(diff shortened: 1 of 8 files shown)


marten:
Code-style wise okay to me, but tests are failing.

camallen:
looking into it right now.

camallen:
updated - had to rework the validation params to switch from strong to json schema.

PR #1754, opened 2016-04-06
Sidekiq 4 upgrade

camallen:
nice to get this to staging and see how it goes. it's a super critical part of the stack. 

@astopy @marten not 100% sure on mergeing to master (we can revert easily). Thoughts on deploying this to staging via tags(?) without having to merge to master directly (will update the PR if we do).


marten:
LGTM. We can just lock lita and disable talk deploys, and either revert or unlock by the end of the day?

astopy:
Yeah that sounds fine to me.

PR #1756, opened 2016-04-06
Users whiteload structure fix

camallen:
add a default value to the users upload whitelist table and fix the out of sync structure.sql file. Not sure we need to have the `NOT NULL` clause anymore since it won't affect the code but happy to add another  migration PR if needed.

Need to rerun the rake task after this is deployed.


marten:
I say we add the NOT NULL back. I took it and the default out so we could deploy without locking the table for too long. I'll double check that I ran that rake task on production, will merge this afterwards.

camallen:
I already ran that rake task yesterday afternoon, hence i added the where clause into the rake task.

camallen:
i'll add the not null constraint back in after this goes out with new users getting a default and re-rerun the rake task for the gap from yesterday...sound ok?

Issue #1640, opened 2016-01-25
"Already seen" banners driving volunteers crazy


mkosmala:
Not sure if this is a back-end or front-end issue, so posting here and mentioning both @camallen and @brian-c (and @aliburchard too).

I've gotten a message on Talk for Season Spotter Questions saying that a user has been seeing "finished" on the images:
https://www.zooniverse.org/projects/kosmala/season-spotter-questions/talk/60/35704

Slightly more than a third of the images are retired, so we are far from being done. I also don't think that any of the workflows is done -- I checked them on Wednesday of last week and none were close.

I just did a recruitment thing on Saturday and would really really really like to not lose volunteers over this. (We need them!) Could you guys look into what might be happening? Thanks.

camallen:
@mkosmala this may happen under certain circumstances. If everything is finihsed or they have seen everything or mostly by chance when the users queues are empty and we just try and show them something (anything), see [here](https://github.com/zooniverse/Panoptes/blob/4d6a34f22950c05db4db4df275a6e5a6e25593ba/lib/subjects/selector.rb#L62).

Can you ask the user if this is still happening for them? I've checked the state of the workflows, it's not finished and they are not a power user so it was most likely a temporary issue.

mkosmala:
"when the users queues are empty and we just try and show them something (anything)"

Just trying to understand this... why would a user's queue be empty? Wouldn't you then grab a new set of subjects? 

(Sorry to harp on this. We lost so much momentum with the problems at launch -- including users getting "finished" messages then -- that I want to make sure this problem is as minimized as possible as an ongoing thing.)

camallen:
When you load the classify page the api returns 10 subjects from the queue. The default queue length is now 20, so two page loads (Ctrl + R) would deplete the queue. We only rebuild the queue on classification receipt to avoid repopulating with what we've just sent out (we haven't got updated state about what images a user has seen in the db until we get a classification for it). 

So under normal conditions the queue should **not** be empty and should always be filled ready to server data. However it can be empty sometimes due to load on the api since we may not run the job to requeue before the a request for more data comes in (under heavy load and a fast workflow the job may not run in time). With this in mind we always default to showing the user something and we have a tiered strategy on that, we attempt to show them something they haven't seen that isn't retired (power users / finished workflows will be empty) and then we fallback to just select some data from the workflow, anything. 

[...]

PR #1759, opened 2016-04-08
split dequeue subject queue workers up

camallen:
linked to #1640

rate limit non-logged in ones (no user) but do not skip the non-logged in ones. Also bump the queue priority for the logged in dequeue operation as we need to get this data out of the queue asap for a user session.

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -129,12 +129,14 @@ module Subjects
 
     def dequeue_ids(queue, sms_ids)
       if queue_user
-        queue.dequeue_update(sms_ids)
+        begin
+          queue.dequeue_update(sms_ids)
+        rescue ActiveRecord::StaleObjectError
+          DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
+        end
       else
-        DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
+        NonLoggedInDequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
       end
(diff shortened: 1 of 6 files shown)


lib/subjects/selector.rb
```diff
@@ -131,7 +131,7 @@ def dequeue_ids(queue, sms_ids)
       if queue_user
         queue.dequeue_update(sms_ids)
       else
-        DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
+        NonLoggedInDequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
       end
     rescue ActiveRecord::StaleObjectError
       DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
```
marten: Because I just now had to think about whether this one should have been changed as well, should we move this rescue block so that the method reads:

```ruby
    def dequeue_ids(queue, sms_ids)
      if queue_user
        begin
          queue.dequeue_update(sms_ids)
        rescue ActiveRecord::StaleObjectError
          DequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
        end
      else
        NonLoggedInDequeueSubjectQueueWorker.perform_async(queue.id, sms_ids)
      end
    end
```

To make it clear that the DequeueSubjectQueueWorker is a fallback in case inline removal for a user queue didn't work the first try?

marten:
LGTM

lib/subjects/selector.rb
camallen: agree it reads better, will do.

PR #1762, opened 2016-04-12
Optimize role control

marten:
Investigating why `/api/v1/user_groups` and `/api/v1/memberships` were so slow to load, I found that the issue could be solved by splitting the query up in two, rather than doing a subquery. PostgreSQL seems to be able to leverage indexes rather than doing seq scans this way.

I'd love to roll this out under Scientist, but we'd have to set that up first and since `scope_for` just returns a scope rather than records, I'm not too sure where I'd start measuring the difference and timing.

# Review checklist


camallen:
LGTM

PR #1764, opened 2016-04-13
Add Scientist

marten:
Adds Scientist which will post to Librato.

TODO:

- [x] Store experiments in database so that we can ramp up and down dynamically
- [ ] Store mismatches somewhere (print to logs?)

# Review checklist


camallen:
Looking good - also re database we could use env vars in the meantime too but i kinda agree the DB state could be nice, especially if it tracks experiment history, etc. 

Is this going to wait around for an experiment to setup?

marten:
I was thinking I'd `find_or_create` experiments by name, and create them with either a 0 or very low run rate. Then we can either ramp up through the Rails/PG console or I could add a controller for this so we could do it through Lita (in theory Lita could also fetch the charts for an experiment from Librato then). Depends on how polished we want to make this on the first go :)

camallen:
LGTM - this is exciting!!

Issue #1748, opened 2016-03-30
Subjects still not retiring correctly


mkosmala:
I thought this was fixed long ago. But I'm now looking at the data generated from last Tuesday, when @mrniaboc sent out the newsletter about the Season Spotter spring challenge. All subjects are supposed to retire with 5 classifications. But they don't necessarily. At first I thought maybe this is because of the way subjects are served in sets -- if classification 5 comes in while someone else is working on the same subject, we might occasionally end up with 6 classifications. But that's not the (only) issue. I also have subjects like subject 589803 for which I have 7 classifications, with created_at stamps of:
2015-12-02 20:17:40 UTC
2015-12-04 15:37:26 UTC
2015-12-22 20:02:42 UTC
2015-12-22 20:48:34 UTC
2015-12-23 15:00:59 UTC
2016-02-02 20:09:06 UTC
2016-03-23 01:11:49 UTC
This subject should have retired in December. Why was it sent out again in February? And again last week?

Please look into this. It's costing us a TON of volunteer effort. (I estimate that of the 30,000 classifications that came in last Tuesday, ~6,000  (i.e. 20%) of them were for subjects that should have been retired.)

mschwamb:
@mkosmala Are you sure those users classifying in February didn't classify the image when it was in a retired state after they've seen everything else? The interface will someone new images that are retired but they haven't seen with the retired flag on it, This might be more related to https://github.com/zooniverse/Panoptes-Front-End/issues/1050

mkosmala:
@mschwamb I thought of this, but it is definitely not what is going on. Our most prolific classifier hasn't even seen 25% of the subjects in our most-complete workflow and a double-check of her classifications shows that 7 of her most recent 10 classifications in that workflow were on subjects that hadn't reached retirement yet. (One of those other 3, sadly, has been classified by 10 people now! And we still have ~10% of our subjects that have fewer than 5 classifications...)

camallen:
Hey @mkosmala 

the subject retired just fine on 23rd Dec 2015
```
workflow.id
=> 174
workflow.finished_at
=> Wed, 23 Mar 2016 08:45:52 UTC +00:00
subject_workflow.retired_at     
=> Wed, 23 Dec 2015 15:00:59 UTC +00:00
```

Email on 3rd March 2016, you asked me to ensure only workflow_id: #174 was active, the rest were set to inactive in preparation for your promotion?

[...]

PR #1767, opened 2016-04-15
fix retired? workflow implementation

### LINKED ISSUE #1482: project subject counts should only respect linked subject sets
Project subject_counts should only summate the linked workflow subject sets and not all the project subject sets. Currently `workflow.subject_counts` respects the unlinked state but `project.subjects_count` doesn't. [Change here](https://github.com/zooniverse/Panoptes/blob/59ffad364842e0f326c5d9d3ac7c0e5f20e5780f/app/models/concerns/subject_counts.rb#L5) to use an `active_subject_sets` method differentiates on workfow / project subjects sets.

### LINKED ISSUE #1572: retirement counters dont' take into account unlinked data
as happened on snapshots as sea, the workflow finished and finished counts matched the subjects count but then the owner unlinked all the active sets and added new ones. This left the retired counts much larger and counting old data that wasn't currently active compared to an arbitrary number of currently linked active subjects.

camallen:
closes #1572 and closes #1482. linked to #1640 and #1748 

Don't use the moving subject counts (new / old data being (un)linked) to determine finished state.
Instead use the workflow.finished_at field set by retirement worker.
Also extract subject count methods to the respective models, add project / workflow specific 
behaviour for the subject counts and add a live_subject_sets relation to project 
via the workflows relation to only count the active subject sets in a project.

# Review checklist


app/models/project.rb
```diff
@@ -127,4 +127,16 @@ def send_notifications
...
+  def subjects_count
+    @subject_count ||= live_subject_sets.sum :set_member_subjects_count
+  end
+
+  def retired_subjects_count
+    @retired_subject_count ||= workflows.sum :retired_set_member_subjects_count
+  end
+
+  def finished?
+    @finished ||= workflows.all? { |w| w.finished_at.present? }
```
marten: Couldn't that just be `.all?(&:finished?)`

camallen: yep

marten:
LGTM

PR #1769, opened 2016-04-15
Export operations

marten:
Extract the handling of various project exports to operation classes. More lines of code, but faster to test because we don't need the whole controller layer.

# Review checklist


spec/operations/projects/create_classifications_export_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe Projects::CreateClassificationsExport do
+  let(:user) { create :user }
+  let(:api_user) { ApiUser.new(user) }
+  let(:operation) { described_class.with(api_user: api_user) }
+  let(:project) { create(:full_project, owner: user) }
+
```
camallen: agree with hound on this

marten: Meh. I had to disable my editor config when I made a change to talk. Will fix.

camallen:
LGTM - slowly shrinking the controllers and getting the services pattern rolled out. Can we hold of merging this till next week?

marten:
Sure.

PR #1771, opened 2016-04-15
serialize the subject_limit attribute name

camallen:
Linked to https://github.com/zooniverse/Panoptes-Front-End/pull/2379

Serialise the attribute name for use in the client. 

# Review checklist


app/serializers/user_serializer.rb
```diff
@@ -15,7 +15,7 @@ class UserSerializer
 
   media_include :avatar, :profile_header
 
-  def max_subjects
+  def subject_limit
```
marten: I assume this override method can just be removed now?

marten:
LGTM apart from that one comment

camallen:
rebased and updated.

PR #1773, opened 2016-04-18
[NO MERGE] DDL migrations in preparation of counting changes

marten:
* Makes sure looking up classifications by subject_id is an indexed query.
* Makes sure we can look up whether a project was live at time of classification using an indexed query.

# Review checklist

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -137,4 +137,13 @@ namespace :migrate do
       result_pages.update_all(url_key: 'results', title: 'Results')
     end
   end
+
+  namespace :workflow do
+    desc 'Fill workflow launch dates using project launch date'
+    task :launch_date => :environment do
+      Project.where("launch_date IS NOT NULL").find_each do |project|
+        project.workflows.where(active: true).update_all(launch_date: project.launch_date)
+      end
+    end
+  end
 end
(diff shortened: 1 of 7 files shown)


lib/tasks/migrate.rake
```diff
@@ -117,6 +117,11 @@ namespace :migrate do
         .update_all(lifecycled_at: Time.current.to_s(:db))
       end
     end
+
+    desc "Fill live_project"
+    task fill_live_project: :environment do
+      Classification.where(live_project: nil).update_all("live_project = (metadata->>'live_project')::boolean")
```
marten: This one takes a bunch of minutes, and I'm not sure if it does a table lock (at least for writes). Any thoughts on how to do it better?

camallen: split it into small batches and update them, ala what you did [here](https://github.com/zooniverse/Panoptes/blob/04d6d77eed29b2e9ced5ff771707851bbb679f11/lib/tasks/migrate.rake#L80)

camallen: What is the use case for the `Classification.where(live_project: true)` query, in the dumps?

marten:
After some discussions IRL, I'm going to rework this.

marten:
I don't know why this is conflating multiple things anymore. Closing this PR and starting over.

PR #1776, opened 2016-04-19
simplify the owner filter join and eager load the identity group users

### LINKED ISSUE #1774: Collections controlled scope returning duplicate records
Linked to https://github.com/zooniverse/Panoptes-Front-End/issues/2460

Collections controller index action return duplicate records for some users `https://www.zooniverse.org/collections/mvgulik`. See the count vs the actual returned records (seems AR is filtering the dups via the join vs the underlying count).

I'm not 100% on the solution apart from removing the dups could be done by adding a distinct scope_context or just making sure the constructed `public_query` does not return duplicates.

camallen:
closes #1774 

the previous query was creating 2 inner joins on the access_control_lists and creating duplicates on the join which was affecting the overal sql count queries, i.e. 1 result but count was 2. This is mostly creating shared helpers to test the filtering (which we weren't before).

TODO:
- [X] See about the N+1 on the owner preload for role filtering...(I failed on this one...)

# Review checklist

### DIFF
--- a/app/controllers/concerns/filter_by_owner.rb
+++ b/app/controllers/concerns/filter_by_owner.rb
@@ -10,11 +10,9 @@ module FilterByOwner
     unless owner_filter.blank?
       groups = UserGroup.where(UserGroup.arel_table[:name].lower.in(owner_filter.map(&:downcase)))
       @controlled_resources = controlled_resources
-        .eager_load(:owner)
-        .joins(:owner)
-        .joins(:access_control_lists)
-        .where(access_control_lists: {user_group: groups})
-        .where.overlap(access_control_lists: { roles: ["owner"] })
+      .eager_load(owner: [:users])
+      .joins(:owner)
+      .where(access_control_lists: { user_group: groups })
     end
   end
 end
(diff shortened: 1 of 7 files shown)


marten:
Devils advocate: how did you know this change fixed the bug? Did you write a (failing) test to see the count being incorrect, and then get that to green?

camallen:
nope - i should have marked this as WIP i wanted to see what failed.

camallen:
@marten added shared filtering specs (missing on collections) and rebased - should be good to go now.

app/controllers/concerns/filter_by_owner.rb
```diff
@@ -10,11 +10,9 @@ def add_owner_ids_to_filter_param!
...
       groups = UserGroup.where(UserGroup.arel_table[:name].lower.in(owner_filter.map(&:downcase)))
       @controlled_resources = controlled_resources
-        .eager_load(:owner)
-        .joins(:owner)
-        .joins(:access_control_lists)
-        .where(access_control_lists: {user_group: groups})
-        .where.overlap(access_control_lists: { roles: ["owner"] })
+      .eager_load(owner: [:users])
+      .joins(:owner)
+      .where(access_control_lists: { user_group: groups })
```
marten: So you removed the join on access_control_lists, but you're still able to query on it. How does that work? And how can we make that more obvious in the code?

camallen: via the owner relation - not sure how to make it more obvious.

camallen: could add a comment on it

PR #1780, opened 2016-04-21
fix evaluation order of class / instance methods

camallen:
Testing existence of instance methods before they were defined was stupid. Do not silently allow invalid method definitions in the cache key DSL, instead raise an error if the method doesn’t exist so it can be fixed.

# Review checklist


zwolf:
Not checking a class for a given association makes sense here. There should also be a spec for the case where the class doesn't have the given association/method and throws an exception.

camallen:
Good call - will add it.

camallen:
@zwolf this should be good to go now.

PR #1787, opened 2016-04-25
Record a reason when a subject is retired

zwolf:
Reasons for retirement are stored in an enum on the SubjectWorkflowCount model. The current list of possible reasons looks like this:

[ :classification_count, :flagged, :blank, :consensus, :other ]

Nil is also still fine. Nero can set this value when it retires a subject after it detects it in a classification.

# Review checklist

### DIFF
--- a/app/models/subject_workflow_count.rb
+++ b/app/models/subject_workflow_count.rb
@@ -31,7 +31,7 @@ class SubjectWorkflowCount < ActiveRecord::Base
     return if retired?
 
     ActiveRecord::Base.transaction(requires_new: true) do
-      update!(retirement_reason: reason, retired_at: Time.now.utc)
+      update!(retirement_reason: reason, retired_at: Time.zone.now)
       Workflow.increment_counter(:retired_set_member_subjects_count, workflow.id)
       yield if block_given?
     end
--- a/app/models/subject_workflow_count.rb
+++ b/app/models/subject_workflow_count.rb
@@ -2,6 +2,8 @@ class SubjectWorkflowCount < ActiveRecord::Base
   belongs_to :subject
   belongs_to :workflow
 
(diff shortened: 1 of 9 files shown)


app/models/subject_workflow_count.rb
```diff
@@ -25,11 +27,11 @@ def retire?
     workflow.retirement_scheme.retire?(self)
   end
 
-  def retire!
+  def retire!(reason=nil)
     return if retired?
 
     ActiveRecord::Base.transaction(requires_new: true) do
-      touch(:retired_at)
+      update!(retirement_reason: reason, retired_at: Time.now.utc)
```
camallen: Might want to use `Time.zone.now`

app/operations/workflows/retire_subjects.rb
```diff
@@ -6,11 +6,12 @@ class RetireSubjects < Operation
     array :subject_ids, default: [] do
       integer
     end
+    string :retirement_reason, default: nil
```
camallen: Are we controlling the values that can be set here, what if i pass in `retirement_reason: "not_a_reason"`? we haven't spec'd this scenario and i assume it'll be set to nil by default. I think this is fine but we should spec this expectation to make sure that behaviour persists OR control the input values for this attribute.

zwolf:
@camallen Added a validator to the RetireSubjects operation that also allows nil. Made that Time call change, too.

camallen:
This is looking good, the final thing is to check the controller will catch invalid operation schema data and raise an `ActiveInteraction::InvalidInteractionError` error via the api_controller, e.g. [here](https://github.com/zooniverse/Panoptes/blob/cbd4cdff8ef6ccf7dd3e1951dc91a052fc880ec8/app/controllers/api/api_controller.rb#L44).

camallen:
LGTM - feel free to merge when those specs are all passed.

PR #1788, opened 2016-04-26
raise project introduction limit to 5K

camallen:
Closes https://github.com/zooniverse/Panoptes-Front-End/issues/2239


zwolf:
Looks good, if 5000 characters is going to be enough. Did they provide a sample chunk o' text to measure?

camallen:
Nope, just taking @vrooje 's suggestion of 5K from https://github.com/zooniverse/Panoptes-Front-End/issues/2239#issuecomment-214777704

zwolf:
LGTM, then. Can always raise it again later in the unlikely event that's necessary.

camallen:
5K should really be enough even with markup. That said you're right, we can always raise if needed.

Issue #1558, opened 2015-12-18
Project state


parrish:
This became an issue in #1552 and #1557.  Currently project state is managed as a combination of boolean fields on the project and role-based permissions.

This allows for all sorts of bungled settings like launch approved, but not requested; private, but launched; etc.

As I understand it, here's the current flows of project state:

Private -> Beta Requested -> Beta Approved -> Launch Requested -> Launch Approved -> Public

I'm not really sure where the "live: live/development" fits into this.

I'm planning to build out a state machine to consolidate and control transitions into a single state field.

@camallen @marten thoughts?

camallen:
I think a state machine is a good idea. Just a few clarifications, The current flows of project state:

Private -> Can never request to be reviewed via the GUI. Private projects are for small teams that have data concerns and will never be "sponsored" by the zooniverse. 

Public -> Beta Requested -> Beta Approved -> Launch Requested -> Launch Approved. You can never request / progress through any of these states unless the project is public.

Finally, live / development is to signal that the project's active workflows should not be touched and to turn on retirement, i.e. the project is live in the system. We use this flag as a general block on workflow changes in the UI too but @marten has a nice idea to deal with this, see #1523. If a project is in dev mode you can't request beta approval to start the review process too.

aliburchard:
One small caveat to @camallen's flow above: projects often cycle a bit in the beta approved --> launch requested stage. After beta is approved, the project goes through beta testing, the project owners then make changes to the interface based upon the beta feedback, and then *currently* have to re-request beta. (This is because the act of taking a project out of "live" to make workflow changes essentially clears the history of beta requested and approved.) A better solution would be to have the beta states remembered so that projects that go through beta, make changes, then reset their project to Live can request launch without havng to re-request beta.

aliburchard:
@parrish relevant conversation to the re-requesting beta problem here: https://github.com/zooniverse/Panoptes-Front-End/issues/2003

parrish:
Closed by #1791

PR #1791, opened 2016-05-03
Project state

zwolf:
Adds "state" as a project attribute and to its serializer. Project#state will, in the case of a nil value, return either "live" or "development" depending on launch_approved?. @marten assisted with the serializer portion on account of Restpack's confusion about the enum and my confusion about Restpack.

# Review checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -42,6 +42,8 @@ class Project < ActiveRecord::Base
 
   has_paper_trail only: [:private, :live, :beta_requested, :beta_approved, :launch_requested, :launch_approved]
 
+  enum state: [:paused, :finished]
+
   cache_by_association :project_contents, :tags
   cache_by_resource_method :subjects_count, :retired_subjects_count, :finished?
 
@@ -137,6 +139,14 @@ class Project < ActiveRecord::Base
   end
 
   def finished?
-    @finished ||= workflows.all?(&:finished?)
+    super ? super : workflows.all?(&:finished?)
+  end
+
+  def state
+    if self[:state]
+      super
+    else
+      live ? "live" : "development"
+    end
   end
 end
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -11,7 +11,7 @@ class ProjectSerializer
     :configuration, :live, :urls, :migrated, :classifiers_count, :slug, :redirect,
     :beta_requested, :beta_approved, :launch_requested, :launch_approved, :launch_date,
     :href, :workflow_description, :primary_language, :tags, :experimental_tools,
-    :completeness, :activity
+    :completeness, :activity, :state
 
   optional :avatar_src
   can_include :workflows, :subject_sets, :owners, :project_contents,
@@ -21,9 +21,14 @@ class ProjectSerializer
     subjects_export: { include: false },
     aggregations_export: { include: false }
   can_filter_by :display_name, :slug, :beta_requested, :beta_approved,
-    :launch_requested, :launch_approved, :private
+    :launch_requested, :launch_approved, :private, :state
   can_sort_by :launch_date, :activity, :completeness, :classifiers_count,
     :updated_at, :display_name
+
+  def self.page(params = {}, scope = nil, context = {})
+    params["state"] = Project.states[params["state"]] if params.key?("state")
+    super(params, scope, context)
+  end
 
   def self.links
     links = super
--- a/db/migrate/20160427150421_add_project_state.rb
+++ b/db/migrate/20160427150421_add_project_state.rb
@@ -1,0 +1,6 @@
+class AddProjectState < ActiveRecord::Migration
+  def change
+    add_column :projects, :state, :integer
+    add_index :projects, :state
+  end
+end
--- a/db/migrate/20160427150421_add_project_state.rb
+++ b/db/migrate/20160427150421_add_project_state.rb
@@ -1,6 +1,5 @@
 class AddProjectState < ActiveRecord::Migration
   def change
     add_column :projects, :state, :integer
-    add_index :projects, :state, where: "(state IS NOT NULL)", algorithm: :concurrently
   end
 end
--- a/db/migrate/20160427150421_add_project_state.rb
+++ b/db/migrate/20160427150421_add_project_state.rb
@@ -1,6 +1,6 @@
 class AddProjectState < ActiveRecord::Migration
   def change
     add_column :projects, :state, :integer
-    add_index :projects, :state
+    add_index :projects, :state, where: "(state IS NOT NULL)", algorithm: :concurrently
   end
 end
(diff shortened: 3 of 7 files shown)


app/models/project.rb
```diff
@@ -139,4 +141,12 @@ def retired_subjects_count
   def finished?
     @finished ||= workflows.all?(&:finished?)
   end
+
+  def state
+    if self[:state]
+      super
+    else
+      launch_approved ? "live" : "development"
```
camallen: hmm - we've already got a live flag on projects...seems we're conflating this flag here. Just because a project is not launch approved (goes on the home page) doesn't mean it's not "live". I think we should probably just use the live flag itself here instead of `launch_approved` (you can't be launch approved without being live).

app/serializers/project_serializer.rb
```diff
@@ -21,10 +21,15 @@ class ProjectSerializer
     subjects_export: { include: false },
     aggregations_export: { include: false }
   can_filter_by :display_name, :slug, :beta_requested, :beta_approved,
-    :launch_requested, :launch_approved, :private
+    :launch_requested, :launch_approved, :private, :state
   can_sort_by :launch_date, :activity, :completeness, :classifiers_count,
     :updated_at, :display_name
 
+  def self.page(params = {}, scope = nil, context = {})
+    params["state"] = Project.states[params["state"]] if params.key?("state")
```
camallen: Question? Did you think about / try extracting the enum out to their own columns? I can't help think about this issue, #1558

app/models/project.rb
```diff
@@ -40,6 +40,8 @@ class Project < ActiveRecord::Base
   has_many :tagged_resources, as: :resource
   has_many :tags, through: :tagged_resources
 
+  enum state: [:paused, :finished]
```
camallen: Does this conflict with the current finished method [here](https://github.com/zooniverse/Panoptes/blob/aa00e646aaa458716220dd2d8fd3542bebb4bc28/app/models/project.rb#L139)?

camallen:
Some minor issues with the implementation. I can't help feel we're missing a chance to rethink the states and we're just adding new ones to conflate the issue...that said we can iterate on this and get the proposed paused / finished stuff out there.

zwolf:
Each of the two (finished, paused) states being separate attributes was the initial idea, but the consensus seemed to be that it was useful to have a more extensible way to add more project states (like complete, someone suggested). The ways that project state is currently calculated in other ways (#finished?, live) could be folded in. But the enum issues discussed in #1558 are relevant and it could get (is already) confusing and allow contradictory states.

camallen:
> that it was useful to have a more extensible way to add more project states

agree. I think we should go ahead with this and tackle #1558 with a some higher review of what is working / what isn't. Can you look to address the issues i've noted inline?

zwolf:
Regarding specifically state[:finished] vs #finished?, which one is more useful? I don't actually see anywhere in Panoptes where #finished? is being called on a project, but that doesn't mean it isn't happening. Could just move that workflows.all?(&:finished?) check to inside #state.

app/models/project.rb
camallen: you should be able to proxy the finished method via a super call and default back to the current implementation by the way.

camallen:
the enum will add a finished? and finished! method won't it, we can just call super in the current method to get the enum state and if it's not set then use the current implementation.

app/serializers/project_serializer.rb
zwolf: This is the piece that @marten helped me with that solved the enum-related int/string issue. Enums are cool but feel kind of squirrely in practice, especially underneath Restpack.

zwolf:
Ok, #finished? will now check super and return the calculated boolean if it isn't "finished". This is instead of returning false because there could be some overlap there. Once we talk about #1558 and get a better map of project states together we can decide if/how to combine them all.

I'm trying to figure out what's up with all those other commits being included with my last. I was unable to push despite being up to date with the branch (I committed + pushed right before the PR), so I rebased to master and --force-with-leased it and all those were added even though they should have been already included. It all looks correct from here, so I'm not sure what I need to do to fix the non-existent conflicts. Sorry about that.

camallen:
@zwolf weird, i'd rebase off an updated master (it should work unless those old commits have new commit ids). 

If that still doesn't work you can move the current branch to something else (`git branch -m project-state-wtf`), then checkout a new one from the updated master with the same name and then finally cherrypick out that latest commits from the renamed branch.

db/migrate/20160427150421_add_project_state.rb
```diff
@@ -0,0 +1,6 @@
+class AddProjectState < ActiveRecord::Migration
+  def change
+    add_column :projects, :state, :integer
+    add_index :projects, :state
```
camallen: I'm wondering if this should be a sparse index? I.e. with a predicate to only index set values and not all:
`add_index :projects, :state, where: "(state IS NOT NULL)", algorithm: :concurrently`

spec/serializers/project_serializer_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe ProjectSerializer do
-  let(:project) { create(:full_project) }
+  let(:project) { create(:full_project, :state => "finished") }
```
camallen: Nitpick, i'd prefer to not use hashrockets instead just `state: "finished"`

spec/serializers/project_serializer_spec.rb
```diff
@@ -28,6 +28,18 @@
...
   end
 
+  describe "#state" do
+    it "includes the state" do
+      expect(serializer.state).to eq project.state
+    end
+
+    it 'can filter by state' do
+      project.finished!
+      serializer = described_class.page({"state" => "finished"})
```
camallen: Also does this actually test the filtering or just the matching factory data? (it probably does but would probably pass without the values passed to page...probably best to test for a matching filter with another non-matching factory in there?

camallen:
This is super close, couple of minor comments.

db/structure.sql
```diff
@@ -2213,6 +2214,13 @@ CREATE INDEX index_projects_on_slug ON projects USING btree (slug);
 
 
 --
+-- Name: index_projects_on_state; Type: INDEX; Schema: public; Owner: -; Tablespace: 
+--
+
+CREATE INDEX index_projects_on_state ON projects USING btree (state);
```
camallen: you'll need to re-run the migrations to recreate the structure dump.

zwolf: Is there a way to force it? db:migrate doesn't run anything in any environment and Travis seems to like the test db.

spec/serializers/project_serializer_spec.rb
```diff
@@ -28,6 +28,20 @@
     end
   end
 
+  describe "#state", :focus do
```
camallen: focus set still active.

camallen:
I think this is mostly there but I'd delay merging / deploying till early next week and get @marten to have a final once over (it's late on a long friday for me), nice work @zwolf 👍

PR #1792, opened 2016-05-04
compare user seens to live data only

### LINKED ISSUE #1786: User finished workflow test should compare the live data
linked to #1767

Currently it's a straight counter comparison that doesn't take into account the unlinking of data. Once work in #1772 is finished this can be a query against the current set of live data and the submitted classification data instead of relying on counters that compare apples and oranges.

camallen:
closes #1786 use a count of the linked / current seen subjects vs what the workflow has linked

# Review checklist


zwolf:
Looks good to me. Lemme know if you'd like to discuss before merging.

Issue #1785, opened 2016-04-22
update csv classification export for new tasks


mcbouslog:
update csv classification export for new tasks - combo, dropdown, maybe text tool within drawing task?

[current code in Panoptes](https://github.com/zooniverse/Panoptes/blob/cbd4cdff8ef6ccf7dd3e1951dc91a052fc880ec8/lib/formatter/csv/annotation_for_csv.rb#L15)

[PFE ReadMe annotation format](https://github.com/zooniverse/Panoptes-Front-End#format-of-annotation-values)

mentioning people that may be interested in this: @zwolf, @simensta, @marten, @vmaidel

mcbouslog:
follow-up comment - Laura and I hope to show the science team a sample export May 19th. sorry for the tight deadline!

@zwolf, @simensta

mcbouslog:
Project #2492 (created on production) includes three workflows with classifications made that will hopefully work well for updating the classification dump worker:

Combo Workflow - 3-task combo task, with a question, text then question
Dropdown Workflow - country, state, and city conditionally related dropdowns
Drawing with Text Tool Workflow - drawing task with text sub-task

Let me know how/if further classifications, workflows or edits would be helpful.

PR #1797, opened 2016-05-10
CSV export to allow new tasks

zwolf:
In order to allow the more complex Combo and Dropdown tasks to be correctly exported, I did some refactoring of Formatter::Csv::AnnotationForCsv. Hopefully this will make it easier to add new tasks in the future. 

~~This is a Work In Progress PR, and while it should all currently be working, not all the upstream specs are passing, it could be prettier/DRYer, and Hound is going to throw a fit. Please let me know, @marten or @camallen, about changes that you'd like to see. I anticipate there will be many, but I wanted to get this into your hands ASAP.~~

Update: ready to be reviewed.

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -10,23 +10,42 @@ module Formatter
       end
 
       def to_h
-        annotation.merge! "task_label" => task_label
-
-        case task["type"]
-        when "drawing"
-          value_with_tool = (annotation["value"] || []).map do |drawn_item|
-            drawn_item.merge "tool_label" => tool_label(drawn_item)
-          end
-          annotation.merge!("value" => value_with_tool)
-        when /single/i
-          annotation.merge!("value" => answer_label)
-        when /multiple/i
-          annotation.merge!("value" => answer_labels)
-        end
-        annotation
+        send(task["type"])
       end
 
       private
+
+      def drawing
+        # annotation.merge! "task_label" => task_label
+        # value_with_tool = (annotation["value"] || []).map do |drawn_item|
+        #  drawn_item.merge "tool_label" => tool_label(drawn_item)
+        # end
+        # annotation.merge!("value" => value_with_tool)
+      end
+
+      def simple
+        {}.tap do |new_annotation|
+          new_annotation['task'] = annotation['task']
+          new_annotation['task_label'] = task_label
+          new_annotation['value'] = task['type'] == 'multiple' ? answer_labels : answer_label
+        end
+      end
+
+      alias_method :single, :simple
+      alias_method :multiple, :simple
+
+      def combo
+        # {}.tap do |new_annotation|
+        #   annotation['value'].each do |subtask|
+        #     binding.pry
+
+            # anno = annotation_by_task(subtask).last
+            # answer_string = anno['answers'][subtask['value']]['label']
+            # new_annotation['value'] = translate(answer_string)
+            # new_annotation['task_label'] = translate(anno['question'])
+          # end
+        # end
+      end
 
       def task_label
         translate(task["question"] || task["instruction"])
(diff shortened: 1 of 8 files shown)


lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -6,34 +6,87 @@ class AnnotationForCsv
       def initialize(classification, annotation, cache)
         @classification = classification
         @annotation = annotation.dup.with_indifferent_access
+        @current = @annotation.dup
         @cache = cache
       end
 
       def to_h
-        annotation.merge! "task_label" => task_label
+        send(task['type'])
```
marten: Extracting those methods is a great idea. But I'm not sure there's a whitelist on this anywhere, might be a remote code execution vulnerability. I'd rather we keep a `case when ...` for the known cases. That also means we can be explicit in the `else` clause about what happens for unknown tools (rather than raising a NotMethodError).

zwolf: I added a whitelist in the form of a private method with an array of valid task types to check against. Also, an error hash to jsonify and include in the csv if the task is invalid. I think that should cover us. If you'd rather go back to case/when, that's cool. This also seemed to help get some of the failing Formatter::Csv::Classification specs passing, woop.

spec/factories/workflow_contents.rb
```diff
@@ -3,16 +3,59 @@
...
+            "interest.tools.1.label" => "Green",
+            "interest.tools.2.label" => "Blue",
+            "shape.question" => "What shape is this galaxy",
+            "shape.help" => "Duh?",
+            "shape.answers.0.label" => "Smooth",
+            "shape.answers.1.label" => "Features",
+            "shape.answers.2.label" => "Star or artifact",
             })
+
+    factory :combo_workflow_content do
```
marten: I've been starting to move from nested factories to [traits](https://robots.thoughtbot.com/remove-duplication-with-factorygirls-traits). Probably a better fit here too?

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -6,34 +6,100 @@ class AnnotationForCsv
...
+
+      def simple(task_info=task)
+        {}.tap do |new_anno|
+          new_anno['task'] = @current['task']
+          new_anno['task_label'] = task_label(task_info)
+          new_anno['value'] = task['type'] == 'multiple' ? answer_labels : answer_label
+        end
+      end
+
+      alias_method :single, :simple
```
camallen: do we need these method aliases still?

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -6,34 +6,100 @@ class AnnotationForCsv
...
-        tool = task["tools"] && task["tools"][drawn_item["tool"]]
+      def dropdown
+        {}.tap do |new_anno|
+          new_anno['task'] = @current['task']
+          new_anno['value'] = []
+          task['selects'].each_with_index do |sel, idx|
+            {}.tap do |tmp|
+              tmp['select_label'] = sel['title']
+              tmp['option'] = sel['allowCreate']
+              sel['options'].each do |key, arr|
```
camallen: these nested blocks do my head in...any chance we can refactor?

camallen:
This is looking pretty good - minor comments.

lib/formatter/csv/annotation_for_csv.rb
zwolf: Definitely my least favorite part, but I was trying to just get it working. The nesting is due to the unfortunate way the data is structured. I'm sure there's a clever way to do this, I'll think on how to make it more readable. Suggestions appreciated.

camallen: i'd rather refactor now so we don't just get this merged and forget about it. Also we can test this on a production (replica) / staging db manually if needed.

zwolf: All about it, just meant that I'm gonna knock out a few of these other issues first.

camallen:
LGTM, @marten any concerns?

marten:
Having a better look, that `@current` variable bothers me. Probably because it makes stuff have mutating state.

zwolf:
In the case of combo tasks, making @current available class-wide prevents having to pass the annotation around to every method. Seemed the cleaner case, is there a better way to allow all methods access to the annotation currently being processed?

marten:
The one I can come up with is splitting up into many classes, which means the combo task can instantiate sub-objects for the subtasks. But I'm okay with merging this for the time being.

zwolf:
Closes #1785

Issue #1798, opened 2016-05-10
make subject id a separate field for classification csv exports


zwolf:
Include an array (or delimited string, or similar) of subject ids in the classification CSV export. Will be redundant with subject_data, but useful for optimization.

Opened by @ggdhines as a PFE issue here: https://github.com/zooniverse/Panoptes-Front-End/issues/2543

zwolf:
Closed by #1801

PR #1801, opened 2016-05-12
Includes subject_ids in the classification CSV export

zwolf:
Requested by @ggdhines in #1798 (and originally in PFE issue #2543). Included as an array, but whoever's manning the aggregation station now that Greg has left is welcome make suggestions. 

# Review checklist

### DIFF
--- a/lib/formatter/csv/classification.rb
+++ b/lib/formatter/csv/classification.rb
@@ -8,7 +8,7 @@ module Formatter
 
       def self.headers
         %w(classification_id user_name user_id user_ip workflow_id workflow_name workflow_version
-           created_at gold_standard expert metadata annotations subject_data)
+           created_at gold_standard expert metadata annotations subject_data subject_ids)
       end
 
       def initialize(project, cache)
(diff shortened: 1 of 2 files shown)


lib/formatter/csv/classification.rb
```diff
@@ -51,6 +51,10 @@ def subject_data
         end.to_json
       end
 
+      def subject_ids
+          classification.subject_ids.map {|id| cache.subject(id).id }.to_json
```
marten: Why go from subject_ids to subjects and back to their ids?

 Also I was wondering: if we just join the ids with something, say a semicolon (so instead of jsonifying), then in most cases (projects with only one subject per classification) this column will simply look like a number that's easy to work with in R or Excel. What do you think of that?

zwolf: Because it's the cache object's list of ids (relevant to that classification) that we really want. I'm still wrapping my head around how the cache is used, though. Could the current classification's list of subjects be out of date?

And delimiting the list would be just fine with me, if it's easier for their tools to parse.

marten: We needed the cache for subjects because `classification.subjects` didn't use to be an actual association (the list of subject_ids was stored as a serialized column). So the cache then loaded all the subjects for a batch of classifications (implementing `find_each` behaviour for this non-association). 

I've since turned that into a normal `has_many :subjects, through: :classification_subjects`, so we could probably just use Rails' eager loading for this now.

camallen:
LGTM

zwolf:
Closes #1798

PR #1805, opened 2016-05-16
Adds 'activity_count_by_workflow' to /project_preferences

### LINKED ISSUE #1796: expose user classification info by workflow
per Notes from Nature conversion request, to show user "badges" (earned by # of classifications) by "expedition" (which are workflows within Panoptes).

per conversation with @camallen:
-will be modifying [this code](https://github.com/zooniverse/Panoptes/blob/master/app/serializers/user_project_preference_serializer.rb#L19)
-maybe as an optional attribute or detecting the presence of a param

zwolf:
Activity counts via UserSeenSubjects are included per workflow to calls made to /project_preferences. There could be the potential for performance issues related to these queries, but the association is already indexed and the non-legacy counts are being generated on the fly already, so this isn't much heavier.

Will close #1796.

# Review checklist

### DIFF
--- a/app/models/user_seen_subject.rb
+++ b/app/models/user_seen_subject.rb
@@ -14,16 +14,16 @@ class UserSeenSubject < ActiveRecord::Base
   end
 
   def self.count_user_activity(user_id, workflow_ids=[])
+    workflow_counts = activity_by_workflow(user_id, workflow_ids)
+    workflow_counts.values.sum
+  end
+
+  def self.activity_by_workflow(user_id, workflow_ids=[])
     workflow_ids = Array.wrap(workflow_ids)
     scope = self.where(user_id: user_id)
     unless workflow_ids.empty?
       scope = scope.where(workflow_id: workflow_ids)
     end
-    scope.sum("cardinality(subject_ids)")
-  end
-
-  def self.activity_by_workflow(user_id)
-    scope = self.where(user_id: user_id)
     scope.group(:workflow_id).sum("cardinality(subject_ids)").as_json
   end
 
--- a/app/models/user_seen_subject.rb
+++ b/app/models/user_seen_subject.rb
@@ -14,13 +14,17 @@ class UserSeenSubject < ActiveRecord::Base
   end
 
   def self.count_user_activity(user_id, workflow_ids=[])
-
     workflow_ids = Array.wrap(workflow_ids)
     scope = self.where(user_id: user_id)
     unless workflow_ids.empty?
       scope = scope.where(workflow_id: workflow_ids)
     end
     scope.sum("cardinality(subject_ids)")
+  end
+
+  def self.activity_by_workflow(user_id)
+    scope = self.where(user_id: user_id)
+    scope.group(:workflow_id).sum("cardinality(subject_ids)").as_json
   end
 
   def subjects_seen?(ids)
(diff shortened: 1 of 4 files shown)


camallen:
specs currently failing

zwolf:
Passing. ID strings vs. ints strikes again, not sure why I didn't catch it.

app/models/user_seen_subject.rb
```diff
@@ -23,6 +22,15 @@ def self.count_user_activity(user_id, workflow_ids=[])
     scope.sum("cardinality(subject_ids)")
   end
 
+  def self.activity_by_workflow(user_id)
+    scope = self.where(user_id: user_id)
```
camallen: we should be able to push this to the DB with something like `scope.group(:workflow_id).sum("cardinality(subject_ids)").as_json`

camallen:
Looks good - i'm a bit worried about power users that work on lots of projects and looping over each one, push that to the DB and i think we're good.

zwolf:
Your suggestion worked, so if you're less concerned about stressing the db then that should help. I amended the commit, though in future I'll probably keep committing so you can more easily see the changes, then squash them when I merge. Cool?

camallen:
The DB is optimized for stuff like this, ruby land code not so much. Feel free to commit the way you like (I more often than not amend). We may ask you to squash them if it looks a bit noisy.

app/models/user_seen_subject.rb
```diff
@@ -23,6 +22,11 @@ def self.count_user_activity(user_id, workflow_ids=[])
     scope.sum("cardinality(subject_ids)")
```
camallen: could we just summate the per workflow call here now instead of running 2 different db calls?

camallen: and cache the one we do call.

camallen: Ignore these - too early :(

app/serializers/user_project_preference_serializer.rb
```diff
@@ -16,6 +16,10 @@ def activity_count
     end
   end
 
+  def activity_count_by_workflow
+    UserSeenSubject.activity_by_workflow(@model.user_id)
```
camallen: So - this is going to return all the workflows for all projects that a user has contributed too, not the ones for the particular project preference, see `user_project_activity` below. 

I'm also wondering if we can't cache the per workflow method here and use it in the below `user_project_activity` via a local cache? They seem to be doing very similar ops.

app/serializers/user_project_preference_serializer.rb
```diff
@@ -16,8 +16,15 @@ def activity_count
...
+    UserSeenSubject.activity_by_workflow(@model.user_id, project_workflows_ids)
+  end
+
   def user_project_activity
-    project_workflows_ids = Workflow.where(project_id: @model.project_id).pluck(:id)
     UserSeenSubject.count_user_activity(@model.user_id, project_workflows_ids)
   end
+
+  def project_workflows_ids
+    Workflow.where(project_id: @model.project_id).pluck(:id)
```
camallen: should we cache this call? the AR call will probably be cached but we are using it possibly twice per serialized UPP.

camallen:
LGTM - minor comment on the method call cache, up to you on this one.

PR #1808, opened 2016-05-17
SubjectSerializer includes seen_all_live_data for each subject

zwolf:
SubjectSerializer includes seen_all_live_data for each subject. Relevant line is 

`!!(user_seen && (user_seen.workflow.subject_ids.uniq.sort == user_seen.subject_ids.uniq.sort))`

If a user has seen all the subjects in the workflow, it'll be true and be included with each sample. If not, then the value will be false for all subjects (including those which have been seen).

# Review checklist

### DIFF
--- a/app/serializers/subject_serializer.rb
+++ b/app/serializers/subject_serializer.rb
@@ -5,7 +5,7 @@ class SubjectSerializer
   attributes :id, :metadata, :locations, :zooniverse_id,
     :created_at, :updated_at, :href
 
-  optional :retired, :already_seen
+  optional :retired, :already_seen, :seen_all_live_data
 
   can_include :project, :collections
 
@@ -25,6 +25,10 @@ class SubjectSerializer
     !!(user_seen && user_seen.subject_ids.include?(@model.id))
   end
 
+  def seen_all_live_data
+    !!(user_seen && (user_seen.workflow.subject_ids.uniq.sort == user_seen.subject_ids.uniq.sort))
+  end
+
   private
 
   def include_retired?
@@ -32,6 +36,10 @@ class SubjectSerializer
   end
 
   def include_already_seen?
+    selected?
+  end
+
+  def include_seen_all_live_data?
     selected?
   end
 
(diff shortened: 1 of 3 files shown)


marten:
Will this run a query for each subject being serialized in the response, or will that get cached somewhere so that it runs just one?

app/serializers/subject_serializer.rb
```diff
@@ -25,6 +25,10 @@ def already_seen
     !!(user_seen && user_seen.subject_ids.include?(@model.id))
   end
 
+  def seen_all_live_data
+    !!(user_seen && (user_seen.workflow.subject_ids.uniq.sort == user_seen.subject_ids.uniq.sort))
```
camallen: Not sure that will take into account the live set of data (the linked subjects are fluid), we could use something like this https://github.com/zooniverse/Panoptes/blob/master/app/models/user.rb#L217

camallen:
Ah - good point @marten.

Maybe we could just inject it into the selection context like we do [here](https://github.com/zooniverse/Panoptes/blob/94b27a1ed55655f33c1202e1eb9b8e7e021cd7cd/lib/subjects/selector.rb#L109) and then just reference that in the above (maybe change the serialized attribute to `finished_worfklow`)?

zwolf:
Would it solve both to just change that line to 

`@finished_workflow ||= !!(user_seen && (user_seen.user.has_finished?(user_seen.workflow)`

? If the class instance is the same for each subject it should only need to ask once, and it seems like User#finished? is doing the same thing I was trying for, anyway.

camallen:
you would have to do it for the context that is passed to the page method on the serializer. Otherwise it'd do that method for each subject and ignore the cached instance var (as it's a different subject being serialized)

app/serializers/subject_serializer.rb
```diff
@@ -3,9 +3,9 @@ class SubjectSerializer
   include FilterHasMany
 
   attributes :id, :metadata, :locations, :zooniverse_id,
-    :created_at, :updated_at, :href
+    :created_at, :updated_at, :href, :finished_workflow
```
camallen: any reason this came out of the optional block - it should only be serializing if the context is set (i.e. via the queued selection end point). Change it back to the way you had it for `seen_all_live_data` but keep the attribute naming you have here.

camallen:
Change the serializer attribute back to optional and this looks good to me. Nice work @zwolf!

PR #1812, opened 2016-05-25
Adds a unique index for user_group_id and user_id.

### LINKED ISSUE #1811: It is possible to have two memberships to the same group
Apparently the `memberships` table is missing unique index on `(user_group_id, user_id)`, which means some people have multiple memberships onto the same group. This causes issues when the roles vary per membership, as the ACL will just pick the one membership out of the database which means you sometimes don't get admin rights to a group you're an admin of.

### Desired behaviour

Joining a group should either merge roles or be a no-op.

simoneduca:
Fix #1811

### DIFF
--- a/db/migrate/20160525103520_ensure_memberships_are_unique.rb
+++ b/db/migrate/20160525103520_ensure_memberships_are_unique.rb
@@ -1,5 +1,5 @@
 class EnsureMembershipsAreUnique < ActiveRecord::Migration
   def change
-      add_index :memberships, [:user_group_id, :user_id], :unique => true
+      add_index :memberships, [:user_group_id, :user_id], unique: true
   end
 end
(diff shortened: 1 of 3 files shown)


db/migrate/20160525103520_ensure_memberships_are_unique.rb
```diff
@@ -0,0 +1,5 @@
+class EnsureMembershipsAreUnique < ActiveRecord::Migration
+  def change
+      add_index :memberships, [:user_group_id, :user_id], :unique => true
```
camallen: use non-hash rocket style if you can, `unique: true`

db/structure.sql
```diff
@@ -2,12 +2,16 @@
 -- PostgreSQL database dump
 --
 
+-- Dumped from database version 9.5.0
```
camallen: seems you've run the migration on a different PG version to the server - we'll need to do this properly (most likely via a the docker-compose pg server as we control that version)

camallen:
This is looking good, thanks for this! it's not a 100% fix for that issue, as it will now 500 with a db index violation but it's a start.

db/structure.sql
simoneduca: @camallen yes I didn't install Docker. I set the db up with Postgres and rvm.

simoneduca:
@camallen I've included your comments and @marten helped me with the missing bit.

camallen:
Thanks @simoneduca :+1:

PR #1813, opened 2016-05-25
[WIP] New emails to user

### LINKED ISSUE #1790: Send email notification when user's email address or password is changed
For security reasons when a user changes their email address we should send a notification letting them know (to both their old address and their new address) in case the change was unauthorised.

zwolf:
An email will be sent to a user when the following events occur:

* User's password is successfully changed
* User's email address is changed
* User is added to a project as a collaborator or expert 

This was done using model callbacks. I understand why this might not be ideal, so if whoever sees this would rather it were refactored using service objects, that's can be done. Might be nice for me to get in the habit.

Also: I thought about DRYing up the user mailer specs, but I got hung up on the fact that the list of arguments is different to each. Is there a simple way to deal with that?

Labelled WIP until copy is confirmed and refactoring question is answered. Resolves #1790.

# Review checklist

### DIFF
--- a/app/mailers/user_added_to_project_mailer.rb
+++ b/app/mailers/user_added_to_project_mailer.rb
@@ -1,0 +1,14 @@
+class UserAddedToProjectMailer < ApplicationMailer
+  layout false
+
+  SITE_NAME = "the Zooniverse"
+  DEFAULT_SUBJECT = "You've been added to a new Zooniverse project!"
+
+  def added_to_project(user, project, roles)
+    @user = user
+    @project = project
+    @roles = roles
+    @email_to = user.email
+    mail(to: @email_to, subject: DEFAULT_SUBJECT)
+  end
+end
--- a/app/mailers/user_info_changed_mailer.rb
+++ b/app/mailers/user_info_changed_mailer.rb
@@ -1,0 +1,21 @@
+class UserInfoChangedMailer < ApplicationMailer
+  layout false
+
+  SITE_NAME = "the Zooniverse"
+
+  def user_info_changed(user, info)
+    @user = user
+    @email_to = user.email
+
+    case info
+    when :email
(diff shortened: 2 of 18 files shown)


marten:
Yeah I think sending emails is pretty much the classic example where callbacks come back to bite you (unless your name is DHH and you manage to get by with `#suppress`). I'd be happy to pair on a service object approach.

camallen:
@zwolf re drying up the specs, you can used a shared helper with let context objects similar to [this](https://github.com/zooniverse/Panoptes/blob/master/spec/workers/subjects_dump_worker_spec.rb#L10)

zwolf:
In heading toward the service object future, I was looking at UserController#update and realized that I could simply add the calls to the mailer workers conditionally there in the same fashion as the (un)subscribe workers are being called. Would this be a better solution than creating operations for each type of mailer?

camallen:
@zwolf yeah - this is the basic concept of the pattern, remove the code from the model and put it where it's required. Ideally this would leave both the controller and model untouched apart from calling a UserUpdateServiceObject.run. I think we could extract that mailer work to a service object as well but if you want to go ahead and start refactoring out of the model to the controller then your half way there..

camallen:
@marten @zwolf I'm thinking we should get this merged and iterate later to extract these to service objects. Thoughts?

marten:
I'm opposed to "we'll fix the rest of the code quality issues later".

zwolf:
This got set aside, but not forgotten. Just adding the calls to the mailer workers in the already-overridden UsersController#update was straightforward enough. But the emails also have to go out when for newly minted project collaborators and experts, which is a little less so. I'm happy to have another look at this (it's been a minute) and see about properly extracting all these mailings into lil' objects. Providing, of course, that there's no significant hurry and you're cool with providing feedback.

zwolf:
PR moved to #1855

PR #1814, opened 2016-05-27
Add foreign keys for all non-polymorphic relations

marten:
Most important things to review are:

* if you agree with my `on_delete:` settings.
* **am i deleting that orphaned data correctly**

No app code changed, just a few specs that tried to set up factories with foreign key constraint failures.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -13,7 +13,7 @@ class User < ActiveRecord::Base
     :recoverable, :rememberable, :trackable, :validatable,
     :omniauthable, omniauth_providers: [:facebook, :gplus]
 
-  has_many :classifications
+  has_many :classifications, dependent: :restrict_with_exception
   has_many :authorizations, dependent: :destroy
   has_many :collection_preferences, class_name: "UserCollectionPreference", dependent: :destroy
   has_many :project_preferences, class_name: "UserProjectPreference", dependent: :destroy
@@ -32,7 +32,7 @@ class User < ActiveRecord::Base
   has_many :project_roles, through: :identity_group
   has_many :collection_roles, through: :identity_group
   has_many :user_seen_subjects, dependent: :destroy
-  has_many :uploaded_subjects, class_name: "Subject", foreign_key: "upload_user_id"
+  has_many :uploaded_subjects, class_name: "Subject", foreign_key: "upload_user_id", dependent: :restrict_with_exception
 
   has_many :subject_queues, dependent: :destroy
 
--- a/app/models/user_group.rb
+++ b/app/models/user_group.rb
@@ -4,12 +4,12 @@ class UserGroup < ActiveRecord::Base
   include Linkable
   include PgSearch
 
-  has_many :memberships
+  has_many :memberships, dependent: :destroy
   has_many :active_memberships, -> { active.where(identity: false) },
            class_name: "Membership"
   has_many :users, through: :memberships
-  has_many :classifications
-  has_many :access_control_lists
+  has_many :classifications, dependent: :restrict_with_exception
+  has_many :access_control_lists, dependent: :destroy
 
   has_many :owned_resources, -> { where(roles: ["owner"]) },
            class_name: "AccessControlList"
--- a/db/migrate/20160527140046_foreign_keys.rb
+++ b/db/migrate/20160527140046_foreign_keys.rb
@@ -63,7 +63,7 @@ class ForeignKeys < ActiveRecord::Migration
 
     Subject.joins("LEFT OUTER JOIN projects ON projects.id = subjects.project_id").where("projects.id IS NULL").update_all(project_id: nil)
     add_foreign_key :subjects, :projects, on_update: :cascade, on_delete: :restrict
-    add_foreign_key :subjects, :users, column: :upload_user_id, on_update: :cascade, on_delete: :nullify
+    add_foreign_key :subjects, :users, column: :upload_user_id, on_update: :cascade, on_delete: :restrict
 
     # add_foreign_key :tagged_resources, :tags
 
@@ -80,7 +80,7 @@ class ForeignKeys < ActiveRecord::Migration
     add_foreign_key :user_seen_subjects, :users, on_update: :cascade, on_delete: :cascade
     add_foreign_key :user_seen_subjects, :workflows, on_update: :cascade, on_delete: :cascade
 
-    add_foreign_key :users, :projects, on_update: :cascade, on_delete: :nullify
+    add_foreign_key :users, :projects, on_update: :cascade, on_delete: :restrict
 
     WorkflowContent.joins("LEFT OUTER JOIN workflows ON workflows.id = workflow_contents.workflow_id").where("workflows.id IS NULL").delete_all
     add_foreign_key :workflow_contents, :workflows, on_update: :cascade, on_delete: :cascade
--- a/db/migrate/20160527140046_foreign_keys.rb
+++ b/db/migrate/20160527140046_foreign_keys.rb
@@ -2,7 +2,7 @@ class ForeignKeys < ActiveRecord::Migration
   class ClassificationSubject < ActiveRecord::Base
   end
 
-  def up
+  def change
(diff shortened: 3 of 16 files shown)


db/migrate/20160527140046_foreign_keys.rb
```diff
@@ -0,0 +1,116 @@
...
+    # that they refer to on the following lines. Apart from that, everything is in alphabetical order, same as the structure.sql
+
+    # Commented out add_foreign_key calls are ones that already exist. Left them in here so that it's
+    # easier to check if I caught all of them.
+    #
+    # add_foreign_key :classification_subjects, :classifications
+    # add_foreign_key :classification_subjects, :subjects
+
+    add_foreign_key :collections_subjects, :subjects, on_update: :cascade, on_delete: :restrict
+    add_foreign_key :collections_subjects, :collections, on_update: :cascade, on_delete: :restrict
```
camallen: will this allow a collection to be deleted if it has linked subjects?

db/migrate/20160527140046_foreign_keys.rb
```diff
@@ -0,0 +1,116 @@
...
+    add_foreign_key :project_contents, :projects, on_update: :cascade, on_delete: :cascade
+
+    ProjectPage.joins("LEFT OUTER JOIN projects ON projects.id = project_pages.project_id").where("projects.id IS NULL").delete_all
+    add_foreign_key :project_pages, :projects, on_update: :cascade, on_delete: :cascade
+
+    # add_foreign_key :recents, :classifications
+    # add_foreign_key :recents, :subjects
+
+    SetMemberSubject.joins("LEFT OUTER JOIN subject_sets ON subject_sets.id = set_member_subjects.subject_set_id").where("subject_sets.id IS NULL").delete_all
+    SetMemberSubject.joins("LEFT OUTER JOIN subjects ON subjects.id = set_member_subjects.subject_id").where("subjects.id IS NULL").delete_all
```
camallen: Out of interest were there many of these kicking around? These are deleted subjects that have orphaned SMS's right?

db/migrate/20160527140046_foreign_keys.rb
marten: No. Other options are `cascade` (which would unlink all the subjects) or `nullify` (which we definitely don't want in this case).

db/migrate/20160527140046_foreign_keys.rb
marten: There's 18 SMS's that refer to a SubjectSet that doesn't exist. There's 24 SMS's that refer to a Subject that doesn't exist.

db/migrate/20160527140046_foreign_keys.rb
marten: My thinking was that if some other user has put effort into adding a subject to a collection, we can no longer remove that subject. But I'm only 2/10 on this, don't really care.

camallen: Yeah - i don't want to remove the subject if it's been collected, etc. I may be thinking about this wrong but how do i remove my collection I don't want anymore that has 10 subjects in it?

db/migrate/20160527140046_foreign_keys.rb
camallen: Not bad at all, thanks.

db/migrate/20160527140046_foreign_keys.rb
marten: Ah, I see now. Yeah, we should `cascade` the deletion of a collection rather than restrict.

db/migrate/20160527140046_foreign_keys.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    # add_foreign_key :subject_sets_workflows, :subject_sets
+    # add_foreign_key :subject_sets_workflows, :workflows
+
+    # add_foreign_key :subject_workflow_counts, :workflows
+    # add_foreign_key :subject_workflow_counts, :subjects
+
+    Subject.joins("LEFT OUTER JOIN projects ON projects.id = subjects.project_id").where("projects.id IS NULL").update_all(project_id: nil)
+    add_foreign_key :subjects, :projects, on_update: :cascade, on_delete: :restrict
+    add_foreign_key :subjects, :users, column: :upload_user_id, on_update: :cascade, on_delete: :nullify
```
camallen: should we look at renaming that `upload_user_id` column to make it a standard FK?

camallen: Also the idea here was to track who uploaded data for provenance...if we nullify on delete we lose this and have to goto archives to figure it out...thoughts?

db/migrate/20160527140046_foreign_keys.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    UserProjectPreference.joins("LEFT OUTER JOIN projects ON projects.id = user_project_preferences.project_id").where("projects.id IS NULL").delete_all
+    add_foreign_key :user_project_preferences, :users, on_update: :cascade, on_delete: :cascade
+    add_foreign_key :user_project_preferences, :projects, on_update: :cascade, on_delete: :cascade
+
+    UserSeenSubject.joins("LEFT OUTER JOIN workflows ON workflows.id = user_seen_subjects.workflow_id").where("workflows.id IS NULL").delete_all
+    add_foreign_key :user_seen_subjects, :users, on_update: :cascade, on_delete: :cascade
+    add_foreign_key :user_seen_subjects, :workflows, on_update: :cascade, on_delete: :cascade
+
+    add_foreign_key :users, :projects, on_update: :cascade, on_delete: :nullify
```
camallen: This is to track which project a user signed up to and is an optional attr via the registration process. We won't want to nullify nor delete this.

camallen:
Looks good, some comments on tracking links. For the FK relations with cascading delets I think we'll have to remove rails association dependent callbacks too.

db/migrate/20160527140046_foreign_keys.rb
marten: User is `activatable` anyways, so we shouldn't ever delete them. I'll change this to a `restrict`.

marten:
I usually prefer leaving those dependent callbacks in just to make it clear from the model definition what's going on, and just have the DB be a fallback for data integrity.

camallen:
right - do they compete at all?

marten:
No. But I'll go over them and make sure they're in sync.

marten:
@camallen Still LTGY?

app/models/user.rb
```diff
@@ -13,7 +13,7 @@ class User < ActiveRecord::Base
     :recoverable, :rememberable, :trackable, :validatable,
     :omniauthable, omniauth_providers: [:facebook, :gplus]
 
-  has_many :classifications
+  has_many :classifications, dependent: :restrict_with_exception
```
camallen: Just confirming, this restriction will effectively never allow a contributing user to be deleted...correct?

app/models/user_group.rb
```diff
@@ -4,12 +4,12 @@ class UserGroup < ActiveRecord::Base
...
   include PgSearch
 
-  has_many :memberships
+  has_many :memberships, dependent: :destroy
   has_many :active_memberships, -> { active.where(identity: false) },
            class_name: "Membership"
   has_many :users, through: :memberships
-  has_many :classifications
-  has_many :access_control_lists
+  has_many :classifications, dependent: :restrict_with_exception
```
camallen: will this mean user_groups with linked classification (classrooms, etc) won't be able to be deleted?

camallen:
This looks good pending the question about cleaning up groups -> classification with restrict_with_exception option

app/models/user_group.rb
marten: Yes. However, we're not using this column (we're stuffing it in metadata right now, since it's a list of user_group_ids rather than a single one). I've been thinking about converting that into a normal join table, shall I create an issue for that?

app/models/user.rb
marten: Yes. But users are Activatable anyways, so shouldn't ever trigger.

PR #1818, opened 2016-06-02
Add feature flags and management tool

camallen:
Provide the ability to turn on features in the API. This does not do per actor (user/groups) feature flipping at the moment but we can add if needed. 

The UI for this is located at `host/flipper`, add features and then code for them in the API..
```
if Panoptes.flipper[:science_things].enabled?
  #do all the science things
end
```

# Review checklist

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -35,5 +35,9 @@ module Panoptes
         end
       end
     end
+
+    config.middleware.use Flipper::Middleware::Memoizer, lambda {
+      Panoptes.flipper
+    }
   end
 end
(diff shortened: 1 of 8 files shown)


marten:
LGTM

config/application.rb
```diff
@@ -35,5 +35,9 @@ class Application < Rails::Application
         end
       end
     end
+
+    config.middleware.use Flipper::Middleware::Memoizer, lambda {
```
camallen: Yeah - this adds a cache per request so we don't keep hitting the db, https://github.com/jnunemaker/flipper/blob/master/docs/Optimization.md. So feel free to litter the code with feature.enabled?

PR #1820, opened 2016-06-06
Add clauses for cleanup of staging data

marten:
I've also made it more conservative with regards to nullable columns, that really aren't nullable due to Rails validations. Better safe than sorry, and FK constraints ignore null values. Gonna make those columns not-null soon, but we'll handle cleanup for that in another migration.


camallen:
LGTM - you happy this doesn't affect prod data?

marten:
Yes, the foreign keys would've hit these additional statements in on my earlier test with production data.

PR #1823, opened 2016-06-07
Test out the speed of the subject selector using scientist

camallen:
Use scientist and feature flipping to test out subject queue selection in band and skip the queue state update problem.

This hopefully will fix a lot of the already seens issues users are having (from delayed queue updates) and this will also allow us to use queue's for targeted subject task allocation.

TODO:
- [ ] Think about dropping all non-logged in queues and enforcing a FK link on users for them.

# Review checklist

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -23,7 +23,15 @@ module Subjects
     end
 
     def selected_subjects(queue)
-      sms_ids = sms_ids_from_queue(queue)
+      sms_ids = CodeExperiment.run "skip_queue_selection" do |e|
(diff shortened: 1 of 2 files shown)


camallen:
For some reason this hasn't passed the CI for the latest commit. Let's get the admin enabled PR in before merging this so i can rebase and force the CI to run.

marten:
Rebase at your convenience.

camallen:
rebased

marten:
LGTM

camallen:
I'm going to move this to 50% of the time instead of 100% just in case the load is too much.

PR #1825, opened 2016-06-08
Include display name when returning user as linked resource

### LINKED ISSUE #1824: Return user display name not just user ID when returning collections, favorites, recents etc - to increase performance
In many situations (such as https://github.com/zooniverse/Panoptes-Front-End/pull/2674 / https://github.com/zooniverse/Panoptes-Front-End/pull/2675) the UI has to carry out multiple GETs to the API to get each user's display name having been given only the ID and wishing to display that user's name on screen in a human readable way.

It would be a massive performance saving (and simplify UI code) if the linked owner resource on the appropriate controllers could include the user display name too.

I would do this myself, but I am not quite sure how the polymorphic owner link gets populated. I think we need to change something here (and maybe equivalent places too) but i'm not sure.: https://github.com/zooniverse/Panoptes/blob/95671795f2790da789c16a35b4bb8fa1f677f637/app/controllers/api/v1/collections_controller.rb#L14

alexbfree:
Fixes #1824.

Include display name when returning user as linked resource

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1117,6 +1117,7 @@ It has the following attributes:
                     "links": {
                         "owner": {
                             "id": "2",
+                            "display_name": "Owner 2",
                             "href": "/users/2",
                             "type": "user"
                         }
@@ -1316,6 +1317,7 @@ Projects are returned as an array under the _projects_ key.
                     "links": {
                         "owner": {
                             "id": "2",
+                            "display_name": "Owner 2",
                             "href": "/users/2",
                             "type": "users"
                         }
@@ -1347,6 +1349,7 @@ Projects are returned as an array under the _projects_ key.
                     "links": {
                         "owner": {
                             "id": "2",
+                            "display_name": "Owner 2",
                             "href": "/users/2",
                             "type": "users"
                         }
@@ -1398,6 +1401,7 @@ returned.
                         "workflows": ["1", "2"]
                         "owner": {
                             "id": 10,
+                            "display_name": "Owner 2",
                             "type": "user_groups",
                             "href": "user_groups/10"
                         }
@@ -2903,6 +2907,7 @@ It has the following attributes:
                     "links": {
                         "owner": {
                             "id": "10",
+                            "display_name": "Owner 10",
                             "href": "/users/10",
                             "type": "users"
                         }
@@ -3052,6 +3057,7 @@ Collections are returned as an array under *collections*.
                     "links": {
                         "owner": {
                             "id": "10",
+                            "display_name": "Owner 10",
                             "href": "/users/10",
                             "type": "users"
                         }
@@ -3065,6 +3071,7 @@ Collections are returned as an array under *collections*.
                     "links": {
                         "owner": {
                             "id": "11",
+                            "display_name": "Owner 11",
                             "type": "user_groups",
                             "href": "/user_groups/11"
                         }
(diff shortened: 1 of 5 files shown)


alexbfree:
Not sure what to do about testing. There don't seem to be any tests for the "concerns" serializers and wouldn't really know where to start in adding one from scratch.

alexbfree:
PS Where is the `apiary.apib` file mentioned in the review checklist. I can't find it:
https://github.com/zooniverse/Panoptes/search?utf8=%E2%9C%93&q=apiary.apib&type=Code

alexbfree:
Think i found the test file to update thanks to the travis results... let's try this...

alexbfree:
Hmm well I managed to fix one of the failing tests, but I am not sure how to fix the 
ProjectRolesController and CollectionRolesController.
I wondered if I needed to modify the test data, but looked in https://github.com/zooniverse/Panoptes/blob/master/spec/factories/users.rb and it does seem to have a display name.

Looked in 
https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/project_roles_controller.rb
https://github.com/zooniverse/Panoptes/blob/master/spec/controllers/api/v1/project_roles_controller_spec.rb
https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/collection_roles_controller.rb
https://github.com/zooniverse/Panoptes/blob/master/spec/controllers/api/v1/collection_roles_controller_spec.rb
https://github.com/zooniverse/Panoptes/blob/master/app/serializers/collection_role_serializer.rb
https://github.com/zooniverse/Panoptes/blob/master/app/serializers/project_role_serializer.rb

Can't see anywhere where `id`, `type`, `href` is enumerated without `display_name`!

Hope someone can advise on how I fix these tests, and on what, if anything else, I need to change!

alexbfree:
@camallen or @parrish or other backend person, could you assist?

camallen:
@alexbfree looks useful for the owned cards...never mind the json api linked resource ideology. Happy to help debug this today.

alexbfree:
@camallen Thanks. See you when you get in!

alexbfree:
@camallen the tests that are owner related are all passing locally now, and on Travis, but there is a strange error from the fifth spec job on Travis only.. 
This does not happen locally when I :focus on that spec, and doesn't even happen consistently (last time it failed with a "flipper" error) so I think it may be a Travis-environment specific issue? I'll kick it off again just in case it magically starts working.... but i think this is actually good to merge despite Travis' output...

alexbfree:
and boom - this time it worked!

good to merge @camallen ?

camallen:
great - i've added another intermittent spec...thanks for raising that. I'll have a look at bisecting and fixing.

camallen:
Looks good, can you please squash these commits down, specifically the fix failing spec ones. 
This may be useful - https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

eatyourgreens:
Re. squashing PR's, this might be handy http://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests

alexbfree:
@camallen Squashed and passing :-)

camallen:
Looks great! Thanks for this :+1:

alexbfree:
No worries, thanks for the guidance with the tests & for the review/merge.

alexbfree:
PS when should I expect this to be live on the production API? (not that it's urgent). Not sure what the deployment process/schedule is?

camallen:
remind one of us to do it tomorrow

camallen:
it's live on staging right now

PR #1830, opened 2016-06-13
test workflows eager loading with scientist

camallen:
get some metrics on the performance impact of this code, going to start with disabled and then turn on for this a percent of the time, and ramp it up to see if it causes the current behaviour.

# Review checklist


marten:
LGTM but wait for the tests to confirm that `super` works there (and isn't broken by instance_eval trickery).

PR #1832, opened 2016-06-17
do not double render in the experiement and add a user context

camallen:
Just call the render method once - i'm not sure exactly how this will be testable since the experiment being recorded will just be adding eager loading to the AR relation object. Though we can use this to ramp up the rate and check response times via new relic.

# Review checklist


zwolf:
I'm still putting all the pieces of this new code experiment stuff together in my brain but the changes in this PR look good to me.

PR #1833, opened 2016-06-20
[WIP] Orphan subject removal worker 

camallen:
Create a background worker to cleanup orphaned subjects.. I've defined an orphan here as one without any classifications and is not linked to a subject set. I couldn't think of a case where a subject could be collected / talked about without a classification but if we can then i'll add something in for these. 

Note: I've cascaded the destroy from subject -> subject media resources too. Happy to remove this if need be.

@parrish thoughts on how a subject can be talked about without any classifications (URL hacking) and not be in a subject set?

TODO:
- [ ] Add talk api client with auth to determine if there are any discussions on an orphaned subject from URL hacking without classifications (https://github.com/zooniverse/panoptes-client.rb/pull/4)

# Review checklist

### DIFF
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -13,7 +13,7 @@ class Subject < ActiveRecord::Base
   has_many :set_member_subjects, dependent: :destroy
   has_many :subject_workflow_counts, dependent: :destroy
   has_many :locations, -> { where(type: 'subject_location') },
-    class_name: "Medium", as: :linked
+    class_name: "Medium", as: :linked, dependent: :destroy
   has_many :recents, dependent: :destroy
   has_many :aggregations, dependent: :destroy
   has_many :tutorial_workflows, class_name: 'Workflow', foreign_key: 'tutorial_subject_id', dependent: :restrict_with_exception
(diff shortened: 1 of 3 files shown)


parrish:
It's possible to end up with references to a deleted subject:

- Subjects can be mentioned by id: `^S123`
- Or by url hacking: `talk/subjects/123`

Last time I checked, postgres_fdw didn't allow for remote triggers, so other than some sort of API call, I'm not sure of the best way to handle it.  It's definitely an edge case though.

camallen:
Right - i'm going to look at checking if there are discussions for a subject focus via the api. I'd to err on the side of caution on this one.

marten:
* Should this also remove them from S3? If yes, how do we deal with subjects that have external_url to existing subjects?
* Should this reset subject_count on users?

marten:
And if we don't remove them from S3, would we be better of leaving these in the DB so that we know why these files are on S3?

camallen:
The medium destroy will cleanup any subject locations media resource (and in turn should remove from s3). 

It should also reset the user counts too once the subject is cleaned up, [here](https://github.com/zooniverse/Panoptes/blob/80160e4f09dc0cd83bf162d30ab4c3bbcb393ad8/app/models/user.rb#L305)

camallen:
closing in favour of #1852

PR #1835, opened 2016-06-21
avoid memoized flipper state in tests

camallen:
Fixes recent intermittent spec failure. Avoid cached flipper state polluting tests

# Review checklist


camallen:
this is fixed now and should fix any intermittent spec failures caused by flipper state

zwolf:
Seems to be passing now, so LGTM. There's no way I can see of re-running the specs in Travis if they all pass (to test intermittence), but if they're good here, locally and they all get run again when staging gets built, that's enough.

camallen:
Cool - i manually recreated the failing travis specs via bisect and the rspec seed to debug this. If you haven't you should check it out, https://relishapp.com/rspec/rspec-core/docs/command-line/bisect

Issue #1836, opened 2016-06-21
Survey tasks causing errors in CSV export


zwolf:
Oh yeah, survey tasks. Forgot about those when the new CSV export went live. They cause the "{error: "task cannot be exported"}" since they don't match a case.

PR #1837, opened 2016-06-21
Adds survey task to CSV export

zwolf:
Oh yeah, survey tasks. This PR adds 'survey' as a case for task type and straight passes the annotation back with no translation of labels or anything else, since that's how they were coming through previously. Also added a spec for this case which seems a little heavy on account of needing everything else built, but works for now.

* Can survey tasks be included in combo tasks? This is easy to accommodate but I'm not sure it makes practical sense.
* Does there need to be any translation? The keys aren't generated and are roughly human readable, so this doesn't seem necessary.
* Drawing, single, multiple, text, combo, dropdown, survey: are there any other task types I should know about?

# Review checklist


zwolf:
Resolves #1836.

PR #1840, opened 2016-06-23
All non-standard tasks should pass unaltered annotations

zwolf:
Tasks are more varied than I thought. If the task isn't of a recognized type, instead of yielding an error, it will just pass the unaltered annotation through. Surveys now fall into this category, as well as whatever Old Weather uses. This was the previous behavior and I figured it was better to be specific, but turns out there's more out there than I thought. 

The exception is if an unsupported task (i.e. *not* drawing/single/multiple/text/dropdown) is inside of a combo task, it will still yield a "task cannot be exported" for the annotation.

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -22,10 +22,8 @@ module Formatter
           combo
         when "dropdown"
           dropdown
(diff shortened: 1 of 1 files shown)


camallen:
LGTM pending specs. Get it out if it all passes.

PR #1844, opened 2016-06-27
Build from official ruby image

astopy:
There's basically no point in us maintaining our own docker builds of Ruby since the official images are almost identical to ours. They're built on Debian rather than Ubuntu, but that shouldn't make any difference.

This also makes the resulting image about 100 MB larger, but that shouldn't be a problem and I think it's worth it to not have to maintain our own builds.

I've confirmed this starts as expected with `docker-compose`, but we'll want to thoroughly test this on staging too.

# Review checklist

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,10 +1,10 @@
-FROM zooniverse/ruby:2.3.0
+FROM ruby:2.3
 
 WORKDIR /rails_app
 
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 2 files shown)


marten:
LGTM but @camallen had been doing optimizations on the image size, he might have thoughts on that size increase?

camallen:
I'm all good on this but it'd be nice to run staging on it for a while via a different dockerfile / trigger build of master in docker hub so we can have test it properly...

astopy:
OK, I've moved this into a separate Dockerfile and configured DockerHub to build it as zooniverse/panoptes:official-ruby

PR #1847, opened 2016-06-28
cascade delete of subject set to subject queues

camallen:
Delete subject queues with a subject_set_id when the subject_set is deleted.

@saschaishikawa once this is deployed you'll be able to remove subject sets for grouped workflows. 

# Review checklist


marten:
Lgtm

PR #1849, opened 2016-06-29
Convert legacy wg survey

marten:


camallen:
LGTM - you tested this out?

marten:
Didn't run it on the entire dataset locally yet, but I did run it for a while and against all the workflow versions that we have classifications for, and everything worked just fine there.

camallen:
Sounds good

PR #1850, opened 2016-06-30
Doorkeeper access tokens can be JWT tokens

marten:
This allows auxiliary services (Nero, EducationAPI, ZooEventStats, ...) to independently verify the validity of OAuth tokens sent to them by clients. Right now they have to make an API call to Panoptes to verify any tokens given to them (`GET /api/me`).

Currently issued tokens will expire at some point (a few hours) and get replaced with JWT-based ones.

### DIFF
--- a/spec/support/controller_helpers.rb
+++ b/spec/support/controller_helpers.rb
@@ -42,7 +42,11 @@ module APIRequestHelpers
   end
 
   def stub_token(scopes: [], user_id: nil)
-    allow(controller).to receive(:doorkeeper_token).and_return(token(scopes, user_id))
+    if user_id
+      allow(controller).to receive(:doorkeeper_token).and_return(token(scopes, user_id))
+    else
+      allow(controller).to receive(:doorkeeper_token).and_return(nil)
+    end
   end
 
   def token(scopes, user_id)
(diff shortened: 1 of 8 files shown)


marten:
@zooniverse/developers Anyone foresee any issues with the OAuth token string (the bit in the `Authorization: Bearer TOKENHERE` header) becoming a bit longer (around 500 chars instead of 64 chars)? I'm thinking it should be fine as long as it fits inside a cookie?

parrish:
I don't see any issues with a longer auth string.  👍 for introducing JWT too.

camallen:
Doesn't seem like there will be problems...seems there are size limits on headers for web servers but it's pretty big

marten:
Can't think of any server-side applications that might be storing access tokens either (which might have too small a DB-column).

spec/support/controller_helpers.rb
```diff
@@ -42,7 +42,11 @@ def set_preconditions
   end
 
   def stub_token(scopes: [], user_id: nil)
-    allow(controller).to receive(:doorkeeper_token).and_return(token(scopes, user_id))
+    if user_id
+      allow(controller).to receive(:doorkeeper_token).and_return(token(scopes, user_id))
+    else
+			allow(controller).to receive(:doorkeeper_token).and_return(nil)
```
camallen: what's up with the spacing?

camallen:
LGTM

PR #1851, opened 2016-06-30
Respect page_size limits in subject selector

camallen:
ensure the page size query param hits the selectors and some minor improvements / refactoring.

# Review checklist

### DIFF
--- a/lib/subjects/postgresql_random_selection.rb
+++ b/lib/subjects/postgresql_random_selection.rb
@@ -1,5 +1,7 @@
 module Subjects
   class PostgresqlRandomSelection
+    using Refinements::RangeClamping
+
     attr_reader :available, :limit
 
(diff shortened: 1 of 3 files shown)


marten:
I found this condition really hard too follow in terms of which number will be chosen in the end. Are you trying to clamp the given limit between a min and a max? If so, could you write it like `[min, limit, max].sort[1]` or (even better) add a clamp refinement?

camallen:
agree, i'll checkout your range clamp refinement.

marten:
I've done it like this before, but that should've been a refinement module these days.

https://github.com/roqua/roqua-support/blob/master/lib/roqua/core_ext/fixnum/clamp.rb

camallen:
updated - thanks for the tip on range claming.

PR #1852, opened 2016-06-30
Deactivate subjects

### LINKED ISSUE #1193: Should we allow subjects to be deleted via the API

camallen:
closes #1193

Don't delete subjects via the API. Mark them as deleted and process the actual deletion in the background so we can check if they are in use / have classifications, etc. 

TODO
- [x] Tie into #1833 for removal OR actually kick off a proper deletion worker that the orphans worker can use as well.
- [x] Look at cellect remove notification (if the workflow is using cellect).
- [x] Ensure that the talk discussions don't need admin access via the client.
- [x] Release new panoptes client gem and use this gem version.
- [x] Update apiary docs showing the workflow_id can be used to filter the subjects end point.
- [x] Wire up to subject_set removal (and sms removal) or look at a timed job to cleanup all orphans.

Talk discussions are public by default but the actual subject data is private so we won't be leaking info and don't need to issue admin calls for the talk client. 

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ gem 'scientist', '~> 1.0.0'
 gem 'flipper'
 gem 'flipper-active_record'
 gem 'flipper-ui'
+gem 'panoptes-client', github: "zooniverse/panoptes-client.rb"
 
 group :production do
   gem 'newrelic_rpm', '~> 3.0', require: false
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ gem 'scientist', '~> 1.0.0'
 gem 'flipper'
 gem 'flipper-active_record'
 gem 'flipper-ui'
-gem 'panoptes-client', github: "zooniverse/panoptes-client.rb"
+gem 'panoptes-client'
 
 group :production do
   gem 'newrelic_rpm', '~> 3.0', require: false
--- a/app/controllers/api/v1/subject_sets_controller.rb
+++ b/app/controllers/api/v1/subject_sets_controller.rb
@@ -21,13 +21,20 @@ class Api::V1::SubjectSetsController < Api::ApiController
 
   def destroy
     resource_class.transaction(requires_new: true) do
-      controlled_resources.each do |resource|
-        remove_linked_set_member_subjects(resource, resource.set_member_subjects)
-        resource.subject_sets_workflows.delete_all
-        #avoid optimisitc locking errors
-        resource.reload
-      end
+      subject_ids = controlled_resource.set_member_subjects.map(&:subject_id)
+      remove_linked_set_member_subjects(
+        controlled_resource,
+        controlled_resource.set_member_subjects
+      )
+      controlled_resource.subject_sets_workflows.delete_all
+      #avoid optimisitc locking errors
+      controlled_resource.reload
+
       super
+
+      subject_ids.each do |subject_id|
+        SubjectRemovalWorker.perform_async(subject_id)
+      end
     end
   end
 
@@ -79,7 +86,8 @@ class Api::V1::SubjectSetsController < Api::ApiController
 
   def destroy_relation(resource, relation, value)
     if relation == :subjects
-      set_member_subjects = resource.set_member_subjects.where(subject_id: value.split(',').map(&:to_i))
+      linked_sms_ids = value.split(',').map(&:to_i)
+      set_member_subjects = resource.set_member_subjects.where(subject_id: linked_sms_ids)
       remove_linked_set_member_subjects(resource, set_member_subjects)
     else
       super
@@ -88,9 +96,9 @@ class Api::V1::SubjectSetsController < Api::ApiController
 
   private
 
-  def remove_linked_set_member_subjects(resource, set_member_subjects)
-    SubjectSetSubjectCounterWorker.perform_in(3.minutes, resource.id)
-    CountResetWorker.perform_async(resource.id)
+  def remove_linked_set_member_subjects(subject_set, set_member_subjects)
     set_member_subjects.delete_all
+    SubjectSetSubjectCounterWorker.perform_in(3.minutes, subject_set.id)
+    CountResetWorker.perform_async(subject_set.id)
   end
 end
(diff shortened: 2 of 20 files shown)


Gemfile
```diff
@@ -48,6 +48,7 @@ gem 'scientist', '~> 1.0.0'
 gem 'flipper'
 gem 'flipper-active_record'
 gem 'flipper-ui'
+gem 'panoptes-client', github: "zooniverse/panoptes-client.rb"
```
camallen: this needs a gem release without breaking older apps.

lib/subjects/remover.rb
```diff
@@ -0,0 +1,61 @@
...
+       .where("subjects.id IS NOT NULL AND collections_subjects.subject_id IS NULL")
+       .first
+    end
+
+    def raise_non_orphan_error
+      raise NonOrphan, "Subject with id: #{subject_id} has linked data and cannot be removed."
+    end
+
+    def no_talk_discussions?
+      talk_client.discussions(focus_id: subject_id, focus_type: 'Subject').empty?
```
camallen: does this need admin permissions for private talk discussions?

Gemfile
marten: What about the gem is breaking older apps?

camallen: nothing per se - just haven't cut a release on this and tested the other apps. It should be fine though.

marten: I've usually just been releasing the gem as patchlevel version bumps, as long as it's just adding more supported API calls.

lib/subjects/remover.rb
camallen: nope - all talk discussions are public by default but the subject data is restricted. 

https://www.zooniverse.org/projects/camallen/cams-amazing-project/talk/subjects/1049231
https://talk.zooniverse.org/comments?discussion_id=61932&page_size=20&page=1

app/models/subject.rb
```diff
@@ -11,7 +13,8 @@ class Subject < ActiveRecord::Base
   has_many :collections, through: :collections_subjects
   has_many :subject_sets, through: :set_member_subjects
   has_many :set_member_subjects, dependent: :destroy
-  has_many :subject_workflow_counts, dependent: :destroy
+  has_many :workflows, through: :set_member_subjects
```
camallen: this will provide the ability to narrow the collection route to the workflow. I should update the apiary docs on this.

lib/subjects/remover.rb
```diff
@@ -0,0 +1,62 @@
+module Subjects
```
marten: Nitpick: I personally prefer things that are overly domain-specific and/or access AR models not to be in `lib/`. I prefer to keep it leaning towards stuff we could potentially extract to gems. Do you have a stance on `lib/` contents?

app/workers/subject_removal_worker.rb
```diff
@@ -0,0 +1,12 @@
+require 'subjects/remover'
+
+class SubjectRemovalWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_medium
+
+  def perform(subject_id)
+    Subjects::Remover.new(subject_id).cleanup
+  rescue Subjects::Remover::NonOrphan
```
marten: Exceptions for flow control?

camallen: This rescue ensures says not an orphaned subject so move along and don't keep trying via sidekiq.

lib/subjects/remover.rb
camallen: Not really - I've always had trouble figuring out how best to organise  `/lib` in a rails app.  I was keeping it in line with subject selection work done earlier which was moved to lib, which in my mind was doing work with the models instead of being concerned, etc. If you like we can move them into app somewhere....

app/models/subject.rb
marten: Did you want to do that in this PR?

camallen: yeah - i use the `subject.workflows` method in the remover `workflow_ids = orphan.workflows.map(&:id)`

That could probably change to just a pluck instead too.

lib/subjects/remover.rb
```diff
@@ -0,0 +1,62 @@
...
+    def can_be_removed?
+      !!orphan && no_talk_discussions?
+    end
+
+    def orphan
+     @orphan ||=
+       Subject
+       .where(id: subject_id)
+       .joins("LEFT OUTER JOIN classification_subjects ON classification_subjects.subject_id = subjects.id")
+       .where("subjects.id IS NOT NULL AND classification_subjects.subject_id IS NULL")
```
marten: Doesn't the first `where` already add the `WHERE subjects.id = #{subject_id}` clause? What does the `WHERE subjects.id IS NOT NULL` add?

camallen: Yeah - this can get rewritten, it was vestigial from a larger scope that I iterated over.

camallen:
@marten i've updated the bits that needed it. Can you please review again when you get a chance?

camallen:
@marten this is ready for review again. Be nice to get it out this week to test cleaning up @aliburchard 's subjects.

app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -21,13 +21,20 @@ def update_links
 
   def destroy
     resource_class.transaction(requires_new: true) do
-      controlled_resources.each do |resource|
```
marten: Are we dropping support for multiple IDs per request?

camallen: i looked up the spec and it looked like the deletes were via the resource URL so we don't need to support it here (and i don't think our rails routes ever did either)
> An individual resource can be deleted by making a DELETE request to the resource’s URL:

app/models/subject.rb
marten: What I meant was: I don't see the update to the API docs here.

app/workers/subject_removal_worker.rb
marten: I know, but it feels dirty to have an exception be inside a common code path. Perhaps `#cleanup` could just return true/false depending on if it removed it?

camallen: Right - i was using it like AR.find that raises if it can't find the subject id (which we use a lot in the workers). I can rework it but from a client of the remover i thought this approach was reasonable.

app/models/subject.rb
camallen: I updated them in #1888

PR #1854, opened 2016-07-01
Workflow is finished when there are no unseens

marten:
Rather than having both a `seen` and an `unseen`, we can consolidate and say that you've finished a workflow if we can't find any unseen subjects.

I'm removing the experiments from SetMemberSubject so that we don't have nested experiments running.

I'm choosing the candidate for `unseen_for_user_by_workflow`, but the control for `seen_for_user_by_workflow`. Candidate in that case seemed much slower hence this alternate approach.


camallen:
LGTM pending your response to hound.

camallen:
@marten this ready to go?

PR #1855, opened 2016-07-01
Mailer service objects

zwolf:
New PR for switching to service objects instead of callbacks for sending emails to users when:

* email address is changed
* password is changed
* user is added to a project as an expert or collaborator

# Review checklist


app/controllers/api/v1/project_roles_controller.rb
```diff
@@ -8,4 +8,12 @@ class Api::V1::ProjectRolesController < Api::ApiController
   def resource_name
     "project_role"
   end
+
+  def update
+    user_added_to_project(user.id, id, params[:roles])
+  end
+
+  def user_added_to_project(user_id, project_id, roles)
+    operation.run!(user_id, project_id, roles)
```
camallen: should this just be inline in the update?

marten: So that'd call `ProjectRoles::UserAddedToProject`

app/controllers/api/v1/project_roles_controller.rb
```diff
@@ -8,4 +8,12 @@ class Api::V1::ProjectRolesController < Api::ApiController
   def resource_name
     "project_role"
   end
+
+  def update
+    user_added_to_project(user.id, id, params[:roles])
```
marten: Shouldn't this also still call super? Probably like:

```
def update
  super do
    user_added_to_project(...)
  end
end
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -98,4 +101,9 @@ def revoke_doorkeeper_request_token!
     token = Doorkeeper.authenticate(request)
     token.revoke
   end
+
+  def user_info_changed(user_id, changed)
+    # Mailers::UserInfoChanged.run!(user_id, changed)
+    operation.run!(user_id, changed)
```
marten: I'd just do it like you commented out in this case. The `operation` helper is more for when we're moving the entire action body into the operation (in my mind at least)

spec/models/access_control_list_spec.rb
```diff
@@ -1,5 +1,5 @@
 require 'spec_helper'
 
 RSpec.describe AccessControlList, :type => :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  # pending "add some examples to (or delete) #{__FILE__}"
```
camallen: just leave this as pending so we see it in spec runs

app/controllers/api/v1/project_roles_controller.rb
zwolf: yeah, definitely.

spec/models/access_control_list_spec.rb
zwolf: ah, missed this when I removed the specs I had in there from before

spec/controllers/api/v1/project_roles_controller_spec.rb
```diff
@@ -71,6 +71,29 @@
...
+    context "mailers" do
+      before(:each) do
+        default_request scopes: scopes, user_id: authorized_user.id
+      end
+
+      # This is the one that should pass, but doesn't
+      it "calls the mailer if user added to project" do
+        params = update_params.merge(id: resource.id)
+        put :update, params
+        #expect(UserAddedToProjectMailerWorker).to receive(:perform_async)
```
zwolf: Thanks, Hound. Would someone mind taking a look at this spec tomorrow? I think I've got the right idea implementation-wise but I'm not sure what the best way to test is. TODO: add similar specs for user controller.

camallen:
Looking good you've got most of the components in this covered. 

I think you should unit test the operation classes in isolation and just make sure the controller calls them. Seems @marten previously still tested the expected behaviour in controllers ala [here](https://github.com/zooniverse/Panoptes/blob/3810d48a8404af01c024e0b48e15bf7c3e12eb5e/spec/controllers/api/v1/workflows_controller_spec.rb#L567) as well. Probably to ensure the controller params piping is correct.

marten:
I did that for pretty much two reasons:

1. I generally try to have one happy-path acceptance test that integrates everything.
2. I was still setting up the operation architecture.

And maybe:

3. Some other cases where I extracted logic to operations, I didn't remove all controller test cases after it was extracted.

app/operations/mailers/user_added_to_project.rb
```diff
@@ -0,0 +1,10 @@
+module Mailers
+  class UserAddedToProject < Operation
+    integer :resource_id
+    array :roles
+
+    def execute
+      UserAddedToProjectMailerWorker.perform_async(api_user.id, resource_id, roles)
```
marten: With how little the operations themselves ended up doing, it might be fine if this line was just directly inside the controller. (and same with other operations). Saves one layer from the Controller -> Operation -> Worker -> Mailer pipeline that just seems to be passing along the method arguments directly.

zwolf:
Can you tell why the ProjectRolesController spec on line 81 is failing? It's trying what you suggest: `expect(Mailers::UserAddedToProject).to receive(:run!)` But a binding.pry call in Mailers::UserAddedToProject#execute makes it clear that it is being called, so I think the problem is somewhere in the operation layer and might just be rspec syntax.

camallen:
you need the expectation before the controller request

camallen:
@zwolf what's the status of this PR?

app/controllers/api/v1/project_roles_controller.rb
```diff
@@ -8,4 +8,20 @@ class Api::V1::ProjectRolesController < Api::ApiController
...
+    end
+  end
+
+  private
+
+  def roles
+    params[:project_roles][:roles]
+  end
+
+  def check_new_roles(roles)
```
camallen: make this method return what you want..i..e push the present into the method and return the boolean, so change the name to `new_roles_present(roles)?` or something...

app/views/user_added_to_project_mailer/added_to_project.text.erb
```diff
@@ -0,0 +1,13 @@
+You have been added to the Zooniverse project <%= @project.display_name %> in the following roles: <%= @roles.join(" and ") %>
+
+Visit <a href="https://zooniverse.org/projects/<%= @project.slug %>"><%= @project.display_name %></a> to get started.
```
camallen: Look at using the [`Panoptes.frontend_configuration.frontend_url`](https://github.com/zooniverse/Panoptes/blob/master/config/initializers/redirect_urls.rb#L11) here if you can as it'll work on staging / prod equally well.

app/views/user_added_to_project_mailer/added_to_project.html.erb
```diff
@@ -0,0 +1,13 @@
+<p>You have been added to the Zooniverse project <%= @project.display_name %> in the following roles: <%= @roles.join(" and ") %></p>
+
+<p>Visit <a href="https://zooniverse.org/projects/<%= @project.slug %>"><%= @project.display_name %></a> to get started.</p>
+
+<% if @roles.include? "collaborator" %>
+  <p>To edit your project, go to <a href="https://zooniverse.org/lab/<%= @project.id %>">
```
camallen: anywhere you've got the hardcoded URL we should be looking to use the frontend URL config instead.

app/workers/user_info_changed_mailer_worker.rb
```diff
@@ -0,0 +1,18 @@
+class UserInfoChangedMailerWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_high
+
+  attr_reader :user
```
camallen: no need for the attr_reader if your using instance vars in the one method.

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -442,7 +442,7 @@
     it_behaves_like "an api response"
   end
 
-  describe "#update" do
+  describe "#update", :focus do
```
camallen: focus set here - full spec suite won't have run.

app/workers/user_added_to_project_mailer_worker.rb
```diff
@@ -0,0 +1,19 @@
+class UserAddedToProjectMailerWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_high
+
+  attr_reader :user
```
camallen: same thing with the reader not being used.

app/operations/mailers/user_added_to_project.rb
camallen: Looking how this turned out i agree with marten but it's not a blocker for me. Just another layer of abstraction that doesn't really do anything since there is no logic in this particular operation.

camallen:
Few issues but it's looking good.

zwolf:
The last spec this needs is a check to ensure the mailer worker gets called when a password is changed. I can't get the controller to allow me to change a password via spec, though, as it always 422s and complains about unpermitted params. Building a request in the spec as close to the one on the home page as I can still doesn't fly, so I must be missing something.

camallen:
@zwolf you'll need to use the passwords controller to change the user password.

spec/operations/mailers/user_added_to_project.rb
```diff
@@ -0,0 +1,12 @@
+require 'spec_helper'
+
+describe Mailers::UserAddedToProject do
```
camallen: As these aren't used, I'm trying to figure out how this operation specs aren't running / failing...

camallen: this filename should change to ` spec/operations/mailers/user_added_to_project_spec.rb `

spec/operations/mailers/user_info_changed.rb
```diff
@@ -0,0 +1,12 @@
+require 'spec_helper'
+
+describe Mailers::UserInfoChanged do
```
camallen: this filename should change to ` spec/operations/mailers/user_info_changed_spec.rb `

camallen:
Once you remove those unused operations specs i think this is good.

camallen:
LGTM

PR #1864, opened 2016-07-08
Set text task answer label to default value

zwolf:
Instead of looking for a nonexistent answer label for a text task, just set it to "text". 

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -52,7 +52,7 @@ module Formatter
         {}.tap do |new_anno|
           new_anno['task'] = @current['task']
           new_anno['value'] = @current['value']
-          new_anno['task_label'] = task_label(task_info)
+          new_anno['task_label'] = "text"
         end
       end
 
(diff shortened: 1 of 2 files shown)


marten:
Why don't text tasks have a question?

zwolf:
They do, it's the task's "instruction". Why it's called that instead, I don't know. They also have "help". They don't have an answer, however, which is why looking for it's label caused the exception.

marten:
I'm sure it's just me being wrong here, but isn't https://github.com/zwolf/Panoptes/blob/85ae87132e0b4afd294b207fb59ca79f6b762eb2/lib/formatter/csv/annotation_for_csv.rb#L124 the line that was being called, which looks up the "instruction"?

zwolf:
Task label is the value for the question/instruction. Answer label is the value for the answer, which does not exist for fill-in-the-blank type Text questions.

marten:
But... aren't you removing task_label here?

zwolf:
yuuuuuuuuup. Looks like I mixed them up from the beginning. Come to think of it, #text doesn't call for answer label translation at all, which is where this error was being thrown.

PR #1867, opened 2016-07-14
Oauth fix

marten:
In case logging in with OAuth tries to generate a conflicting user account, try to generate a unique value for the login field.

We can't deduplicate emails, and we definitely can't be sure that foo@bar.com on our site is the same person as foo@bar.com on Facebook, so that duplication is still a failure. But now we're telling the user about it and chance are it is the same person who just forgot they already had a normal account.

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -5,7 +5,8 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     sign_in @user, event: :authentication
     redirect_to sign_in_redirect
   rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotUnique
-    # TODO Redirect to a page to edit user options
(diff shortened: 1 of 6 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -5,7 +5,8 @@ def callback
     sign_in @user, event: :authentication
     redirect_to sign_in_redirect
   rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotUnique
-    # TODO Redirect to a page to edit user options
+    flash[:alert] = "Sorry, an account with your email address already exists. Please log in normally or attempt a password reset if you've forgotten your password."
```
astopy: Will the user definitely know what "normally" means?

marten: Changed to "with your Zooniverse account"

zwolf:
LGTM, merging

PR #1876, opened 2016-07-25
Active workflows completeness

### LINKED ISSUE #1873: Project completeness score should only include active workflows
We're including the inactive workflows in the project completeness score (see wildcam and the inactive education api assignments workflows). This [here](https://github.com/zooniverse/Panoptes/blob/2f454e36e711fe9e8ec641cde6fd01b08a9e0d2d/app/workers/calculate_project_completeness_worker.rb#L22) should change to filter on the active workflows only.

camallen:
closes #1873

Exclude inactive workflows from the project completeness score as we should be reflecting the actual current workload for this metric...not any previous / testing workflows.

# Review checklist

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -106,7 +106,7 @@ describe Project, type: :model do
       end
     end
 
-    it "should have only one active workflow" do
+    it "should only return the active workflow" do
       expect(project.active_workflows.size).to eq(1)
       expect(project.active_workflows).to all( be_a(Workflow) )
     end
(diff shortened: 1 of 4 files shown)


spec/models/project_spec.rb
```diff
@@ -98,6 +98,20 @@
...
 
+  describe "#active_workflows" do
+    let(:project) do
+      create(:project) do |p|
+        create(:workflow, project: p, active: true)
+        create(:workflow, project: p, active: false)
+      end
+    end
+
+    it "should have only one active workflow" do
```
zwolf: I think that projects can have more than one active workflow. Seems that it's just this single spec that seems to disagree.

camallen: yeah - the intention isn't clear, i'll reword it.

zwolf:
LGTM

PR #1877, opened 2016-07-26
Password reset errors

camallen:
Provide useful error messages when resetting passwords (e.g. it's too short, token invalid, etc) for the UI.

# Review checklist

### DIFF
--- a/app/controllers/passwords_controller.rb
+++ b/app/controllers/passwords_controller.rb
@@ -26,7 +26,12 @@ class PasswordsController < Devise::PasswordsController
     def update_from_json
       resource = resource_class.reset_password_by_token(resource_params)
       yield resource if block_given?
-      respond_to_request(resource.errors.empty?)
+      resource.valid? if resource.errors.empty?
+      response_body = {}
+      unless resource.errors.empty?
+        response_body.merge!(error_response(resource.errors.full_messages.join(", ")))
+      end
+      respond_to_request(response_body.empty?, response_body)
     end
 
     def create_from_json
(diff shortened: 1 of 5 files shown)


app/controllers/passwords_controller.rb
```diff
@@ -26,7 +26,12 @@ def edit
     def update_from_json
       resource = resource_class.reset_password_by_token(resource_params)
       yield resource if block_given?
-      respond_to_request(resource.errors.empty?)
+      resource.valid? if resource.errors.empty?
```
camallen: so this was confusing for me but devise `reset_password_by_token` will either return the correct instance or initiate a new empty object with a reset_token error, [see here](http://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Recoverable/ClassMethods). Basically i'm using the error object as a proxy for the find by token working...

config/initializers/devise.rb
```diff
@@ -9,7 +9,7 @@
   config.skip_session_storage = [:http_auth]
   config.stretches = Rails.env.test? ? 1 : 10
   config.reconfirmable = true
-  config.password_length = 4..128
+  config.password_length = 8..128
```
camallen: @parrish as discussed I don't think this will interfere with any legacy / ouroboros login but just when updating the user model via panoptes API if the password field is present (change), [see here](https://github.com/plataformatec/devise/blob/6720f93d2c54904f10b6e6d6beb51d0b5b330976/lib/devise/models/validatable.rb#L35)

parrish:
Matching change for Ouroboros is on the way

PR #1879, opened 2016-07-26
Added /classifications/offset?project_id=Y&last_id=X

zwolf:
Added a route to grab new classifications by sending it the last id you've got. Uses a simple where clause to find them. I'm having trouble with my local setup that breaks requests outside of rspec. I'll try to solve that tomorrow, but in the mean time I'm making this PR to run all the specs and see what happens.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -2,7 +2,7 @@ require 'classification_lifecycle'
 
 class Api::V1::ClassificationsController < Api::ApiController
   skip_before_filter :require_login, only: :create
-  require_authentication :show, :index, :destroy, :update, :incomplete, :project, :offset,
+  require_authentication :show, :index, :destroy, :update, :incomplete, :project, :project_offset,
     scopes: [:classification]
 
   resource_actions :default
@@ -34,16 +34,15 @@ class Api::V1::ClassificationsController < Api::ApiController
     index
   end
 
-  def offset
-    @controlled_resources =
-      Classification.where("project_id = ? AND id > ?", params[:project_id], params[:last_id])
-                    .includes(:project)
-                    .includes(:user)
-    render json_api: serializer.page(params, controlled_resources, context),
-           generate_response_obj_etag: true
+  def project_offset
+    index
   end
 
   private
+
+  def scope_context
+    params
+  end
 
   def filter_by_subject_id
     subject_ids = (params.delete(:subject_ids) || params.delete(:subject_id)).try(:split, ',')
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -2,7 +2,7 @@ require 'classification_lifecycle'
 
 class Api::V1::ClassificationsController < Api::ApiController
   skip_before_filter :require_login, only: :create
-  require_authentication :show, :index, :destroy, :update, :incomplete, :project, :project_offset,
+  require_authentication :show, :index, :destroy, :update, :incomplete, :project,
     scopes: [:classification]
 
   resource_actions :default
@@ -31,10 +31,6 @@ class Api::V1::ClassificationsController < Api::ApiController
   end
 
   def project
-    index
-  end
-
-  def project_offset
     index
   end
 
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -2,7 +2,7 @@ require 'classification_lifecycle'
 
 class Api::V1::ClassificationsController < Api::ApiController
   skip_before_filter :require_login, only: :create
-  require_authentication :show, :index, :destroy, :update, :incomplete, :project,
+  require_authentication :show, :index, :destroy, :update, :incomplete, :project, :offset,
     scopes: [:classification]
 
   resource_actions :default
@@ -35,9 +35,10 @@ class Api::V1::ClassificationsController < Api::ApiController
   end
 
   def offset
-    @controlled_resources = Classification.where("project_id = ? AND id > ?", params[:project_id], params[:last_id])
-    .includes(:project)
-    .includes(:user)
+    @controlled_resources =
+      Classification.where("project_id = ? AND id > ?", params[:project_id], params[:last_id])
+                    .includes(:project)
+                    .includes(:user)
     render json_api: serializer.page(params, controlled_resources, context),
            generate_response_obj_etag: true
   end
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -34,6 +34,14 @@ class Api::V1::ClassificationsController < Api::ApiController
     index
   end
 
+  def offset
+    @controlled_resources = Classification.where("project_id = ? AND id > ?", params[:project_id], params[:last_id])
+    .includes(:project)
+    .includes(:user)
+    render json_api: serializer.page(params, controlled_resources, context),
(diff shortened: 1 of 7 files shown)


spec/controllers/api/v1/classifications_controller_spec.rb
```diff
@@ -154,6 +154,27 @@ def create_gold_standard
     end
   end
 
+  describe "#offset" do
+    let!(:first) { create(:classification, user: user, project: project) }
+    let!(:second) { create(:classification, user: user, project: project) }
+
+    before(:each) do
+      default_request user_id: authorized_user.id, scopes: scopes
```
camallen: Can you check that a non-auth user cannot use this route too, ala [here](https://github.com/zooniverse/Panoptes/blob/50566215d7d263df52a307a2b433d2185623153f/app/controllers/api/v1/classifications_controller.rb#L5)

app/controllers/api/v1/classifications_controller.rb
```diff
@@ -34,6 +34,14 @@ def project
     index
   end
 
+  def offset
+    @controlled_resources = Classification.where("project_id = ? AND id > ?", params[:project_id], params[:last_id])
+    .includes(:project)
+    .includes(:user)
+    render json_api: serializer.page(params, controlled_resources, context),
```
camallen: I think you'll want to replicate what i did with incomplete and use the index action with the `scope_for` setup on the Classification model to fit into the default authorization rules to make sure non-project owners can't access other project classifications.

camallen: specifically the project scope for rules in this [line](https://github.com/zooniverse/Panoptes/blob/50566215d7d263df52a307a2b433d2185623153f/app/models/classification.rb#L39)

zwolf: I didn't change this to use a scope because of the fact that project and classification ids are required as params. Not sure how to provide these to a scope...could use a lambda in the scope def and pass it some params, probably. Would that be worth doing?

camallen: you should just be able to chain the where clause here. e.g. 
``` ruby
Classification
.where("project_id = ? AND id > ?", params[:project_id], params[:last_id])
.joins(:project)
.merge(Project.scope_for(:update, user)
.includes(:subjects)) 
```
if you don't include the project scope for any authenticated user will be able to use this to harvest classifications for a project_id. The Project.scope_for ensure the user has update privs on the project (collab / owner)

zwolf:
I didn't change this to use a scope because of the fact that project and classification ids are required as params. Not sure how to provide these to a scope...could use a lambda in the scope def and pass it some params, probably. Would that be worth doing?

app/controllers/api/v1/classifications_controller.rb
```diff
@@ -34,8 +34,16 @@ def project
     index
   end
 
+  def project_offset
```
camallen: the naming here is bugging me...i think the reason is that this is really just a modified project index route with some extra params.

app/models/classification.rb
```diff
@@ -31,12 +32,14 @@ def self.scope_for(action, user, opts={})
...
       created_by(user).includes(:subjects)
-    when :update, :destroy
+  when :update, :destroy
       incomplete_for_user(user)
     when :incomplete
       incomplete_for_user(user).includes(:subjects)
     when :project
       joins(:project).merge(Project.scope_for(:update, user)).includes(:subjects)
+    when :project_offset
+      after_id(opts[:last_id]).joins(:project).merge(Project.scope_for(:update, user))
```
camallen: seems project and project_offset should collapse into one (project in my mind) and just merge the extra offset scope if the param is set. Sound ok? Also does the last_id param make sense without the project_id...maybe. If not then perhaps ensure the project_id is a req'd param with last_id.

zwolf:
I am not certain where the logic is for filtering classifications by project, so made it explicit in the project scope alongside the filter for last_id. This might be DRYable, but it's at least clear this way.

camallen:
Primary and foreign key filtering is provided by the serialized if the param is set, see https://github.com/RestPack/restpack_serializer/blob/master/README.md#by-foreign-key

zwolf:
Looks that way, all the classification model/controller specs still pass with that where clause removed. Recommitted without it.

app/models/classification.rb
```diff
@@ -36,7 +37,16 @@ def self.scope_for(action, user, opts={})
     when :incomplete
       incomplete_for_user(user).includes(:subjects)
     when :project
-      joins(:project).merge(Project.scope_for(:update, user)).includes(:subjects)
+      return none if opts[:last_id] && !opts[:project_id]
+      if opts[:project_id]
+        project_scope = joins(:project)
+                        .includes(:subjects)
+                        .merge(Project.scope_for(:update, user))
+        project_scope = project_scope.after_id(opts[:last_id]) if opts[:last_id]
```
camallen: seems we can just have append the last_id to the main project scope if the param is set instead of the if branch.
```
project_scope = joins(:project).merge(Project.scope_for(:update, user)).includes(:subjects)
project_scope = project_scope.after_id(opts[:last_id]) if opts[:last_id]
project_scope
```

app/models/classification.rb
```diff
@@ -36,7 +37,16 @@ def self.scope_for(action, user, opts={})
     when :incomplete
       incomplete_for_user(user).includes(:subjects)
     when :project
-      joins(:project).merge(Project.scope_for(:update, user)).includes(:subjects)
+      return none if opts[:last_id] && !opts[:project_id]
```
camallen: do we want to raise an error about missing params here instead of returning nothing? E.g. [missing selector params](https://github.com/zooniverse/Panoptes/blob/e4873028032d06c1dab709ef69f34d1157c139a6/lib/subjects/selector.rb#L85)

camallen:
Looking good just a couple of minor comments on this one, def remove WIP.

app/models/classification.rb
zwolf: Did this because that's what unauthorized users are getting, but it deserves a better, less confusing response.

app/models/classification.rb
zwolf: yup, this was from back when I didn't realize that the use of project_id was implicit

zwolf:
Do you think `last_id` is the best name for that param? I've considered changing it to `offset_id` but I'm not sure that's better.

spec/controllers/api/v1/classifications_controller_spec.rb
```diff
@@ -152,6 +152,24 @@ def create_gold_standard
...
+        get :project, last_id: first.id, project_id: project.id
+        resources = json_response[api_resource_name]
+        expect(resources.length).to eq 1
+        expect(resources).not_to include Classification.first
+      end
+
+      it 'returns nothing if project id' do
+        get :project, last_id: first.id
+        resources = json_response[api_resource_name]
+        expect(resources.length).to eq 0
```
camallen: final piece is to test the error message is correctly serialized, see [here](https://github.com/zooniverse/Panoptes/blob/219854367f346a0376906cf4389e078f57a6a80c/app/controllers/api/api_controller.rb#L35) for the api controller error handling for good error messages.

spec/controllers/api/v1/classifications_controller_spec.rb
```diff
@@ -119,7 +119,7 @@ def create_gold_standard
     end
   end
 
-  describe "#project" do
+  describe "#project", :focus do
```
camallen: remove focus

camallen:
Nice work, LGTM after you remove the focus set and rerun the specs.

PR #1882, opened 2016-07-27
allow virtual field attributes to be specified for serializer response

### LINKED ISSUE #1881: respect virtual attribute on workflow fields request
Requesting the workflow `subjects_count` virtual attribute via the fields param should return the field. So [here](https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/workflows_controller.rb#L58) should probably check the serializer attributes instead of just the column hash.

camallen:
fixes #1881 - allows virtual attributes to be serialized via fields param.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -54,11 +54,10 @@ class Api::V1::WorkflowsController < Api::ApiController
 
   def field_context
     if params[:fields].present?
-      included_keys = params[:fields].split ','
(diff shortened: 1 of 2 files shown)


parrish:
Good catch -- hadn't considered virtual attributes.

parrish:
This [spec failure](https://travis-ci.org/zooniverse/Panoptes/builds/147815288) is unrelated to this, right?

camallen:
yeah - that is an intermittent failure i haven't been able to debug :cry:

PR #1888, opened 2016-08-03
update docs for subject filtering by association

camallen:
make a note that workflow_id is only used for when the sort=queued param is set, i.e. via subject selection and subject_set_id is required for subject selection when the workflow is grouped. Also subject_set_id can be used on normal index routes to narrow the result set scope.

# Review checklist


zwolf:
looks good to me.

marten:
LGTM

PR #1892, opened 2016-08-03
respect workflow for retired filtering in selection

### LINKED ISSUE #1891: Retried subjects filter selector filtering across all workflows?
@marten I'm seeing some unexpected behaviour via subject selection. 

I've linked some data retired in one workflow to a new workflow but it's being filtered in selection (via retired) for the brand new workflow.

Seems [here](https://github.com/zooniverse/Panoptes/blob/4f4fcd6398a1cf1a0d92483057b28827f53e6cca/lib/subjects/selector.rb#L57) is not respecting the workflow for selection. Is it intentional that the selector's retired subject filter just looks for all SWC's for that sms_id regardless of the selection workflow?

camallen:
Closes #1891 - Do not filter out subjects that have been retired for another workflow. 

Also added in an explicit spec to ensure the subjects are returned in the order the selector chooses them. used mostly for priority selection but will preserve ordering for random too.

Describe your change here.

# Review checklist

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -58,6 +58,7 @@ module Subjects
       retired_ids = SetMemberSubject
         .joins("INNER JOIN subject_workflow_counts ON subject_workflow_counts.subject_id = set_member_subjects.subject_id")
         .where("subject_workflow_counts.retired_at IS NOT NULL")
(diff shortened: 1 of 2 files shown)


marten:
Nice catch.

PR #1893, opened 2016-08-03
do not destroy location relations with invalid location updates

### LINKED ISSUE #1885: "Error: Validation failed: Locations is invalid" will reset subject locations to []
Came across this today while playing around with updates to metadata for Diagnosis London subjects in subject set 3492, which have `text/plain` files in the locations array.

Updating a subject without setting the admin flag returns `Error: Validation failed: Locations is invalid`, which I'm pretty sure is correct because we don't officially support text subjects yet?

Anyway, inspecting those subjects after the failed update, the locations array had been reset to [], which I'm pretty sure is a bug. I had a back up of the subject set data, so I was able to update the subject locations manually in admin mode, but this seems like a dangerous bug for any project that needs to update subject metadata.

camallen:
closes #1885 - avoid invalid location updates on the media resource. Make sure these validations pass before updating the relation to reflect the updated location resource.

Describe your change here.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -42,25 +42,19 @@ class Api::V1::SubjectsController < Api::ApiController
     subject
   end
 
-  def build_update_hash(update_params, id)
+  def build_update_hash(update_params, resource)
     locations = update_params.delete(:locations)
-    subject = Subject.find(id)
-    subject.locations = add_locations(locations, subject)
+    new_locations = add_locations(locations, resource)
     subject.save!
-    super(update_params, id)
+    subject.locations = new_locations if new_locations
+    super(update_params, resource)
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/subjects_controller.rb
```diff
@@ -42,12 +42,16 @@ def build_resource_for_create(create_params)
...
     locations = update_params.delete(:locations)
-    subject = Subject.find(id)
-    subject.locations = add_locations(locations, subject)
-    subject.save!
-    super(update_params, id)
+    locations_update = !locations.blank?
+    if locations_update
+      new_locations = add_locations(locations, resource)
+    end
+    resource.save!
```
camallen: actually - think i'll move this to a block for the locations update method.

camallen:
thank you hound for barking at me for a simpler solution...much better now.

PR #1897, opened 2016-08-09
Feature flag to activate / deactivate cellect

camallen:
Provide a faster way to turn this feature on / off but allow it to fall back to env vars / defaults as well. If the feature is defined it'll use the enabled? setting on it, if it's non-existinant in the db (must be manually removed) it'll fallback to ENV vars and then the false flag.

# Review checklist


camallen:
@marten do you think this is a useful mechanism to quickly modify the behaviour of the API? Particularly links to non-critical external services?

marten:
Yeah, was thinking the same thing. Still think ENV is useful for configuration like connection URLs (database, other services, API keys) but probably less so for feature enabling.

PR #1901, opened 2016-08-10
correct user projet preference activity counts

camallen:
when the project has no workflows ensure we dont' return all the workflow counts (across projects).

Also removed an old experiment while i was there...

# Review checklist


spec/models/user_seen_subject_spec.rb
```diff
@@ -104,7 +104,7 @@ def created_uss
       end
     end
 
-    describe '::activity_by_workflow' do
+    describe '::activity_by_workflow', :focus do
```
marten: remove focus?

camallen: thanks for the catch...done.

marten:
LGTM

PR #1902, opened 2016-08-10
add api routes for cellect

camallen:
Provide cellect panoptes with api endpoints to enum the workflow to preload. Long term we can look to decouple the db connector and try api end points.

# Review checklist

### DIFF
--- a/app/controllers/cellect_controller.rb
+++ b/app/controllers/cellect_controller.rb
@@ -1,0 +1,10 @@
+class CellectController < ApplicationController
+  def workflows
+    respond_to do |format|
+      format.json do
+        cellect_workflows = Workflow.all
+        render json: { workflow_ids: cellect_workflows.pluck(:id).as_json }
+      end
+    end
+  end
+end
--- a/app/controllers/cellect_controller.rb
+++ b/app/controllers/cellect_controller.rb
@@ -1,6 +1,8 @@
 class CellectController < ApplicationController
   before_filter :html_to_json_override
 
+  EXPIRY = 10.minutes.freeze
+
   def workflows
     respond_to do |format|
       format.json do
@@ -25,7 +27,11 @@ class CellectController < ApplicationController
   end
 
   def all_cellect_workflow_ids
-     workflow_ids_using_cellect | Workflow.using_cellect.pluck(:id)
+    cache_key = "#{Rails.env}#all_cellect_workflow_ids"
+    Rails.cache.fetch(cache_key, expires_in: EXPIRY) do
+      workflow_ids_using_cellect | Workflow.using_cellect.pluck(:id)
+    end
+  end
 
   def html_to_json_override
     request.format = :json if request.format == :html
--- a/app/controllers/cellect_controller.rb
+++ b/app/controllers/cellect_controller.rb
@@ -43,7 +43,7 @@ class CellectController < ApplicationController
 
   def cellect_workflow_subject_ids
     if workflow = cellect_workflow_from_param
-      workflow.set_member_subjects.pluck(:subject_id)
+      SetMemberSubject.non_retired_for_workflow(workflow).pluck(:subject_id)
     else
       []
     end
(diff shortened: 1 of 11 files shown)


camallen:
I think this is ready to get it out and see how it goes. I will add more routes (and better caching via redis) if it works well so we can decouple the database adapter for cellect

db/structure.sql
```diff
@@ -2741,13 +2749,17 @@ CREATE INDEX users_idx_trgm_login ON users USING gin ((COALESCE((login)::text, '
 
 CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE ON projects FOR EACH ROW EXECUTE PROCEDURE tsvector_update_trigger('tsv', 'pg_catalog.english', 'display_name');
 
+ALTER TABLE projects DISABLE TRIGGER tsvectorupdate;
```
parrish: Are these meant to be in the structure?

db/structure.sql
```diff
@@ -2741,13 +2749,17 @@ CREATE INDEX users_idx_trgm_login ON users USING gin ((COALESCE((login)::text, '
...
+
 
 --
 -- Name: tsvectorupdate; Type: TRIGGER; Schema: public; Owner: -
 --
 
 CREATE TRIGGER tsvectorupdate BEFORE INSERT OR UPDATE ON users FOR EACH ROW EXECUTE PROCEDURE tsvector_update_trigger('tsv', 'pg_catalog.english', 'login');
 
+ALTER TABLE users DISABLE TRIGGER tsvectorupdate;
```
parrish: They seem likely to be accidental

parrish:
Looks like a few [failing specs](https://travis-ci.org/zooniverse/Panoptes/jobs/151591903) to resolve first

camallen:
sorry for the early call - this should be good now.

app/controllers/cellect_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  def all_cellect_workflow_ids
+    cache_key = "#{Rails.env}#all_cellect_workflow_ids"
+    Rails.cache.fetch(cache_key, expires_in: EXPIRY) do
+      workflow_ids_using_cellect | Workflow.using_cellect.pluck(:id)
+    end
+  end
+
+  def cellect_workflow_subject_ids
+    if workflow = cellect_workflow_from_param
+      workflow.set_member_subjects.pluck(:subject_id)
```
marten: Hmm, when cellect loads these it filters out retired subjects in the query: https://github.com/zooniverse/cellect_panoptes/blob/bdb16f0bc6a375847718afc832402b3aff29fc37/lib/cellect/server/adapters/panoptes.rb#L63

camallen: good one - i'll remove them too.

camallen:
@marten updated - now only returning non-retired subject ids for a cellect workflow

app/controllers/cellect_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def launched_workflows
+    Workflow
+      .joins(:project)
+      .where("projects.launch_approved IS TRUE")
+      .select(:id, :use_cellect)
+  end
+
+  def workflow_ids_using_cellect
+    launched_workflows.collect do |w|
```
parrish: Any reason to not move this into the query rather than iterate on them?

parrish:
This looks good for merge.  [This comment](https://github.com/zooniverse/Panoptes/pull/1902#discussion_r74614595) is pretty minor.  Happy to get this up if everybody else is.

app/controllers/cellect_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      w.using_cellect? ? w.id : nil
+    end.compact
+  end
+
+  def all_cellect_workflow_ids
+    cache_key = "#{Rails.env}#all_cellect_workflow_ids"
+    Rails.cache.fetch(cache_key, expires_in: EXPIRY) do
+      workflow_ids_using_cellect | Workflow.using_cellect.pluck(:id)
```
marten: Why do this `|` operator rather than just `launched_workflows.using_cellect.pluck(:id)` on this line?

PR #1903, opened 2016-08-10
Added new ProjectPreferences endpoint to update UPP

zwolf:
UserProjectPrefs need to be updated by a Doorkeeper app that's got project ownership permissions. After a couple of false starts, here's how I (with an assist from @parrish) figured it can be done. Main takeaways are that there's an extra attribute on UPP that we can give project owners full access to and there's a new route to go along with it. The specs throw a 401 but it works authing as an app and hitting it directly, so there's something squirrely there.  

Let's talk about it at standup.

# Review checklist


app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -4,8 +4,36 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+        render status: :unprocessable_entity, nothing: true
+      end
+    else
+      render json: { error: 'not authorized' }, status: :not_authorized
+    end
+  end
+
   private
 
+  def required_params
```
camallen: i don't think you need this method in here, instead use the api controller `params_for` method (normally accessed via the action alias [here](https://github.com/zooniverse/Panoptes/blob/a89ff672762f5b403f85c1c007f18db134bf62f0/lib/json_api_controller/json_schema_validator.rb#L41)) instead.

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -4,8 +4,36 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+    params_to_check = [ params[:project_id], params[:user_id], params[:settings] ]
+    params_to_check.all? { |param| !param.blank? }
+  rescue JsonSchema::ValidationError,
```
camallen: these will be taken care of normally via the [api_controller](https://github.com/zooniverse/Panoptes/blob/a89ff672762f5b403f85c1c007f18db134bf62f0/app/controllers/api/api_controller.rb#L14)

config/routes.rb
```diff
@@ -56,7 +56,9 @@
 
       json_api_resources :project_roles
 
-      json_api_resources :project_preferences
+      json_api_resources :project_preferences do
+        post "/update_project_settings", to: "project_preferences#update_project_settings", on: :collection
```
camallen: double check but i think this can be cleaned up and written like the other resource collection routes
``` ruby
collection do 
  post :update_project_settings
end
```

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -4,8 +4,36 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
   require_authentication :all, scopes: [:project]
   schema_type :json_schema
 
+  def update_project_settings
+    binding.pry
+    project = Project.find params[:project_id]
+    unless required_params
+      render json: { error: 'missing or invalid params' }, status: :unprocessable_entity
+    end
+    if project.owner?(api_user.user)
```
camallen: Neat - didn't think of this. As long as it's just updating the settings and it doesn't change / leak to other attributes i'm ok with this. Only gripe is that I'd prefer that this was extracted to a `before_action` filter.

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -4,8 +4,36 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
   require_authentication :all, scopes: [:project]
   schema_type :json_schema
 
+  def update_project_settings
+    binding.pry
+    project = Project.find params[:project_id]
+    unless required_params
```
camallen: If you wire up the params_for via the json schema validator you get this for free.

camallen:
Good stuff @zwolf, this looks good and it'll work. Happy to help wire up the internal schema validation to cleanup the code a bit.

app/controllers/api/v1/project_preferences_controller.rb
marten: Agreed but i'd like to see an explicit test for this. Do we keep email subscription in this model? Don't want to let owners mass subscribe everyone

zwolf: By adding a new "settings" attribute that's intended to be fully editable by project owners, we can leave "preferences" locked to the user, along with sensitive stuff like mail prefs.

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -3,35 +3,28 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
-    end
-    if project.owner?(api_user.user)
-      upp = UserProjectPreference.find_or_initialize_by(project: project, user_id: api_user.user)
-      upp.settings.merge! params[:settings]
-      if upp.save!
-        render status: :ok, nothing: true
-      else
-        render status: :unprocessable_entity, nothing: true
-      end
+    upp = UserProjectPreference.find_or_initialize_by(project: project, user_id: api_user.user)
```
camallen: hmm won't this only pull the requesting user's UPP instead of the desired upp of a user to update? Seem like you'll have to provide a `user_id` param to this route as well to load the correct upp.

app/schemas/project_preference_update_project_settings_schema.rb
```diff
@@ -0,0 +1,21 @@
+class ProjectPreferenceUpdateSchema < JsonSchema
```
camallen: this will need the DSL config stuff for the json_schema validator we talked about.

zwolf: Which stuff do you mean? I'm missing that stuff.

app/controllers/api/v1/project_preferences_controller.rb
zwolf: 🎉  derp 🎉  

(param was there)

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -3,9 +3,30 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
 
   require_authentication :all, scopes: [:project]
   schema_type :json_schema
+  before_action :require_project_ownership, only: [:update_project_settings]
+
+  def update_project_settings
+    upp = UserProjectPreference.find_or_initialize_by(project: project, user_id: api_user.user)
+    upp.settings.merge! params[:settings]
```
camallen: you have to user `params_for` to make the json schema validator run and validate the schema

app/schemas/project_preference_update_project_settings_schema.rb
camallen: this stuff, https://github.com/camallen/Panoptes/commit/94b915db3bf0340a8c2e097a038ddc5393a462b7

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
 
   private
 
+  def project
+    Project.find params_for[:project_id]
+  end
+
+  def require_project_ownership
+    unless project.owner?(api_user.user)
+      render json: { error: 'not authorized' }, status: :not_authorized
```
camallen: raise a custom error here instead and catch it in the [api_controller](https://github.com/zooniverse/Panoptes/blob/18a44e0c7e6775c4a9ae1b794ce87ed86e485492/app/controllers/api/api_controller.rb#L30) where the rest are.

camallen: Your custom error can contain a message as per normal and the controller should put it into the response body correctly [here](https://github.com/zooniverse/Panoptes/blob/18a44e0c7e6775c4a9ae1b794ce87ed86e485492/app/controllers/concerns/json_api_render.rb#L24)

spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -137,4 +137,30 @@
 
     it_behaves_like "is creatable"
   end
+
+  describe "#update_project_settings", :focus do
+    let!(:api_user) { create(:user) }
+    let!(:project) { create(:project, owner: api_user) }
+    let!(:upps) { create(:user_project_preference, user: authorized_user, project: project) }
```
camallen: don't set the user here as you want to test an authorized project owner updating another user's UPP.

spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -137,4 +137,30 @@
...
+    let(:settings_params) do
+      {
+        user_id: authorized_user.id,
```
camallen: make this the `upp.user_id`

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
 
   require_authentication :all, scopes: [:project]
+  extra_schema_actions :update_settings
   schema_type :json_schema
+  before_action :require_project_ownership, only: [:update_project_settings]
+
+
+  def update_settings
+    upp = UserProjectPreference.find_or_initialize_by(project_id: params_for[:project_id],
+                                                      user_id: params_for[:user_id])
```
camallen: some weird formatting here...

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+    upp.settings.merge! params_for[:settings]
+    if upp.save!
```
camallen: the `save!` method will raise an error if it's not valid so the if / else block won't fire...instead the api_controller will catch the raised error with the correct body response via (i think) [here] (https://github.com/zooniverse/Panoptes/blob/18a44e0c7e6775c4a9ae1b794ce87ed86e485492/app/controllers/api/api_controller.rb#L20).

camallen: so i'd just rewrite it to be 
``` ruby
  upp.save!
  json_api_render(:ok, nothing: true) 
  #or if you want to conform to json_api json_api_render(:ok, serializer.resource({}, controlled_resources, context) but i'm not sure about this....
```

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+  end
 
   private
 
+  def project
+    Project.find params_for[:project_id]
+  end
+
+  def require_project_ownership
+    unless project.owner?(api_user.user)
```
camallen: what'll happen if we pass in a project_id that doesn't resolve to a project?

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
   include PreferencesController
 
   require_authentication :all, scopes: [:project]
+  extra_schema_actions :update_settings
   schema_type :json_schema
+  before_action :require_project_ownership, only: [:update_project_settings]
+
+
```
camallen: as for the failing spec - try and see if you get to the before_action filter or not..if you do then trace it down...if not it might be a thing with controlled resources...

app/controllers/api/v1/project_preferences_controller.rb
zwolf: just trying to stay ahead of the hound 🐩

zwolf:
Thanks for all the suggestions, @camallen, I think I got them all. The spec request is still responding with a 403, and a breakpoint in the before_action is never reached. Seems like it's something deeper...

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
   include PreferencesController
 
   require_authentication :all, scopes: [:project]
+  extra_schema_actions :update_settings
   schema_type :json_schema
+  before_action :validate_records, only: [:update_settings]
+
+  def update_settings
+    upp = UserProjectPreference.find_or_initialize_by(project_id: params_for[:project_id],
```
camallen: Question, the UPP should exist if we are setting data for it right (i.e. they've classified)? If so should we be initializing up a UPP here?

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+                                                      user_id: params_for[:user_id])
+    upp.settings.merge! params_for[:settings]
+    upp.save
```
camallen: This should be save! to raise and error on invalid record.

app/models/user_project_preference.rb
```diff
@@ -1,6 +1,9 @@
 class UserProjectPreference < ActiveRecord::Base
   include Preferences
 
+  class Unauthorized < StandardError; end
+  class RecordNotFound < StandardError; end
```
camallen: This can be removed completely if you just use the AR `.find` lookups.

app/models/user_project_preference.rb
```diff
@@ -1,6 +1,9 @@
 class UserProjectPreference < ActiveRecord::Base
   include Preferences
 
+  class Unauthorized < StandardError; end
```
camallen: This one should move the the controller layer in [api_errors.rb](https://github.com/zooniverse/Panoptes/blob/master/app/controllers/concerns/api_errors.rb), in the controller you can then just raise the api error itself.

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,34 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+    render status: :ok, nothing: true
+  end
 
   private
 
+  def find_project
+    Project.find params_for[:project_id] || nil
```
camallen: I was wrong before, these `.find` methods will raise `ActiveRecord::RecordNotFound` that the API Controller will deal with by default. What you do with this will depend on how you treat the upp find above, if the upp has to exist before you update it then you can proxy the find by user / project id into this lookup, e.g.

``` ruby
UserProjectPreference.find(project_id: params_for[:project_id], user_id: params_for[:user_id])
```
If the upp doesn't have to exist then yeah you should make sure the user / project's can be found, maybe via 
``` ruby
def validate_records
  Project.find(params_for[:project_id])
  User.find(params_for[:user_id]
  unless project.owner?(api_user.user)
    raise Api::Unauthorized, "you must be the project owner"
  end
end
```

spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -137,4 +137,34 @@
...
+    before(:each) do
+      default_request user_id: authorized_user.id, scopes: scopes
+    end
+
+    it "responds with a 200" do
+      post :update_settings, settings_params
+      expect(response.status).to eq(200)
+    end
+
+    xit "updates the UPP settings attribute" do
```
camallen: xit typo here

spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -137,4 +137,34 @@
...
+    let!(:upp) { create(:user_project_preference, project: project) }
+    let(:settings_params) do
+      {
+        user_id: upp.user_id,
+        project_id: project.id,
+        settings: { workflow_id: 1234 }
+      }
+    end
+
+    let(:scopes) { %w(project_preference) }
```
camallen: this is the problem why you can't get into the before_action method block, basically this has to marry the scopes in the require_authentication dsl block define in the controller [here](https://github.com/zooniverse/Panoptes/blob/18a44e0c7e6775c4a9ae1b794ce87ed86e485492/app/controllers/api/v1/project_preferences_controller.rb#L4)

lib/json_api_controller/json_schema_validator.rb
```diff
@@ -7,7 +7,8 @@ module JsonSchemaValidator
 
     included do
       @action_params = Hash.new
-      (@actions & %i(update create)).each do |action|
+      schema_actions = %i(update create) | Array.wrap(@extra_schema_actions)
+      (@actions | schema_actions).each do |action|
```
camallen: don't change this to be the set union it is the set intersection on purpose. 
`(@actions & schema_actions)`

camallen:
@zwolf this is super close, just a bit more controller work to iron out.

spec/controllers/api/v1/project_preferences_controller_spec.rb
zwolf: not a typo, just skipping the test because the "successful request" one is still broken.

spec/controllers/api/v1/project_preferences_controller_spec.rb
zwolf: Without that let(:scopes), it's a 422. Remove the "scopes:" from the default_request and it's a 403. Remove default_request entirely and it's a 422.

lib/json_api_controller/json_schema_validator.rb
zwolf: Changing this | to & (union back to intersection) causes `JsonApiController::NoValidatorForActionError` (from #params_for). Why should it be intersection? Don't I want to include the extra actions defined in @extra_schema_actions (in the ProPrefs controller) to be passed to #schema_class? Using & throws it away, hence the error.

app/controllers/api/v1/project_preferences_controller.rb
zwolf: Good question, I wasn't sure. Since the user ids will be presumably coming from classifications (in gravityspy's case, at least) then the UPPs will exist (...right?). In general, though, it does seem reasonable to generally require the UPP to exist before you can mess with it via this route. Changed to a .where, which should bubble up a RecordNotFound like the others (project/user).

camallen: I think the intention is revealed in the name of the method on the resource, `update`. Just build this update out for now as it's the use case you have for Gravity Spy

lib/json_api_controller/json_schema_validator.rb
camallen: because there is no associated wired up action from the `resource_actions` DSL, you'll need to override the preferences controller shared module resource_actions to include your update_settings method too. 

And yes this should be the intersection otherwise it'll run these for GET req's etc and it only need to validate the schema on POST / PUT methods.

spec/controllers/api/v1/project_preferences_controller_spec.rb
camallen: Remove the `let(:scopes) { %w(project_preference) }` def here and you'll be allowed into the controller before_action filter. I'm pretty sure that the scopes here is 422'ing and not entering the controller (due to scopes restriction via doorkeeper). Removing this and still 422 is due to the internal controller, most likely the json schema (check the response body).

spec/controllers/api/v1/project_preferences_controller_spec.rb
camallen: neat - did not know that.

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,29 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+    render status: :ok, nothing: true
+  end
 
   private
 
+  def validate_records
+    User.find params_for[:user_id]
+    p = Project.find params_for[:project_id]
+    unless p.owner?(api_user.user)
+      raise Api::Unauthorized.new ("You must be the project owner")
```
camallen: Pretty sure these are just the space between the method name and the ()

app/controllers/api/v1/project_preferences_controller.rb
```diff
@@ -2,10 +2,29 @@ class Api::V1::ProjectPreferencesController < Api::ApiController
...
+                                      user_id: params_for[:user_id]).first
+    upp.settings.merge! params_for[:settings]
+    upp.save!
+    render status: :ok, nothing: true
+  end
 
   private
 
+  def validate_records
+    User.find params_for[:user_id]
```
camallen: It's be good to not look these up unless we need to, instead of the current implementation we could use the `.find_by!` for the upp.

``` def upp_settings_update
  @upp = UserProjectPreference.find_by!(user_id: params_for[:user_id], project_id: params_for[:project_id])
  unless upp.project.owner?(api_user.user)
    raise ....
  end
```

And then access the upp via the instance param `@upp` in the update_settings action.

marten: I actually like that the authorization is in a `before_action`. Don't know why you need to find the user (you're not even storing it in a variable), but storing the project in `@project` and then doing `@project.user_project_preferences.find_by!(user_id: params[:user_id])` should work and keep the authorization here in a before filter.

marten:
Minor comments but LGTM

app/controllers/api/v1/project_preferences_controller.rb
zwolf: Call to User was just to confirm its existence and raise an AR exception if it didn't. Seems like I can have it both ways and save @upp in the validation before_action and not have to make more than the one call, since I'm not potentially initializing a UPP anymore.

spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -138,7 +138,7 @@
     it_behaves_like "is creatable"
   end
 
-  describe "#update_settings" do
+  describe "#update_settings", :focus do
```
marten: If you get rid of this focus here you're good to merge.

camallen:
LGTM

PR #1905, opened 2016-08-12
Subject workflow status resource

### LINKED ISSUE #1886: New API Route to retrieve retired subjects
I need a route that lets me retrieve the subjects that have been retired in a given workflow, optionally since a given datetime. Zach and I discussed this a while ago and I forgot to create an issue about it for tracking purposes.

camallen:
closes #1886 - expose the SubjectWorkflowStatus resource via the api

# Review checklist


marten:
Renaming the table can't really be done while the app is running. Are we sure we want to do that, vs just setting self.table_name inside the AR model?

camallen:
Much better idea - i'll do that.

camallen:
@astopy do we need this out this week?

astopy:
@camallen As soon as we can, but it's not vital for it to be this week.

marten:
LGTM

PR #1906, opened 2016-08-12
Update CellectController responses to be more usable

parrish:
Instead of just returning ids, it returns the necessary fields for cellect.

### DIFF
--- a/app/controllers/cellect_controller.rb
+++ b/app/controllers/cellect_controller.rb
@@ -45,7 +45,7 @@ class CellectController < ApplicationController
     if workflow = cellect_workflow_from_param
       SetMemberSubject
         .non_retired_for_workflow(workflow)
-        .select(:id, :priority)
+        .select('set_member_subjects.subject_id as id', :priority)
     else
       []
     end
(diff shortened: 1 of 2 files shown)


parrish:
There's something odd going on in the specs with set_member_subjects vs subjects.  Any thoughts?

app/controllers/cellect_controller.rb
```diff
@@ -25,33 +25,35 @@ def launched_workflows
...
     end
   end
 
-  def cellect_workflow_subject_ids
+  def cellect_workflow_subjects
     if workflow = cellect_workflow_from_param
-      SetMemberSubject.non_retired_for_workflow(workflow).pluck(:subject_id)
+      SetMemberSubject
+        .non_retired_for_workflow(workflow)
+        .select(:id, :priority)
```
camallen: This should be `.select(:subject_id, :priority)` instead of the sms ids.

camallen:
This is great - thanks @parrish. Yeah re the specs, it'll be the sms.id attribute select vs the sms.subject_id.

app/controllers/cellect_controller.rb
parrish: Yep, that got it.  Thanks!

PR #1910, opened 2016-08-16
cellect api caching

camallen:
Add some caching headers to these end points (i wasn't sure about the max-age values)

# Review checklist

### DIFF
--- a/app/controllers/cellect_controller.rb
+++ b/app/controllers/cellect_controller.rb
@@ -6,7 +6,7 @@ class CellectController < ApplicationController
   def workflows
     respond_to do |format|
       format.json do
-        expires_in EXPIRY, public: true
+        cache_response(EXPIRY)
         render json: { workflows: all_cellect_workflows.as_json }
       end
     end
@@ -60,5 +60,11 @@ class CellectController < ApplicationController
 
   def html_to_json_override
     request.format = :json if request.format == :html
+  end
+
+  def cache_response(expiration_time)
+    if Panoptes.flipper[:cellect_controller_caching].enabled?
+      expires_in expiration_time, public: true
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


parrish:
Looks good, but I'm not 100% on the caching, particularly with a file cache-store.  Multiple requests per-minute should only occur when something goes wrong with cellect (or we're bringing up multiple nodes simultaneously).  Now that the loading can't be re-entrant, we should be good without it -- I imagine caching large sets have a fair bit of overhead.

What do you think?

camallen:
These cache directives should just be telling the nginx reverse proxy (and intermediate ones) that it can cache the response and thus avoid hitting the rails app. It's currently only using the rails file store cache (probably should move to redis/memcache) for the workflow details as i thought the larger / slower subjects responses could swamp the internal file store (moving to redis cache should help with this).

screen shot 2016-08-17 at 09 50 30

I figured anything to avoid re-calculating / speeding these up would be beneficial for the rails app. IMHO these responses should be cacheable at the HTTP layer.

astopy:
I'm pretty sure that we don't have caching turned on in nginx or cloudfront for panoptes, so this won't have any effect without also setting those up.

camallen:
you are right re the nginx config but we certainly could be adding nginx cache directives for the these cellect routes.

parrish:
Yeah, that's a pretty gnarly request.  Since a large chunk of that time is just data transfer, maybe we should enable GZIP on it? `Rack::Deflater` provides middleware for it: https://github.com/rack/rack/blob/master/lib/rack/deflater.rb

camallen:
I'll look into adding compression, probably via the nginx reverse proxy though and report back.

marten:
Keep in mind too that this would be used mostly from internal traffic. Compression would still help, but by a much smaller fraction than calling it from the office.

parrish:
👍 Ready to go, or does nginx still need setup?

camallen:
Ready to roll, i'll add nginx later.

PR #1911, opened 2016-08-16
Added settings to UPP serializer

zwolf:
Hopefully this is all that needs to be included.

# Review checklist

### DIFF
--- a/app/serializers/user_project_preference_serializer.rb
+++ b/app/serializers/user_project_preference_serializer.rb
@@ -1,6 +1,6 @@
 class UserProjectPreferenceSerializer
   include RestPack::Serializer
-  attributes :id, :email_communication, :preferences, :href, :activity_count, :activity_count_by_workflow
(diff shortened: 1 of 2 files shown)


app/serializers/user_project_preference_serializer.rb
```diff
@@ -1,6 +1,6 @@
 class UserProjectPreferenceSerializer
   include RestPack::Serializer
-  attributes :id, :email_communication, :preferences, :href, :activity_count, :activity_count_by_workflow
+  attributes :id, :email_communication, :preferences, :href, :activity_count, :activity_count_by_workflow, :settings
```
camallen: you can test this via the controller using [this dsl method](https://github.com/zooniverse/Panoptes/blob/6da6f2a31b6d8ccf7398c6a8a3f3dc7bba6c60dc/spec/controllers/api/v1/project_preferences_controller_spec.rb#L12)

camallen:
LGTM pending specs

PR #1914, opened 2016-08-19
[WIP] use post to update user_preference settings

camallen:
since we don't know the resource URI we should not use PUT, instead we should use POST with the user_id & project_id params to update the resource. Not sure how well this will play with the json api client as i couldn't figure out if it'll work via a resource create. If it does then it'll require the resource to be serialized in full and not just return nothing.

# Review checklist

### DIFF
--- a/spec/controllers/api/v1/project_preferences_controller_spec.rb
+++ b/spec/controllers/api/v1/project_preferences_controller_spec.rb
@@ -150,24 +150,25 @@ RSpec.describe Api::V1::ProjectPreferencesController, type: :controller do
         }
       }
     end
+    let(:run_update) { post :update_settings, settings_params }
 
     before(:each) do
       default_request user_id: authorized_user.id, scopes: scopes
     end
 
     it "responds with a 200" do
-      put :update_settings, settings_params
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -150,63 +150,85 @@
...
+      it "only updates settings of owned project" do
+        run_update
+        expect(response.status).to eq(403)
+      end
     end
 
-    it "responds with a 404 if UPP not found" do
-      unused_project = create(:project)
-      silly_params =
+    context "with silly params" do
```
zwolf: With the restructuring you did here, it might be good to make this a little more descriptive.

zwolf:
The apiary docs should also be updated to reflect this change

zwolf:
LGTM

zwolf:
Is this ready to go out? I've got a feature I'd love to throw on top of this one.

camallen:
Yep - just waiting to get the prev PR out on staging before i merge this one.

PR #1915, opened 2016-08-19
Break up subjects default add_column migration

camallen:
Let's try and avoid locking the large table so this will remove the addtional non-null constraint with the default value. On Staging - I'm going to undo the prev migration so this can run correctly this time.

# Review checklist


marten:
Looks like the Activatable concern doesn't like it if there's no default value?

marten:
LGTM

PR #1916, opened 2016-08-19
Re add default activated_state value to subjects table

camallen:
Merge / deploy only after #1915 has been deployed and the default values filled in by the rake task.

# Review checklist


marten:
LGTM. Don't know if you have any dependent branches so won't merge+squash for you.

PR #1917, opened 2016-08-19
Background counting work and remove increment_counter's

### LINKED ISSUE #1685: remove all increment_counter && counter_cache properties
move the counter jobs to rate limited background workers
remove the counter_cache properties but leave the fields for use in the count workers.

camallen:
closes #1685 - Background the counters instead of trying to do these in band and push the logic to some PORO's for counting. Also remove all the `.increment_counter` calls & `counter_cache` defs that still kick around (left the one on `TaggedResource` intact).

# Review checklist


marten:
LGTM

PR #1925, opened 2016-08-25
skip the subject removal temporarily

camallen:
cleaning up subjects is causing really high cpu on the db, skipping the cleanup in the meantime instead just leaving them as inactive.

# Review checklist

### DIFF
--- a/spec/workers/subject_removal_worker_spec.rb
+++ b/spec/workers/subject_removal_worker_spec.rb
@@ -3,10 +3,17 @@ require 'spec_helper'
 RSpec.describe SubjectRemovalWorker do
   let(:subject_id) { 1 }
 
-  it 'should call the orphan remover cleanup' do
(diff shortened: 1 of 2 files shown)


spec/workers/subject_removal_worker_spec.rb
```diff
@@ -3,10 +3,17 @@
...
+  # TODO: reinstate this when the cleanup code is not killing the db
+  # it 'should call the orphan remover cleanup' do
+  #   remover = instance_double(Subjects::Remover)
+  #   expect(Subjects::Remover).to receive(:new).with(subject_id).and_return(remover)
+  #   expect(remover).to receive(:cleanup)
+  #   subject.perform(subject_id)
+  # end
+
+  # remove this when the above is fixed
+  it 'should not call the orphan remover cleanup', :focus do
```
zwolf: :focus left here

PR #1926, opened 2016-08-26
Gravity Spy subject selection service

marten:
Describe your change here.

# Review checklist

### DIFF
--- a/app/services/aggregation_client.rb
+++ b/app/services/aggregation_client.rb
@@ -2,7 +2,11 @@ class AggregationClient
   include Configurable
 
   self.config_file = "aggregation_api"
-  self.api_prefix = "aggregation"
+  self.api_prefix = "aggregation_api"
+
+  configure :host
+  configure :user_id, file_field: :user, type: :integer
+  configure :application_id, file_field: :application, type: :integer
 
   attr_reader :connection
 
--- a/lib/subjects/strategy_selection.rb
+++ b/lib/subjects/strategy_selection.rb
@@ -39,11 +39,12 @@ module Subjects
     private
 
     def cellect_strategy?
-      return nil unless Panoptes.cellect_on
+      return nil unless Panoptes.flipper.enabled?("cellect")
       strategy_param == :cellect || workflow.using_cellect?
     end
 
     def cellect_ex_strategy?
+      return nil unless Panoptes.flipper.enabled?("cellect_ex")
       strategy_param == :cellect_ex || workflow.subject_selection_strategy == 'cellect_ex'
     end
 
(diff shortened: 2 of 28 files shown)


app/services/aggregation_client.rb
```diff
@@ -2,7 +2,11 @@ class AggregationClient
   include Configurable
 
   self.config_file = "aggregation_api"
-  self.api_prefix = "aggregation"
+  self.api_prefix = "aggregation_api"
```
camallen: did you mean to update the aggregation client?

camallen: and yes you did!

lib/subjects/strategy_selection.rb
```diff
@@ -27,26 +27,38 @@ def select
...
 
     private
 
     def cellect_strategy?
+      return nil unless Panoptes.cellect_on
       strategy_param == :cellect || workflow.using_cellect?
     end
 
+    def cellect_ex_strategy?
+      strategy_param == :cellect_ex || workflow.subject_selection_strategy == 'cellect_ex'
```
camallen: Do you want to add a feature flag here to enable / disable this at run time?

lib/subjects/strategy_selection.rb
```diff
@@ -27,26 +27,38 @@ def select
...
       case strategy
       when :cellect
         cellect_params = [ workflow.id, user.try(:id), subject_set_id, limit ]
         subject_ids = Subjects::CellectClient.get_subjects(*cellect_params)
         sms_scope = SetMemberSubject.by_subject_workflow(subject_ids, workflow.id)
         sms_scope.pluck("set_member_subjects.id")
+      when :cellect_ex
+        cellect_ex_params = [ workflow.id, user.try(:id), subject_set_id, limit ]
+        subject_ids = Subjects::CellectExSelection.get_subjects(*cellect_ex_params)
+        sms_scope = SetMemberSubject.by_subject_workflow(subject_ids, workflow.id)
```
camallen: any desire to extract the common code to a method for the plucking of sms ids for cellect clients?

camallen:
minor comments LGTM

lib/subjects/strategy_selection.rb
marten: Yeah, might be a good idea since we're using this to see if it's stable ;)

lib/subjects/strategy_selection.rb
marten: Not yet, I've got a bigger follow-up branch with refactorings on the whole selection stuff planned.

marten:
@camallen quick catch-up review on that last commit?

PR #1927, opened 2016-08-26
Disable Hound multiline method chains indentation warnings

marten:
Stops hound complaints about:

```ruby
Foo.bar
  .baz
  .quux
```


camallen:
thank you

PR #1930, opened 2016-08-30
Save user project_id attr on first UPP save

### LINKED ISSUE #1794: Save first project info for all users
There is a field in the users table for recording the first project a user tries, but it is not always recording the info. It would be great to record this properly for all new users. The simplest would be to take the info from the first classification they submit. We update the user's stats page as soon as they make a classification on a project, so we must be able to grab this information easily and send the project id to the project_id field in the users table.

@zwolf is this something you fancy having a go at?

zwolf:
Closes #1794. Saves a user's project_id when the UPP is created for the first time so that we know which project was the first that a user contributed to.

# Review checklist

### DIFF
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -40,7 +40,10 @@ module Api
         was_new = upp.new_record?
         upp.legacy_count[create_params[:workflow]] = create_params[:count]
         if upp.save
-          ProjectClassifiersCountWorker.perform_async(upp.project_id) if was_new
+          if was_new
+            ProjectClassifiersCountWorker.perform_async(upp.project_id)
+            upp.user.update( {project_id: upp.project.id} )
+          end
           :ok
         else
           :unprocessable_entity
--- a/app/controllers/api/events_controller.rb
+++ b/app/controllers/api/events_controller.rb
@@ -42,7 +42,7 @@ module Api
         if upp.save
           if was_new
             ProjectClassifiersCountWorker.perform_async(upp.project_id)
-            upp.user.update( {project_id: upp.project.id} )
+            upp.user.update_column(:project_id, upp.project.id)
           end
           :ok
         else
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -37,7 +37,7 @@ class ClassificationLifecycle
       if first_classifcation = upp.email_communication.nil?
         ProjectClassificationsCountWorker.perform_async(project.id)
         upp.email_communication = user.project_email_communication
-        upp.user.update( {project_id: project.id} )
+        upp.user.update_column(:project_id, project.id)
       end
       upp.changed? ? upp.save! : upp.touch
     end
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -37,6 +37,7 @@ class ClassificationLifecycle
       if first_classifcation = upp.email_communication.nil?
         ProjectClassificationsCountWorker.perform_async(project.id)
         upp.email_communication = user.project_email_communication
+        upp.user.update( {project_id: project.id} )
       end
       upp.changed? ? upp.save! : upp.touch
     end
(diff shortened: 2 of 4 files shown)


spec/controllers/api/events_controller_spec.rb
```diff
@@ -18,7 +18,7 @@ def overridden_params(new_params)
       let(:workflow) { create(:workflow) }
       let(:project) do
         p = workflow.project
-        p.update_columns(migrated: true, configuration: { zoo_home_project_id: zoo_home_project_id })
+        p.update_columns(migrated: true, id: zoo_home_project_id, configuration: { zoo_home_project_id: zoo_home_project_id })
```
camallen: why the update to the auto assigned id of the model instance?

spec/controllers/api/events_controller_spec.rb
```diff
@@ -144,7 +144,7 @@ def overridden_params(new_params)
 
       context "when the project id refers to a non-legacy migrated project" do
         let(:unknown_legacy_project) do
-          overridden_params(project_id: project.id)
+          overridden_params(project_id: 9999)
```
camallen: Not sure why this needs to be change along with the project id as the created factory project id should not match the config => zoo home project id from the let(:project) block above in prev work. 

We could look at ensuring this via the 
``` ruby
let(:zoo_home_project_id) { (10..20).to_a.sample }
#could become
let(:zoo_home_project_id) { project.id + 1 }
```
The lookups for `event_params_check` calls the `legacy_zoo_project` method to find the legacy project via the config.

spec/controllers/api/events_controller_spec.rb
zwolf: Because the UPP is initialized in the events controller with `project_id: legacy_zoo_project.id`. If this and the project id below aren't changed, then many of the specs will fail due to foreign key restraints on upp.project_id not referencing existing projects.

camallen: aha!

lib/classification_lifecycle.rb
```diff
@@ -37,6 +37,7 @@ def process_project_preference
       if first_classifcation = upp.email_communication.nil?
         Project.increment_counter :classifiers_count, project.id
         upp.email_communication = user.project_email_communication
+        upp.user.update( {project_id: project.id} )
```
camallen: can this just be a `user.update_column(:project_id, upp.project_id)` instead of running validations, callbacks etc.

app/controllers/api/events_controller.rb
```diff
@@ -36,7 +36,10 @@ def process_incoming_event
           was_new = upp.new_record?
           upp.legacy_count[create_params[:workflow]] = create_params[:count]
           if upp.save
-            Project.increment_counter :classifiers_count, upp.project_id if was_new
+            if was_new
+              Project.increment_counter :classifiers_count, upp.project_id
+              upp.user.update( {project_id: create_params[:project_id]} )
```
camallen: same thing here, i don't think we need to run the validations if we're just updating a FK relation.

lib/classification_lifecycle.rb
zwolf: Should a user's project_id not have to reference an actual project? If not, would it make more sense to remove the FK constraint?

camallen: it should for sure

app/controllers/api/events_controller.rb
camallen: @zwolf the source of the FK violations are due to using the legacy project id her and not the id of the project itself. if you chnange this to `upp.update_column(:project_id, upp.project_id)` it should not violate the FK constraint.

zwolf:
Did my most recent commit/comments adequately answer your questions, @camallen, or did I miss what you were suggesting?

camallen:
Somewhat, any reason why we're still running update with validations, callbacks on the user model (i don't think we need them here)? I'm about a 2/10 on this ;)

zwolf:
Thought I got those! Got 'em now.

camallen:
LGTM pending specs

PR #1936, opened 2016-09-01
Fix TalkApiClient random test failure

marten:
Normally you'd scratch your head about why things randomly fail. In this case, I'm actually not sure why it sometimes passes...

This will fix it in any case.


camallen:
LGTM

PR #1937, opened 2016-09-01
speed up doorkeeper lookups

camallen:
add index for doorkeeper authorisation lookup now with cleanup code that will remove old tokens / grants before running the migration. We should add (happy to do for this PR but can come in future) a rake task / scheduled worker for the cleanup

Also once these get cleanup up it'll help with the issue of the video showing on the home page for PFE. 

# Review checklist

### DIFF
--- a/db/migrate/20160901141903_add_oauth_tokens_index.rb
+++ b/db/migrate/20160901141903_add_oauth_tokens_index.rb
@@ -2,6 +2,8 @@ class AddOauthTokensIndex < ActiveRecord::Migration
   disable_ddl_transaction!
 
   def change
+    Doorkeeper::AccessCleanup.new.cleanup!
+
     unless index_exists?(:oauth_access_tokens, [:application_id, :resource_owner_id])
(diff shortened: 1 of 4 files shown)


marten:
When I looked at this before I realised this will also remove the refresh tokens (as mentioned in the issue you linked in your commit). I don't think anyone is using those right now, but wanted to make sure we're aware of it. Personally I think long term we shouldn't be removing them, but since panoptes-javascript-client has just been getting new tokens all the time (every page load) I think we can do with a cleanup while nobody will notice.

camallen:
yep - agree completely, i don't know anyone using refresh tokens right now. And they can always get a new token if their grant allows it. I still think it can cleanup revoked tokens but we might want an audit trail on these.

astopy:
Just noting that the python client will use refresh tokens if needed https://github.com/zooniverse/panoptes-python-client/blob/master/panoptes_client/panoptes.py#L353 but I doubt a one off cleanup will cause problems for anyone

PR #1938, opened 2016-09-01
summate the counters from the lowest level

camallen:
Rely on the subject workflow status counts being the authoritative source of classification counts that respect the launch dates, just aggregate these up the chain via simple sum queries to avoid counting these again.

# Review checklist

### DIFF
--- a/app/counters/workflow_counter.rb
+++ b/app/counters/workflow_counter.rb
@@ -7,10 +7,6 @@ class WorkflowCounter
   end
 
   def classifications
-    scope = workflow.classifications
-    if launch_date = workflow.project.launch_date
-      scope = workflow.classifications.where("created_at >= ?", launch_date)
-    end
-    scope.count
+    SubjectWorkflowStatus.where(workflow: workflow).sum(:classifications_count)
   end
 end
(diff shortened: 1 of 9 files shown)


app/counters/workflow_counter.rb
```diff
@@ -7,10 +7,6 @@ def initialize(workflow)
   end
 
   def classifications
-    scope = workflow.classifications
-    if launch_date = workflow.project.launch_date
-      scope = workflow.classifications.where("created_at >= ?", launch_date)
-    end
-    scope.count
+    SubjectWorkflowStatus.where(workflow: workflow).sum(:classifications_count)
```
marten: Should we care about subjects and sets that get unlinked after they have had classifications here?

marten: I'm fine with including those in the counts (due to speed, and because the classifications happened) but it does keep confusing people.

marten:
LGTM

app/counters/workflow_counter.rb
camallen: Yeah - i tend to agree about the numbers not validating with the current linked set. I'll iterate on this in another PR.

PR #1944, opened 2016-09-05
Fix early retire workflows

camallen:
Only count the current set of linked subjects to determine the retired subject count for the workflow and respect live workflows for the project counts.

# Review checklist

### DIFF
--- a/app/workers/workflow_retired_count_worker.rb
+++ b/app/workers/workflow_retired_count_worker.rb
@@ -3,23 +3,20 @@ class WorkflowRetiredCountWorker
 
   sidekiq_options queue: :data_high
 
-  attr_reader :workflow
+  sidekiq_options congestion: {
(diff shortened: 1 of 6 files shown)


app/workers/workflow_retired_count_worker.rb
```diff
@@ -3,23 +3,20 @@ class WorkflowRetiredCountWorker
 
   sidekiq_options queue: :data_high
 
-  attr_reader :workflow
+  sidekiq_options congestion: {
+    interval: 10,
```
camallen: It seems this worker can be fired multiple times from the `workflows_controller#retire_subjects` route (think aggregation engine). I assume once every 10 seconds with a cancel if exceeding will be ok...thoughts?

marten:
LGTM

PR #1946, opened 2016-09-05
Avoid multiple classifcation counts on classification receipt

camallen:
Try and not spam the db, instead use a increment counter on hopefully? non-deadlocking subject workflow status records and a rate limited worker to resolve the correct count.

# Review checklist

### DIFF
--- a/app/workers/subject_workflow_status_count_worker.rb
+++ b/app/workers/subject_workflow_status_count_worker.rb
@@ -1,7 +1,7 @@
 class SubjectWorkflowStatusCountWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: :data_low
+  sidekiq_options queue: :really_high
 
   sidekiq_options congestion: {
     interval: ENV.fetch("sws_count_worker_interval", 30),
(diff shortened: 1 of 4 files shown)


app/workers/subject_workflow_status_count_worker.rb
```diff
@@ -0,0 +1,21 @@
+class SubjectWorkflowStatusCountWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :really_high
+
+  sidekiq_options congestion: {
+    interval: ENV.fetch("sws_count_worker_interval", 30),
```
camallen: happy to modify the defaults here, thought this was sensible (for random dist on selection) and added ability to override with ENV vars.

marten: I just thought about that this afternoon. I'd be cool to integrate configuration values like this with Flipper, and have a live admin panel for them. In general we should probably be putting more non-critical things under flags anyways, so we can toggle them off under load.

marten:
LGTM

app/workers/subject_workflow_status_count_worker.rb
camallen: Indeed it would, we'd have to flesh out the model for it but I agree. You want me to rip the ENV configs out? They may not get used, due the cost of rebuilding the ami's...

marten: Nah, they're a nice greppable reminder to pull these out if/when we build such a thing.

PR #1947, opened 2016-09-05
Remove queue cruft in selection

camallen:
Queues are moving for task assignment and not as a buffer for subject selection. This sets the scene to have simple queue workers and removes the queues from selection. We'll need to add them back in for task assignment and figure out how that works, in the meantime this will remove the code paths. 

# Review checklist


marten:
This is LGTM btw, but didn't want to merge it while we were working on the counters.

PR #1950, opened 2016-09-06
Fix Workflow#finished_at assignment

### LINKED ISSUE #1571: workflows are not auto finishing
Workflow counts show the project is finished but the `finished_at` flag is not being set, most likely due to the transaction block [here](https://github.com/zooniverse/Panoptes/blob/4db3fca4e7fb9ab644826437c7ffc057fe36d241/app/workers/retirement_worker.rb#L11).

parrish:
Closes #1571 

Basically, the `finished?` check is called before `WorkflowRetiredCountWorker` can update the retired counters used by `finished?`.  This just moves the `finished?` call to `WorkflowRetiredCountWorker`.


camallen:
LGTM - nice catch @parrish

PR #1952, opened 2016-09-06
Unfinish when adding new data

### LINKED ISSUE #1639: Workflows should auto-unfinish
* Adding subjects to a subject_set should trigger a check on all associated workflows.
* Adding subject_sets to a workflow should trigger a check on that workflow.
* Checking that workflow should basically do: `if workflow.retired?; workflow.finished_at ||= Time.zone.now; else; workflow.finished_at = nil; end`

Labelling this as a bug because we're basically forgetting to invalidate caches here.

camallen:
closes #1639 unset the workflow finished_at flag from classification events and adding new data.

The only caveat is that it won't unfinish a workflow if someone adds new subjects to an existing subject set as we'd need to chain the unfinish worker to the set count worker and possibly check against counts. Happy to review if we think this is how people will upload data (we could also expose the finished_at field to remove it via the api / front-end).

# Review checklist

### DIFF
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -11,7 +11,6 @@ class RetirementWorker
       workflow = count.workflow
       WorkflowRetiredCountWorker.perform_async(workflow.id)
       PublishRetirementEventWorker.perform_async(workflow.id)
-      UnfinishWorkflowWorker.perform_async(workflow.id)
 
       if Panoptes.use_cellect?(workflow)
         RetireCellectWorker.perform_async(count.subject_id, workflow.id)
--- a/app/workers/retirement_worker.rb
+++ b/app/workers/retirement_worker.rb
@@ -11,6 +11,7 @@ class RetirementWorker
       workflow = count.workflow
       WorkflowRetiredCountWorker.perform_async(workflow.id)
       PublishRetirementEventWorker.perform_async(workflow.id)
+      UnfinishWorkflowWorker.perform_async(workflow.id)
 
       if Panoptes.use_cellect?(workflow)
         RetireCellectWorker.perform_async(count.subject_id, workflow.id)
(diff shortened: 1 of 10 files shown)


app/workers/retirement_worker.rb
```diff
@@ -11,6 +11,7 @@ def perform(count_id)
       workflow = count.workflow
       WorkflowRetiredCountWorker.perform_async(workflow.id)
       PublishRetirementEventWorker.perform_async(workflow.id)
+      UnfinishWorkflowWorker.perform_async(workflow.id)
```
parrish: If this needs to be checked here, it'll need to be called from `WorkflowRetiredCountWorker` (like #1950) to get the current counts.

parrish:
The only thing I'm a bit confused about is that the `UnfinishWorkflowWorker` only seems to be called by the `RetirementWorker`.  So a workflow would only be marked as unfinished when it receives a classification.  That's contentious with `finished_at` getting set in `WorkflowRetiredCountWorker` too.

Would it make more sense to queue this job when a subject is added to the workflow?

camallen:
@parrish yeah - it was a first attempt to see if i could allow new data coming in to remove the finished flag. Ended up agreeing with you that relying on count syncing via workers was a can of worms. Instead I've tried to simplify and just unfinish when linking new sets of data to a workflow instead.

parrish:
Looks good.  Ready for this to go?

marten:
This feels quite stateful though. Did you consider an alternative like:

```
if retired_count >= subject_count
  workflow.finished_at ||= Time.now
else
  workflow.finished_at = nil
end
```

camallen:
@marten yeah, we can't rely on those counts being correct.

camallen:
@marten not sure i'm following, the `WorkflowRetiredCountWorker` corrects / sets the value for `retired_subjects_count` that this functionality you propose would rely on.

marten:
Yeah, but rather than setting workflow.finished it could reuse my proposal. Anyways, I'm 2/10 on this. Go ahead and merge and get it out.

PR #1955, opened 2016-09-07
Pull worker enqueues out of transactions

marten:
Describe your change here.

# Review checklist


lib/classification_lifecycle.rb
```diff
@@ -1,82 +1,106 @@
...
-      # as the count worker won't fire if the seens are set, see #should_count_towards_retirement
-      instance_eval(&block) if block_given?
+      update_counters
+      process_project_preference
       create_recent
       update_seen_subjects
     end
+
     publish_data
+    notify_cellect
```
camallen: can this be above the publish data to get into the active sidekiq slots (hopefully / depends on queue priority) before the publish jobs.

lib/classification_lifecycle.rb
```diff
@@ -1,82 +1,106 @@
...
+
+    subject_ids.each do |subject_id|
+      SeenCellectWorker.perform_async(workflow.id, user.try(:id), subject_id)
     end
   end
 
   def refresh_queue
     subjects_workflow_subject_sets.each do |set_id|
       queue = SubjectQueue.by_set(set_id).find_by(user: user, workflow: workflow)
+
```
camallen: I think we should just remove this stuff and i'll rebase my queue removal PR later.

lib/classification_lifecycle.rb
```diff
@@ -1,82 +1,106 @@
...
+  def self.queue(classification, action)
     unless classification.persisted?
       message = "Background process called before persisting the classification."
       raise ClassificationNotPersisted.new(message)
     end
     ClassificationWorker.perform_async(classification.id, action.to_s)
   end
 
-  def transact!(&block)
+  def self.perform(classification, action)
```
marten: I liked the duality of `self.queue` + `self.perform`, but I don't care too much. Will change.

lib/classification_lifecycle.rb
marten: Is just removing the `refresh_queue` method enough?

lib/classification_lifecycle.rb
marten: Ah, actually, removing this would make it harder to test, since the spec needs to set up a stub on `.new` to return a spy so we can test that `#execute` is called on the instance.

Ideally the lifecycle would just be the worker itself, but I don't know if that's easy.

camallen: leave it as is then if it allows neat testing interface.

lib/classification_lifecycle.rb
camallen: Removing any queue stuff is ideal but happy if you want to leave it as is.

marten: It's probably better if we remove queue cruft in one go throughout. Nicer history for future debugging.

app/workers/classification_count_worker.rb
```diff
@@ -3,16 +3,19 @@ class ClassificationCountWorker
...
 
-      count.class.increment_counter(:classifications_count, count.id)
-      SubjectWorkflowStatusCountWorker.perform_async(count.id)
+      Workflow.transaction do
+        count = SubjectWorkflowStatus.find_or_create_by!(subject_id: subject_id, workflow_id: workflow_id)
+        count.class.increment_counter(:classifications_count, count.id) unless was_update
+      end
 
       ProjectClassificationsCountWorker.perform_async(workflow.project.id)
+      SubjectWorkflowStatusCountWorker.perform_async(count.id)
```
camallen: if the count `find_or_create_by!` raises and error and the `count = nil` this will raise an error.

camallen: ignore this, i managed to not delete it during the review approval.

camallen:
LGTM

app/workers/classification_count_worker.rb
marten: If `find_or_create_by!` raises, AR's `.transaction` catches that exception to perform the rollback, but it also reraises so that the exception itself bubbles up. So in case of exception, this line will never be reached.

PR #1959, opened 2016-09-09
Process CellectEx HTTP 500s correctly

marten:


zwolf:
Gotta catch 'em all. LGTM.

PR #1961, opened 2016-09-09
speed up and fix some csv annotation errors

camallen:
add cache for classification -> subjects lookup (massive speed up here) and fixed some missing annotation tool label information in classifications. 

some classifications are missing the tool label index e.g. {"x"=>X, "y"=>Y} VS {"x"=>X, "y"=>Y, "tool"=>0, "frame"=>0, "details"=>[]} ~~most likely a borked PFE deploy that lost some frame info...~~ Seems to be the gold standard classification uploads after chatting to the researcher.

No will just return nil label as there is no information on the tool index that was being used.

# Review checklist

### DIFF
--- a/app/workers/classifications_dump_worker.rb
+++ b/app/workers/classifications_dump_worker.rb
@@ -8,33 +8,49 @@ class ClassificationsDumpWorker
 
   sidekiq_options queue: :data_high
 
+  attr_accessor :cache
+
   def perform_dump
     CSV.open(csv_file_path, 'wb') do |csv|
-      cache = ClassificationDumpCache.new
+      @cache = ClassificationDumpCache.new
       formatter = Formatter::Csv::Classification.new(project, cache)
-
-      csv <<  formatter.class.headers
-
-      completed_project_classifications.find_in_batches do |group|
-        subject_ids = group.flat_map(&:subject_ids).uniq
-        workflow_ids = group.map(&:workflow_id).uniq
-
-        cache.reset_subjects(Subject.where(id: subject_ids).load)
-        retired_counts = SubjectWorkflowStatus.retired.where(
-          subject_id: subject_ids,
-          workflow_id: workflow_ids
-        ).load
-        cache.reset_subject_workflow_statuses(retired_counts)
-
-        group.each { |classification| csv << formatter.to_array(classification) }
+      csv << formatter.class.headers
+      completed_project_classifications.find_in_batches do |batch|
+        subject_ids = setup_subjects_cache(batch)
+        setup_retirement_cache(batch, subject_ids)
+        batch.each do |classification|
+          csv << formatter.to_array(classification)
+        end
       end
     end
   end
 
   def completed_project_classifications
-    project.classifications
-    .complete
-    .joins(:workflow)
-    .includes(:user, workflow: [:workflow_contents])
+    project
+      .classifications
+      .complete
+      .joins(:workflow)
+      .includes(:user, workflow: [:workflow_contents])
+  end
+
+  private
+
+  def setup_subjects_cache(classifications)
+    classification_ids = classifications.map(&:id).join(",")
+    sql = "SELECT classification_id, subject_id FROM classification_subjects where classification_id IN (#{classification_ids})"
+    c_s_ids = ActiveRecord::Base.connection.select_rows(sql)
+    cache.reset_classification_subjects(c_s_ids)
+    subject_ids = c_s_ids.map { |cs| cs.last }
+    cache.reset_subjects(Subject.unscoped.where(id: subject_ids).load)
+    subject_ids
+  end
+
+  def setup_retirement_cache(classifications, subject_ids)
+    workflow_ids = classifications.map(&:workflow_id).uniq
+    retired_counts = SubjectWorkflowStatus.retired.where(
+      subject_id: subject_ids,
+      workflow_id: workflow_ids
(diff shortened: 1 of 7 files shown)


lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -125,7 +125,7 @@ def task_label(task_info)
       end
 
       def tool_label(task_info, drawn_item)
-        tool = task_info["tools"] && task_info["tools"][drawn_item["tool"]]
+        tool = task_info["tools"] && task_info["tools"][drawn_item.fetch("tool", 0)]
```
camallen: Here instead of returning the 0 (first tool label) should we just not return a label?

app/workers/classifications_dump_worker.rb
```diff
@@ -8,33 +8,49 @@ class ClassificationsDumpWorker
 
   sidekiq_options queue: :data_high
 
+  attr_accessor :cache
+
   def perform_dump
     CSV.open(csv_file_path, 'wb') do |csv|
-      cache = ClassificationDumpCache.new
+      @cache = ClassificationDumpCache.new
```
marten: Maybe just a method `def cache; @cache ||= ClassificationDumpCache.new; end`? I tend to prefer that (sets it up on first use) over instance variables that may or may not be set up.

app/workers/classifications_dump_worker.rb
```diff
@@ -8,33 +8,49 @@ class ClassificationsDumpWorker
...
+      .complete
+      .joins(:workflow)
+      .includes(:user, workflow: [:workflow_contents])
+  end
+
+  private
+
+  def setup_subjects_cache(classifications)
+    classification_ids = classifications.map(&:id).join(",")
+    sql = "SELECT classification_id, subject_id FROM classification_subjects where classification_id IN (#{classification_ids})"
```
marten: Should we just create that `ClassificationSubject` model? I've needed it on other places too.

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -145,8 +145,8 @@ def answer_labels
               error_class:   "Task export error",
               error_message: "The task cannot be exported",
               context: {
-                classification: @classification,
-                annotation: @annotation,
+                classification: @classification.inspect,
```
marten: Perhaps just the classification ID? This inspect might include details like `user_ip`.

lib/classification_dump_cache.rb
```diff
@@ -15,10 +16,20 @@ def reset_subject_workflow_statuses(subject_workflow_statuses)
     @subject_workflow_statuses = subject_workflow_statuses.group_by(&:subject_id)
   end
 
+  def reset_classification_subjects(classification_subjects)
+    @classification_to_subjects = classification_subjects.map do |cs|
```
marten: This won't work for classifications with more than one subject, you'll need to group `classification_subjects` by `classification_id`, which will also make this entire body simpler:

```ruby
@classification_to_subjects = classification_subjects.group_by {|classification_id, _| classification_id }
```

Of course, if you go the model route that'll need to be different than my example.

app/workers/classifications_dump_worker.rb
```diff
@@ -8,33 +8,49 @@ class ClassificationsDumpWorker
...
+    c_s_ids = ActiveRecord::Base.connection.select_rows(sql)
+    cache.reset_classification_subjects(c_s_ids)
+    subject_ids = c_s_ids.map { |cs| cs.last }
```
marten: In this case I'd prefer a more explicit `.map {|_, subject_id| subject_id }` over a `last` which doesn't tell you what the "last" actually is, but if we go the model route this will just be `.map(&:subject_id)`.

app/workers/classifications_dump_worker.rb
camallen: Perhaps, though not super straightforward. What's the use case your end?

lib/formatter/csv/annotation_for_csv.rb
camallen: good point, this was fixing some older logging that zach put in that dumped the object instance data. I may just remove this as i am pretty sure this will fix it.

app/workers/classifications_dump_worker.rb
marten: Counters-stuff, like https://github.com/zooniverse/Panoptes/blob/master/app/models/classification.rb#L55

camallen: I think we should do this but maybe in a diff PR

camallen: I'd like to get this out and iterate as it fixes a bug + optimises performance.

app/workers/classifications_dump_worker.rb
camallen: agreed - done

lib/formatter/csv/annotation_for_csv.rb
camallen: agreed - done.

lib/classification_dump_cache.rb
camallen: great catch! i've modified the code and added a spec for this case.

camallen:
@marten should be ready for re-review

camallen:
@marten calling id on the hash resulting in a NoMethodError, i've updated.

PR #1976, opened 2016-09-26
Add explicit timeouts

marten:


camallen:
LGTM

camallen:
Will this raise an error that the Strategy selector can catch and fallback on?

PR #1984, opened 2016-09-28
correctly serialize the retired status of subjects

camallen:
Closes: https://github.com/zooniverse/Panoptes-Front-End/issues/3077

Rename the enum field to not override the instance `blank?` method.
``` ruby 
irb(main):041:0> SubjectWorkflowStatus.retired.by_subject_workflow(subject.id, workflow.id).present?
=> false
irb(main):042:0> SubjectWorkflowStatus.retired.by_subject_workflow(subject.id, workflow.id)         
=> #<SubjectWorkflowStatus id: .....>
irb(main):053:0> SubjectWorkflowStatus.retired.by_subject_workflow(subject.id, workflow.id).blank?  
=> true
```

# Review checklist

this was returning false even though an AR instance was returned.

### DIFF
--- a/app/operations/workflows/retire_subjects.rb
+++ b/app/operations/workflows/retire_subjects.rb
@@ -1,5 +1,7 @@
 module Workflows
   class RetireSubjects < Operation
+    set_callback :validate, :before, :rewrite_blank_reason
+
     validates :retirement_reason, inclusion: {
       in: SubjectWorkflowStatus.retirement_reasons.keys,
       allow_nil: true
@@ -34,6 +36,14 @@ module Workflows
 
     def subject_ids
       Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+    end
+
+    private
+
+    def rewrite_blank_reason
+      if @retirement_reason == "blank"
+        @retirement_reason = "nothing_here"
+      end
     end
   end
 end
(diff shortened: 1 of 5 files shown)


marten:
Interesting, any idea why that is? 

A test case to make sure we don't refactor it back might be a good idea.

camallen:
yeah - just figured it out, it's due to an enum field :blank declared on that model. I'll update my PR to rename the enum field so that it doesn't overwrite the blank? method.

marten:
How awesome that Rails doesn't do a nice warning that you probably don't want to shoot that foot. \_(ツ)_/¯

app/operations/workflows/retire_subjects.rb
```diff
@@ -1,6 +1,9 @@
 module Workflows
   class RetireSubjects < Operation
```
marten: This operation should (for now) rewrite incoming "blank" reasons to "nothing_here". Nero sends that, so we'd break the API if we don't. It's not a public API method as far as I can tell from the docs, so this is just so we decouple the Nero update from this one.

camallen: Got it, will update.

camallen: updated

PR #1990, opened 2016-10-05
CSV Export by Workflow

zwolf:
This is a ~~rather ridiculous~~ ___totally reasonable___ pull request

# Review checklist

- [ ] First, the most important one: is this PR small enough that you can actually review it? lol. Feel free to just reject a branch if the changes are hard to review due to the length of the diff.


camallen:
This seems pretty reasonable to me you as the impact of the instance type difference is pretty minimal. You could have implemented it a couple of other ways, perhaps:

1. Use composition and Extract the common dump worker code to a module that is included each ClassificationsExportWorker and provide the any custom behaviour in the specific classes. 
2. Use an adapter to take either a Project / Workflow ClassificationsExportWorker instance and fit the current code.
3. As you said duplicate all the code without refactoring into reusable components, though i don't really think its simpler to maintain more code long term.

zwolf:
Culprit ended up being a one-line fix to add :create_classifications_exports to the require_login prepend_before_action. Added a couple other improvements as well. Pair programming++.

zwolf:
It should be resource all over, but ActiveInteraction requirements force the use of `object: object` or else it complains. You can't have an operation with a shared input unless it's a class they both inherit from.

camallen:
intriguing

zwolf:
see: https://github.com/orgsync/active_interaction/issues/169 and https://github.com/orgsync/active_interaction/issues/178

marten:
My opinion is that project-based exports should go away as soon as PFE is over to the workflow ones anyway. Shouldn't be an eyesore for too long

camallen:
@zwolf :shipit:

PR #1992, opened 2016-10-11
[WIP] add index to users valid_email column

### LINKED ISSUE #1967: Add index for valid_email column
It looks like there's no index for this field (in the `users` table). We use this to avoid including invalid addresses on our mailing lists, so it should really be indexed.

camallen:
Closes #1967 

Concurrently adds an index to the boolean valid_email column on users table. I'm not sure this is going to avoid table scans as it depends on the query's where clause and the table scan cost used in the query planner.

Note: ~13K / ~1.5M have the value set to false and it's not null. I haven't actually tested this but I'm pretty sure a query like `select * from users where valid_email is true` will be a table scan even if we add this index. If that query is the only use case and it is slower than a table scan then we should not add the index. 

http://dba.stackexchange.com/questions/27681/unexpected-seq-scan-when-doing-query-against-boolean-with-value-null

Thoughts?
# Review checklist

### DIFF
--- a/db/migrate/20161011134340_add_valid_email_index_to_users.rb
+++ b/db/migrate/20161011134340_add_valid_email_index_to_users.rb
@@ -1,0 +1,7 @@
+class AddValidEmailIndexToUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    add_index :users, :valid_email, algorithm: :concurrently
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2663,6 +2663,13 @@ CREATE UNIQUE INDEX index_users_on_unsubscribe_token ON users USING btree (unsub
 
 
 --
+-- Name: index_users_on_valid_email; Type: INDEX; Schema: public; Owner: -; Tablespace: 
+--
+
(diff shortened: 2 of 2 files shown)


astopy:
For now I've run `update users set global_email_communication=False, project_email_communication=False where valid_email is False` to make this less urgent, so we have time to test this.

camallen:
I'm going to do some testing and report back.

camallen:
```
[2] pry(main)> User.where(valid_email: nil).explain
....
 Index Scan using index_users_on_valid_email on users 
   Index Cond: (valid_email IS NULL)
(2 rows)
```
```
[3] pry(main)> User.where(valid_email: false).limit(10000).explain
....
  Index Scan using index_users_on_valid_email on users
         Index Cond: (valid_email = false)
         Filter: (NOT valid_email)
(4 rows)
```
```
[4] pry(main)> User.where(valid_email: true).limit(10000).explain
...
  Seq Scan on users
         Filter: valid_email
(3 rows)
```

camallen:
Looks like our use case of `valid: true` will be a full table scan regardless of the index.

PR #2011, opened 2016-10-21
handle persistent task type in csv annotation exports

camallen:
**MERGE DEPENDENT** on https://github.com/zooniverse/Panoptes-Front-End/pull/2914

Adds a new task type (persistent) that is always displayed for a workflow in PFE. This task can not be wired up in normal task flow. It's use case is for constant answers irrespective of specific tasks, e.g. nothing in the image? and fire? Similar to the way snapshot serengeti works with fire / nothing.

image
# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -14,7 +14,7 @@ module Formatter
         case task['type']
         when "drawing"
           drawing
-        when /single|multiple|persistent/
+        when /single|multiple|shortcut/
           simple
         when "text"
           text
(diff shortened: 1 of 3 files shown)


lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -14,7 +14,7 @@ def to_h
         case task['type']
         when "drawing"
           drawing
-        when "single", "multiple"
+        when /single|multiple|persistent/
```
camallen: I'm not convinced about the persistent task type name here and am happy to change it.

PR #2012, opened 2016-10-21
use linked subject / retired ratio for project completion

camallen:
linked to https://github.com/zooniverse/Panoptes-Front-End/issues/3145

ignore the metrics based on classification counts and dealing with unlinked sets, just use the current linked set of subjects and retired counts to provide a ratio for completeness. 

# Review checklist

### DIFF
--- a/app/workers/calculate_project_completeness_worker.rb
+++ b/app/workers/calculate_project_completeness_worker.rb
@@ -24,22 +24,14 @@ class CalculateProjectCompletenessWorker
   end
 
   def workflow_completeness(workflow)
-    return 0.0 if workflow.subjects.count == 0
+    workflow_subjects_count = workflow.subjects_count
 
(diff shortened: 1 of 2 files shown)


camallen:
@marten @zwolf you both ok with just using a ratio between retired / subjects to provide a workflow completion metric? It takes into account the current linked state of subjects this way and is a simple calc.

mschwamb:
even if this a temp fix, it would be really helpful to have a more useful completion plot showing and especially for projects that are doing smarter retiring of subjects going by the number of classifications completed/total number of needed assuming  a constant retirement rate will be wrong. So going by the retired/subjects will work better for those projects as well.

marten:
I'm happy with this. "Number of classifications needed" means less and less as we get more intelligent retirement anyways.

PR #2016, opened 2016-10-26
smarter fallback subject selection when using cellect

camallen:
I'm not sure this is the best solution as it could swamp the database. 

The reason i'm proposing this is that cellect can be out of sync re loaded data so try and return useful data. This happened today with a project, for some reason cellect wasn't told / hadn't reloaded the new data that was linked up (not sure why) but there was plenty of data left to classify and the fallback selector was just showing any data...sad times. 
# Review checklist

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -38,7 +38,19 @@ module Subjects
       fallback_selector = PostgresqlSelection.new(workflow, user, opts)
 
       sms_ids = []
-      sms_ids = fallback_selector.select if workflow.using_cellect?
+      if workflow.using_cellect?
+        sms_ids = fallback_selector.select
+        if data_available = !sms_ids.empty?
+          ReloadCellectWorker.perform_async(workflow.id)
+          Honeybadger.notify(
+            error_class:   "Cellect data sync error",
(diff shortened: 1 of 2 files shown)


camallen:
@marten i'd appreciate your thoughts on this.

lib/subjects/selector.rb
```diff
@@ -35,7 +35,16 @@ def active_subjects_in_selection_order(sms_ids)
 
     def fallback_selection
       opts = { limit: subjects_page_size, subject_set_id: subject_set_id }
-      PostgresqlSelection.new(workflow, user, opts).any_workflow_data
+      fallback_selector = PostgresqlSelection.new(workflow, user, opts)
+
+      sms_ids = []
+      sms_ids = fallback_selector.select if workflow.using_cellect?
```
marten: Yeah, but it'll be hard not to just start swamping Cellect with reloads near the actual end of a project.

marten:
I'd be okay with this. Maybe we can get notifications when Cellect returned `[]` while a subsequent `selector.select()` did return some data?

camallen:
Updated to reload cellect and report to honeybadger when this is occurring. We can turn the cellect reload off if it's causing issues but lets see how it goes.

PR #2017, opened 2016-10-26
eager & preload project data for the serialized relations

camallen:
Avoid N+1 queries in the project serializer. 

I chose to use the `scope_for` to pre/eager load for projects as that is how data is loaded by default in the controller. I had to take care with the owner relation to access the relation that was preloading correctly. Also had to modify the parental controlled `scope_for` to not attempt eager loading and attempting to join with conflicting parent relations.

# Review checklist


app/models/project.rb
```diff
@@ -92,6 +106,27 @@ class Project < ActiveRecord::Base
...
 
+  def self.scope_for(action, user, opts={})
+    experiment_name = "eager_load_projects"
+    scope = super
+    CodeExperiment.run "#{experiment_name}" do |e|
+      e.run_if { Panoptes.flipper[experiment_name].enabled? }
+      e.context user: user
+      e.use { scope }
+      e.try do
+        if opts[:skip_eager_load]
```
camallen: I can see the clarity of using a keyword arg but moving to a keyword arg means i have to modify the method sigs of a lot of classes that don't use the eager load code (yet). The options hash currently allows to modify runtime behaviour without touching method sigs.

If we refactor all the controlled resources to have eager loading then yeah we should refactor the method sigs as well. 

I'll update to use the positive voice.

marten: There's already other overrides that use keyword arguments btw. But I'm 2/10.

PR #2020, opened 2016-10-31
handle incorrectly formatted annotation values in classification dumps

camallen:
Return the raw data instead of crashing on invalid annotation values. Also removed unnecessary db lookups in the annotation cache as we had the resource already.

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -149,9 +149,6 @@ module Formatter
             translate(answer_string)
           rescue TypeError
             "unknown answer label"
-          rescue NoMethodError => e
-            report_to_honey_badger(answer_idx)
-            raise e
           end
         end
       end
@@ -195,22 +192,6 @@ module Formatter
          key == annotation["task"]
         end
         @task = task_annotation.try(:last) || {}
-      end
-
-      def report_to_honey_badger(answer_idx)
-        Honeybadger.notify(
-          error_class:   "Task export error",
-          error_message: "The task cannot be exported",
-          context: {
-            classification: @classification.id,
-            annotation: @annotation,
-            workflow_at_version: workflow_at_version.id,
-            workflow_version: workflow_version,
-            content_version: content_version,
-            current_task: @current['task'],
-            answer_idx: answer_idx
-          }
-        )
       end
     end
   end
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -134,11 +134,8 @@ module Formatter
       end
 
       def tool_label(task_info, tool)
-        begin
-          tool_index = tool["tool"]
-        rescue TypeError => e
-          raise ToolFormattingError.new(e.message)
-        end
+        raise ToolFormattingError unless tool.is_a?(Hash)
+        tool_index = tool["tool"]
         have_tool_lookup_info = !!(task_info["tools"] && tool_index)
         known_tool = have_tool_lookup_info && task_info["tools"][tool_index]
         translate(known_tool["label"]) if known_tool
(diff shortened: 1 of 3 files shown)


lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -33,10 +35,16 @@ def drawing(task_info=task)
...
-            tool_label = tool_label(task_info, drawn_item["tool"])
-            drawn_item.merge("tool_label" => tool_label)
+
+          value_with_tool = begin
+            Array.wrap(@current["value"]).map do |drawn_item|
+              tool_label = tool_label(task_info, drawn_item)
+              drawn_item.merge("tool_label" => tool_label)
+            end
+          rescue ToolFormattingError => e
+            @current["value"]
```
marten: Should we do any error reporting besides making sure we can continue the export?

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -125,7 +133,12 @@ def task_label(task_info)
         translate(task_info["question"] || task_info["instruction"])
       end
 
-      def tool_label(task_info, tool_index)
+      def tool_label(task_info, tool)
+        begin
+          tool_index = tool["tool"]
+        rescue TypeError => e
```
marten: Are we particularly interested in the specific `e.message` arising from duck typing in this case. If not, perhaps `raise ToolFormattingError unless Hash === tool` as the first line of the method would be more readable.

camallen: No - it's really a this isn't a hash error only. Will update.

lib/formatter/csv/annotation_for_csv.rb
marten: Mostly I was wondering if we're not going to use exceptions for that purpose, we should probably not use them for flow controlish things either, and rather just have tool_label return nil in case of incompatible config style.

camallen:
this fixes a bug in the dump workers anyone want to sign off on this?

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -125,7 +131,8 @@ def task_label(task_info)
         translate(task_info["question"] || task_info["instruction"])
       end
 
-      def tool_label(task_info, tool_index)
+      def tool_label(task_info, tool)
+        tool_index = tool["tool"]
         have_tool_lookup_info = !!(task_info["tools"] && tool_index)
         known_tool = have_tool_lookup_info && task_info["tools"][tool_index]
         translate(known_tool["label"]) if known_tool
```
zwolf: What does this return if the tool is unknown (but still a hash) and the translation doesn't happen? `False`, since `known_tool` was the last thing computed?

camallen: this will return nil if `known_tool` is nil

camallen: and nil if the `known_tool = {}` since translate will just do a lookup on a nil string.

zwolf: If the hash is correctly formatted but the value is not translatable, will then `nil` be used as the tool_label value instead of the raw value?

camallen: That case should not happen as we have a correct tool label (it matches a lookup value in the task tools) and those tool labels should have been setup correctly when creating the tasks object for the workflow. 

If the tool label does not match up to a know index / tool then it'll return nil and no tool label will be recorded in the annotation. Basically if it's a weird annotation and the tool label doesn't match then we don't try and make a guess. I could return something like 'unknown_tool' but i don't think that helps the researcher (no traceability) and it just bloats the export.

zwolf:
LGTM

PR #2021, opened 2016-10-31
Organizations

zwolf:
Probably a WIP PR, but I'd like some direction on where I should take this next and what needs to be done before it's safe to merge so that the PFE folks can start working on the UI component. 

Organization models (with content), controllers, and ACL associations are in there and are specced out. What's next?

# Review checklist

### DIFF
--- a/app/controllers/api/v1/organization_contents_controller.rb
+++ b/app/controllers/api/v1/organization_contents_controller.rb
@@ -1,0 +1,12 @@
+class Api::V1::OrganizationContentsController < Api::ApiController
+  include Versioned
+
+  require_authentication :all, scopes: [:project]
+  resource_actions :default
+  schema_type :strong_params
+
+  allowed_params :create, :language, :title,
+    *Api::V1::OrganizationsController::CONTENT_PARAMS, links: [:organization]
+
+  allowed_params :update, :title, *Api::V1::OrganizationsController::CONTENT_PARAMS
+end
--- a/app/controllers/api/v1/organization_contents_controller.rb
+++ b/app/controllers/api/v1/organization_contents_controller.rb
@@ -1,12 +1,11 @@
 class Api::V1::OrganizationContentsController < Api::ApiController
   include Versioned
 
-  require_authentication :all, scopes: [:project]
+  require_authentication :all, scopes: [:organization]
   resource_actions :default
   schema_type :strong_params
 
-  allowed_params :create, :language, :title,
-    *Api::V1::OrganizationsController::CONTENT_PARAMS, links: [:organization]
+  allowed_params :create, :language, *Api::V1::OrganizationsController::CONTENT_PARAMS, links: [:organization]
 
-  allowed_params :update, :title, *Api::V1::OrganizationsController::CONTENT_PARAMS
+  allowed_params :update, *Api::V1::OrganizationsController::CONTENT_PARAMS
 end
--- a/app/controllers/api/v1/organizations_controller.rb
+++ b/app/controllers/api/v1/organizations_controller.rb
@@ -1,0 +1,35 @@
+class Api::V1::OrganizationsController < Api::ApiController
+  include FilterByOwner
+  include FilterByCurrentUserRoles
+  # include TranslatableResource
+  include IndexSearch
+  include AdminAllowed
+
+  require_authentication :update, :create, :destroy, scopes: [:organization]
+
+  resource_actions :show, :index, :create, :update, :deactivate
+  schema_type :json_schema
+
+  prepend_before_action :require_login,
+    only: [:create, :update, :destroy]
+
+  def create
+    organizations = Organization.transaction do
+      Array.wrap(params[:organizations]).map do |organization_params|
+        Organizations::Create.with(api_user: api_user).run!(organization_params)
+      end
+    end
+
+    created_resource_response(organizations)
+  end
+
+  def update
+    Organization.transaction do
+      Array.wrap(resource_ids).zip(Array.wrap(params[:organizations])).map do |organization_id, organization_params|
+        Organizations::Update.with(api_user: api_user, id: organization_id).run!(organization_params)
+      end
+    end
+
+    updated_resource_response
+  end
+end
--- a/app/controllers/api/v1/organizations_controller.rb
+++ b/app/controllers/api/v1/organizations_controller.rb
@@ -32,4 +32,14 @@ class Api::V1::OrganizationsController < Api::ApiController
 
     updated_resource_response
   end
+
+  def destroy
+    Organization.transaction do
+      Array.wrap(resource_ids).zip(Array.wrap(params[:organizations])).map do |organization_id, organization_params|
+        Organizations::Destroy.with(api_user: api_user, id: organization_id).run!
+      end
+    end
+
+    deleted_resource_response
+  end
 end
--- a/app/controllers/api/v1/organizations_controller.rb
+++ b/app/controllers/api/v1/organizations_controller.rb
@@ -1,7 +1,6 @@
 class Api::V1::OrganizationsController < Api::ApiController
   include FilterByOwner
   include FilterByCurrentUserRoles
-  # include TranslatableResource
   include IndexSearch
   include AdminAllowed
 
@@ -12,6 +11,10 @@ class Api::V1::OrganizationsController < Api::ApiController
 
   prepend_before_action :require_login,
     only: [:create, :update, :destroy]
+
+  CONTENT_PARAMS = [:description,
+                    :title,
+                    :introduction].freeze
 
   def create
     organizations = Organization.transaction do
--- a/app/controllers/concerns/filter_by_current_user_roles.rb
+++ b/app/controllers/concerns/filter_by_current_user_roles.rb
@@ -5,6 +5,8 @@ module FilterByCurrentUserRoles
     before_action :add_roles_to_filter_params!, only: :index
   end
 
+  # Filter controlled_resources by what role you have onto them,
+  # e.g., "only show me projects that I am a collaborator in".
   def add_roles_to_filter_params!
     roles_filter = params.delete(:current_user_roles).try(:split, ",")
     if !roles_filter.blank? && api_user.logged_in?
--- a/app/models/concerns/translated_content.rb
+++ b/app/models/concerns/translated_content.rb
@@ -3,7 +3,7 @@ module TranslatedContent
   include RoleControl::ParentalControlled
 
   included do
-    has_paper_trail skip: [:langauge]
+    has_paper_trail skip: [:language]
     validates :language, format: {with: LanguageValidation.lang_regex}
     belongs_to translated_for
 
--- a/app/models/concerns/translated_content.rb
+++ b/app/models/concerns/translated_content.rb
@@ -3,7 +3,7 @@ module TranslatedContent
   include RoleControl::ParentalControlled
 
   included do
-    has_paper_trail skip: [:langauge]
+    has_paper_trail ignore: [:language]
     validates :language, format: {with: LanguageValidation.lang_regex}
     belongs_to translated_for
 
--- a/app/models/concerns/translated_content.rb
+++ b/app/models/concerns/translated_content.rb
@@ -3,7 +3,7 @@ module TranslatedContent
   include RoleControl::ParentalControlled
 
   included do
-    has_paper_trail skip: [:language]
+    has_paper_trail skip: [:langauge]
     validates :language, format: {with: LanguageValidation.lang_regex}
     belongs_to translated_for
 
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1,0 +1,24 @@
+class Organization < ActiveRecord::Base
+  include RoleControl::Owned
+  include RoleControl::Controlled
+  include Linkable
+  include Translatable
+
+  scope :public_scope, -> { where.not(listed_at: nil) }
+  scope :private_scope, -> { where(listed_at: nil) }
+
+  has_many :projects
+  has_many :organization_contents, dependent: :destroy
+  has_many :acls, class_name: "AccesControlList", as: :resource, dependent: :destroy
+
+  accepts_nested_attributes_for :organization_contents
+
+  can_by_role :destroy, :update, :update_links, :destroy_links, roles: [ :owner, :collaborator ]
+
+  can_by_role :show, :index, :versions, :version, public: true,
+    roles: [ :owner, :collaborator, :tester, :translator, :scientist, :moderator ]
+
+  can_be_linked :project, :scope_for, :update, :user
+  can_be_linked :access_control_list, :scope_for, :update, :user
+
+end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -19,6 +19,8 @@ class Organization < ActiveRecord::Base
   can_by_role :show, :index, :versions, :version, public: true,
     roles: [ :owner, :collaborator, :tester, :translator, :scientist, :moderator ]
 
+  can_by_role :translate, roles: [ :owner, :translator, :collaborator ]
+
   can_be_linked :project, :scope_for, :update, :user
   can_be_linked :access_control_list, :scope_for, :update, :user
 
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1,6 +1,7 @@
 class Organization < ActiveRecord::Base
   include RoleControl::Owned
   include RoleControl::Controlled
+  include Activatable
   include Linkable
   include Translatable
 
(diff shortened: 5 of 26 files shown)


zwolf:
I added a spec for OrganizationContent and now I've got some specific questions:

- The ProjectContent spec indicates that translatables should not version track changes to language. However, that spec was only passing because the language in the factory and the "new_lang" were the same, both "en". Changing the language on a ProjectContent does trigger a version save and both specs are now failing.

- Possibly related: the line `has_paper_trail skip: [:language]` in models/concerns/translated_content.rb misspelled the word "language". This didn't seem to affect the outcome of the specs, though it seems like it should have.

- The ProjectContent model does not include `belongs_to :project`. Is that correct? It def has a foreign key assigned so it seems like it's missing. I left it off of the org content model just to keep them similar for now. 

Thanks!

camallen:
@zwolf i think points 1 & 2 are a bug..be good to extract these from this PR and fix that. 

Re point 3, the `ProjectContent` model does define the `belongs_to` through the `TranslatedContent` module [here](https://github.com/zooniverse/Panoptes/blob/922b5d0548a9362584d0ce592089089d75d8f6d7/app/models/concerns/translated_content.rb#L8)

app/controllers/api/v1/organizations_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+        Organizations::Create.with(api_user: api_user).run!(organization_params)
+      end
+    end
+
+    created_resource_response(organizations)
+  end
+
+  def update
+    Organization.transaction do
+      Array.wrap(resource_ids).zip(Array.wrap(params[:organizations])).map do |organization_id, organization_params|
```
camallen: Pretty sure our v1 routes require an id for update / delete operations, e.g. `/organisations/1`. I think ths means that `resource_ids == resource_id` and so we can probably clean up the outer loop on this.

app/controllers/api/v1/organizations_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+        Organizations::Update.with(api_user: api_user, id: organization_id).run!(organization_params)
+      end
+    end
+
+    updated_resource_response
+  end
+
+  def destroy
+    Organization.transaction do
+      Array.wrap(resource_ids).zip(Array.wrap(params[:organizations])).map do |organization_id, organization_params|
```
camallen: same as the update above with the outer loop.

app/controllers/concerns/filter_by_current_user_roles.rb
```diff
@@ -5,6 +5,8 @@ module FilterByCurrentUserRoles
     before_action :add_roles_to_filter_params!, only: :index
   end
 
+  # Filter controlled_resources by what role you have onto them,
```
camallen: I'd prefer to rename the method to make this more implicit than add the comments...but i'm like a 2/10 on this.

app/models/concerns/translated_content.rb
```diff
@@ -3,7 +3,7 @@ module TranslatedContent
   include RoleControl::ParentalControlled
 
   included do
-    has_paper_trail skip: [:langauge]
+    has_paper_trail skip: [:language]
```
camallen: I think we should fix this in another PR unless you need this here.

app/models/organization.rb
```diff
@@ -0,0 +1,25 @@
+class Organization < ActiveRecord::Base
+  include RoleControl::Owned
+  include RoleControl::Controlled
+  include Activatable
+  include Linkable
+  include Translatable
+
+  scope :public_scope, -> { where.not(listed_at: nil) }
```
camallen: you got indexes on these attribute scopes?

app/operations/organizations/destroy.rb
```diff
@@ -0,0 +1,12 @@
+module Organizations
+  class Destroy < Operation
+    integer :id
+
+    def execute
+      Organization.transaction do
+        organization = Organization.find(id)
+        Activation.disable_instances!([organization])
```
camallen: the activation module just adds a wrapper around calling `.disable!` and is generalized in the `DeactivatableResource` module, you could just do `Organization.find(id).disable!` here instead.

app/controllers/api/v1/organization_contents_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  include Versioned
+
+  require_authentication :all, scopes: [:project]
+  resource_actions :default
+  schema_type :strong_params
+
+  allowed_params :create, :language, :title,
+    *Api::V1::OrganizationsController::CONTENT_PARAMS, links: [:organization]
+
+  allowed_params :update, :title, *Api::V1::OrganizationsController::CONTENT_PARAMS
```
camallen: `Api::V1::OrganizationsController::CONTENT_PARAMS` is this constant defined?

app/operations/organizations/create.rb
```diff
@@ -0,0 +1,21 @@
...
+    string :title
+    string :description
+    string :introduction, default: ''
+
+    def execute
+      Organization.transaction do
+        organization = Organization.new owner: api_user.user, name: name, display_name: display_name, primary_language: primary_language
+        organization.organization_contents.build title: title, description: description, introduction: introduction, language: primary_language
+        organization.save!
+        organization
```
camallen: Out of interest, do you have to return the instance here for use in the response object?

app/serializers/organization_serializer.rb
```diff
@@ -0,0 +1,32 @@
...
+
+ def content
+   @content ||= _content
+ end
+
+ def fields
+   %i(title description introduction)
+ end
+
+ def _content
```
camallen: should we extract this common method (project / org serializer) to a module?

spec/controllers/api/v1/organizations_controller_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+          owned_unlisted_organization.save
+        end
+
+        it 'returns unlisted organizations that I own' do
+          default_request scopes: scopes, user_id: authorized_user.id
+          get :index
+          expect(json_response['organizations'].map { |o| o['id'] }).to include(owned_unlisted_organization.id.to_s)
+        end
+
+        it "doesn't return unlisted organizations for unauthorized users" do
```
camallen: shouldn't this spec belong on the non-logged in user context?

app/controllers/api/v1/organizations_controller.rb
```diff
@@ -0,0 +1,45 @@
+class Api::V1::OrganizationsController < Api::ApiController
+  include FilterByOwner
+  include FilterByCurrentUserRoles
+  # include TranslatableResource
```
zwolf: ✔️

app/controllers/api/v1/organizations_controller.rb
zwolf: @marten, what do you think? It was explained to me that the schema says the API should support an array of params, but it isn't totally clear to me from the routes. Is this to maintain the standard, is this an upgrade, or is the wrap n' zip here unnecessary?

app/controllers/concerns/filter_by_current_user_roles.rb
zwolf: I'm inclined to leave it, I don't have the aversion to comments that the rest of the crew seems to feel. @marten wrote this one, anyway. :D

app/models/concerns/translated_content.rb
zwolf: This typo fix doesn't seem to fix the bug, anyway, and it'd feel weird to revert to an incorrect spelling. I'll create an issue for the versioning bug.

camallen: agree it's weird not to change it while you can but i think it's out of scope for this PR. Do the tests fail if you revert it?. 5/10 on this one.

app/models/organization.rb
zwolf: ✔️

app/operations/organizations/create.rb
zwolf: I think so, since #save! returns a boolean. The return object should be an org (or an array thereof) to pass to `created_resource_response`.

app/serializers/organization_serializer.rb
zwolf: ✔️ , noticed that I didn't add a spec for the org serializer, but I'm not sure that's necessary as it just now doesn't include anything unexpected.

spec/controllers/api/v1/organizations_controller_spec.rb
zwolf: Goal was to ensure that logged in users didn't see other users' unlisted orgs. Fixed.

app/models/concerns/translated_content.rb
zwolf: ✔️

app/controllers/api/v1/organization_contents_controller.rb
camallen: seems title is defined in the CONST and the dsl here.

app/operations/organizations/create.rb
camallen: Right - thought so. I'm wondering why create & update operations differ in this regard. Seems like the orgs update action relies on the controlled_resources method and not the operation results. Might want to check how that is setup.

db/migrate/20161101175911_add_organization_listed_at_index.rb
```diff
@@ -0,0 +1,10 @@
+class AddOrganizationListedAtIndex < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    add_index :organizations,
+    :listed_at,
+    where: "(listed_at IS NOT NULL)",
+    algorithm: :concurrently
```
camallen: The index can be made with a lock on the table as it's is a brand new table (no data). 

You should double check that both the scope where clauses satisfy this partial index clause otherwise one may be a table scan (probs not a big deal right now though). If it's not a partial index then you could also extract the index def to the table migration (i think).

spec/controllers/api/v1/organizations_controller_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  let(:authorized_user) { create(:user) }
+  let(:organization) { build(:organization, listed_at: Time.now, owner: authorized_user) }
+  let(:unlisted_organization) { build(:organization, listed_at: nil) }
+
+  let(:scopes) { %w(public organization) }
+
+  describe "when not logged in" do
+    describe "#index" do
+
+      it 'returns only the listed organization' do
```
camallen: i wondered if we should be testing that a non-logged in can't access non-public resources? You may have covered this with the unauthorized user though.

app/controllers/api/v1/organizations_controller.rb
camallen: Yeah - it was a JSON-API spec requirement from memory. We do have [code](https://github.com/zooniverse/Panoptes/blob/922b5d0548a9362584d0ce592089089d75d8f6d7/lib/role_control/roled_controller.rb#L44) to handle the multi ids, however I'm still not sure they make it to the controller layer from the routes we've specified, e.g.
`PUT  |  /api/users/:id  |  api/v1/users#update  |  {:id=>/[0-9]*/}`

Agree we could look to take these out but not for this PR, so happy for this to go through as is and we'll revisit later.

zwolf: gonna call this a ✔️  then

app/operations/organizations/create.rb
zwolf: I presume that this is due to the fact that `created_resource_response` takes the created object as an argument and grabs its updated_at, while `updated_resource_response` does not bother and just goes for the max: https://github.com/zooniverse/Panoptes/blob/master/app/controllers/concerns/json_api_responses.rb

Changing the way this works seems OOS for this PR.

app/controllers/api/v1/organization_contents_controller.rb
zwolf: ✔️  again

db/migrate/20161101175911_add_organization_listed_at_index.rb
zwolf: Removed `algorithm :concurrently` and moved index definition to table migration.

As far as the rest goes:

```
[4] pry(main)> Organization.where(listed_at: nil).explain
  Organization Load (3.3ms)  SELECT "organizations".* FROM "organizations" WHERE "organizations"."listed_at" IS NULL
=> EXPLAIN for: SELECT "organizations".* FROM "organizations" WHERE "organizations"."listed_at" IS NULL
                           QUERY PLAN
----------------------------------------------------------------
 Seq Scan on organizations  (cost=0.00..14.20 rows=2 width=160)
   Filter: (listed_at IS NULL)
(2 rows)

CREATE INDEX index_organizations_on_listed_at ON organizations USING btree (listed_at) WHERE (listed_at IS NOT NULL);
```

I think that should be an index scan instead of a seq scan, but that's how all the other indices are being built. What would you change?

app/operations/organizations/destroy.rb
zwolf: As indicated, I reverted this change since it broke the shared spec. Updating this will be included when we have to rewrite all the controller specs during the Great Controller Refactoring that's perpetually just beyond the horizon.

spec/controllers/api/v1/organizations_controller_spec.rb
zwolf: I added an extra expect line to make sure the non-logged in user can't see an unlisted org which is unowned by them. Same spec since it's testing the same thing, that only listed orgs are returned.

app/operations/organizations/create.rb
camallen: Good points - i think it's me not reasoning well, and the created routes will return the newly minted objects...update already has them.

app/operations/organizations/destroy.rb
camallen: seems fair if it meets the shared helper.

db/migrate/20161017135917_create_organizations.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :display_name
+      t.string :slug, default: ""
+      t.string :primary_language, null: false
+      t.timestamp :listed_at, default: nil
+      t.integer  :activated_state, default: 0, null: false
+
+      t.timestamps null: false
+    end
+
+    add_index :organizations, :listed_at, where: "(listed_at IS NOT NULL)"
```
camallen: not sure this partial index will satisfy both scope defs and thus be used for both, you might want to test it out using a normal btree index (no partial) in your test env without sequential scans disabled.

zwolf:
The final non-passing tests are the ones that test if a change to language on project/org contents triggers a version save. The project content spec was changing language from "en" to "en", which is why that was passing before, also. 

However: even after fixing that typo, new versions are created when the language is change and the tests fail. Can you tell why that might be?

db/migrate/20161017135917_create_organizations.rb
zwolf: I don't completely understand what you mean here, we should talk next week

camallen: with that index in place you should be able to run `Organisation.public.explain` & `Organisation.private.explain` and see if they use the indexes you've got. If you're in a testing env then the query planner may want to use a table scan as it'll be cheaper. 

You can force pg to not table scan for the session with `SET enable_seqscan = OFF;`

zwolf:
For some reason, it seems that :skip and :ignore do the opposite of what the docs say they should do. When I changed the `skip` in `has_paper_trail skip: [:language]` to `ignore`, it began working as expected. There's a few old issues on their repo about this kind of thing, but it doesn't seem to be outstanding.

camallen:
@zwolf we're probably using an older version, check the gemfile.lock and compare the docs for that version.

zwolf:
According to the v3.0.9 docs (our version), it should work with skip (minus typo). I've been playing with it in the console and using ignore definitely increments versions on update while skip does not. Changing the spec to test `expect{subject.update!(language: new_lang)}.to_not change{subject.versions.count}` instead of that the prev version is nil doesn't help.

db/migrate/20161017135917_create_organizations.rb
zwolf: ✔️ , changed the index to the whole column, not just where listed_at is null.

zwolf:
I realized I had missed the OrganizationContent controller specs, so I modeled them after the ProjectContents specs and they mostly failed. It's varyingly 404s and 422s, which leads me to believe that the permissions aren't cascading from the Organization (via scope) as I expect them to. Spent all day trying to figure out why, is it obvious to someone?

camallen:
@zwolf you were missing the parental controlled (Organization) translation role dsl setup. see https://github.com/camallen/Panoptes/commit/e1b8276838372af63bebbfadd089cdb011ee7f6c

zwolf:
Great, thanks a ton @camallen. Since this was approved a while back, I'll wait for a lgtm to merge this.

PR #2025, opened 2016-11-02
use the reired_at timestamp not created_at

### LINKED ISSUE #2023: Project 2025: retired subject count and % completion incorrectly reported
For the MWP (Project 2025), we are seeing incorrect stats on our workflow stats page.

Retired subject count was approaching 130 but a few days back, it was reset to 3. Just today, the % completion rate was reset to 0% from 17%. 

Anyone know what's going on?

Thanks for your help!

camallen:
fixes #2023 compare apples and apples here as the SubjectWorkflowStatus will be created the first time it's classified not retired.

# Review checklist

### DIFF
--- a/spec/counters/workflow_counter_spec.rb
+++ b/spec/counters/workflow_counter_spec.rb
@@ -30,9 +30,13 @@ describe WorkflowCounter do
     end
 
     context "with workflow counts" do
+      let(:sws_params) do
+        { workflow: workflow, retired_at: DateTime.now, created_at: DateTime.now + 2.days }
+      end
+
       before do
         workflow.subjects.each do |subject|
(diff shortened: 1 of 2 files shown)


spec/counters/workflow_counter_spec.rb
```diff
@@ -30,9 +30,13 @@
     end
 
     context "with workflow counts" do
+      let(:sws_params) do
+        { workflow: workflow, retired_at: DateTime.now, created_at: DateTime.now + 2.days }
```
marten: Perhaps more realistic would be `DateTime.now - 2.days` ;)

camallen: i explicitly want to test that it's not going to catch these based on created at dates

PR #2027, opened 2016-11-02
Include non-enum states in serialized projects

zwolf:
Serializes non-enum (presently `paused` and `finished`) states. If your params include `{ state: [live] }` (or any other boolean on Project, all of which are indexed), your results will be filtered by that attribute. Might have made this _too_ flexible, since any `state` sent that isn't in the enum will be set to `true` and thrown into the serializer, but we'll see if you think so.

# Review checklist

### DIFF
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -21,12 +21,20 @@ class ProjectSerializer
     subjects_export: { include: false },
     aggregations_export: { include: false }
   can_filter_by :display_name, :slug, :beta_requested, :beta_approved,
-    :launch_requested, :launch_approved, :private, :state
+    :launch_requested, :launch_approved, :private, :state, :live
   can_sort_by :launch_date, :activity, :completeness, :classifiers_count,
     :updated_at, :display_name
 
   def self.page(params = {}, scope = nil, context = {})
-    params["state"] = Project.states[params["state"]] if params.key?("state")
+    if params.key?("state")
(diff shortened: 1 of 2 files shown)


app/serializers/project_serializer.rb
```diff
@@ -21,12 +21,20 @@ class ProjectSerializer
...
   can_sort_by :launch_date, :activity, :completeness, :classifiers_count,
     :updated_at, :display_name
 
   def self.page(params = {}, scope = nil, context = {})
-    params["state"] = Project.states[params["state"]] if params.key?("state")
+    if params.key?("state")
+      if Project.states.include?(params["state"])
+        params["state"] = Project.states[params["state"]]
+      else
+        params[params["state"]] = true
```
camallen: I think this could result in unexpected responses from valid query params, e.g. 
`?launch_approved=false&state=launch_approved` 

This would negate the specific query param resulting in `params[launch_approved]=true`

zwolf: it is, w/o a doubt, a hack. If you'd rather, I can add `if project["state"] == "live"` and only replace `params["state"]` with `params["live"] == true`. Less generalized, but the other booleans are all included in `can_filter_by` anyway.

camallen: you make the call

zwolf:
k, I'll make it alter just the "live" attribute for now. Don't much care for the nested conditionals, but I'm fine with this until we can make the full state refactor.

PR #2028, opened 2016-11-04
No more paused projects in index

zwolf:
Allows for the confusing situation where you could have a `live: true` project that was also `state: paused`. This adds a scope to the serializer call to force `Project.where(state: nil)` if `?state=live` is passed as a param. Hacky also, but this is exactly why we need to consolidate all of these.

# Review checklist

### DIFF
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -32,7 +32,7 @@ class ProjectSerializer
       elsif params["state"] == "live"
         params["live"] = true
         params.delete("state")
-        scope = Project.where(state: nil)
+        scope = scope.where(state: nil)
       end
     end
 
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -32,6 +32,7 @@ class ProjectSerializer
       elsif params["state"] == "live"
         params["live"] = true
         params.delete("state")
+        scope = Project.where(state: nil)
       end
     end
 
--- a/spec/serializers/project_serializer_spec.rb
+++ b/spec/serializers/project_serializer_spec.rb
@@ -34,7 +34,8 @@ describe ProjectSerializer do
     end
 
     describe 'can filter by state' do
-      let(:paused_project) { create(:full_project, state: "paused", live: true) }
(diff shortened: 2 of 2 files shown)


app/serializers/project_serializer.rb
```diff
@@ -32,6 +32,7 @@ def self.page(params = {}, scope = nil, context = {})
       elsif params["state"] == "live"
         params["live"] = true
         params.delete("state")
+        scope = Project.where(state: nil)
```
camallen: don't reassign the scope..instead chain it, `scope = scope.where(state:nil)`

spec/serializers/project_serializer_spec.rb
```diff
@@ -34,7 +34,7 @@
     end
 
     describe 'can filter by state' do
-      let(:paused_project) { create(:full_project, state: "paused", live: false) }
+      let(:paused_project) { create(:full_project, state: "paused", live: true) }
```
camallen: you may want to add a failing test case to ensure the scope re-assignment doesn't pass these tests.

app/serializers/project_serializer.rb
zwolf: The default is nil. Not way into ```        if scope
          scope = scope.where(state: nil)
        else
          scope = Project.where(state: nil)
        end```
What do you suggest?

camallen: the scope is passed into the serializer by the controller and will propogate down, the method signature you've got there is to match the override to inject your behaviour and then call super.

zwolf: Right, but scope isn't defined in the serializer specs. I guess I can just pass it the class to call #where on. Doesn't seem like the default should be nil if it's being altered, what class should scope be?

spec/serializers/project_serializer_spec.rb
zwolf: ✔️

simoneduca:
Is this good to merge?

camallen:
@simoneduca 
> camallen approved these changes 3 days ago

PR #2038, opened 2016-11-11
Added task to copy pan-and-zoom config from projects to workflows

zwolf:
# Review checklist

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -175,7 +175,7 @@ namespace :migrate do
     task :set_pan_and_zoom => :environment do
       workflows = Workflow.joins(:project).where("projects.experimental_tools @> ?", '{pan and zoom}')
       workflows.map do |w|
-        unless w.configuration.key?(:pan_and_zoom)
+        unless w.configuration.key?("pan_and_zoom")
           new_config = w.configuration.merge({'pan_and_zoom': true})
           w.update_column(:configuration, new_config)
         end
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -169,4 +169,13 @@ namespace :migrate do
       puts ' done'
     end
   end
+
+  namespace :projects do
+    desc "Set pan_and_zoom on linked workflows if in experimental_tools array"
+    task :set_pan_and_zoom => :environment do
+      workflows = Workflow.joins(:project).where("projects.experimental_tools @> ?", '{pan and zoom}')
+      workflows.map { |w| w.update_column(:configuration, w.configuration.merge({'pan_and_zoom': true})) unless w.configuration.has_key?(:pan_and_zoom) }
+    end
+  end
+
 end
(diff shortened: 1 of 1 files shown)


lib/tasks/migrate.rake
```diff
@@ -169,4 +169,13 @@ namespace :migrate do
       puts ' done'
     end
   end
+
+  namespace :projects do
+    desc "Set pan_and_zoom on linked workflows if in experimental_tools array"
+    task :set_pan_and_zoom => :environment do
+      workflows = Workflow.joins(:project).where("projects.experimental_tools @> ?", '{pan and zoom}')
+      workflows.map { |w| w.update_column(:configuration, w.configuration.merge({'pan_and_zoom': true})) unless w.configuration.has_key?(:pan_and_zoom) }
```
camallen: i'd break this up for readability but i'm like a 1/10 on this

``` ruby
workflows.map do |w| 
  unless w.configuration.has_key?(:pan_and_zoom) }
    new_config = w.configuration.merge({'pan_and_zoom': true})
    w.update_column(:configuration, new_config)
  end
end
```

zwolf:
Will test this on Monday before setting loose on live data.

zwolf:
Need some input. Testing this code on staging leads to odd diffs between symbols and strings used as keys in the configuration hash. `old_workflow` is the first result that 

irb(main):035:0> workflow_update.configuration
=> {}
irb(main):038:0> new_config = workflow_update.configuration.merge({"pan_and_zoom": true})
=> {:pan_and_zoom=>true}
irb(main):045:0> workflow_update.update_column(:configuration, new_config)
  SQL (2.3ms)  UPDATE "workflows" SET "configuration" = '{"pan_and_zoom":true}' WHERE "workflows"."id" = 2406  [["id", 2406]]
=> true
irb(main):046:0> workflow_update.configuration
=> {:pan_and_zoom=>true}

irb(main):054:0> old_workflow.configuration [PREVIOUSLY SET BY PFE]
=> {"pan_and_zoom"=>true}
irb(main):050:0> workflow_update.configuration
=> {:pan_and_zoom=>true}
irb(main):052:0> old_workflow.configuration == workflow_update.configuration
=> false

Sorry about formatting, code blocks are apparently unusable in comments. But you get the idea: is this a problem for PFE? What did we do last time to get it in there correctly, @camallen, or is that sufficient?

camallen:
@zwolf seems like you're not reloading the data from AR so it's not stringified the jsonb hash keys. I can't find the link right now but postgres will convert the symbol keys to strings so reloading the data will fix what your are seeing. Try: `old_workflow.configuration == workflow_update.reload.configuration`

zwolf:
OK, that makes sense. In that case, existing configs need to be searched by string instead of by symbol, so here's an update.

camallen:
Re https://github.com/zooniverse/Panoptes-Front-End/pull/3105#issuecomment-260708257 so this doesn't need to be a rake task anymore?

zwolf:
Depends on when that PR finally gets merged. Happy to let this drop if it's never going to be an issue again.

camallen:
Fancy pushing the merge on the front end repo and making sure the data is up to date in the backend? That way we can close this and the ui : backend should be in sync.

zwolf:
It was merged, so I'm going to just close this as this should never be necessary again.

PR #2039, opened 2016-11-11
Rework the project / workflow eager load experiments

camallen:
Actually test the serialized response objects using eager / preloading. Move from the scope_for to the serializer and add experimental testing in them.

Project eager loading queries were stupidly big on my dev machine ~500ms per serialized page so i've dropped these and will try the owner / roles preload (not much difference on dev). Workflows shows as speed up though with a warm cache so this could be good. 

Also added a migration to ensure the workflow / workflow_contents have their version number cached on the model to avoid these lookups on the fly. 

# Review checklist


db/structure.sql
```diff
@@ -2218,13 +2218,27 @@ CREATE INDEX index_collections_display_name_trgrm ON collections USING gin ((COA
 
 
 --
+-- Name: index_collections_on_activated_state; Type: INDEX; Schema: public; Owner: -; Tablespace: 
+--
+
+CREATE INDEX index_collections_on_activated_state ON collections USING btree (activated_state);
```
camallen: I've borked these...got a fix coming in

PR #2057, opened 2016-11-23
Use custom logger object for logging, forego Rails integration for now

marten:
Removes the Rails part of SemanticLogger. Sets up a completely separate logger that will log to Graylog only. Add module that you can include if you want to get access to the logger.

This restores the controller/action as logged by logstasher.

# Review checklist


lib/logging.rb
```diff
@@ -0,0 +1,5 @@
+module Logging
+  def eventlog
```
camallen: Should this be a class method instead of a require and instance method call?

Issue #1841, opened 2016-06-24
subject_ids column in classification exports reporting double ids


vrooje:
First, many thanks for reporting `subject_ids` as a column a-la #1637 - it's much appreciated.

I did notice that for my project (id=3) where each classification only deals with 1 subject, it's reporting the subject id twice in that column:
![screen shot 2016-06-23 at 6 16 40 pm](https://cloud.githubusercontent.com/assets/2357126/16325053/5bc934e8-396f-11e6-8b1a-ccbd1a2c0ee4.jpg)

I can easily work around this but just wanted to note it.

marten:
Interesting, I had noticed that with @chrislintott's .astro hack day project but since they were submitting classifications via custom API calls I assumed the error was there. I'll investigate.

marten:
Awesome. We're bitten by one of those Rails features everyone always says you shouldn't use: `default_scope`. We automatically do an OUTER JOIN between subjects and their images, so subjects with multiple images will be reported multiple times.

camallen:
urgh - linking #1614

PR #2060, opened 2016-11-23
remove subjects default scope

### LINKED ISSUE #1614: remove subject.rb default scope
see if we can get rid of this...it complicates query plans on all sorts of associations where it's not used.

camallen:
closes #1614 #1841 

don't conflate the subject queries with eager loads on media resources

# Review checklist

### DIFF
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -3,8 +3,6 @@ class Subject < ActiveRecord::Base
   include Linkable
   include Activatable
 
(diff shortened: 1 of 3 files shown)


marten:
There wasn't anywhere where we actually needed that eager load?

camallen:
Not for the tests ;) - i'm going to open a follow up PR for eager/preloading in the subject serializer.

PR #2061, opened 2016-11-25
Subject remove default scope and optimize the selector scope queries

### LINKED ISSUE #1614: remove subject.rb default scope

camallen:
supersedes #2060 and closes #1614 #1841 

- Remove the Subject locations default scope
- Add public/private & active scope indexes for controlled resources
- Add preloading to the subject serializer to avoid N+1 lookups and added the ability to turn off extra selection lookups (seen / retired / user_finished) in the serializer for peak load events (SGL) to avoid multiple extra db queries.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -72,10 +72,7 @@ class Api::V1::SubjectsController < Api::ApiController
   end
 
   def selector
-    @selector ||= Subjects::Selector.new(api_user.user,
-      workflow,
-      params,
-      Subject.all)
+    @selector ||= Subjects::Selector.new(api_user.user, workflow, params)
   end
 
   def location_params(locations)
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -75,7 +75,7 @@ class Api::V1::SubjectsController < Api::ApiController
     @selector ||= Subjects::Selector.new(api_user.user,
       workflow,
       params,
-      controlled_resources)
+      Subject.all)
   end
 
   def location_params(locations)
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/subjects_controller.rb
```diff
@@ -75,7 +75,7 @@ def selector
     @selector ||= Subjects::Selector.new(api_user.user,
       workflow,
       params,
-      controlled_resources)
+      Subject.all)
```
camallen: Do we need to pass a scope at all here? I think we can remove this from the interface.

lib/subjects/selector.rb
```diff
@@ -104,15 +107,19 @@ def subjects_page_size
...
-      }
+      @context ||= { workflow: workflow, user: user }
     end
 
     def subject_set_id
       params[:subject_set_id]
     end
+
+    def subject_selection_context
+      if Panoptes.flipper[:skip_subject_selection_context].enabled?
```
camallen: This will make a db call per selection request, we can look at adding a redis cache layer in to Flipper as well if this is too costly.

app/serializers/subject_serializer.rb
```diff
@@ -9,6 +9,11 @@ class SubjectSerializer
 
   can_include :project, :collections, :subject_sets
 
+  def self.page(params = {}, scope = nil, context = {})
+    scope = scope.preload(:locations, :project, :collections, :subject_sets)
```
camallen: Do we need this context for the subject selector results? Should we use a serializer that doesn't link these associations e.g. cards serializer for projects..?

marten: Can we check if those relations were actually requested to be included?

marten: Yeez. Let's not do that for API v2.

PR #2062, opened 2016-11-28
Add function to respond to user removal requests

marten:
Can someone sanity-check if there might be more places we need to strip information?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -344,4 +344,20 @@ class User < ActiveRecord::Base
   def subject_count_cache_expiry
     ENV.fetch("UPLOADED_SUBJECTS_COUNT_CACHE_EXPIRY", 1.hour)
   end
+
+  def anonymize
+    self.email = "noreply-#{SecureRandom.hex(4)}@zooniverse.org"
+    self.current_sign_in_ip = nil
+    self.last_sign_in_ip = nil
+    self.display_name = self.id
+    self.login = self.id
+    self.credited_name = nil
+    self.password = self.password_confirmation = SecureRandom.hex(16)
+    self.global_email_communication = false
+    self.project_email_communication = false
+    self.beta_email_communication = false
+    self.api_key = nil
+    self.tsv = nil
+    self
+  end
 end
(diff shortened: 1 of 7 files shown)


app/models/user.rb
```diff
@@ -318,4 +318,20 @@ def no_other_logins_exist?(login)
...
+    self.email = "noreply-#{SecureRandom.hex(4)}@zooniverse.org"
+    self.current_sign_in_ip = nil
+    self.last_sign_in_ip = nil
+    self.display_name = self.id
+    self.login = self.id
+    self.credited_name = nil
+    self.password = self.password_confirmation = SecureRandom.hex(16)
+    self.global_email_communication = false
+    self.project_email_communication = false
+    self.beta_email_communication = false
```
astopy: Could you also set `self.valid_email = false`?

app/models/user.rb
```diff
@@ -318,4 +318,20 @@ def no_other_logins_exist?(login)
   def uploaded_subjects_count
     Subject.unscoped.where(upload_user_id: self.id).count
   end
+
+  def anonymize
+    self.email = "noreply-#{SecureRandom.hex(4)}@zooniverse.org"
+    self.current_sign_in_ip = nil
+    self.last_sign_in_ip = nil
+    self.display_name = self.id
+    self.login = self.id
```
astopy: I know it's unlikely, but what if there's already a user with this login?

app/models/user.rb
astopy: Maybe also set `self.private_profile = true`

app/models/user.rb
```diff
@@ -318,4 +318,20 @@ def no_other_logins_exist?(login)
   def uploaded_subjects_count
     Subject.unscoped.where(upload_user_id: self.id).count
   end
+
+  def anonymize
+    self.email = "noreply-#{SecureRandom.hex(4)}@zooniverse.org"
+    self.current_sign_in_ip = nil
+    self.last_sign_in_ip = nil
+    self.display_name = self.id
```
astopy: This might be visible publicly in some places, so I wonder if something like `"Deleted user #{self.id}"` would be better.

camallen:
https://github.com/zooniverse/Panoptes/blob/master/lib/user_info_scrubber.rb

The original plan was to use the above and write it up to the users#destroy method. We should probably merge these two bits of code

camallen:
Talk comments should stay, we should just show the user name as deleted or the like

app/models/user.rb
marten: Validation and indexes would stop us from saving that. Not too worried about that, unlikely to happen and it's a behind the scenes management function anyways.

PR #2063, opened 2016-11-28
Adds classification export info to serialized workflows

zwolf:
Also adds specs for workflows + media links.

# Review checklist

### DIFF
--- a/spec/controllers/api/v1/media_controller_spec.rb
+++ b/spec/controllers/api/v1/media_controller_spec.rb
@@ -301,8 +301,7 @@ RSpec.describe Api::V1::MediaController, type: :controller do
 
       describe "classifications_exports #index" do
         let!(:resources) do
-          create_list :medium, 2, linked: parent, content_type: "text/csv",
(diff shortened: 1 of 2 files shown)


spec/controllers/api/v1/media_controller_spec.rb
```diff
@@ -297,6 +297,25 @@
       let(:parent) { create(:workflow, project: project) }
 
       it_behaves_like "has_many media", :workflow, :attached_images, %i(index create show destroy), 'image/jpeg'
+      it_behaves_like "has_one media", :workflow, :classifications_export, %i(index), 'text/csv'
+
+      describe "classifications_exports #index" do
+        let!(:resources) do
+          create_list :medium, 2, linked: parent, content_type: "text/csv",
```
camallen: Any reason to build 2 exports here instead of just one?

zwolf: That's what projects were doing before I got there. The both still pass when there's only one, though, so updated.

PR #2064, opened 2016-11-28
Adds researcher_quote to project contents

### LINKED ISSUE #2035: Allow Words from the Researcher in Project Resource
In preparation for the new project landing page, there should be a tag on the project resource that allows project builders to insert text as "Words from the Researcher." This text will be added under the "project details" section of the project builder on the front end. See attached image (lower left) for how the text will later appear on the new landing page.

desktop copy 21

zwolf:
Adds "researcher_quote" field to project contents, including associated serializers and constants. Character min 255. Probably didn't need the index but I thought better to include than not. No default included, nil values should be dealt with on the front end since many projects will not have this quote. That'll also mean we won't have to backfill.

Closes #2035 .

# Review checklist

### DIFF
--- a/db/migrate/20161128193435_add_researcher_quote_to_project_contents.rb
+++ b/db/migrate/20161128193435_add_researcher_quote_to_project_contents.rb
@@ -1,0 +1,7 @@
+class AddResearcherQuoteToProjectContents < ActiveRecord::Migration
+  def change
+    add_column :project_contents, :researcher_quote, :text
+
+    add_index :project_contents, :researcher_quote
+  end
+end
--- a/db/migrate/20161128193435_add_researcher_quote_to_project_contents.rb
+++ b/db/migrate/20161128193435_add_researcher_quote_to_project_contents.rb
@@ -1,7 +1,5 @@
 class AddResearcherQuoteToProjectContents < ActiveRecord::Migration
   def change
-    add_column :project_contents, :researcher_quote, :text
-
-    add_index :project_contents, :researcher_quote
+    add_column :project_contents, :researcher_quote, :text, default: ""
   end
 end
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -18,12 +18,14 @@ class Api::V1::ProjectsController < Api::ApiController
   CONTENT_PARAMS = [:description,
                     :title,
                     :workflow_description,
+                    :researcher_quote,
                     :introduction].freeze
 
   CONTENT_FIELDS = [:title,
                     :description,
                     :workflow_description,
                     :introduction,
+                    :researcher_quote,
                     :url_labels].freeze
 
   CARD_FIELDS = [:id,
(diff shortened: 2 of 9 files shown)


db/migrate/20161128193435_add_researcher_quote_to_project_contents.rb
```diff
@@ -0,0 +1,7 @@
+class AddResearcherQuoteToProjectContents < ActiveRecord::Migration
+  def change
+    add_column :project_contents, :researcher_quote, :text
+
+    add_index :project_contents, :researcher_quote
```
camallen: we won't need this unless we actually query on this directly. indexes take up resources on the database so we should only add them if we have an actual query path that uses them.

db/migrate/20161128193435_add_researcher_quote_to_project_contents.rb
```diff
@@ -0,0 +1,7 @@
+class AddResearcherQuoteToProjectContents < ActiveRecord::Migration
+  def change
+    add_column :project_contents, :researcher_quote, :text
```
camallen: This should have a default value for consistency with the other attributes on that table. We should serialize the empty so we are consistent for all clients (not just pfe use cases)

zwolf: If that's the case, what needs to be done to backfill the data? Can they be left nil or do they need to be backfilled with a blank-string default? If so, can that be done in the migration, or is that too heavy for the db?

camallen: adding the column with a default will backfill, https://www.postgresql.org/docs/9.3/static/ddl-alter.html. However it'll lock the table to add that col and depending on table size it might be a while. If it's too long for normal ops then we'd have to add without default, backfill with rake task and then add default col val.

zwolf: Projects table isn't huge, so I don't think it'll be an issue.

zwolf:
Uh, I don't know. Is that somehow measurable before running the migration? There are only ~3300 records in projects and project_contents each and that doesn't seem large enough to cause a problem.

camallen:
Using a restored backup snapshot db to run the migration for testing against. If that's all there is then it should be pretty quick to add that column and staging would probably be a proxy benchmark to see how long the actual migration runs in.

zwolf:
Before I merge this, then, how do I make sure the migration takes on staging? That part's automatic, right?

camallen:
It'll run on staging auto deploy after you merge. Just lookup the deploy logs to see how long the migration takes.

zwolf:
I don't see any Jenkins jobs for migrating the staging database. That build of the staging AMI took 22 minutes. More and less recent builds took 33, 17, 10, and 14 minutes. If that's the metric to go by, it should be fine.

camallen:
The jenkins staging deploy jobs runs the migration on the new instance. Check these jenkins logs as one of them will include the `rake db:migrate` task output.

zwolf:
`== 20161128193435 AddResearcherQuoteToProjectContents: migrated (0.0890s) =====` 

Doesn't seem like it's gonna be a problem.

PR #2068, opened 2016-11-30
Send emails to the right people

zwolf:
When a user is added to a project as a collaborator or expert, they should get an email, not the person who added them. Just looking for the first person in the ACL's user group, since 100% of the time they're identity groups.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/project_roles_controller.rb
+++ b/app/controllers/api/v1/project_roles_controller.rb
@@ -10,13 +10,13 @@ class Api::V1::ProjectRolesController < Api::ApiController
   end
 
   def update
-    UserAddedToProjectMailerWorker.perform_async(acl_user_id, controlled_resource.resource.id, roles) if new_roles_present?(roles)
     super
+    UserAddedToProjectMailerWorker.perform_async(acl_user_id, controlled_resource.resource.id, roles) if new_roles_present?(roles)
   end
 
   def create
-    UserAddedToProjectMailerWorker.perform_async(new_user_id, project_id, roles) if new_roles_present?(roles)
     super
+    UserAddedToProjectMailerWorker.perform_async(new_user_id, project_id, roles) if new_roles_present?(roles)
   end
 
   private
--- a/app/controllers/api/v1/project_roles_controller.rb
+++ b/app/controllers/api/v1/project_roles_controller.rb
@@ -10,15 +10,31 @@ class Api::V1::ProjectRolesController < Api::ApiController
   end
 
   def update
-    super do
-      UserAddedToProjectMailerWorker.perform_async(api_user.id, controlled_resource.resource.id, roles) if new_roles_present?(roles)
-    end
+    UserAddedToProjectMailerWorker.perform_async(acl_user_id, controlled_resource.resource.id, roles) if new_roles_present?(roles)
+    super
+  end
+
+  def create
+    UserAddedToProjectMailerWorker.perform_async(new_user_id, project_id, roles) if new_roles_present?(roles)
+    super
   end
 
   private
 
   def roles
     params[:project_roles][:roles]
+  end
+
+  def acl_user_id
+    controlled_resource.user_group.users.first.id.to_i
+  end
+
+  def new_user_id
+    params[:project_roles][:links][:user].to_i
+  end
+
+  def project_id
+    params[:project_roles][:links][:project].to_i
   end
 
   def new_roles_present?(roles)
--- a/spec/controllers/api/v1/project_roles_controller_spec.rb
+++ b/spec/controllers/api/v1/project_roles_controller_spec.rb
@@ -74,7 +74,7 @@ RSpec.describe Api::V1::ProjectRolesController, type: :controller do
 
     context "mailers" do
       let(:user) do
-        create(:user).tap do |user|
+        create(:user) do |user|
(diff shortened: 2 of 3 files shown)


app/controllers/api/v1/project_roles_controller.rb
```diff
@@ -10,9 +10,13 @@ def resource_name
   end
 
   def update
-    super do
-      UserAddedToProjectMailerWorker.perform_async(api_user.id, controlled_resource.resource.id, roles) if new_roles_present?(roles)
-    end
+    UserAddedToProjectMailerWorker.perform_async(acl_user_id, controlled_resource.resource.id, roles) if new_roles_present?(roles)
```
camallen: Do we want to send emails before we update / create the resource?

spec/factories/access_control_lists.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryGirl.define do
   factory :access_control_list do
-    user_group
+    association :user_group, factory: :user_group_with_users
```
camallen: I don't think this needs to be the default factory? Best to just build the one you need when you need it. 

For ref, if this is the default, everywhere the `access_control_list` factory is used it'll go an build 2 extra users, identity groups & memberships. This chained factory building just adds time to the spec runs and most likely isn't needed for the rest of the specs.

spec/controllers/api/v1/project_roles_controller_spec.rb
```diff
@@ -73,19 +73,25 @@
     it_behaves_like "is updatable"
 
     context "mailers" do
+      let(:user) do
+        create(:user).tap do |user|
```
camallen: just a note, pretty sure you can just do 

```
create(:user).do |user|
```

app/controllers/api/v1/project_roles_controller.rb
camallen: Also i'm not sure i'm following the `acl_user_id`, don't we just want to send an email to the linked user in the params?

zwolf: First, no. Should come after, but the one in #create doesn't get called in the spec with `super do`. I tried `super do |acl|` as well, since other examples of this behavior in Panoptes passed a block parameter as well, but that #create one still never gets called. Still looking into this.

zwolf: Second, also no. There is no linked user in the params of a call to `/project_roles/:id`. PFE makes a few calls after the event, culminating in a final PUT to the above route with a body containing only an array of roles. Hence the roundabout way of getting at the user ID.

camallen: Got you now on the ACL lookup [this](https://github.com/zooniverse/Panoptes/blob/7c0106054b20b4e13dc8e5f345784d8679b60e58/app/serializers/concerns/acl_serializer.rb#L31) is how we serialize the owner relation for that resource.

camallen: re the yield order, create will run after the [save!](https://github.com/zooniverse/Panoptes/blob/7c0106054b20b4e13dc8e5f345784d8679b60e58/lib/json_api_controller/creatable_resource.rb#L17) but update will run before the [save!](https://github.com/zooniverse/Panoptes/blob/7c0106054b20b4e13dc8e5f345784d8679b60e58/lib/json_api_controller/updatable_resource.rb#L20). 

I'd move the update to be after super to make sure an update error blocks the email going out and leverage the yield block in create. Double check it's not the if statement stopping the code executing though a debug statement in the yielded block should run regardless.

spec/factories/access_control_lists.rb
camallen: It should be that you can add any user_group you want when creating the acl, E.g. 
`create(:access_control_list, user_group: my_group, resource: project, roles: ["collaborator"])`

app/controllers/api/v1/project_roles_controller.rb
zwolf: Ah hah. It looks like there's an exception being raised in CreatableResource that isn't bubbling: `#<ActiveRecord::RecordInvalid: Validation failed: User group . Roles have already been set for this user or group>`. Because I was manually creating the ACLs in the let block. Which of course I wanted--nay, _needed_ to have for #update, but were severely getting in the way on a #create.

spec/factories/access_control_lists.rb
zwolf: Due to the way the factories are currently set up, this results in `ActiveRecord::RecordInvalid: Validation failed: User group . Roles have already been set for this user or group`. As I said, validations are preventing me from adding factory-built users to an existing ACL. Because the actual specs are in the various shared examples, I can't figure out how to define them in such a way that the specs will pass. Building them in the ACL factory works, though, even if it may cause a bit of unnecessary building elsewhere.

camallen: I get they are broken, i just don't want the default factory creating unnecessary DB objects, long term that leads to very slow test suites (slower than what we have already). 

Could you just maybe add another named factory (or a trait) instead of the default that is used everywhere else?

``` ruby
factory :access_control_list_with_user_group do
  association :user_group, factory: :user_group_with_users
end
```

zwolf:
Looks like that was a pretty simple workaround, thought I tried that already. thanks @camallen.

PR #2076, opened 2016-12-05
Adds linkage between orgs and projects

zwolf:
Includes a new join table for maximum flexibility (projects can be in many orgs, orgs can have many projects), and new specs to test the link and the controller method.

# Review checklist

### DIFF
--- a/app/models/organizations_project.rb
+++ b/app/models/organizations_project.rb
@@ -1,4 +1,0 @@
-class OrganizationsProject < ActiveRecord::Base
-  belongs_to :project
-  belongs_to :organization
-end
--- a/app/models/organizations_project.rb
+++ b/app/models/organizations_project.rb
@@ -1,0 +1,4 @@
+class OrganizationsProject < ActiveRecord::Base
+  belongs_to :project
+  belongs_to :organization
+end
--- a/db/migrate/20161205203956_add_organization_id_to_projects.rb
+++ b/db/migrate/20161205203956_add_organization_id_to_projects.rb
@@ -1,0 +1,6 @@
+class AddOrganizationIdToProjects < ActiveRecord::Migration
+  def change
+    add_reference :projects, :organization, index: true
+    add_foreign_key :projects, :organizations
+  end
+end
--- a/db/migrate/20161205203956_create_join_table_organization_project.rb
+++ b/db/migrate/20161205203956_create_join_table_organization_project.rb
@@ -1,8 +1,0 @@
-class CreateJoinTableOrganizationProject < ActiveRecord::Migration
-  def change
-    create_join_table :organizations, :projects do |t|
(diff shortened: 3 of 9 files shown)


zwolf:
Just saw this: https://github.com/zooniverse/Panoptes-Front-End/issues/2948#issuecomment-264976409

So I guess be refactoring to just add an org ID to projects.

db/migrate/20161205203956_create_join_table_organization_project.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableOrganizationProject < ActiveRecord::Migration
+  def change
+    create_join_table :organizations, :projects do |t|
+      t.index [:organization_id, :project_id]
```
camallen: Do you think the query planner use two single column indexes instead of the two compound indexes?

marten:
I'm semi-strongly against turning this into a many-to-many join association. I don't think we'll ever need it, and Chris seemed to agree back in Sept. Do we now have a concrete use case for a project that belongs to many organizations?

In my mind, "this project belongs to the NFN organization" is much simpler than "this project belongs to a list of organizations that includes the NFN organization, and in practise that is the only organization, but in theory it could have many". 

I'm thinking back to all the weird edge-cases we have to explain to any new developer joining our company, where we draw a big lie on a whiteboard and then say, "well, technically a classification can be for multiple subjects, but we've never tested it for that and it will break everywhere". I worry we might get something like this again, where various parts of the UI either need to be complicated because they need to account for the case where a project has many organizations, or they gloss over that and then break if someone comes along and thinks it's already supported. For instance, you couldn't just display "NFN -> Pollinator Plants of Virginia" in a navbar, if it could belong to both NFN and FooBar.

A simple tree structure is easier to reason about that an interconnected graph, and that is a cost we pay until the end of days. Our app is already hard enough, let's keep new stuff as simple as could possibly work.

camallen:
Sounds like a reasoned comment @marten - agree with simplicity but I can envisage the edge case. Most likely this is building stuff we don't need though and happy to defer @zwolf and yourself on this one.

db/migrate/20161205203956_add_organization_id_to_projects.rb
```diff
@@ -0,0 +1,6 @@
+class AddOrganizationIdToProjects < ActiveRecord::Migration
```
zwolf: bad dog, very bad dog

app/models/organizations_project.rb
```diff
@@ -0,0 +1,4 @@
+class OrganizationsProject < ActiveRecord::Base
```
marten: I think this file can be deleted too now right?

marten:
Happy for you to go merge if you remove that left-over file.

PR #2078, opened 2016-12-06
Store duration on classification

camallen:
Calculate (if available) the classification duration during the lifecycle event and store it in a summable first class column on the classification table.

I see this as a building block for an stream processor / offline worker towards https://github.com/zooniverse/rfc/issues/7. Thoughts?

**TODO**
- [ ] Ensure the duration column is included in the internal classification stream fields.
- [ ] Rake task to backfill this info where appropriate

# Review checklist


camallen:
@marten I think only if we want to use those columns in a sum as we can index them in the jsonb object for querying, etc. What do you think?

marten:
We can always do it later.

camallen:
I'm going to close this for now and investigate a better solution via the internal classification stream.

PR #2080, opened 2016-12-07
Collection & UserProjectPreference order by

### LINKED ISSUE #2077: Sort alphabetically project list in email preference in account settings
I noticed the project subscription list is not in a weird order, so you have to search for the project you want to uncheck or check. It would be easier if it was in alphabetical order. The user after 5 entries is going to expect that this will be alphabetical, and it would make it a lot easier to take find the project you're looking to remove yourself from.  This was originally on the front end by @parrish says it is a backend issue (https://github.com/zooniverse/Panoptes-Front-End/issues/1489#issuecomment-265269756)

1ee874f4-4ee6-11e5-8f12-4ec39f977135

camallen:
linked to https://github.com/zooniverse/Panoptes-Front-End/issues/2455 Allow collections to be sorted by display_name

closes #2077 allow ordering by project display_name

# Review checklist


app/models/concerns/preferences.rb
```diff
@@ -16,14 +16,7 @@ def preferences_for(preference, opts={})
...
       return all if user.is_admin?
-      case action.to_s
-      when "show", "index"
-        self.joins(:user, @preferences_for).where(:users => {private_profile: false},
-          @preferences_for.to_s.pluralize => {private: false})
-          .or(user.send("#{@preferences_for}_preferences"))
-      else
-        user.send("#{@preferences_for}_preferences")
-      end
+      where(user_id: user.id)
```
camallen: No, it'll now just return the preferences for this user with FK filtering in the seriailzer via params. Note our doc's explicitly state that 'preferences are only visible to user they belong to'. So we aren't using it client side (to my knowledge) and I don't think we need this public project / private profile joining here at all. If we do need it reinstated we should figure out a better way then the joins here.

camallen:
@zwolf The UPPs are used to track the email subscription settings for a project and would like to be sorted by project display name as requested by #2077

PR #2083, opened 2016-12-07
Fixes argument error on other mailer workers

zwolf:
The classification dump mailer worker needs the resource type ("project", "workflow") along with the other arguments, so the dump worker concern was sending it along. The other mailers, however, didn't know what to do with the extra argument. 

This is currently not tested for because the DumpMailerWorkerSpec stubs out a worker class and doesn't test sending #perform_async to each mailer worker individually, so changing the number of args didn't send up any red flags. Trying to add some there began to feel extremely hacky and I get why it was done the way it was, but it'd have been nice for tests to catch this.

That, and I can't really be sure this'll fix the issue. See https://app.honeybadger.io/projects/40595/faults/31953245 for details.

# Review checklist

### DIFF
--- a/app/workers/subject_data_mailer_worker.rb
+++ b/app/workers/subject_data_mailer_worker.rb
@@ -3,8 +3,8 @@ class SubjectDataMailerWorker
 
   sidekiq_options queue: :data_high
 
-  def perform(project_id, s3_url, emails)
+  def perform(resource_id, resource_type, s3_url, emails)
     return unless emails.present?
-    SubjectDataMailer.subject_data(Project.find(project_id), s3_url.to_s, emails).deliver
+    SubjectDataMailer.subject_data(Project.find(resource_id), s3_url.to_s, emails).deliver
   end
 end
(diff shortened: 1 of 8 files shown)


camallen:
> This is currently not tested for because the DumpMailerWorkerSpec stubs out a worker class and doesn't test sending #perform_async to each mailer worker individually, so changing the number of args didn't send up any red flags. 

This should have been caught with explicit specs for the mailer worker class e.g. `subject_data_mailer_worker_spec.rb`. Can you please add them in even if just to test the perform interface?

> Trying to add some there began to feel extremely hacky and I get why it was done the way it was, but it'd have been nice for tests to catch this.

Agree but I think this is the wrong spot to test as this just ensures the mailer workers are called correctly and not that they do their specific jobs. I did modify that spec a little to test the worker params are being passed correctly (patch here https://gist.github.com/camallen/2ef62ff4299e4ea2562a663e39f9bfaa), feel free to update that spec if you like.

zwolf:
Great, thanks for the patch. Updated.

PR #2084, opened 2016-12-12
Adds researcher quote to project update schema

zwolf:
It's already in ProjectCreateSchema, but #update was throwing a 422 when researcher quote was included in an update. 

# Review checklist


zwolf:
I added `researcher_quote` to the update params in the spec so that they will fail if it's not included in the schema, even if the resultant errors aren't immediately clear. There aren't any specs per schema, so maybe this is what you meant.

PR #2085, opened 2016-12-12
Adds display_name to tutorials

zwolf:
Blank by default, but there are only 220 rows in production so there shouldn't be any backfill issues.

# Review checklist

### DIFF
--- a/db/migrate/20161212205412_add_display_name_to_tutorials.rb
+++ b/db/migrate/20161212205412_add_display_name_to_tutorials.rb
@@ -1,9 +1,6 @@
 class AddDisplayNameToTutorials < ActiveRecord::Migration
   def change
     add_column :tutorials, :display_name, :text, default: ""
-    Tutorial.find_each do |tut|
-      tut.display_name = ""
-      tut.save!
-    end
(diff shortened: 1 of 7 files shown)


db/migrate/20161212205412_add_display_name_to_tutorials.rb
```diff
@@ -0,0 +1,9 @@
+class AddDisplayNameToTutorials < ActiveRecord::Migration
+  def change
+    add_column :tutorials, :display_name, :text, default: ""
+    Tutorial.find_each do |tut|
```
camallen: For what it's worth, when doing a bulk update like this you can use the `update_all` method to avoid the batch / validations, round trips, etc `Tutorial.update_all(display_name: "")`

zwolf: I was guessing there was a method like that but I couldn't track it down in the docs. #find_each was used in other migrations so I went with it. Updating now.

zwolf:
Closes https://github.com/zooniverse/Panoptes-Front-End/issues/2959

PR #2086, opened 2016-12-13
handle missing location metadata for older subjects

camallen:
Recent deploy breaks when serialising older subjects that have no location index metadata

# Review checklist

### DIFF
--- a/lib/formatter/csv/subject.rb
+++ b/lib/formatter/csv/subject.rb
@@ -39,9 +39,8 @@ module Formatter
       end
 
       def locations
-        subject_locs = subject.locations.sort_by { |loc| loc.metadata["index"] }
         {}.tap do |locs|
-          subject_locs.each_with_index.map do |loc, index|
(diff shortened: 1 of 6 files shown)


lib/formatter/csv/subject.rb
```diff
@@ -39,9 +39,8 @@ def workflow_ids
       end
 
       def locations
-        subject_locs = subject.locations.sort_by { |loc| loc.metadata["index"] }
```
zwolf: Your refactor is cleaner, but to make this work as-is it would have only needed to have this`sort_by` chopped off, since the `locations.loaded?` is the relevant question, right?

camallen: Not quite, it'll still need to return the sorted indexes when the association is loaded. Though, good point on natural sort order of the association, i'll modify it to catch any error and just revert back the natural ordering (which should be id / insertion order).

PR #2092, opened 2016-12-14
Fix broken subject specs

camallen:
#2086 had an active focus set checked in, sorry, so the formatter spec failed to run properly on that PR. 

1. I had the column ordering wrong...fixed. 
2. The spec was relying on database ordering to build up an expectation, I've fixed it to be consistent with the expectations.

# Review checklist

### DIFF
--- a/spec/models/subject_spec.rb
+++ b/spec/models/subject_spec.rb
@@ -102,7 +102,7 @@ describe Subject, :type => :model do
       expect(expected.map(&:id)).to eq(lone_subject.ordered_locations.map(&:id))
     end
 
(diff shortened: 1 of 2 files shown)


spec/models/subject_spec.rb
```diff
@@ -102,7 +102,7 @@
       expect(expected.map(&:id)).to eq(lone_subject.ordered_locations.map(&:id))
     end
 
-    context "subject without location metadata", :focus do
```
zwolf: https://gist.github.com/zwolf/a2f20f7b673792bda95a156619d34acf

I haven't done it since.

PR #2093, opened 2016-12-15
Add memory cache for testing out caching in staging

marten:
This adds a feature-flagged cache for the `as_json` hashes for models. 

Issues:

* Since we update counters without touching the `updated_at` timestamp, this will effectively keep them static until TTL (which I've set to 5 minutes for now).

### DIFF
--- a/app/serializers/concerns/cached_serializer.rb
+++ b/app/serializers/concerns/cached_serializer.rb
@@ -1,0 +1,17 @@
+module CachedSerializer
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def as_json(model, context)
+      if Panoptes.flipper["cached_serializer"].enabled?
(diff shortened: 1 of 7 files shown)


app/serializers/concerns/cached_serializer.rb
```diff
@@ -0,0 +1,17 @@
+module CachedSerializer
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def as_json(model, context)
+      if Panoptes.flipper["cached_serializer"].enabled?
+        cache_key = "#{model.class.to_s}/#{model.id}/#{model.updated_at.to_i}/context-#{Digest::MD5.hexdigest(context.to_json)}"
```
camallen: any reason not to create a smaller cache key from using the hexdigest of the whole string?

marten: No real reason, just came out of adding that digest later on in the development. Though once we use some external cache store we might enjoy being able to see which objects have caches, which would only really be possible with human readable keys.

PR #2098, opened 2016-12-19
Add avatars and background images to organizations

zwolf:
Adds avatars and background image Medium associations to Organizations.

Also adds a `can_filter_by :organization_id` to the ProjectSerializer, for which there is already an index.

Also adds some missing model specs to Organizations (translatable, activatable) and reorganizes them a bit.

# Review checklist


app/serializers/organization_serializer.rb
```diff
@@ -1,26 +1,37 @@
 class OrganizationSerializer
   include RestPack::Serializer
   include ContentSerializer
+  include MediaLinksSerializer
 
-  attributes :id, :name, :display_name, :description, :introduction, :title, :href
+  attributes :id, :name, :display_name, :description, :introduction, :title, :href, :avatar_src, :background
```
camallen: do we want avatar_src to be an optional attribute, we use this in projects with the cards param only.

app/serializers/organization_serializer.rb
```diff
@@ -1,26 +1,37 @@
...
+  media_include :avatar, :background
```
camallen: do you want background declared as attributes as well?

app/serializers/project_serializer.rb
```diff
@@ -37,7 +37,7 @@ class ProjectSerializer
     subjects_export: { include: false },
     aggregations_export: { include: false }
   can_filter_by :display_name, :slug, :beta_requested, :beta_approved,
-    :launch_requested, :launch_approved, :private, :state, :live
+    :launch_requested, :launch_approved, :private, :state, :live, :organization_id
```
camallen: you get FK filtering for free via restpack, https://github.com/RestPack/restpack_serializer#filtering so no need for the explicit setup

spec/controllers/api/v1/organizations_controller_spec.rb
```diff
@@ -146,5 +146,27 @@
         it_behaves_like "supports update_links"
       end
     end
+
+    describe "include avatar and background" do
```
camallen: can we test this in the serializer spec instead of the controllers? I've been working on extracting them recently when i see them.

spec/models/workflow_spec.rb
```diff
@@ -15,6 +15,8 @@
     let(:translatable) { create(:workflow_with_contents, build_extra_contents: true) }
     let(:translatable_without_content) { build(:workflow, build_contents: false) }
     let(:primary_language_factory) { :workflow }
+    let(:private_project) { create(:project, private: true) }
```
camallen: are these factories used?

app/serializers/organization_serializer.rb
zwolf: Cards are on my list of things to add, so I thought I'd include this while I was adding avatars.

app/serializers/organization_serializer.rb
zwolf: nope!

spec/controllers/api/v1/organizations_controller_spec.rb
zwolf: ✔️

spec/models/workflow_spec.rb
zwolf: https://github.com/zooniverse/Panoptes/blob/master/spec/support/translatable.rb#L109

camallen: that link is a ref's a private_model a few lines above and not the private_project here.

app/serializers/organization_serializer.rb
```diff
@@ -1,26 +1,37 @@
 class OrganizationSerializer
   include RestPack::Serializer
   include ContentSerializer
+  include MediaLinksSerializer
 
-  attributes :id, :name, :display_name, :description, :introduction, :title, :href
+  attributes :id, :name, :display_name, :description, :introduction, :title, :href, :avatar_src
```
camallen: It would be nice to have the avatar_src as an optional attribute e.g. [here](https://github.com/zooniverse/Panoptes/blob/bcfc0b2d659a533e71ee7d3f7774379e78e4d36d/app/serializers/project_serializer.rb#L32) instead of on each request. Mostly as this is duplicate info via the `media_include :avatar` call below

spec/models/workflow_spec.rb
zwolf: That's where it's used. Here's where it's built:
https://github.com/zwolf/Panoptes/blob/f220f2000a67f806d040e83eb92c43a1d421d5ff/spec/models/workflow_spec.rb#L18

Unless you mean that there's another way to ensure the translatable support has everything it needs?

camallen: got it - my bad on this one, sorry for the noise.

zwolf:
the hound has defeated me, i don't even know what it wants anymore.

i am vanquished.

PR #2099, opened 2016-12-20
parallel gem installs

camallen:
try and speed up the gem installs when building images.

# Review checklist


Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN mkdir config && curl "https://ip-ranges.amazonaws.com/ip-ranges.json" > conf
 ADD ./Gemfile /rails_app/
 ADD ./Gemfile.lock /rails_app/
 
+RUN bundle config --global jobs `cat /proc/cpuinfo | grep processor | wc -l | xargs -I % expr % - 1`
```
camallen: this could just use the BUNDLE_JOBS env var.

PR #2103, opened 2016-12-21
reset sws cache for each subject

### LINKED ISSUE #2101: subject export misreporting retirement and workflow attachment
For project #2125, the subject export incorrectly indicates that all of the project's subjects are:
-  retired
-  attached to the same workflow

camallen:
Fixes #2101 - reset the subject workflow status cache for each new subject

# Review checklist


camallen:
@marten agreed. I've updated to pass in the context on initialize and call to_array like a serializer

PR #2105, opened 2017-01-03
record unknown answer label for out of bounds lookups

### LINKED ISSUE #2104: not receiving comet hunters classification export
I'm not getting the Comet Hunters classification data export. I do get Planet Four: Terrains. I waited for 18 hours with no email. I tried again today and it's on the 5 hour mark and no export has been received. I think something's wrong/breaking for Comet Hunters. Can someone take a look?

camallen:
Closes #2104 - some annotations have out of bounds index lookups for the workflows answers array, this will mark them as invalid but still record the data in the export for the researcher.

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -145,9 +145,10 @@ module Formatter
       def answer_labels
         Array.wrap(@current["value"]).map do |answer_idx|
           begin
-            answer_string = workflow_at_version.tasks[@current['task']]['answers'][answer_idx]['label']
+            task_answer = workflow_at_version.tasks[@current['task']]['answers'][answer_idx]
+            answer_string = task_answer['label']
             translate(answer_string)
-          rescue TypeError
+          rescue TypeError, NoMethodError
             "unknown answer label"
           end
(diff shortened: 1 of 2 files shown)


lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -145,9 +145,10 @@ def answer_label
       def answer_labels
         Array.wrap(@current["value"]).map do |answer_idx|
           begin
-            answer_string = workflow_at_version.tasks[@current['task']]['answers'][answer_idx]['label']
+            task_answer = workflow_at_version.tasks[@current['task']]['answers'][answer_idx]
```
marten: Could we use `Hash#dig` instead here to fetch nested values (works across hashes with arrays with hashes just fine)? That way we don't have to capture all `NoMethodError`s. But we'd get a `nil` back, and we wouldn't be able to distinguish between a label that is configured to be `nil` and one that is not configured. Don't think that's an issue, but sanity check me here.

zwolf: Agreed that there needs to be something in the invalid label's place for researchers.

PR #2112, opened 2017-01-04
custom preloading serializer code experiment

camallen:
Added a custom panoptes wrapper over restpack serialzier to allow preloading on the page method.
This allows any valid include params to preload their associations and it provides a DSL to wire up preloads for each serializer. 

I've created a code experiment to test this out on a custom route to see the performance of doing this. 

# Review checklist


.rubocop.yml
```diff
@@ -208,3 +208,8 @@ Style/MultilineMethodCallIndentation:
   Enabled: false
 Style/FrozenStringLiteralComment:
   Enabled: false
+Style/StringLiterals:
```
camallen: happy to extract this to it's own PR

PR #2114, opened 2017-01-06
Use loaded associations to avoid sum db queries

camallen:
Avoid calling sum N+1 queries in the project serializer if the model is preloaded (paging).

The subject_set & workflow calls here are the sum queries (note preloading was turned on @ 16:18):

image

# Review checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -138,7 +138,13 @@ class Project < ActiveRecord::Base
   end
 
   def subjects_count
-    @subject_count ||= live_subject_sets.sum :set_member_subjects_count
+    @subject_count ||= if live_subject_sets.loaded?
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -138,11 +138,23 @@ def send_notifications
   end
 
   def subjects_count
-    @subject_count ||= live_subject_sets.sum :set_member_subjects_count
+    @subject_count ||= if live_subject_sets.loaded?
+      live_subject_sets.inject(0) do |sum,set|
```
marten: Alternatively: `live_subject_sets.to_a.sum(&:set_member_subjects_count)`. The `to_a` is a little bit hidden in there though.

PR #2116, opened 2017-01-09
Add tests for applications controller

marten:
Follow-up to #2115

# Review checklist


spec/controllers/applications_controller_spec.rb
```diff
@@ -9,4 +57,54 @@
       expect(Doorkeeper::Application.first.owner).to eq(user)
     end
   end
+
+  describe '#update' do
+    let(:application) { create(:application, name: 'test app') }
+
+    it 'shows your own application' do
```
zwolf: This and lower test description strings should probably be "updates" instead of "show"

PR #2117, opened 2017-01-10
Adds URLs to organizations

zwolf:
I've added URLs to organizations! This involved duping some of the URL label translation code from the projects controller to the create and update org operations. It currently works, but that duped code is bugging me.

It seems like I should be able to inherit from an Organization class (within the same Organizations module) and put those methods in that class, but when I do, the operation doesn't run correctly. 

This is what I tried: I created `organization.rb`, which is in the Organizations module and starts with `class Organization < Operation`. Therein, `self.content_from_params` and `self.extract_url_labels` are defined. The error when the spec runs is this: `NoMethodError: undefined method `run!' for #<Organizations::Update::ActiveRecord_Relation:0x007fadbfe118c8>`

Not the end of the world, but there's no other examples of the type of service object inheritance I was trying to I figured I'd ask.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/organizations_controller.rb
+++ b/app/controllers/api/v1/organizations_controller.rb
@@ -16,6 +16,11 @@ class Api::V1::OrganizationsController < Api::ApiController
                     :title,
                     :introduction].freeze
 
+  CONTENT_FIELDS = [:description,
+                    :title,
+                    :introduction,
+                    :url_labels].freeze
+
   def create
     organizations = Organization.transaction do
       Array.wrap(params[:organizations]).map do |organization_params|
--- a/app/operations/concerns/url_labels.rb
+++ b/app/operations/concerns/url_labels.rb
@@ -1,18 +1,20 @@
-module Organizations
-  class Organization < Operation
-    def self.content_from_params(ps)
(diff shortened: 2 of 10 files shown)


camallen:
@zwolf i'd create a concerns folder under operations (see models, etc) and in there create a shared code module that can be included in the organisation create and update operation definitions. That include should let you re-use the code. 

As a side note, as a general rule I always try and use composition instead of inheritance.

app/controllers/api/v1/organizations_controller.rb
```diff
@@ -16,6 +16,11 @@ class Api::V1::OrganizationsController < Api::ApiController
                     :title,
                     :introduction].freeze
 
+  CONTENT_FIELDS = [:description,
```
camallen: should these be moved to the shared operations module if they aren't used in the controller?

zwolf: The shared concern isn't organization-specific, which is why I'm passing the `CONTENT_FIELDS` array in via argument to the shared method in the operation. That way, if we do ever migrate Projects to use service objects too, they can use the same concern.

PR #2126, opened 2017-01-12
simplify the recents schema for query optimization

camallen:
move the has_one :through :classifications associations to belongs_to relations with indexes. Adds the belongs_to fkeys to recent schema and provides a rake task to backfill the missing information.

Once this is done we can remove the Recent has_one associations and modify the costly / complex recents query [here](https://github.com/zooniverse/Panoptes/blob/586f239d16a00fd75ef22df8498952e778684072/app/controllers/concerns/recents.rb#L9) to just use local indexes without the joins. 

# Review checklist

### DIFF
--- a/app/models/recent.rb
+++ b/app/models/recent.rb
@@ -32,7 +32,7 @@ class Recent < ActiveRecord::Base
 
   def copy_classification_fkeys
     if classification
-      %W(project_id workflow_id user_id user_group_id).each do |key|
+      %w(project_id workflow_id user_id user_group_id).each do |key|
         send("#{key}=", classification.send(key))
       end
     end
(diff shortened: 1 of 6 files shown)


app/models/recent.rb
```diff
@@ -9,11 +9,32 @@ class Recent < ActiveRecord::Base
...
+  # TODO: modify the following validation to run all the time
+  # once the recents schema has been migrated from has_one through associations
+  validates_presence_of :project_id, :workflow_id, :user_id, only: :create
+
+  before_validation :copy_classification_fkeys
+
   def self.create_from_classification(classification)
-    classification.subject_ids.each do |sid|
-      create!(subject_id: sid, classification: classification)
+    classification.subject_ids.map do |sid|
```
marten: Maybe just call that block parameter `subject_id`, to make sure any grepping will find it.

camallen:
Yep - i've got PR #2127 to modify the recent scope selector to the belongs_to fk columns.

PR #2130, opened 2017-01-13
Cache uploaded subjects count

### LINKED ISSUE #2129: speed up users#me route
avoid counting the user's subjects each time they request the me route data. Adding caching to the subjects_count method and invalidation / increment after each successful subject create.

Pinging @srallen - pretty sure this is the cause of slow user me routes on first page loads

camallen:
closes #2129 - avoid recounting the subjects_count where we can, instead cache the value and increment it on each successful subject create event.

# Review checklist

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -694,6 +694,19 @@ describe User, type: :model do
     end
   end
 
+  describe "#increment_subjects_count_cache", :with_cache_store, :focus do
+    it 'should return nil if no cache entry' do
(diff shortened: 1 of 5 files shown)


spec/models/user_spec.rb
```diff
@@ -694,6 +694,19 @@
     end
   end
 
+  describe "#increment_subjects_count_cache", :with_cache_store, :focus do
```
marten: There's a `focus` left here.

PR #2132, opened 2017-01-13
Export user email lists to s3

### LINKED ISSUE #1889: Exporting email subscription lists
The new mailing list server needs to be able to grab lists of people with `global_email_communication` and `beta_email_communication` turned on. The simplest option is probably to export both lists as plain text files, one address per line, to somewhere on S3 once per day.

camallen:
closes #1889 - adds workers to export  global, beta and launch approved projects email lists on a daily at 3am schedule. I have added a simple spread offset when the dumps run but i haven't tested this on real data. So the spread values may need to be adjusted / configured with env vars (happy to change).

# TODO

- [ ] Ensure the bucket config is setup for exports
- [ ] Merge migrated project lists, e.g. `zooniverse/milky-way-project` & `povich/milky-way-project`

# Review checklist


app/workers/concerns/dump_emails.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def write_to_s3
+    storage_path = MediaStorage.stored_path("application/x-gzip", "email_exports")
+    prefix = File.dirname(storage_path)
+    file_paths = File.basename(storage_path).split(".")
+    file_paths.shift
+    exts = file_paths.join(".")
+    file_name = "#{export_type}_email_list"
+    s3_path = "#{prefix}/#{file_name}.#{exts}"
+    MediaStorage.put_file(s3_path, gzip_file_path, storage_opts(file_name))
```
camallen: This may need to go into another bucket path (not static) here. If so we'd need to initialize the AwsAdapter with the specific bucket path for the put operations.

mrniaboc:
@marten @zwolf if you guys are able to review this, it would really make my life a much happier place!

app/workers/concerns/dump_commons.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def gzip_file_path
+    @gzip_tempfile ||= Tempfile.new(['export', '.gz'])
+    @gzip_tempfile.path
+  end
+
+  def upload_dump
+    to_gzip
+    write_to_s3
```
marten: I'd prefer if modules don't depend om methods they don't themselves define (or include automatically).

In fact, I'm thinking maybe this can just be a class that inside `def initialize` makes the csv tempfile, and then has a `to_gzipped` method that opens a gzip tempfile, zips it up and returns the tempfile object. And of course it should have a `close` method that closes both of those.

It could even handle the CSV opening, and just expose methods like "set_headers" and "write_row", and completely encapsulate the CSV generation.

app/workers/emails_users_export_worker.rb
```diff
@@ -0,0 +1,40 @@
...
+  include DumpEmails
+
+  sidekiq_options queue: :data_low
+
+  attr_reader :export_type
+
+  def perform(export_type=:global)
+    @export_type = export_type
+    begin
+      emails_to_csv_file
```
marten: Why is this using inheritance instead of e.g. a generic object that takes an ActiveRecord query scope as input, and dumps it?

mrniaboc:
@camallen @marten @zwolf I want this to happen almost as much as world peace or Scotland reaching a World Cup finals tournament in my lifetime. What has to happen to make this dream a reality?

camallen:
@marten how resolute are you in me reworking the code to address the changes? It would be nice to get this out even though it's poorly written. Can i add it to my other "debt" in the codebase?

marten:
@camallen maybe like 6-7 out 10? I find the code as in this PR hard to follow, mostly because of the way I have to keep 5 different modules all in my head. 

It's not just the DumpCommons, it's things like the EmailsProjectsExportWorker calling `emails_to_csv_file`, which is defined in DumpEmails, which in turn calls `user_emails`. After that some invisible state has changed, which means that it's now valid to call `upload_dump`, which is defined in DumpCommons.

That said, if you don't have any time, I'm happy to go refactor this directly on your branch?

camallen:
7/10 is a tough call, i'll add it to my list todo, though i know how much you love a good refactor.

app/models/csv_dumps/generic_dump_process.rb
```diff
@@ -0,0 +1,62 @@
+module CsvDumps
+  class GenericDumpProcess
```
marten: Rename this, there is only one

app/models/csv_dumps/full_email_list.rb
```diff
@@ -0,0 +1,31 @@
+module CsvDumps
+  class FullEmailList
```
marten: Should inhert from Dumpscope

app/workers/aggregations_dump_worker.rb
```diff
@@ -1,29 +1,9 @@
...
-  def cleanup_dump
-    # aggregation engine will write directly to medium when done, no need to cleanup
-  end
-
-  def load_medium
-    m = Medium.find(@medium_id)
-    m.update!(path_opts: resource_file_path, private: true, content_type: "application/x-gzip")
-    m
+  def perform(resource_id, resource_type, medium_id=nil, requester_id=nil, *args)
+    raise "No longer supported, nobody should be triggering this."
```
marten: Last one was generated months ago. Button in PFE is gone afaik. @camallen happy with this change?

app/workers/emails_projects_export_worker.rb
```diff
@@ -0,0 +1,31 @@
...
+  include DumpEmails
+
+  sidekiq_options queue: :data_low
+
+  def perform(project_id)
+    @project = Project.find(project_id)
+    @scope = CsvDumps::ProjectEmailList.new(project_id)
+
+    begin
+      perform_dump
```
marten: Isn’t this method gone?

marten:
@camallen Okay, needs a review from your side now.

app/models/csv_dumps/dump_processor.rb
```diff
@@ -0,0 +1,62 @@
...
+    def set_ready_state
+      medium.metadata["state"] = 'ready'
+      medium.save!
+    end
+
+    def write_to_s3(gzip_file_path)
+      medium.put_file(gzip_file_path, compressed: true)
+    end
+
+    def read_from_database(&block)
```
camallen: Seems like this is duplicating the underlying scope each methods (e.g. `CsvDumps::DumpScope`) that specify to `read_from_database` too. Maybe one or the other to ensure we read from the replica

app/workers/aggregations_dump_worker.rb
camallen: :tada: So happy

app/workers/aggregations_dump_worker.rb
```diff
@@ -1,29 +1,9 @@
...
-  include RateLimitDumpWorker
 
   sidekiq_options queue: :data_high
 
-  def perform_dump
-    raise ApiErrors::FeatureDisabled unless Panoptes.flipper[:dump_worker_exports].enabled?
-    # Set expiry time of signed_url to one day from now
-    medium.update!(put_expires: 1.day.from_now.to_i - Time.now.to_i)
-
-    AggregationClient.new.aggregate(resource, medium)
```
camallen: Maybe we should look at purging the aggregation client if we aren't using it? can come later but best not to leave it laying around.

app/workers/concerns/dump_worker.rb
```diff
@@ -4,128 +4,27 @@ module DumpWorker
...
-    "attachment; filename=\"#{name}-#{dump_target}.csv\""
   end
 
-  def get_resource
-    @resource_type.camelize.constantize.find(@resource_id)
+  def dump_target
+    @dump_target ||= self.class.to_s.underscore.match(/\A(\w+)_dump_worker\z/)[1]
   end
 
   def read_from_database(&block)
```
marten: This can go away

app/workers/concerns/dump_worker.rb
```diff
@@ -4,128 +4,27 @@ module DumpWorker
   included do
     include ActiveSupport::Callbacks
     define_callbacks :dump
-    attr_reader :resource
+    attr_reader :resource, :medium, :scope, :processor
   end
 
   def perform(resource_id, resource_type, medium_id=nil, requester_id=nil, *args)
     raise ApiErrors::FeatureDisabled unless Panoptes.flipper[:dump_worker_exports].enabled?
     @resource_type = resource_type
```
marten: This can go away (i think)

app/workers/concerns/dump_worker.rb
```diff
@@ -4,128 +4,27 @@ module DumpWorker
...
     @resource_id = resource_id
```
marten: This can go away (i think)

app/workers/emails_projects_export_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class EmailsProjectsExportWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_low
+
+  def perform(project_id)
+    @project = Project.find(project_id)
+
+    # TODO: use zoo wide email export media resources once bucket paths
```
camallen: Can you move this comment to the `DirectToS3` class as it's only there since we can't use the medium resource to put these files to a different bucket path.

app/workers/emails_users_export_worker.rb
```diff
@@ -0,0 +1,18 @@
+require 'csv'
+
+class EmailsUsersExportWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_low
+
+  def perform(export_type=:global)
+    # TODO: use zoo wide email export media resources once bucket paths
```
camallen: same - remove once it's at the class level.

lib/formatter/csv/user_email.rb
```diff
@@ -0,0 +1,17 @@
...
+    class UserEmail
+      def headers
+        false
+      end
+
+      def to_rows(user)
+        [to_array(user)]
+      end
+
+      def to_array(user)
```
camallen: seems like overkill

spec/lib/formatter/csv/subject_spec.rb
```diff
@@ -68,7 +68,7 @@
       it "should match the expected output" do
         sms.destroy
         subject.reload
-        expect(result).to match_array(expected.map { |row| row.merge(subject_set_id: nil) })
+        expect(result).to match_array(expected.map { |row| row[3] = nil; row })
```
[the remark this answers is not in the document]
camallen: maybe extract out the expectation transformation to a named variable

spec/models/csv_dumps/classification_scope_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:subject) { create(:subject, project: project, subject_sets: [create(:subject_set, workflows: [workflow])]) }
+  let(:classifications) do
+    create_list(:classification, 2, project: project, workflow: workflow, subjects: [subject])
+  end
+  let(:cache) { ClassificationDumpCache.new }
+
+  let(:scope) { described_class.new(project, cache) }
+
+  it "should find all the classifications" do
+    expect { |b| scope.each(&b) }.to yield_successive_args(*classifications)
```
camallen: nice

lib/formatter/csv/user_email.rb
marten: Ah, yeah with the addition of `to_rows` to all formatters (to maintain the same interface between everything including subjects) this could just be inlined into `to_rows`.

`to_array` should just be a private method anyway.

app/workers/concerns/dump_worker.rb
camallen: yep - this is called from the each block of the scopes

app/workers/concerns/dump_worker.rb
camallen: yeah - these instance vars can go

app/workers/concerns/dump_worker.rb
camallen: yeah - these instance vars can go

app/workers/aggregations_dump_worker.rb
marten: Yeah I want to make a second PR removing AggregationClient and the API to trigger the worker.

camallen: 👍

camallen:
:shipit: when the specs all pass 🎉

PR #2141, opened 2017-01-17
Adds a new ActiveInteraction filter for Organizations

zwolf:
Updates to organizations were failing due to missing parameters. ActiveInteraction doesn't have an "optional" flag for filters, so when the JS client only sends the changed attributes, AI validation would fail. So instead, now there's an `organization_params` filter (alongside `string`, `integer`, etc) that does it's own validation against the standard format JSON schema. 

Few things:
* Not sure why the `require_relative` is...required. Seems like it should just be loaded automatically since it's in /lib.
* Most of what's in the filter could easily be abstracted out for all Org ops or even as a generic adapter for JSONSchema-->ActiveInteraction. Don't really need that now, though.
* Opted to put the filter into `lib/filters` instead of `lib/gem_ext/active_interaction` since I'm not overriding anything.
* I found a closed (#wontfix) issue relating to this which basically said that the suggested fix was to set a nil or blank default for all your optional params, then sort through and remove the ones that aren't `given?()` from the inputs. I couldn't get this to work, though, as the request would 422 before the op's execute block was run.

# Review checklist

### DIFF
--- a/lib/filters/organization_filter.rb
+++ b/lib/filters/organization_filter.rb
@@ -13,7 +13,7 @@ module ActiveInteraction
     end
 
     def valid?(value)
-      OrganizationUpdateSchema.new.validate!(value) == nil
+      OrganizationUpdateSchema.new.validate!(value).nil?
     end
   end
 end
--- a/lib/filters/organization_filter.rb
+++ b/lib/filters/organization_filter.rb
@@ -4,7 +4,7 @@ module ActiveInteraction
 
     def cast(value, context)
       schema = OrganizationUpdateSchema.schema[:properties]
-      included_keys = value.keys.map {|z| z.to_sym}
+      included_keys = value.keys.map(&:to_sym)
       if included_keys.all? {|s| schema.key? s } && valid?(value)
         value
       else
--- a/lib/filters/organization_filter.rb
+++ b/lib/filters/organization_filter.rb
@@ -13,7 +13,7 @@ module ActiveInteraction
     end
 
     def valid?(value)
-      JSON::Validator.validate(OrganizationUpdateSchema.schema, value)
+      OrganizationUpdateSchema.new.validate!(value) == nil
     end
   end
 end
--- a/lib/filters/organization_filter.rb
+++ b/lib/filters/organization_filter.rb
@@ -1,0 +1,19 @@
+module ActiveInteraction
+  class OrganizationFilter < Filter
+    register :organization
+
+    def cast(value, context)
+      schema = OrganizationUpdateSchema.schema[:properties]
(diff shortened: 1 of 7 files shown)


lib/filters/organization_filter.rb
```diff
@@ -0,0 +1,19 @@
+module ActiveInteraction
+  class OrganizationFilter < Filter
+    register :organization
+
+    def cast(value, context)
+      schema = OrganizationUpdateSchema.schema[:properties]
+      included_keys = value.keys.map(&:to_sym)
+      if included_keys.all? {|s| schema.key? s } && valid?(value)
```
marten: Does this work for nested hashes and arrays? My own thought had been to actually call the `OrganizationSchema.validate!(value)`, similarly to how `JsonSchemaValidator` does it (which provides the `params_for` that is used by the create and update actions.

I'm fine with your approach, but do want to check if it works in all cases. You could simply add tests for that on the `Organizations::Create` operation.

zwolf: Yeah oops, it won't work with anything nested. Keep forgetting that's gotta be supported.

`#valid?`is just calling `JSON::Validator.validate(OrganizationUpdateSchema.schema, value)` (down on line 16), which is I believe what you're suggesting? (`OrganizationUpdateSchema` doesn't have a #validate method) 

The other bit is just to ensure that no params that aren't in the schema are included. I didn't actually change how create works, since all params are required. This would for sure be a next step, I just wanted to get this out so updates worked as expected.

lib/filters/organization_filter.rb
```diff
@@ -0,0 +1,19 @@
...
+        value
+      else
+        raise InvalidValueError, "#{name}: #{describe(value)}"
+      end
+    end
+
+    def valid?(value)
+      JSON::Validator.validate(OrganizationUpdateSchema.schema, value)
```
marten: This does something different than the controller counterpart at https://github.com/zooniverse/Panoptes/blob/master/lib/json_api_controller/json_schema_validator.rb#L43.

That `validate!` method is defined here: https://github.com/zooniverse/Panoptes/blob/master/lib/json_schema.rb#L89, and again, that's different than what's here. I'm not sure what the difference between `validate` and `fully_validate` is. Also, the implementation of that `validate!` suggests that `.fully_validate` might not return true/false depending on validation status?

Tests for this would make things clear (and keep stuff working if we upgrade the validator gem.

zwolf: `fully_validate` returns a hash of errors rather than just true/false. But since JsonSchema's validate just raises an exception if it's not empty (and changing that is out of scope for this PR), I'll just compare it to nil.

PR #2145, opened 2017-01-18
avoid count queries on classifications end point that timeout / stress the db

camallen:
Work around very query count calls that consume a lot of db resources / timeout api side buy manually set the classification serializer page values based on paging params to ensure we don't break clients that use the next/prev hrefs.

Long term this should be replaced with a modified Restpack::Serializer that uses the kaminari without_count page scopes, https://github.com/kaminari/kaminari#paginating-without-issuing-select-count-query

# Review checklist


camallen:
@marten i've updated to add a note in the api docs

PR #2152, opened 2017-01-31
[NO MERGE] Remove subject metadata from classifications dump

marten:


marten:
@mrniaboc @chrislintott This column is JSON and super repetitive. It also hardly ever changes, and it's the stuff you gave us when you uploaded the subjects in the first place. This data will still be in the subjects export, so if you want you can get that one and join the two exports on subject ID. You can even do this in Excel ([with the `VLOOKUP` function](https://support.office.com/en-us/article/VLOOKUP-function-0BBC8083-26FE-4963-8AB8-93A18AD188A1?ui=en-US&rs=en-US&ad=US&fromAR=1)).

Do either of you foresee any huge issues with dropping the subject_metadata from the classifications dump?

chrislintott:
I think this is a terrible idea. I think you've overestimating how easy our community will find doing this, and it has implications for long-term storage. Please talk to me before merging.

PR #2153, opened 2017-02-01
Collaborative Collections

zwolf:
Collections and projects can be filtered with the `editor` param. Index GET requests to either resource that includes that param will return projects/collections that the included user name/group name is either an owner or collaborator on.

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1259,7 +1259,7 @@ this action.
 
 + Response 204
 
-## Project Collection [/projects{?page,page_size,sort,owner,include}]
+## Project Collection [/projects{?page,page_size,sort,owner,editor,include}]
 A collection of _Panotpes Project_ resources.
 
 All collections add a *meta* attribute hash containing
@@ -1383,7 +1383,8 @@ Projects are returned as an array under the _projects_ key.
   + page (optional, integer) ... the index of the page to retrieve default is 1
   + page_size (optional, integer) ... number of items to include on a page default is 20
   + sort (optional, string) ... field to sort by
-  + owner (optional, string) ... string owner name of either a user or a user group to filter by.
+  + owner (optional, string) ... string owner name of either a user or a user group to filter by owner
+  + editor (optional, string) ... string name of either a user or user group to filter by owner and collaborator
   + include (optional, string) ... comma separated list of linked resources to include in the response
 
 + Request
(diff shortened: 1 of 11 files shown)


camallen:
@zwolf before i review this is there any reason not to use the existing [FilterByCurrentUserRoles](https://github.com/zooniverse/Panoptes/blob/d02556888a50671d77e04e6eb4952577d604a0fc/app/controllers/concerns/filter_by_current_user_roles.rb) module? 

Currently projects, orgs and collections index routes support it and it could be extended to some show routes and I think just adding the roles query params should work, e.g. 
`api/projects?current_user_roles=collaborator,owner` should work out of the box.

zwolf:
If I'm looking at `/collections/zwolf`, current user roles works fine. If I'm looking at `/collections/camallen`, it doesn't, and I can only see collections/projects you own, and not ones you are a collaborator on.

camallen:
Shouldn't you just see my collections that are public and my ones i've added you as a collaborator on? I don't think you should see the ones that I am collaborator on unless you yourself have collaborator rights on them?

zwolf:
The requested feature was a tab in the collections area that shows a list of "collaborative collections" that are "shared". The use case is for Gravity Spy, where the researchers want users to see that a "New and exciting glitch" collection is shared and can then be added to it and contribute.

Is that about right, @trouille ?

camallen:
Cool - thanks for the heads up re the use case. Regardless of implementation, I think you'd still have to add a collaborator acl for each user to the collection resource and how does that work for new people joining a project? 

Does the use case seem more like a publicly writable collection for logged in users, so anyone logged in user can write to it?

zwolf:
Each of which users? I don't think a publicly writable collection was what they had in mind, just a way to see that it was there. They wouldn't be able to browse it without adding a new ACL, is that what you meant?

Message me on slack if you're around and you'd like to talk to @trouille and I about this directly.

camallen:
@zwolf what's the status of this PR? Do we still need it?

zwolf:
Depends on whether this ends up being a "version 2" feature for gravity spy or abandoned entirely. I'm happy to just keep the branch and delete the PR if that's your preference.

camallen:
Just close it for now and keep your branch around if you like.

PR #2156, opened 2017-02-02
Remove redundant name column from Org

zwolf:
Removes `name` attribute from db, serializer, and schema. It is redundant with `display_name` and will cause confusion.

# Review checklist

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -637,7 +637,6 @@ ALTER SEQUENCE organization_contents_id_seq OWNED BY organization_contents.id;
 
 CREATE TABLE organizations (
     id integer NOT NULL,
-    name character varying,
     display_name character varying,
     slug character varying DEFAULT ''::character varying,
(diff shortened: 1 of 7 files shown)


db/structure.sql
```diff
@@ -2534,13 +2533,34 @@ CREATE INDEX index_recents_on_classification_id ON recents USING btree (classifi
 
 
 --
+-- Name: index_recents_on_project_id; Type: INDEX; Schema: public; Owner: -; Tablespace: 
+--
+
+CREATE INDEX index_recents_on_project_id ON recents USING btree (project_id);
```
marten: Where do these come from?

zwolf: That's a pretty good question. The migration is super simple and wouldn't have added those. Guess I'll remove them?

```class RemoveNameFromOrganization < ActiveRecord::Migration
  def change
    remove_column :organizations, :name
  end
end```

PR #2160, opened 2017-02-06
Nero config on workflow

marten:
Adds a dedicated place for Nero config to go.


db/migrate/20170206161946_add_nero_to_workflow.rb
```diff
@@ -0,0 +1,5 @@
+class AddNeroToWorkflow < ActiveRecord::Migration
+  def change
+    add_column :workflows, :nero_config, :jsonb
```
camallen: do we want a default value here?

marten: Could be defaulting to an empty hash

PR #2165, opened 2017-02-08
production only classification heartbeat warning

camallen:
don't cry wolf in staging system, provide better copy with rails env and only run in staging. Happy to change the recipient list from a yaml config to env vars if peeps want.

Describe your change here.

# Review checklist

### DIFF
--- a/app/views/classification_heartbeat_mailer/missing_classifications.text.erb
+++ b/app/views/classification_heartbeat_mailer/missing_classifications.text.erb
@@ -1,5 +1,5 @@
-Alert: all hands on deck!
+Team Alert!
 
-Eeek! No classification data received via panoptes in the last <%= (@window_period / 60).to_i %> minutes!
+No classification data received in the last <%= (@window_period / 60).to_i %> minutes!
 
-Panoptes API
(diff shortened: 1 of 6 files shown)


marten:
Should the recipient just be one of those mailing lists I keep hearing good things about?

camallen:
Indeed - it's using env vars now that have been setup on prod / staging with a list email address.

app/views/classification_heartbeat_mailer/missing_classifications.text.erb
```diff
@@ -1,5 +1,5 @@
-Alert: all hands on deck!
+Team Alert!
 
-Eeek! No classification data received via panoptes in the last <%= (@window_period / 60).to_i %> minutes!
+No classification data received in the last <%= (@window_period / 60).to_i %> minutes!
```
marten: How will that work with the new default of 15 seconds? "0 minutes"?

camallen: good point, i'll update the default to something sensible.

PR #2168, opened 2017-02-08
Added "contributor" role to Collections

zwolf:
You can now `update_links` on Collections if you have the "contributor" role. This was extremely simple on the model side. The controller specs were more proving it to myself than anything else.

Most of the effort on this feature will be on the front end, which I'm gonna give a shot soon.

# Review checklist

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -19,8 +19,8 @@ class Collection < ActiveRecord::Base
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
 
-  can_by_role :destroy, :update, :update_links, :destroy_links,
-              roles: [ :owner, :collaborator ]
+  can_by_role :destroy, :update, :destroy_links, roles: [ :owner, :collaborator ]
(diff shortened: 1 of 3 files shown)


app/models/collection.rb
```diff
@@ -19,8 +19,8 @@ class Collection < ActiveRecord::Base
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
 
-  can_by_role :destroy, :update, :update_links, :destroy_links,
-              roles: [ :owner, :collaborator ]
+  can_by_role :destroy, :update, :destroy_links, roles: [ :owner, :collaborator ]
+  can_by_role :update_links, roles: [ :owner, :collaborator, :contributor ]
   can_by_role :index, :show, public: true, roles: [ :owner, :collaborator, :viewer ]
```
camallen: I assume we want to have the :contributor to be able to show / index the collections too? Or just rely on the collection being made public to provide view rights?

zwolf: I figured since the only view restrictions on collections are public vs. private that it didn't matter, but it looking again it does seem cleaner to add it there.

PR #2169, opened 2017-02-09
Touch model updated_at vs cache key include

camallen:
Instead of including the associated resources with an N+1 in the extended cache_key call for ETAG generation this will ensure the relevant linked model gets a touched updated_at timestamp so the etag gets changed and the browser doesn't use a cached version of the resource. 

PFE - doesn't use this anymore so the original bugs that the extended cache key fixed don't necessarily apply (index vs show routes) but we're still paying the cost of lookup on the associated models here. 

# Review checklist


marten:
This seems reasonable.

PR #2172, opened 2017-02-10
Recents schema, index and serialzier updates

camallen:
Cleanup invalid recent data to allow the fk migrations to pass. Rebuild the indexes as they are bloated and referencing old removed rows. Preload the recent media locations by reusing the extracted subject#ordered_locations code and finally add an index for created_at sorting without scans. 

# Review checklist

### DIFF
--- a/db/migrate/20170210163241_recents_index_updates.rb
+++ b/db/migrate/20170210163241_recents_index_updates.rb
@@ -1,0 +1,10 @@
+class RecentsIndexUpdates < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    %i(classification_id project_id subject_id user_id workflow_id).each do |col|
+      remove_index :recents, col
+      add_index :recents, col, algorithm: :concurrently
+    end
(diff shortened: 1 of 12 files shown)


db/migrate/20170210163241_recents_index_updates.rb
```diff
@@ -0,0 +1,12 @@
+class RecentsIndexUpdates < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    %i(classification_id project_id subject_id user_id workflow_id).each do |col|
+      remove_index :recents, column: col
+      add_index :recents, col, algorithm: :concurrently
```
marten: What's the point of this remove-then-add? Doesn't seem to have change the index algorithm in the structure.sql, so ... ?

camallen: It rebuilds the index to remove the bloat from deleted removed rows. Currently ~1GB but should be around 700MB

PR #2176, opened 2017-02-14
Clean up subject selection

marten:
I've removed all of the checks to see if we need to call out to Cellect/CellectEx before enqueueing the jobs. Instead, it'll just do that when it runs. However, I can see a case to be made for keeping it. If you like, I would suggest doing that in an override to `perform_async` in the worker, rather than littering the rest of the codebase with it.


app/models/subject_selector.rb
```diff
@@ -0,0 +1,25 @@
+class SubjectSelector
+  def self.for(workflow)
+    case strategy_for(workflow)
```
camallen: could we push responsibility to the workflow to hide the implementation? perhaps
``` ruby
case 
when workflow.strategy?(:cellect)
...
```

app/workers/notify_subject_selector_of_seen_subject_worker.rb
```diff
@@ -1,6 +1,4 @@
-require 'subjects/cellect_client'
-
-class SeenCellectWorker
+class NotifySubjectSelectorOfSeenSubjectWorker
```
camallen: the second subject seems redundant. not super keen on these names but maybe `NotifySubjectSelectorOfSeenWorker`

app/workers/notify_subject_selector_of_subjects_change_worker.rb
```diff
@@ -1,6 +1,4 @@
-require 'subjects/cellect_client'
-
-class ReloadCellectWorker
+class NotifySubjectSelectorOfSubjectsChangeWorker
```
camallen: `NotifySubjectSelectorOfChangeWorker` ?

lib/subjects/panoptes_client.rb
```diff
@@ -0,0 +1,23 @@
+module Subjects
```
camallen: Should this be in `app/services`?

marten: Ah, it's no longer referenced due to refactorings. Deleting.

marten:
I'll incorporate your comments and merge. I've got more refactorings coming, but let's get the notifications stuff out the door.

PR #2184, opened 2017-02-17
Fix calls to CellectExSelector

marten:


marten:
No, it doesn't support them. Plus, we never retire subjects in a subject_set, we retire them in a workflow. I don't really know what normal Cellect does with that group parameter.

camallen:
remove them from an internal data structured key'd to the subject_set id, https://github.com/zooniverse/Cellect/blob/master/lib/cellect/server/grouped_workflow.rb

marten:
Right, seems like Cellect just needs to know it because of it's internal structure. With CellectEx, this doesn't actually remove them, it just triggers a full reload. In the future, I'm going to have it maintain a temporary list of items to remove while streaming, that gets emptied upon DB reload.

PR #2186, opened 2017-02-17
set default retirement reason to classifications count if not supplied

### LINKED ISSUE #2182: add in an entry for  retirement_reason  in subject export if subject is retired due to normal classification limit hit
It's really confusing that the  retirement_reason column is empty if the project doesn't have custom retirement rules. It's possible a project may switch from one type of retirement to another, so it would be good to have an entry for all retirement types including normal non-custom ones.

camallen:
closes #2182 

# Review checklist

### DIFF
--- a/app/models/subject_workflow_status.rb
+++ b/app/models/subject_workflow_status.rb
@@ -36,7 +36,7 @@ class SubjectWorkflowStatus < ActiveRecord::Base
     !retired? && workflow.retirement_scheme.retire?(self)
   end
 
-  def retire!(reason=:classification_count)
+  def retire!(reason=nil)
     unless retired?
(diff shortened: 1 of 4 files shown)


app/models/subject_workflow_status.rb
```diff
@@ -36,7 +36,7 @@ def retire?
     !retired? && workflow.retirement_scheme.retire?(self)
   end
 
-  def retire!(reason=nil)
+  def retire!(reason=:classification_count)
```
marten: I'm wondering if we should do this higher-up (in `ClassificationLifecycle`) so that if some API user comes and retires without giving a reason, we don't assume it's due to classification count in that case.

camallen:
@marten updated to record from classification count worker (via lifecycle).

PR #2193, opened 2017-02-22
respect the activated_state of the active workflows

### LINKED ISSUE #2191: Project `active_workflows` ignored deletedness (i.e. `activated_state`)
## Steps to reproduce

* Create project
* Create two workflows and give each of them a distinct subject set with some images
* Mark both workflows as "Active" on the Visibility tab.
* Delete one workflow

*Expected result*: Number of subjects on the project page goes down
*Actual result*: Number of subjects stays the same.

This seems to be the result of `Project#active_workflows` ignoring the `activated_state` of workflows (ignoring the soft-delete).

@camallen pretty sure this isn't intentional, but wanted to double check with you since you added that `Project#active_workflows` scope.

camallen:
Fixes #2191 - don't serialize deleted workflows

# Review checklist

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -114,7 +114,7 @@ describe Project, type: :model do
     end
   end
 
-  describe "#active_workflows" do
+  describe "#active_workflows", :focus do
     let(:project) do
       create(:project) do |p|
         create(:workflow, project: p, active: true)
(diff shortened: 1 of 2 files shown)


spec/models/project_spec.rb
```diff
@@ -114,7 +114,7 @@
     end
   end
 
-  describe "#active_workflows" do
+  describe "#active_workflows", :focus do
```
marten: focus left in here

PR #2194, opened 2017-02-22
simplify the subject selector serialized response

camallen:
Avoid counting AR scopes when selecting subjects and only serialize the required data, e.g. the subjects and their locations.

# Review checklist


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -11,24 +11,6 @@ class Api::V1::WorkflowsController < Api::ApiController
 
   prepend_before_action :require_login, only: [:create, :update, :destroy, :create_classifications_export]
 
-  # temp action to test perf of Panoptes::RestpackSerializer
```
marten: Did you mean to do all this on this branch?

app/serializers/subject_selector_serializer.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def include_already_seen?
+    select_context?
+  end
+
+  def include_finished_workflow?
+    select_context?
+  end
+
+  def select_context?
```
marten: We can probably get rid of this here too, it should just always be true in this serializer (just like it's false in the other).

I can't quite find where that context is getting passed in from the controller. The tests talk about a "feature flag" but I don't see it.

spec/serializers/subject_selection_serializer_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    describe "seen, retired, finished selection contexts" do
+      it "should run the lookups if the feature flag is off" do
+        expect_any_instance_of(SubjectSelectorSerializer).to receive(:retired)
+        expect_any_instance_of(SubjectSelectorSerializer).to receive(:already_seen)
+        expect_any_instance_of(SubjectSelectorSerializer).to receive(:finished_workflow)
+        run_serializer
+      end
+
+      context "when skip select context lookup feature flag is on" do
+        let(:selection_context) { { select_context: false } }
```
marten: That's not really a feature flag (at least we're using that term elsewhere for Flipper things, so it's a little confusing).

app/controllers/api/v1/workflows_controller.rb
camallen: I've branched of off remove_prp_serializer_experiment and was going to rebase, probably should have put that in the description.

app/serializers/subject_selector_serializer.rb
camallen: It's coming from the `Subjects::Selector` and is derived from a flipper feature flag. We need to keep this in to allow the optional attributes to be turned off unless you want to have 2 SubjectSelector serializers and put that behind a feature flag in the controller?

spec/serializers/subject_selection_serializer_spec.rb
camallen: it's a derived boolean to modify the serializer behaviour at run time, see [here](https://github.com/zooniverse/Panoptes/blob/master/lib/subjects/selector.rb#L120). Maybe splitting up the serializers makes more sense than using optional attributes in 1 serializer.

app/serializers/subject_selector_serializer.rb
marten: Why not just reference that flipper directly in the SubjectSelectorSerializer? I'd much prefer that. And actually, I think the rest of the context should be coming from the controller. Keeps the Subjects::Selector dedicated to selecting subjects, not configuring a serializer.

spec/serializers/subject_selection_serializer_spec.rb
marten: Huh? Haven't you already split up the serializers?

app/serializers/subject_selector_serializer.rb
camallen: I'll extract the selector context to the controller

spec/serializers/subject_selection_serializer_spec.rb
camallen: yeah and the new selector serializer this PR introduces has optional attributes that are configured via the serializer context boolean. If we want to no serialize the user has seen this subject / it's finished in the workflow, we can either use optional attributes in the selector serializer OR split this new serializer into 1 with all attributes and another without the optional ones.

camallen:
@marten extracted the selector serializer context to the controller.

PR #2195, opened 2017-02-23
Update API documentation wrt /subjects/queued

marten:


apiary.apib
```diff
@@ -3159,7 +3171,7 @@ Resources related to _Panotpes SubjectQueues_
 ## SubjectQueue [/subject_queues/{id}{?include}]
 A subject queue is a list of subjects for classification by a
 user that is created by a project owner or editor. When a user
-requests subjects to classify using the `?sort=queued` param they we
+requests subjects to classify using the `/subjects/queued` API they we
```
camallen: This should be removed as it's no longer valid. Maybe leave a note about custom task allocation instead?

PR #2200, opened 2017-02-23
Move project/org linking to project controller

zwolf:
Linking a project to an organization would makes sense if we had a more standard JSON API spec implementation. As it is, the behavior of all controllers when unlinking is to destroy the unlinked resource. This only happens with workflows that were copied when linked originally (between projects) or to SetMemberSubjects when a subject is removed from a set or collection.

It's probably not a good idea to delete projects when we remove them from organizations. Rather than changing how this works or overriding a bunch of methods, it makes mores sense to just remove the `can_be_linked :project [etc]` from orgs and just update the `organization_id` attribute on the project directly. This means that you need to have permission to edit the project, which makes sense too.

# Review checklist

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -23,7 +23,6 @@ class Organization < ActiveRecord::Base
 
   can_by_role :translate, roles: [ :owner, :translator, :collaborator ]
 
-  can_be_linked :project, :scope_for, :update, :user
   can_be_linked :access_control_list, :scope_for, :update, :user
 
 end
--- a/app/schemas/project_update_schema.rb
+++ b/app/schemas/project_update_schema.rb
@@ -109,6 +109,11 @@ class ProjectUpdateSchema < JsonSchema
       end
     end
 
+    property "organization_id" do
+      type "string", "integer", nil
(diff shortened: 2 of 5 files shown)


marten:
I think @camallen was suggesting the linked behaviour should be considered a bug, and we could potentially fix it rather than throwing it out completely for organizations.

zwolf:
I know, but this is a blocker for the lab app. If @amyrebecca can continue to wait I can ditch this, though I'm in the middle of SGL-related work just now.

amyrebecca:
We aren't actively working on organizations this week anyways; what kind of timeline are we looking at for fixing the linking behavior?

camallen:
@amyrebecca It should just be some reflection on the association type and unset the fk vs call destroy on the join table. I'd imagine a few hours to a day if issues pop up. Related code is [here](https://github.com/zooniverse/Panoptes/blob/c6067cc758e3b275f31bbb7bec864caa488f16f9/lib/json_api_controller/relation_manager.rb#L25)

zwolf:
Fixed in #2217 instead.

PR #2205, opened 2017-02-28
Adds 'favorite' to subject serializer

zwolf:
Adds a Flipper flag to toggle inclusion of the new `favorite` attribute to the subject serializer. It should only be included when the toggle is on and the user is logged in. It indicates whether the serialized subject is in the user's Favorites collection.

# Review checklist

### DIFF
--- a/app/serializers/subject_serializer.rb
+++ b/app/serializers/subject_serializer.rb
@@ -3,7 +3,7 @@ class SubjectSerializer
   include FilterHasMany
 
   attributes :id, :metadata, :locations, :zooniverse_id,
-    :created_at, :updated_at, :href
+    :created_at, :updated_at, :href, :favorite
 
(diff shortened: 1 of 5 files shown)


app/serializers/subject_serializer.rb
```diff
@@ -3,7 +3,7 @@ class SubjectSerializer
   include FilterHasMany
 
   attributes :id, :metadata, :locations, :zooniverse_id,
-    :created_at, :updated_at, :href
+    :created_at, :updated_at, :href, :favorite
```
camallen: This should be an optional attribute and turned on / off via the existing context setup and feature flag for the extra selection context.

zwolf:
I went with @marten's suggestion from Panoptes-Front-End/issues/3494 of using a feature flag to determine whether to return it, not a param. Happy to switch if that's the preference.

PR #2210, opened 2017-03-03
Optional 'favorite' attr now available on subject

zwolf:
Now with includes @camallen's updated subject serializer! `favorite` is an optional subject attribute along with the others, all behind a single flipper. 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -22,10 +22,11 @@ class Api::V1::SubjectsController < Api::ApiController
   def queued
     selector = Subjects::Selector.new(api_user.user, workflow, params)
     non_filterable_params = params.except(:project_id, :collection_id)
+    selected_subjects = selector.get_subjects
     render json_api: SubjectSelectorSerializer.page(
       non_filterable_params,
-      selector.get_subjects,
-      selector_context
+      selected_subjects,
+      selector_context(selected_subjects.map(&:id))
     )
   end
 
@@ -108,15 +109,22 @@ class Api::V1::SubjectsController < Api::ApiController
     end
   end
 
-  def selector_context
+  def selector_context(selected_subject_ids)
     if Panoptes.flipper[:skip_subject_selection_context].enabled?
       {}
     else
+      if api_user.logged_in?
+        fav_collections = api_user.user.favorite_collections_for_project(workflow.project)
+        fav_subjects = fav_collections.joins(:subjects).where(subjects: {id: selected_subject_ids}).pluck("subjects.id")
+      else
+        fav_subjects = []
+      end
       {
         workflow: workflow,
         user: api_user.user,
         user_seen: user_seen,
         url_format: :get,
+        favorite_subject_ids: fav_subjects,
         select_context: true
       }.compact
     end
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -113,18 +113,12 @@ class Api::V1::SubjectsController < Api::ApiController
     if Panoptes.flipper[:skip_subject_selection_context].enabled?
       {}
     else
-      if api_user.logged_in?
-        fav_collections = api_user.user.favorite_collections_for_project(workflow.project)
-        fav_subjects = fav_collections.joins(:subjects).where(subjects: {id: selected_subject_ids}).pluck("subjects.id")
-      else
-        fav_subjects = []
-      end
       {
         workflow: workflow,
         user: api_user.user,
         user_seen: user_seen,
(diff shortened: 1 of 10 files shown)


app/serializers/subject_selector_serializer.rb
```diff
@@ -31,6 +31,10 @@ def already_seen
     !!(user_seen&.subjects_seen?(@model.id))
   end
 
+  def favorite
+    user&.collections&.where(favorite: true)&.first&.subjects&.include? @model
```
zwolf: This line makes me kind of cringe, but it works.

app/serializers/subject_serializer.rb
```diff
@@ -7,7 +7,7 @@ class SubjectSerializer
   attributes :id, :metadata, :locations, :zooniverse_id,
     :created_at, :updated_at, :href
 
-  optional :retired, :already_seen, :finished_workflow
+  optional :retired, :already_seen, :finished_workflow, :favorite
```
camallen: Rip these out, i've left them in and they shouldn't be there, apologies.

spec/controllers/api/v1/subjects_controller_spec.rb
```diff
@@ -209,6 +213,32 @@
             expect(retired).to all( be false )
           end
 
+          describe "favorited subjects" do
+            context "user has no favorites collection" do
+              it 'should return favorite as false' do
+                user.collections.destroy_all
```
camallen: Any reason why we pay the cost of making collections only to destroy them? Best just to remove the auto-create ! in let and call them when (before block) you need them.

spec/controllers/api/v1/subjects_controller_spec.rb
```diff
@@ -209,6 +213,32 @@
...
+                get :index, request_params
+                favorites = json_response["subjects"].map{ |s| s['favorite']}
+                expect(favorites).to all( be nil )
+              end
+            end
+
+            context "user has a favorites collection" do
+              before(:each) { get :index, request_params }
```
camallen: e.g. add collections here or move the let block down here entirely if this is where they are used.

spec/controllers/api/v1/subjects_controller_spec.rb
```diff
@@ -387,6 +417,30 @@
             expect(retired).to all( be false )
           end
 
+          describe "favorited subjects" do
```
camallen: If this is a straight copy of the above, then maybe extract to a shared helper in this spec to avoid the duplication? Up to you

e.g. in that file somewhere
``` ruby
shared_examples "a favorited subject" do
  ...
end

it_behaves_like "a favorited subject"

```

app/serializers/subject_selector_serializer.rb
camallen: Does this just compare the first user favourite collection for inclusion even though they will have many, i.e. multiple projects? the serializer module to filter the has many's uses a join query to determine the favorite collections that a subject belongs to. That is they type of query this needs

I think this will result in N+1's in the serializer, i.e. each model in the serializer page will request the user's first collection and test the include subjects. It would be better to pass in the predefined set of fav subject ids to compare the model.id against. 

However I am not sure exactly how this will work but one solution could be in the controller behind the feature flag, run subject selection, get the subject ids, run the join query with the subject_ids then you could reduce the result to have a { subject_id: fav_or_not_bool } or [ fav_subject_ids_array ] that can be tested in the serializer here. This would ensure the query is only run once and not every time, thus saving some computation cost on a critical end point that we really need to be as fast as possible.

app/controllers/api/v1/subjects_controller.rb
```diff
@@ -108,15 +109,22 @@ def location_params(locations)
     end
   end
 
-  def selector_context
+  def selector_context(selected_subject_ids)
     if Panoptes.flipper[:skip_subject_selection_context].enabled?
       {}
     else
+      if api_user.logged_in?
```
camallen: This would be great if it was just `FavoritesFinder.new(api_user, project).ids` and that handled the internal logic.

app/models/user.rb
```diff
@@ -328,6 +328,10 @@ def increment_subjects_count_cache
     Rails.cache.increment(subjects_count_cache_key)
   end
 
+  def favorite_collections_for_project(project)
+    collections.where(favorite: true, projects: [project])
```
camallen: This seems like it has feature envy with collections. I think it would be great to extract all this to `FavoritesFinder.new(api_user, project).ids` (lib / model concern?) that deals with this logic.

app/serializers/subject_serializer.rb
camallen: again - can you rip these out?

PR #2214, opened 2017-03-06
Rename CellectEx to Designator, and start sending authentication

marten:
This pushes the service rename through to Panoptes. Additionally, this adds support for the authentication that will soon be required on reload/retire endpoints of Designator.

# Review checklist

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -32,7 +32,7 @@ class Workflow < ActiveRecord::Base
   cache_by_association :workflow_contents
   cache_by_resource_method :subjects_count, :finished?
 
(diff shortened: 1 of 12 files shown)


app/models/workflow.rb
```diff
@@ -32,7 +32,7 @@ class Workflow < ActiveRecord::Base
   cache_by_association :workflow_contents
   cache_by_resource_method :subjects_count, :finished?
 
-  enum subject_selection_strategy: [:default, :cellect, :cellect_ex]
+  enum subject_selection_strategy: [:default, :cellect, :designator]
```
marten: This keeps using the same integer in the DB, which is what we want.

PR #2217, opened 2017-03-07
Destroy should be delete in the relation manager

zwolf:
Calling this a WIP just now because I'd like to see what color the other specs will be with just this change. Can't really think of a situation where we'd want to do the opposite of what `delete` does here, i.e. actually delete the join table row OR set the relation's foreign key on the resource to nil.

update: refactored `destroy_relation` specs a bit and added a few more to cover all association types. Seems to behave as expected/desired.

# Review checklist

### DIFF
--- a/spec/lib/json_api_controller/relation_manager_spec.rb
+++ b/spec/lib/json_api_controller/relation_manager_spec.rb
@@ -154,11 +154,11 @@ describe JsonApiController::RelationManager do
     context "has_many" do
       let!(:resource) { create(:project_with_workflows) }
 
-      it 'should destroy the unlinked items' do
+      it 'should not destroy the unlinked items' do
         expect do
           del_string = resource.workflows.map(&:id).join(",")
           test_instance.destroy_relation(resource, :workflows, del_string)
-        end.to change{ Workflow.count }.from(2).to(0)
+        end.to_not change{ Workflow.count }
       end
     end
   end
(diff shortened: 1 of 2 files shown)


spec/lib/json_api_controller/relation_manager_spec.rb
```diff
@@ -135,30 +135,70 @@ def resource_class
...
+      let(:relation) { :subjects }
+      let(:del_string) { resource_to_remove.map(&:id).join(",") }
+
+      it "does not destroy the unlinked items" do
+        expect{ destroy_helper }.to_not change{Subject.count}
+      end
+
+      it "deletes the linked relations" do
+        destroy_helper
+        expect(resource.subjects).to_not include(*resource_to_remove)
```
camallen: is this testing the actual delete / destroy of the join model resources?

spec/lib/json_api_controller/relation_manager_spec.rb
```diff
@@ -135,30 +135,70 @@ def resource_class
...
+      let(:resource) { create(:project_with_workflows) }
+      let(:resource_to_remove) { resource.workflows[0..2] }
+      let(:relation) { :workflows }
+      let(:del_string) { resource.workflows.map(&:id).join(",") }
+
+      it "does not destroy the unlinked items" do
+        expect{ destroy_helper }.to change{ Workflow.count }.by(0)
+      end
+
+      it "deletes the linked resource" do
```
camallen: I think the naming of this spec is misleading, do we want to remove the linked resource or just the intermittent join models?

zwolf: First one is "deletes the linked relation", which I forgot to change the other three to. Is that better enough?

camallen: first one seems good, i understood it as removing the relation from the resource.

PR #2219, opened 2017-03-08
Include state in classifications export metadata on update

zwolf:
Yup.

# Review checklist


app/operations/create_or_update_medium.rb
```diff
@@ -13,6 +13,7 @@ class CreateOrUpdateMedium < Operation
 
   def execute
     if medium = object.public_send(type)
+      media['metadata']["state"] = 'creating'
```
camallen: seems this runs on both branches of the if..move to outside the conditional?

PR #2222, opened 2017-03-08
Add settings to project prefs update schema

zwolf:
The settings property should be in the ProjectPreferencesUpdate schema so that PFE can update a user's project preferences. 

# Review checklist


zwolf:
Eh, not really, but I'll add the settings object to the update params in the controller spec and that should be good.

PR #2224, opened 2017-03-09
compare weak etags for update / destroy actions precondition checks

camallen:
Allow weak etag comparison if we receive a weak_etag server side for modifications, etc. This is due to adding gzip compression to nginx as it modifes the etag to be weak as this is correct etag spec. Rails is off spec with strong etags (changing in ver 5 to be weak) and this change allows the etags to not be rejected api side.

# Review checklist

### DIFF
--- a/lib/json_api_controller/precondition_check.rb
+++ b/lib/json_api_controller/precondition_check.rb
@@ -21,13 +21,17 @@ module JsonApiController
 
     def precondition_fails?
       query = resource_class.where(id: resource_ids)
-      current_etag = gen_etag(query)
-      current_etag = "W/#{current_etag}" if weak_etag?(precondition)
-      !(current_etag == precondition)
+      run_etag_validation(query)
     end
 
     def precondition_error_msg
       "Request requires #{HEADER_NAME} header to be present"
+    end
+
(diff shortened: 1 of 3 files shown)


lib/json_api_controller/precondition_check.rb
```diff
@@ -21,11 +21,17 @@ def precondition
 
     def precondition_fails?
       query = resource_class.where(id: resource_ids)
-      !(gen_etag(query) == precondition)
+      current_etag = gen_etag(query)
+      current_etag = "W/#{current_etag}" if weak_etag?(precondition)
+      !(current_etag == precondition)
```
astopy: The format for the weak etags is `W/"31877a31e40bd799084519e302a5aac3"` vs `"31877a31e40bd799084519e302a5aac3"` -- I'm probably being overly cautious but are we sure it'll handle the quotes correctly?

camallen: internally the etag strings look like
```
current_etag
=> "\"7f9f1615b7dd80924345f5da779536f8\""
request.headers[HEADER_NAME]
=> "W/\"7f9f1615b7dd80924345f5da779536f8\""
```

PR #2226, opened 2017-03-10
Adds index to media.linked_id

zwolf:
So that asking for included images with a tutorial takes less than 60 seconds.

# Review checklist


zwolf:
This now includes both inverted indices for media and tagged resources (type before id) rather than an extra, since postgresql will use that same index for id-only lookups w/o a whole extra stored in memory. Incoming is another PR that will delete the old indices once these are built, since subjects is just shy of 10M rows and this'll take a while.

PR #2227, opened 2017-03-10
Add a feature flag for completed subject removal

zwolf:
Only remove complete subjects from the set returned from the selector if the feature flag is enabled. 

# Review checklist

### DIFF
--- a/lib/subjects/strategy_selection.rb
+++ b/lib/subjects/strategy_selection.rb
@@ -19,9 +19,13 @@ module Subjects
       eventlog.info("Selected subjects", desired_strategy: strategy, used_strategy: used_strategy, subject_ids: selected_ids)
 
       selected_ids = selected_ids.compact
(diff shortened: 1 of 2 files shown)


marten:
Protip: go to https://panoptes.zooniverse.org/flipper/features and make sure to Add and Enable them in advance. Otherwise it'll default to disabled upon next deploy.

zwolf:
Added and enabled.

camallen:
you could have just inverted the boolean check....

zwolf:
Still needs to be in the UI, and that last click wasn't a lot of trouble.

marten:
Somehow it feels more logical to have Flipper be a list of features that are positively named and normally enabled (as opposed to `disable_X`). But maybe that's just me.

zwolf:
I agree, mixing in `:skip_feature` flags seems to just be confusing.

PR #2231, opened 2017-03-14
count the user uploaded subjects for longer to speed up login refreshes

camallen:
cache this value for a decent session time (1 hour) to avoid paying this cost on page reloads

# Review checklist

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -318,7 +318,7 @@ class User < ActiveRecord::Base
   end
 
   def uploaded_subjects_count
-    count = Rails.cache.fetch(subjects_count_cache_key) do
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -337,4 +337,8 @@ def favorite_collections_for_project(project)
   def subjects_count_cache_key
     @subjects_count_cache_key ||= "User/#{id}/uploaded_subjects_count"
   end
+
+  def subject_count_cache_expiry
+    ENV.fetch("UPLOADED_SUBJECTS_COUNT_CACHE_EXPIRY", 1.hour)
```
zwolf: Where does this get set? Is there a corresponding infrastructure repo PR or docker image update to add this?

camallen: the env file that docker-compose exposes to the container at runtime, it's in prod configs.

PR #2232, opened 2017-03-14
Shortcut task type value is an array, not a string

### LINKED ISSUE #2228: Open Shortcut Task Export to Multiple Answers
In anticipation for the Stellar Watch launch on Wednesday (3/15) the Shortcut task needs to allow the possibility of multiple shortcut answers. Currently, only one shortcut answer is allowed. The logic has been hooked up on the front-end with this [PR](https://github.com/zooniverse/Panoptes-Front-End/pull/3532). 

I believe the only backend change needed is with the exporter? If so, I believe [this](https://github.com/zooniverse/Panoptes/blob/0871ff599d7f11c4715acd3746eebca6efad272e/lib/formatter/csv/annotation_for_csv.rb#L54) is the line that needs to be addressed so the `shortcut` type behaves more like `multiple` by receiving an array of index values. However, I could be missing something.

zwolf:
Shortcut tasks, like multiples, should export as their values an array instead of a string. This is the only difference between these two tasks. This, along with the rest of the exporter, is on the list of things to refactor.

Fixes #2228 

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -51,7 +51,7 @@ module Formatter
         {}.tap do |new_anno|
           new_anno['task'] = @current['task']
           new_anno['task_label'] = task_label(task_info)
-          new_anno['value'] = task_info['type'] == 'multiple' ? answer_labels : answer_label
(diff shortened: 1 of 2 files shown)


zwolf:
No one's using the shortcut task yet, no changes to anything else.

PR #2244, opened 2017-03-17
Optimize gold standard queries

camallen:
Avoid the join on public workflows, use an in query for filtering the gold standard index hits. Also use the filter_has_many serializer for proper paging urls in #2239. Will need a rebase after that is in.

# Review checklist

### DIFF
--- a/spec/controllers/api/v1/classifications_controller_spec.rb
+++ b/spec/controllers/api/v1/classifications_controller_spec.rb
@@ -214,7 +214,7 @@ describe Api::V1::ClassificationsController, type: :controller do
       end
 
       it 'should be filterable by a workflow id' do
-        another_gs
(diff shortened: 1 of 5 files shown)


spec/controllers/api/v1/classifications_controller_spec.rb
```diff
@@ -228,7 +228,7 @@ def create_gold_standard
         end
 
         it 'should be filterable by a list of subject ids' do
-          get :gold_standard, subject_ids: gs.subject_ids.join(',')
```
marten: Won't this break the front-end (or potentially other API users)?

camallen: I though the same thing and then wondered if anyone was using it as it's not documented. Perhaps i'll just add a before filter in the controller to map it to the right key for that filter.

PR #2248, opened 2017-03-20
Check for retirements all at once, remove n+1

### LINKED ISSUE #2236: Remove retired N+1 in subject selector serializer
Modify the lookup [here](https://github.com/zooniverse/Panoptes/blob/942f74d1fe23b1667cb1e37f7544a22cdee30780/app/serializers/subject_selector_serializer.rb#L25) for the retired subject status to not be an N+1. Probably push it into the selection context like the favourites finder did to consolidate the db work.

zwolf:
Moves the checking of whether a subject is retired for a given workflow to the selector context and asks about them all at once rather than one at a time in the serializer. A one liner that might not need it's own lib, but looks nice next to FavoriteFinder and I got to call a method `#find_retirees`. 

Good eye, btw. Is there a good way to test for these? I tried testing this with Bullet but it didn't detect this one in particular. Didn't expect it to but it's probably just the non-obvious ones left.

Fixes #2236 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -119,6 +119,7 @@ class Api::V1::SubjectsController < Api::ApiController
         user_seen: user_seen,
         url_format: :get,
         favorite_subject_ids: FavoritesFinder.new(api_user.user, workflow.project, selected_subject_ids).find_favorites,
+        retired_for_workflow: SubjectWorkflowRetirements.new(workflow, selected_subject_ids).find_retirees,
         select_context: true
       }.compact
     end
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/subjects_controller.rb
```diff
@@ -119,6 +119,7 @@ def selector_context(selected_subject_ids)
         user_seen: user_seen,
         url_format: :get,
         favorite_subject_ids: FavoritesFinder.new(api_user.user, workflow.project, selected_subject_ids).find_favorites,
+        retired_for_workflow: SubjectWorkflowRetirements.new(workflow, selected_subject_ids).find_retirees,
```
camallen: maybe name `retired_subject_ids` for consitency with `favorite_subject_ids`

camallen:
> Is there a good way to test for these?

Yeah - with real data and logs in the serializer or check out new relic transactions specifically the `Avg calls (per txn)` anything above 1 for each action means we're looking them up multiple times. Check the `Api::V1::SubjectsController#queued` pre and post this deploy for details.

Nice work 👍

zwolf:
Fixes https://github.com/zooniverse/Panoptes/issues/2236

PR #2249, opened 2017-03-20
Duplicate gold standard classifications to own model for faster lookups

camallen:
Searching for gold standard classifications is not efficient over the classifications table space. Long term we're moving to a different summary mechanism for the UI. Isolate the existing gold standard classifications in their own table for fast index lookup 

TODO
- [x] Hook the classification scope_for to use this model
~- [ ] Remove sparse gold standard classifications index~ do this in a future PR after data is migrated
- [x] Add rake task to migrate classification data
- [x] Test migration rake task

This can be deployed by 
1. running the migration to add the table
2. run the rake task to migrate the gold standard data
3. deploy the code to start serving the migrated gold standard data 

# Review checklist

### DIFF
--- a/spec/models/classification_spec.rb
+++ b/spec/models/classification_spec.rb
@@ -239,7 +239,7 @@ describe Classification, :type => :model do
           project: project, user: project.owner
         )
         gsa.workflow.update_column(:public_gold_standard, true)
-        expect(Classification.scope_for(:gold_standard, user)).to match_array(gsa)
+        expect(Classification.scope_for(:gold_standard, user)).to match_array([gsa])
       end
     end
   end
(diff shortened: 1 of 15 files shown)


spec/models/classification_spec.rb
```diff
@@ -235,11 +235,11 @@
...
     describe "#gold_standard" do
       it 'should return all gold_standard project data' do
-        gsc = create(:gold_standard_classification,
+        gsa = create(:gold_standard_annotation,
           project: project, user: project.owner
         )
-        gsc.workflow.update_column(:public_gold_standard, true)
-        expect(Classification.scope_for(:gold_standard, user)).to match_array(gsc)
+        gsa.workflow.update_column(:public_gold_standard, true)
+        expect(Classification.scope_for(:gold_standard, user)).to match_array(gsa)
```
zwolf: This seems awkward because `gsa` isn’t an array, consider using `to contain_exactly` instead

camallen: This should match an array and not an GoldStandardAnnotation

PR #2250, opened 2017-03-21
test no join parental controlled scopes

camallen:
avoid using joins for parental controlled scope access and paying the merge costs at the db. Instead use the ANY parent foreign_key index lookup (all parental controlled models have a belongs_to FK to the parent). This should decrease query execution time and speed up count queries from the serializer and use less database resources dealing with these joins.

~~For proper testing [these](https://github.com/zooniverse/Panoptes/blob/dd24acd0d28893434f928a61780cc532a9cb6ca1/lib/role_control/parental_controlled.rb#L7) public / private scopes (used in the fallback scope_for) will need to use this method as well. I'll add these in if we think this is worthwhile.~~ I've added these.

# Review checklist


lib/role_control/parental_controlled.rb
```diff
@@ -22,7 +22,13 @@ def parent_class
 
       def scope_for(action, user, opts={})
         parent_scope = parent_class.scope_for(action, user, opts)
-        joins(@parent).merge(parent_scope)
+
+        if Panoptes.flipper["no_join_parental_scope"].enabled?
```
zwolf: non-blocking: still think it's kind of confusing to have a negative toggle and it should still be created in the Flipper UI so we've got a list of toggles.

camallen: Agree, initially i was writing this as a code experiment but realised (again) that creating scopes doesn't actually execute the code, then i kept the naming of the conditional branch as it's not default behaviour...i.e. what i want to test. 

Flipper is great in that you don't have to make a feature for it to be tested against, it'll happily load something that doesn't exist, it'll just never be enabled until you do so via the UI.

Issue #2252, opened 2017-03-21
Add feature flag to dump workers


zwolf:
CSV exporting should have a flipper toggle to store the job in sidekiq but not run it immediately. Raise an error instead.

Issue #2118, opened 2017-01-11
Feature flags for all non-essential features


marten:
For instance, most workers should have something like `reschedule_until_later if disabled_by_flipper?` as their first line, so that we can disable them easily if a specific one causes issues (locking, ...) under high loads.

There are probably plenty of other things that aren't workers.

zwolf:
superset that includes https://github.com/zooniverse/Panoptes/issues/2252

zwolf:
and https://github.com/zooniverse/Panoptes/pull/2262 and https://github.com/zooniverse/Panoptes/pull/2261 and https://github.com/zooniverse/Panoptes/pull/2263

PR #2258, opened 2017-03-21
Adds feature flag to dump workers

zwolf:
Adds a feature flag check to the dump workers for classifications, workflows, subjects, and aggregations. The API will return an error that says that exports are disabled. The sidekiq jobs will hit the flag, raise an exception, and fail however many (25?) times and then be put in the dead queue where we can run them later...or not.

I have been messing with sidekiq middleware all afternoon trying to conditionally set the retry amount but it doesn't really matter. If we want to reschedule these jobs instead of basically ignoring them, I'll need to know what that schedule looks like. **If we do flip this switch, we should ensure that PFE is going to do something smart with the response.**

Closes https://github.com/zooniverse/Panoptes/issues/2252 and https://github.com/zooniverse/Panoptes/issues/2118 (more or less, depending on rescheduling thoughts).

# Review checklist

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -17,7 +17,8 @@ module Api
       Subjects::Selector::MissingSubjectSet,
       Subjects::Selector::MissingSubjects,                 with: :not_found
     rescue_from ActiveRecord::RecordInvalid,               with: :invalid_record
-    rescue_from Api::LiveProjectChanges,                   with: :forbidden
+    rescue_from Api::LiveProjectChanges,
+      Api::ExportDisabled,                                 with: :forbidden
     rescue_from Api::NotLoggedIn,
       Api::UnauthorizedTokenError,
       Operation::Unauthenticated,                          with: :not_authenticated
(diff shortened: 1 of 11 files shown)


zwolf:
Forgot to rescue from the raised exception and have the controller respond with something. I don't immediately see a place to test this.

app/controllers/api/api_controller.rb
```diff
@@ -17,7 +17,8 @@ class ApiController < ApplicationController
       Subjects::Selector::MissingSubjectSet,
       Subjects::Selector::MissingSubjects,                 with: :not_found
     rescue_from ActiveRecord::RecordInvalid,               with: :invalid_record
-    rescue_from Api::LiveProjectChanges,                   with: :forbidden
+    rescue_from Api::LiveProjectChanges,
+      Api::ExportDisabled,                                 with: :forbidden
```
camallen: I don't think you need to expose this error as an api response. This error should only be detected in sidekiq after scheduling a dump to run and will just force sidekiq to retry and backoff over time, they'll just sit in the retry queue until we turn them back on.

zwolf: What happens on the front end when someone hits the button and expects to get an export? If we turn these off we should have a way of letting PFE as well as people who are hitting the API directly know that they're not gonna get anything.

camallen: it just gets queued up in the backend. It's a black box process already, the feedback then will be the same as it is already. Maybe we just put some text up there during SGL saying they may be delayed?

zwolf:
Closes https://github.com/zooniverse/Panoptes/issues/2252

PR #2259, opened 2017-03-22
move to IN sub select queries instead of JOINS on parental controlled

camallen:
remove the feature flag conditional behaviour and implement the IN avoiding join scopes for queries.

~**DO NOT MERGE THIS** until we've tested the behaviour and determined that we want to move away from the join scopes on all parental controlled resources.~

I've tested this live without incident.

# Review checklist

### DIFF
--- a/lib/role_control/parental_controlled.rb
+++ b/lib/role_control/parental_controlled.rb
@@ -20,6 +20,10 @@ module RoleControl
         @parent_class ||= @parent.to_s.camelize.constantize
       end
 
+      def parent_relation
+       @parent
+     end
+
       def parent_foreign_key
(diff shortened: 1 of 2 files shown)


lib/role_control/parental_controlled.rb
```diff
@@ -40,14 +24,12 @@ def parent_foreign_key
         reflect_on_association(@parent).foreign_key
       end
 
-      def scope_for(action, user, opts={})
-        parent_scope = parent_class.scope_for(action, user, opts)
+      def parent_scope(scope)
+        where(parent_foreign_key => scope.pluck(:id))
```
camallen: Don't use plucks here, use select and let rails create a sub -select query

PR #2262, opened 2017-03-22
Adds feature flag for subject uploading

zwolf:
When DISABLED, subject uploading will be disallowed. The flags already exist and are enabled on production and staging.

Also: Renamed the `ExportDisabled` error to `FeatureDisabled` and am sending it back with a 503. So far, subject uploads will be the only one to see it.

Also, I refactored the Flipper stuff a bit. Since we're accumulating more than a couple flags, they are now all enabled in a spec helper instead of at the top of each spec. Disables still go in their context block.

# Review checklist


zwolf:
Yeah, since the that PR depends on changes that I made to this one. If this is merged first, I'll rebase the other one.

PR #2263, opened 2017-03-22
Adds feature flag for classification lifecycle processing

zwolf:
Builds on the subject uploading feature flag and adds one for classification lifecycle processing. When it is on, lifecycle processing is done normally: thrown into sidekiq via `perform_async` and left in the queue. When the flag is turned off, the processing will be done immediately at the cost of performance but preventing a race condition on postgresql subject selection.

Flag exists and enabled on production and staging.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -64,15 +64,13 @@ class Api::V1::ClassificationsController < Api::ApiController
   end
 
   def lifecycle(action, classification)
-    begin
-      if Panoptes.flipper[:cls_lifecycle_queue].enabled?
-        ClassificationLifecycle.queue(classification, action)
-      else
-        ClassificationLifecycle.perform(classification, action.to_s)
-      end
+    if Panoptes.flipper[:cls_lifecycle_queue].enabled?
+      ClassificationLifecycle.queue(classification, action)
+    else
+      ClassificationLifecycle.perform(classification, action.to_s)
+    end
     rescue Redis::CannotConnectError, Redis::TimeoutError, Timeout::Error => e
       Honeybadger.notify(e)
-    end
   end
 
   def completed_error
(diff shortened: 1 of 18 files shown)


app/controllers/api/v1/classifications_controller.rb
```diff
@@ -64,9 +64,13 @@ def build_resource_for_create(create_params)
   end
 
   def lifecycle(action, classification)
-    ClassificationLifecycle.queue(classification, action)
+    if Panoptes.flipper[:cls_lifecycle_queue].enabled?
```
camallen: could this be named a little more clearly? makes it easier for users to understand it implications of the feature from the UI without needing to look at the code. maybe `classification_lifecycle_in_background` or invert it and do `classification_lifecycle_in_request`

spec/support/flipper.rb
```diff
@@ -0,0 +1,13 @@
+module Flipper
+  RSpec.configure do |config|
+    config.before(:each) do |example|
+      allow(Panoptes).to receive(:flipper).and_return(Flipper.new(Flipper::Adapters::Memory.new))
+      Panoptes.flipper["cellect"].enable
+      Panoptes.flipper["designator"].enable
+      Panoptes.flipper[:remove_complete_subjects].enable
```
camallen: are globally enabling these a good idea? I think the state is a little hidden in this file, the reactor makes sense so it's probably fine.

zwolf: That's what Flipper is doing for us already. I imagined this as setting the default state we expect from Flipper, rather than all being disabled by default and enabled with a before(:each) at the top of every file.

camallen: For some they are always on but they could be off...here we're just saying these are always on for the specs. If it becomes an issue (and i don't think it will) we can change it.

zwolf: They're turned off manually in the "This feature is disabled" context blocks. They are always on for the purposes of every other spec.

app/controllers/api/v1/classifications_controller.rb
zwolf: Yeah that name sucks ✔️

PR #2274, opened 2017-03-25
remove unused indexes to free up db memory

camallen:
Remove indexes that aren't being used in query paths. 

I ran this sql to find them, https://gist.github.com/camallen/4d46cc0853efefbf55fb7ed0a4a4afa3 You can modify the idx_scan value to find less used queries too.

# Review checklist

### DIFF
--- a/db/migrate/20170325135953_remove_unused_indexes_sgl2017.rb
+++ b/db/migrate/20170325135953_remove_unused_indexes_sgl2017.rb
@@ -1,0 +1,10 @@
+class RemoveUnusedIndexesSgl2017 < ActiveRecord::Migration
+  def change
+    # PRE SGL 2017 index cleanup
+    remove_index :recents, :classification_id
+    remove_index :set_member_subjects, :priority
+    remove_index :oauth_applications, [:owner_id, :owner_type]
+    remove_index :subject_queues, [:workflow_id, :user_id]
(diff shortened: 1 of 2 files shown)


db/migrate/20170325135953_remove_unused_indexes_sgl2017.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveUnusedIndexesSgl2017 < ActiveRecord::Migration
+  def change
+    # PRE SGL 2017 index cleanup
+    remove_index :recents, :classification_id
+    remove_index :set_member_subjects, :priority
+    remove_index :oauth_applications, [:owner_id, :owner_type]
+    remove_index :subject_queues, [:workflow_id, :user_id]
+    remove_index :subjects, :activated_state
```
camallen: This one seems to not being hit in the selector query [here](https://github.com/zooniverse/Panoptes/blob/91ec95d77b682dd4e0715d9bdff64b225e4fb949/lib/subjects/selector.rb#L79) as the db is most likely running a filter over the small results instead. 

Also this could be actively used if we start looking at rake task to clean up orphaned subjects. It's currently 196MB. I'm keen to remove it and add it back in if we start cleaning up orphans on a schedule.

zwolf:
I concur! LGTM.

Issue #2251, opened 2017-03-21
Add `?sort=random` to workflows end point


srallen:
As a follow up to zooniverse/Panoptes-Front-End#3556:

If a project does not have a set default workflow and if the user is not logged in or if the user logged in does not have any stored workflow ids in their project preferences, the fallback behavior for the front end is to randomly select a workflow. Previously, PFE would request for all active workflows with an arbitrarily large page size and then select a random index out of the returned array of resources. The above mentioned PR changes this behavior to instead randomly select an index out of a project's `links.workflows` array and then attempts to request this workflow with `active=true` query param. If it returns undefined, this means the workflow is inactive and PFE will attempt again. What we've found is that there are projects with lots of inactive workflows, so it may take some time for PFE to randomly select an active workflow. 

Instead it would be nice if the API had a `?sort=random` query param for the workflows end point to return a random active workflow from the project links so that there is no attempt to request an inactive workflow at all. This also means we wouldn't have to depend on projects to clean up inactive workflows and/or make sure a workflow is set as default in their projects too.

camallen:
Yeah this will save a bunch of requests from the client, we should try and do this pre SGL.

camallen:
@srallen could we instead serialize the `active_workflows` links on the project that you can randomly select from? e.g. 
```
links: {
  workflows: ["3609", "1687", "2227", "3541", "3522"],
  active_workflows: [ "3609" ]
}
```

I think this may be more useful long term and isn't much extra work for the api.

srallen:
@camallen Yeah that would work for solving this specific issue. @zwolf and I talked a bit yesterday and thought that at some point in the future a random sort might be useful for other resources, but having something in place prior to SGL for this would be good. 👍

PR #2277, opened 2017-03-26
Include list of active workflows in serialized projects

zwolf:
Only bummer here might be if the `@model.workflows.where(active: true)` call in the project serializer is making an extra db query, but hopefully it's just filtering what it's already pulled. Could have put that method on the model, but made more sense this way to me just now.

Closes https://github.com/zooniverse/Panoptes/issues/2251

# Review checklist

### DIFF
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -29,7 +29,7 @@ class ProjectSerializer
     :completeness, :activity, :state, :researcher_quote
 
   optional :avatar_src
-  can_include :workflows, :subject_sets, :owners, :project_contents,
+  can_include :workflows, :active_workflows, :subject_sets, :owners, :project_contents,
     :project_roles, :pages, :organization
   media_include :avatar, :background, :attached_images,
     classifications_export: { include: false},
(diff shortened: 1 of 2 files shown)


app/serializers/project_serializer.rb
```diff
@@ -126,6 +126,10 @@ def avatar_src
     end
   end
 
+  def active_workflows
+    @model.workflows.where(active: true)
```
camallen: This may result in a query and an N+1 in the serializer page method. You should just be able to use the project#active_workflows relation [here](https://github.com/zooniverse/Panoptes/blob/91ec95d77b682dd4e0715d9bdff64b225e4fb949/app/models/project.rb#L21) instead of manually running the query. If you do that then you should also preload that relation in the serializer page [here](https://github.com/zooniverse/Panoptes/blob/91ec95d77b682dd4e0715d9bdff64b225e4fb949/app/serializers/project_serializer.rb#L54)

Maybe even migrate that to use the https://github.com/zooniverse/Panoptes/blob/91ec95d77b682dd4e0715d9bdff64b225e4fb949/lib/serialization/panoptes_restpack.rb as well. You may still need to keep the page class method override to get the cards preload sorted too.

PR #2280, opened 2017-03-28
cache more serialiser outputs where we can

camallen:
avoid regenerating the json for the resources, do it once, cache them and serve that. 

This should help with the extra data base N+1 loads happening with all the new `?include=` sideloading that PFE is requesting. Long term we should either fix restpack to respect the model preloads when sideloading or move to another serializer lib. 

# Review checklist

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -2,6 +2,7 @@ class UserSerializer
   include Serialization::PanoptesRestpack
   include RecentLinkSerializer
   include MediaLinksSerializer
+  include CachedSerializer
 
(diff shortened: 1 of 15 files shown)


app/serializers/user_serializer.rb
```diff
@@ -2,6 +2,7 @@ class UserSerializer
   include Serialization::PanoptesRestpack
   include RecentLinkSerializer
   include MediaLinksSerializer
+  include CachedSerializer
```
camallen: This one will mean we don't get the correct uploaded_subjects count being serialized until the cache resets. If this is an issue i'll look at either touching the timestamp on the resource for the upload subject event or remove the cache.

PR #2284, opened 2017-03-29
only return resource under control

camallen:
make sure we don't test on polymorphic resource id's but only on the controlled type we're testing.

This will need some testing in staging to make sure there aren't any complications as it's been like this for a while. We could in theory be selecting the wrong resources based on other table id's in the private query.

# Review checklist


marten:
Good catch

PR #2296, opened 2017-04-06
Add default subject to Collections

zwolf:
Collections now have a default subject field. It's serialized as either the media's first location's URL or just the collection's first subject's if nil. This creates an extra db query for the medium but prevents an entire extra API call from PFE for the first linked subject just to get its URL.

I need to test this on staging with the JS client and make sure it's set up to handle however it builds requests (via `/api/collections/:id` and `/api/collections/:id/links/default_subject`)

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2917,6 +2917,7 @@ It has the following attributes:
 - updated_at
 - name
 - display_name
+- default_subject
 
 *id*, *created_at*, and *updated_at* are set by the API.
 
@@ -2937,6 +2938,7 @@ It has the following attributes:
                     "updated_at": "2014-04-20T06:23:12Z",
                     "name" : "flowers",
                     "display_name": "Lots of Pretty flowers",
+                    "default_subject_src": "panoptes-uploads.zooniverse.org/production/subject_location/hash.jpeg",
                     "links": {
                         "owner": {
                             "id": "10",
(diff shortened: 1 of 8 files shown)


db/migrate/20170403194826_add_default_subject_to_collections.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultSubjectToCollections < ActiveRecord::Migration
+  def change
+    add_column :collections, :default_subject_id, :integer
```
marten: Should we get a foreign key on this too?

apiary.apib
```diff
@@ -3047,6 +3050,33 @@ owner or have edit permissions.
 
 + Response 204
 
+## Add default subject link [/collections/{id}/links/default_subject]
+Links a default subject to a collection. This subject's first media
+location URL will be included in the serialized collection and used
+as the thumbnail. Update this attribute with `null` to use the first
```
marten: Should the backend be taking care of that default? It doesn't seem to be in this PR.

zwolf: This? https://github.com/zwolf/Panoptes/blob/6a1f7806ad2c5bf2e4a0de2b8c98d9ae7ec98d56/app/serializers/collection_serializer.rb#L26

db/migrate/20170403194826_add_default_subject_to_collections.rb
zwolf: ✔️

apiary.apib
zwolf: Or if you mean the route, it's handled here

https://github.com/zooniverse/Panoptes/pull/2296/files#diff-21497849d8f00507c9c8dcaf6288b136R124

and setting is allowed via this:
https://github.com/zooniverse/Panoptes/pull/2296/files#diff-e5f000780a616ca07cd53cdbbfdbfd59R14

marten: Yeah like that serializer. Missed that, 👍

PR #2303, opened 2017-04-19
Add :created_at attribute to UPP serializer

srallen:
This adds `:created_at` attribute to the user project preferences serializer. The first a/b split that Gravity Spy ran was to study behavior with users new to the project and the only way we can determine that is by the existence of the user project preferences for the project. If they want to run any future experiments with new users, it will be useful to have the `created_at` timestamp available via the API and possibly log it to geordi for the researcher's use in data analysis.

I looked through the spec files and didn't see anything that I would need to update in the tests, but do let me know if I missed something. 

# Review checklist

### DIFF
--- a/app/serializers/user_project_preference_serializer.rb
+++ b/app/serializers/user_project_preference_serializer.rb
@@ -3,7 +3,8 @@ class UserProjectPreferenceSerializer
   include CachedSerializer
 
   attributes :id, :email_communication, :preferences, :href,
-    :activity_count, :activity_count_by_workflow, :settings
+    :activity_count, :activity_count_by_workflow, :settings,
+    :created_at
   can_include :user, :project
   can_sort_by :updated_at, :display_name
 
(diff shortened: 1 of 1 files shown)


srallen:
Cool, I'll take a look at the travis failure. Definitely missed something.

marten:
To me that failure seemed like a random failure that you didn't cause. This one looks like the test didn't take into account that Time.now can tick over to the next second while a test runs (it was expecting a timestamp like 12:34:56 but got 12:34:57). But again, I think it's not your fault. I told Travis to go retry the test suite, we'll see if it magically passes next time.

app/serializers/user_project_preference_serializer.rb
```diff
@@ -3,7 +3,8 @@ class UserProjectPreferenceSerializer
   include CachedSerializer
 
   attributes :id, :email_communication, :preferences, :href,
-    :activity_count, :activity_count_by_workflow, :settings
+    :activity_count, :activity_count_by_workflow, :settings,
+    :created_at
```
marten: Might want to add `:updated_at` while you're at it.

srallen:
Hm, good to know about the test. I only briefly looked over the logs before I left yesterday.

PR #2304, opened 2017-04-20
Add project/workflow mobile_friendly

marten:
This lets the mobile app filter the projects listing to just projects that have been marked as mobile-friendly. This can be set by owners via the API since we don't want to have to manage this long-term. 

Short term we probably will be setting this ourselves without any UI, we'll think about adding it to the project builder at some point.

### DIFF
--- a/app/schemas/project_create_schema.rb
+++ b/app/schemas/project_create_schema.rb
@@ -38,6 +38,10 @@ class ProjectCreateSchema < JsonSchema
       type "boolean"
     end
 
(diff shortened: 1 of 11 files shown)


marten:
@robinschaaf Once this merges/deploys you can add `?mobile_friendly=true` to the API call fetching the projects.

robinschaaf:
Great - thanks!!

marten:
@zwolf Ready for review

PR #2306, opened 2017-04-21
Collection Default Subject updates

zwolf:
* Adds proper URL formatting to the default subject src URL returned from the collection serializer.
* If a subject is being unlinked from a collection, check to see if it's the current default. If it is, null the default subject before unlinking.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/collections_controller.rb
+++ b/app/controllers/api/v1/collections_controller.rb
@@ -33,7 +33,7 @@ class Api::V1::CollectionsController < Api::ApiController
   private
 
   def check_default_subject(collection)
-    collection.update({ default_subject: nil }) if collection.default_subject&.id.to_s == params["link_ids"]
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/collections_controller.rb
```diff
@@ -19,6 +19,10 @@ class Api::V1::CollectionsController < Api::ApiController
     query.search_display_name(name.join(" "))
   end
 
+  def destroy_links
+    super { |collection| check_default_subject(collection) }
```
marten: Is there a test asserting that it does nullify?

zwolf:
Noticed that what I wrote wouldn't work if an array of subjects was given instead of a string. Fixed that and added tests, too.

app/controllers/api/v1/collections_controller.rb
zwolf: there are now!

PR #2309, opened 2017-04-26
Organization pages

zwolf:
Adds pages to organizations, just like projects. This is mostly duplicated code from the projects version of pages. I would like to extract it all out and DRY it all up, but with only these two uses it seems a bit like a waste of effort. This was everyone's take when orgs were first dropped in (new controller scheme aside) and it's still true. When something comes along that can make use of this code again, I think that'll be the time.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/organization_pages_controller.rb
+++ b/app/controllers/api/v1/organization_pages_controller.rb
@@ -1,0 +1,33 @@
+class Api::V1::OrganizationPagesController < Api::ApiController
+  require_authentication :update, :create, :destroy, scopes: [:organization]
+
+  resource_actions :default
+
+  schema_type :strong_params
+
+  allowed_params :create, :url_key, :title, :content, :language
+  allowed_params :update, :url_key, :title, :content, :language
+
+  before_filter :set_language_from_header, only: [:index]
+
+  def controlled_resources
+    @controlled_resouces ||= super.where(organization: params[:organization_id])
+  end
+
+  protected
+
+  def build_resource_for_create(create_params)
+    create_params[:links] ||= {}
+    create_params[:links][:organization] = params[:organization_id]
+    super create_params
+  end
+
+  def link_header(resource)
+    resource = resource.first
+    send(:"api_#{ resource_name }_url", id: resource.id, organization_id: resource.organization_id)
+  end
+
+  def set_language_from_header
+    params[:language] = "en"
+  end
+end
--- a/app/serializers/organization_page_serializer.rb
+++ b/app/serializers/organization_page_serializer.rb
@@ -13,7 +13,7 @@ class OrganizationPageSerializer
 
     organization_id = options.filters.delete :organization_id
 
-    url = "#{self.href_prefix}/organizations/#{organization_id.try(:first)}/pages"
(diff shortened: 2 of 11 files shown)


camallen:
can you fix the hound warnings and leave the whitespace in the structure.sql (regenerate it via the migration task) as it's auto-generated so it'll have whitespace the next time we create a migration.

app/controllers/api/v1/organization_pages_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    super create_params
+  end
+
+  def link_header(resource)
+    resource = resource.first
+    send(:"api_#{ resource_name }_url", id: resource.id, organization_id: resource.organization_id)
+  end
+
+  def set_language_from_header
+    params[:language] = "en"
```
camallen: this is a straight copy but we should probably note that it's a working solution. we need this to actually respect the headers with a default back to english.

app/serializers/organization_page_serializer.rb
```diff
@@ -0,0 +1,36 @@
+class OrganizationPageSerializer
+  include RestPack::Serializer
```
camallen: can you use the restpack serializer wrapper here instead and setup preloading to avoid N+1 for linked resources.

app/serializers/organization_page_serializer.rb
```diff
@@ -0,0 +1,36 @@
...
+  include CachedSerializer
+
+  attributes :id, :href, :created_at, :updated_at, :url_key, :title,
+    :language, :content, :type
+
+  can_include :organization
+  can_filter_by :url_key, :language
+
+  def self.page_href(page, options)
```
camallen: this is pretty much a straight copy coudn't we extract the code to a module that reflected on the resource class / id and just include that to get the behaviour? understand if you'd prefer not to

zwolf: That's true for a lot of this, as I said up top. Obviously it'd be better if all the duped code was cleaned up. This was faster, and short of adding another level of organization here we won't need it again. I can extract this all out for the serializers, controllers, etc., if you'd prefer.

zwolf: Might be better in a separate PR, to let the lab app devs use this in the meantime.

zwolf:
Guess the projects code for same predates hound, fixed both.

app/serializers/organization_page_serializer.rb
camallen: no problems.

db/structure.sql
```diff
@@ -2568,14 +2632,14 @@ CREATE INDEX index_projects_on_migrated ON projects USING btree (migrated) WHERE
 
 
 --
--- Name: index_projects_on_mobile_friendly; Type: INDEX; Schema: public; Owner: -
+-- Name: index_projects_on_mobile_friendly; Type: INDEX; Schema: public; Owner: -; Tablespace:
```
camallen: Any ideas why these tablespace directives crept in? Not a blocker.

zwolf: Happened when I updated the structure, presumably because I had already updated my database from the master branch and a more recent migration. Should merge fine.

camallen: OOI what postgres server version is it?

PR #2310, opened 2017-05-01
Set organization title by display_name param

zwolf:
Title is no longer a required param for create and is removed from the update schema. It is set to the display name for the content's language.

listed_at is now serialized

# Review checklist


zwolf:
It is updated to be the same as the display name for the primary (or specified) language, so that there is consistency with how projects do the same thing.

PR #2313, opened 2017-05-09
Export classifications in segments

marten:
Describe your change here.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/classifications_export_segments_controller.rb
+++ b/app/controllers/api/v1/classifications_export_segments_controller.rb
@@ -1,14 +1,14 @@
 class Api::V1::ClassificationsExportSegmentsController < Api::ApiController
   require_authentication :all, scopes: [:project]
 
+  before_action :require_access_to_project, only: [:create]
+
   resource_actions :default
 
   def create
     classifications_export_segments = ClassificationsExportSegment.transaction do
       Array.wrap(params[:classifications_export_segments]).map do |segment_params|
-        segment_params[:links] ||= {}
-        segment_params[:links][:project] = params[:project_id]
-        ClassificationsExportSegments::Create.with(api_user: api_user).run!(segment_params)
+        ClassificationsExportSegments::Create.with(api_user: api_user, project: @project).run!(segment_params)
       end
     end
 
@@ -21,10 +21,12 @@ class Api::V1::ClassificationsExportSegmentsController < Api::ApiController
 
   protected
 
-  def build_resource_for_create(create_params)
-    create_params[:links] ||= {}
-    create_params[:links][:project] = params[:project_id]
-    super create_params
+  def require_access_to_project
+    @project = Project.scope_for(:update, api_user, {}).where(id: params[:project_id]).first
+
+    unless @project
+      raise Api::Unauthorized
+    end
   end
 
   def link_header(resource)
--- a/app/controllers/api/v1/classifications_export_segments_controller.rb
+++ b/app/controllers/api/v1/classifications_export_segments_controller.rb
@@ -5,10 +5,20 @@ class Api::V1::ClassificationsExportSegmentsController < Api::ApiController
 
   schema_type :strong_params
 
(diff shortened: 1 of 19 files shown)


db/migrate/20170509105754_create_classifications_export_segments.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.integer :last_classification_id, null: false
+      t.integer :requester_id, null: false
+
+      t.datetime :started_at
+      t.datetime :finished_at
+
+      t.timestamps null: false
+    end
+
+    add_foreign_key :classifications_export_segments, :classifications, column: :first_classification_id
```
marten: Add `NOT VALID` as per https://joinhandshake.com/engineering/2017/02/27/database-migrations-and-locks.html

app/operations/classifications_export_segments/create.rb
```diff
@@ -0,0 +1,42 @@
...
+        segment
+      else
+        nil
+      end
+    end
+
+    private
+
+    def workflow
+      @workflow ||= Workflow.find(links[:workflow])
```
marten: This doesn't do any permissions checking!

spec/controllers/api/v1/classifications_export_segments_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    let(:index_params) { {project_id: project.id} }
+    let(:n_visible) { 1 }
+
+    it_behaves_like "is indexable", false
+
+    it 'should not allow access to projects the user is not a collaborator on' do
+      project2 = create :project
+      segment2 = create :classifications_export_segment, project: project
+      get :index, project_id: project2.id
+      expect(response.status).to eq(401)
```
camallen: Is this 401 coming from the create operation?

spec/factories/classifications.rb
```diff
@@ -14,8 +14,9 @@
     workflow_version "15.15"
     completed true
     user
-    project
     workflow
+    project { workflow&.project || build(:project) }
```
camallen: nice

app/models/classifications_export_segment.rb
```diff
@@ -0,0 +1,56 @@
...
+    segment = self.class.new(project_id: project_id, workflow_id: workflow_id, requester: requester)
+    segment.set_first_last_classifications(last_classification_id)
+    segment
+  end
+
+  def set_first_last_classifications(after)
+    scope = complete_classifications
+    scope = scope.where("id > ?", after) if after.present?
+
+    res = scope.select("min(id), max(id)").limit(1).load[0]
```
marten: Not performant, really:

```
[6] pry(main)> z.set_first_last_classifications(nil)
  Classification Load (24700.2ms)  SELECT  min(id), max(id) FROM "classifications" WHERE "classifications"."workflow_id" = $1 AND "classifications"."completed" = $2 LIMIT 1  [["workflow_id", 338], ["completed", "t"]]
```

I'm open to better solutions if you can think of any?

app/operations/classifications_export_segments/create.rb
marten: Don't think so, I'll write a test to make sure.

app/controllers/api/v1/classifications_export_segments_controller.rb
```diff
@@ -0,0 +1,36 @@
+class Api::V1::ClassificationsExportSegmentsController < Api::ApiController
+  require_authentication :all, scopes: [:project]
+
+  before_action :require_access_to_project, only: [:create]
```
marten: Yeah, index and show are scoped through the `can_through_parent` stuff. But the `create` isn't using the normal stuff.

zwolf:
For currently problematic workflows/projects, we'll want to be able to create smallish segments to get through the backlog of huge annotations/tasks. Is that something we might want to be able to do via rake task?

PR #2316, opened 2017-05-10
update retired counts and completeness metrics when linking new data sets

### LINKED ISSUE #2201: completion bar doesn't update very quickly when you swap subject sets
I've added subjects to P4: Ridges and nearly 12 hours later the project still says 100% complete. Is there a way that changing subject sets could trigger a recalculation of the progress bar? 

ba71526c-fa14-11e6-837b-267fea150e53

camallen:
closes #2201 - recount the retired currently linked subjects and recalculate the completeness metrics. This should ensure the stats are updated and the finished banner doesn't show when new data is uploaded after a project finished.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -87,6 +87,7 @@ class Api::V1::WorkflowsController < Api::ApiController
           NotifySubjectSelectorOfRetirementWorker.perform_async(subject_id, workflow.id)
         end
       when :subject_sets, 'subject_sets'
+        CalculateProjectCompletenessWorker.perform_async(workflow.project_id)
         NotifySubjectSelectorOfChangeWorker.perform_async(workflow.id)
       end
     end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -82,15 +83,13 @@ def post_link_actions(workflow)
     if workflow.set_member_subjects.exists?
       case relation
       when :retired_subjects, 'retired_subjects'
-        WorkflowRetiredCountWorker.perform_async(workflow.id)
-
         params[:retired_subjects].each do |subject_id|
           NotifySubjectSelectorOfRetirementWorker.perform_async(subject_id, workflow.id)
         end
       when :subject_sets, 'subject_sets'
+        CalculateProjectCompletenessWorker.perform_async(workflow.project_id)
```
camallen: There is no guarantee that this will run before the retired count worker unfortunately.

PR #2320, opened 2017-05-17
Loading constants

camallen:
reworked the code paths and require statements to use rails auto / eager loading where we can. This should fix random errors with unloaded constants and load errors.

Might be best to pull this down and check it works for your dev / test envs too before we merge.

This code was super helpful looking into the autoload stuff
```
d = ActiveSupport::Dependencies
d.logger = Logger.new(STDOUT)
d.log_activity = true

Formatters::Csv::Classification

d.autoloaded_constants.select { |t| t.match(/BelongsTo/) }
```

# Review checklist

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -13,11 +13,22 @@ Bundler.require(*Rails.groups)
 
 module Panoptes
   class Application < Rails::Application
-    config.autoload_paths += Dir[Rails.root.join('app', 'models', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('app', 'serializers', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('app', 'workers', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('app', 'operations', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('lib', '**/')]
+    # config.eager_load_paths += Dir[Rails.root.join('app', 'models', '*/')]
+    # config.eager_load_paths += Dir[Rails.root.join('app', 'workers', '*/')]
+    # config.eager_load_paths += Dir[Rails.root.join('app', 'operations', '*/')]
+    # config.eager_load_paths += Dir[Rails.root.join('app', 'serializers', '*/')]
+    # config.eager_load_paths += Dir[Rails.root.join('app', 'formatters', '*/')]
+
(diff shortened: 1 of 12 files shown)


config/application.rb
```diff
@@ -13,11 +13,13 @@
 
 module Panoptes
   class Application < Rails::Application
-    config.autoload_paths += Dir[Rails.root.join('app', 'models', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('app', 'serializers', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('app', 'workers', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('app', 'operations', '**/')]
-    config.autoload_paths += Dir[Rails.root.join('lib', '**/')]
+    config.eager_load_paths += Dir[Rails.root.join('app', 'models', '*/')]
```
camallen: eager load the code where we can instead of just providing the paths for lookup when searching for constants.

Note: the '*/' to only load the sub dirs!

lib/gem_ext/gem_ext.rb
```diff
@@ -0,0 +1,10 @@
+require_dependency 'gem_ext/doorkeeper/application'
```
camallen: I could not for the life of me avoid a dup enum warning without using the require_dependency on initialize. So that's why this is here...

config/application.rb
marten: Does that break reloading in development mode?

marten:
Wondering about the reloading, but LGTM

config/application.rb
camallen: Seems to be working fine, Ialso tested a server view too
``` ruby
[1] pry(main)> User
=> User (call 'User.connection' to establish a connection)
[2] pry(main)> User.test_call
NoMethodError: undefined method `test_call' for User (call 'User.connection' to establish a connection):Class
from /home/allenc/.rvm/gems/ruby-2.3.3@panoptes/gems/activerecord-4.2.5.2/lib/active_record/dynamic_matchers.rb:26:in `method_missing'
[3] pry(main)> reload!
Reloading...
=> true
[4] pry(main)> User.test_call
:Reloading: is working
=> nil
[5] pry(main)> 
```

PR #2323, opened 2017-05-19
Adds a description text field to collections

zwolf:
The only thing here is the `default: ""` and subsequent `Collection.update_all(display_name: "")`. There are 60k of them and so I think this should be fine, but if anyone's worried, maybe there's in a more forgiving way.

# Review checklist

### DIFF
--- a/db/migrate/20170519181110_add_description_to_collections.rb
+++ b/db/migrate/20170519181110_add_description_to_collections.rb
@@ -1,6 +1,7 @@
 class AddDescriptionToCollections < ActiveRecord::Migration
   def change
-    add_column :collections, :description, :text, default: ""
-    Collection.update_all(display_name: "")
+    add_column :collections, :description, :text
+    Collection.update_all(description: "")
+    change_column_default(:collections, :description, "")
   end
 end
--- a/db/migrate/20170519181110_add_description_to_collections.rb
+++ b/db/migrate/20170519181110_add_description_to_collections.rb
@@ -1,0 +1,6 @@
+class AddDescriptionToCollections < ActiveRecord::Migration
+  def change
+    add_column :collections, :description, :text, default: ""
+    Collection.update_all(display_name: "")
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20170519181110_add_description_to_collections.rb
```diff
@@ -0,0 +1,6 @@
+class AddDescriptionToCollections < ActiveRecord::Migration
+  def change
+    add_column :collections, :description, :text, default: ""
```
marten: How many collections do we have, and will this block the database too long?

zwolf: see description of this PR. Is there a better way?

camallen: Check "the bad" section of this blog post for details, https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/

camallen: Easiest way is to add the column without a default, backfill it using a rake task / migration and then enforce a default value in a different migration. My understanding is that adding a column with a default will force the table to be copied to a new location on disk (with the new col defaults) and lock while it does so.

zwolf: Is 60k "high volume"? I knew about backfilling and was just avoiding a multiple-deploy resolution. I guess any table lock on collections is bad and I'll do the above, but what's the cutoff for this?

camallen: It's all about timing, if that table is locked for a while and people are trying to write to it then it'll impact on the site. I guess test the run time on a snapshot for metrics. 

You can always backfill and add the default in 1 migration / transaction too. I believe this will avoid avoid any locks.
```
add_column :collections, :description, :text
Collection.update_all(description: "")
change_column_default :collections :description ""
```

zwolf: I love this.

Issue #2318, opened 2017-05-15
Organization pages return empty array


srallen:
Calling `GET` on organization pages returns an empty array when there should be one page. With the panoptes-javascript-client `GET` on `https://panoptes-staging.zooniverse.org/api/organizations/6/pages?http_cache=true` returns a `200` response and the following:
```
{
  "organization_pages": [],
  "links": {
    "organization_pages.organization": {
      "href": "/organizations/{organization_pages.organization}",
      "type": "organizations"
    }
  },
  "meta": {
    "organization_pages": {
      "page": 1,
      "page_size": 20,
      "count": 0,
      "include": [],
      "page_count": 0,
      "previous_page": null,
      "next_page": null,
      "first_href": "/organizations/6/pages?language=en",
      "previous_href": null,
      "next_href": null,
      "last_href": "/organizations//pages?page=0&language=en"
    }
  }
}
```
This would be fine, except I had previously created a page and would expect that one to be in the array of the response of the GET request. If I make a `POST` for the page I had thought I had created, I get a `400` with the response `errors: [{message: "Validation failed: Url key has already been taken"}]`.

camallen:
@srallen i assume the page was linked to org 6? Thoughts @zwolf ?

srallen:
Yep, id `6` is the org I'm working with on staging.

zwolf:
This is because the `/pages#index` endpoint (of both projects and orgs) is open and doesn't check authentication. As such, it's filtering out all the private orgs and projects. For project, it's checking the `private` boolean, and for orgs it's checking `where(listed_at: nil)`. 

I haven't looked, but I'm guessing the way PFE is actually getting the project pages is with a call to `/projects/:id?include=pages`. Adding that to orgs should get you where you wanna be. PR incoming.

srallen:
We found this is actually happening because of a mismatch of the language codes between the organization resource's `primary_language` and the contents' `language` property (`en-US` vs the hardcoded `en` language on pages right now). The API probably shouldn't 500 for this.

eatyourgreens:
I'm not sure about the best place to note this, but a request for a dialect should really fall back to the root language if the dialect isn't available. So a request for `en-US` should return resources flagged `en` instead of nothing.

We can possibly handle this in the front end by making a second request for the general language, if the request for a specific dialect fails.

[...]

PR #2324, opened 2017-05-19
Optionally include Pages in serialized organizations

zwolf:
Reference: #2318 

Just stuck pages into the org serializer's can_include. Index methods aren't authenticated and therefore won't show you pages from private projects or orgs.

EXTRA CREDIT: I took the opportunity to actually extract out the page controller stuff from the org and project controllers into it's own concern. 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/organization_pages_controller.rb
+++ b/app/controllers/api/v1/organization_pages_controller.rb
@@ -1,17 +1,4 @@
 class Api::V1::OrganizationPagesController < Api::ApiController
+  PARENT_RESOURCE = :organization
   include Pages
-
-  require_authentication :update, :create, :destroy, scopes: [:organization]
-
-  def parent_resource
-    :organization
-  end
-
-  def resource_id
-    params[:organization_id]
-  end
-
-  def resource_name
-    "organization_page"
-  end
 end
--- a/app/controllers/api/v1/organization_pages_controller.rb
+++ b/app/controllers/api/v1/organization_pages_controller.rb
@@ -1,4 +1,4 @@
 class Api::V1::OrganizationPagesController < Api::ApiController
-  PARENT_RESOURCE = :organization.freeze
+  PARENT_RESOURCE = :organization
   include Pages
 end
--- a/app/controllers/api/v1/organization_pages_controller.rb
+++ b/app/controllers/api/v1/organization_pages_controller.rb
@@ -1,4 +1,4 @@
 class Api::V1::OrganizationPagesController < Api::ApiController
-  PARENT_RESOURCE = :organization
+  PARENT_RESOURCE = :organization.freeze
   include Pages
 end
--- a/app/controllers/api/v1/organization_pages_controller.rb
+++ b/app/controllers/api/v1/organization_pages_controller.rb
@@ -1,33 +1,17 @@
 class Api::V1::OrganizationPagesController < Api::ApiController
+  include Pages
+
   require_authentication :update, :create, :destroy, scopes: [:organization]
 
-  resource_actions :default
-
-  schema_type :strong_params
-
-  allowed_params :create, :url_key, :title, :content, :language
-  allowed_params :update, :url_key, :title, :content, :language
-
-  before_filter :set_language_from_header, only: [:index]
-
-  def controlled_resources
-    @controlled_resouces ||= super.where(organization: params[:organization_id])
+  def parent_resource
+    :organization
   end
 
-  protected
-
-  def build_resource_for_create(create_params)
-    create_params[:links] ||= {}
-    create_params[:links][:organization] = params[:organization_id]
-    super create_params
+  def resource_id
+    params[:organization_id]
   end
 
-  def link_header(resource)
-    resource = resource.first
-    send(:"api_#{ resource_name }_url", id: resource.id, organization_id: resource.organization_id)
-  end
-
-  def set_language_from_header
-    params[:language] = "en"
+  def resource_name
+    "organization_page"
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/pages.rb
```diff
@@ -0,0 +1,43 @@
...
+    @controlled_resouces ||= super.where("#{parent_resource}": resource_id)
+  end
+
+  protected
+
+  def set_language_from_header
+    params[:language] = "en"
+  end
+
+  def serializer
```
zwolf: doh, fixed

app/controllers/api/v1/organization_pages_controller.rb
```diff
@@ -1,33 +1,17 @@
...
+  include Pages
 
-  before_filter :set_language_from_header, only: [:index]
-
-  def controlled_resources
-    @controlled_resouces ||= super.where(organization: params[:organization_id])
-  end
-
-  protected
+  require_authentication :update, :create, :destroy, scopes: [:organization]
```
camallen: this could be extracted to the shared module too
`require_authentication :update, :create, :destroy, scopes: [self.model_class.to_sym]` or something?

app/controllers/api/v1/organization_pages_controller.rb
```diff
@@ -1,33 +1,17 @@
...
-    resource = resource.first
-    send(:"api_#{ resource_name }_url", id: resource.id, organization_id: resource.organization_id)
+  def resource_id
+    params[:organization_id]
   end
 
-  def set_language_from_header
-    params[:language] = "en"
+  def resource_name
+    "organization_page"
```
camallen: Should we use a class constant for this instead? It's been a while with mem improvements but don't string allocations add up over time?

app/controllers/concerns/pages.rb
```diff
@@ -0,0 +1,39 @@
...
+    resource_actions :default
+
+    schema_type :strong_params
+
+    allowed_params :create, :url_key, :title, :content, :language
+    allowed_params :update, :url_key, :title, :content, :language
+
+    before_filter :set_language_from_header, only: [:index]
+
+    define_method(:link_header) do |resource|
```
camallen: OOI - why not just extract the instance method def outside the included block with the same protected?

zwolf: If those methods are just defined normally in the concern, they're pulled in but then overridden later by the creatable_resource lib. If they're defined in the included block they're the ones doing the overriding, which is what I want (and expected in the first place). I chalked it up to just order-of-operations weirdness in the hierarchy of includes and inheritance.

camallen: Interesting, the inheritance provides the DSL to wire up the resource actions and thus the call the include createable resource code only happens when that DSL is run. So i can see that running after that call is made in the include block can work. Any ruby magic to avoid this? If not maybe leave a comment on those lines for future.

app/controllers/api/v1/organization_pages_controller.rb
zwolf: Everything else is extracted (to the Pages concern). That particular line seems to run in a different context (self=concern vs. self=controller). I've got it working, but I'm going to have another look first thing tomorrow and see if I can't de-awkward it a bit.

zwolf:
Didn't find a more magical way to define those methods, did add a comment. The `require_authentication` line seems to run in a slightly different context than the rest of the included block (self = controller vs. the `Pages` include itself), making using the constants bit squirrely.

It's all working now, though moving forward I might rather have two similar lines than one awkward one.

app/controllers/api/v1/organization_pages_controller.rb
```diff
@@ -1,33 +1,4 @@
...
-
-  def link_header(resource)
-    resource = resource.first
-    send(:"api_#{ resource_name }_url", id: resource.id, organization_id: resource.organization_id)
-  end
-
-  def set_language_from_header
-    params[:language] = "en"
-  end
+  PARENT_RESOURCE = :organization
```
camallen: .freeze the constant but the api_controller has a method to get the model name from the controller [here](https://github.com/zooniverse/Panoptes/blob/d63945945be7849391e5754710fd11721d3ad66a/lib/json_api_controller.rb#L68)

app/controllers/api/v1/project_pages_controller.rb
```diff
@@ -1,33 +1,4 @@
...
-
-  def link_header(resource)
-    resource = resource.first
-    send(:"api_#{ resource_name }_url", id: resource.id, project_id: resource.project_id)
-  end
-
-  def set_language_from_header
-    params[:language] = "en"
-  end
+  PARENT_RESOURCE = :project
```
camallen: drop this if you can get it working with the `resource_class` method

app/controllers/api/v1/organization_pages_controller.rb
zwolf: Those methods aren't defined yet when during that included block, since that JsonApiController stuff gets included after the this class is defined. 

It kind of makes sense in my head but the order of inclusion/inheritance has been tripping me up a bit. Also, I thought constants were frozen by default, they are totally not. Thanks.

app/controllers/api/v1/organization_pages_controller.rb
```diff
@@ -1,33 +1,4 @@
...
-
-  def link_header(resource)
-    resource = resource.first
-    send(:"api_#{ resource_name }_url", id: resource.id, organization_id: resource.organization_id)
-  end
-
-  def set_language_from_header
-    params[:language] = "en"
-  end
+  PARENT_RESOURCE = :organization.freeze
```
zwolf: @camallen ahahahahahaaha @camallen

camallen: My bad, I thought they were strings :(...fail.

PR #2325, opened 2017-05-22
Includes slug concern on orgs

zwolf:
Also removes redundant code: The only difference between the included `#slugged_name` and the one in the actual Project model is is the lack of a `#try`, on display_name, which is enforced via the model/schema anyway and shouldn't be needed anywhere. 

Removed also the param-massaging for slugs, since they're computed and are not updatable via the API.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -7,6 +7,7 @@ class Api::V1::ProjectsController < Api::ApiController
   include Versioned
   include UrlLabels
   include ContentFromParams
+  include Slug
 
   require_authentication :update, :create, :destroy, :create_classifications_export,
     :create_subjects_export, :create_aggregations_export,
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -33,7 +33,6 @@ class Api::V1::ProjectsController < Api::ApiController
                  :updated_at].freeze
 
   before_action :filter_by_tags, only: :index
-  before_action :downcase_slug, only: :index
 
   prepend_before_action :require_login,
     only: [:create, :update, :destroy, :create_classifications_export,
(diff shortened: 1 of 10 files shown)


app/serializers/organization_serializer.rb
```diff
@@ -5,10 +5,11 @@ class OrganizationSerializer
   include MediaLinksSerializer
   include CachedSerializer
 
-  attributes :id, :display_name, :description, :introduction, :title, :href, :primary_language, :listed_at
+  attributes :id, :display_name, :description, :introduction, :title, :href, :primary_language, :listed_at, :slug
   optional :avatar_src
   media_include :avatar, :background
   can_include :organization_contents, :organization_roles, :projects, :owners
+  can_filter_by :display_name, :slug, :listed_at
```
camallen: you got an index on slug? Might not be a problem if the table space is small enough

app/controllers/api/v1/projects_controller.rb
```diff
@@ -33,7 +33,6 @@ class Api::V1::ProjectsController < Api::ApiController
                  :updated_at].freeze
 
   before_action :filter_by_tags, only: :index
-  before_action :downcase_slug, only: :index
```
camallen: will this still work if I pass in a slug with caps for filtering?

zwolf: oh heck, forgot about filtering by slug in the same commit I added filtering by slug. I'll add this back.

app/serializers/organization_serializer.rb
zwolf: it's, like, non-existent at the moment. But I suppose an index on a text field couldn't hurt?

camallen: Check the query path if there is a more restrictive where clause this may not be needed. If ti's a straight match on this then without an index it'll probably be a table scan.

PR #2329, opened 2017-05-23
schedule cleanup of the expired / old access tokens & grants

camallen:
schedule a daily worker to cleanup old (> 30.days) and revoked doorkeeper access tokens & grants. 
These should all be expired or revoked unless we're using large expiry times which i don't think we are. So should not impact on any oauth flows or token access.

this may **impact on refresh tokens** - i have to look into this. 

# Review checklist

### DIFF
--- a/spec/workers/cleanup_expired_access_tokens_worker_spec.rb
+++ b/spec/workers/cleanup_expired_access_tokens_worker_spec.rb
@@ -1,0 +1,60 @@
+require 'spec_helper'
+
+describe CleanupExpiredAccessTokensWorker do
+  let(:worker) { described_class.new }
+  let(:user) { create(:user) }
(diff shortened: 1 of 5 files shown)


spec/workers/cleanup_expired_access_tokens_worker_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'spec_helper'
+
+describe CleanupExpiredAccessTokensWorker do
```
marten: Should we be testing the behaviour of Doorkeeper::AccessCleanup here? The worker doesn't actually do any of this stuff, all it does is call another method.

camallen:
Good call. I've updated to now only test the worker behaviour with clearer naming as well.

PR #2331, opened 2017-05-24
Adds boolean to Organizations that touches/nulls listed_at on update

zwolf:
Having a timestamp field for public/private seemed like a good idea, but Restpack isn't having it. You can't filter by anything you can't put after an equals sign. So I added a `listed` boolean to Organizations and set up the controller to touch the `listed_at` timestamp on update when `listed` is present in the params and true and nil it when false.

# Review checklist

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -5,8 +5,8 @@ class Organization < ActiveRecord::Base
   include Linkable
   include Translatable
 
-  scope :public_scope, -> { where.not(listed_at: nil) }
-  scope :private_scope, -> { where(listed_at: nil) }
+  scope :public_scope, -> { where(listed: true) }
+  scope :private_scope, -> { where(listed: false) }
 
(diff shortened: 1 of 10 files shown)


zwolf:
That would have worked, too, though I would have guessed that you'd have been against using a before_action. Is a callback here because it's controller and not model?

camallen:
> Is a callback here because it's controller and not model?

I don't like model callbacks that modify persisted state (their own / associations) for a few reasons. Controller before_actions are fine but agree it's fighting restpack here. It was mostly this information is already stored in the timestamp col but we duplicate it in the boolean column for the query use (which is valid). That can add up with row counts if we index both columns.

zwolf:
Agree, and that's why I had it the other way to begin with, but I don't think there will ever be enough orgs to feel the performance hit. Consider the callback suggestion absorbed, though.

PR #2364, opened 2017-06-22
Add param to include avatar in serializer context

zwolf:
Avatar URL is an optional attribute on the project serializer. However, there was no way to indicate via param that it should be included. Now there is.

Sarah and I were working on reduxifying watch.zooniverse.org and we need the avatar--but we also need classification, which isn't included in the 'card' serialization method.

I thought about taking the opportunity to switch to PanoptesRestpack on the project serializer, but it occurred to me that I don't really know what the difference is supposed to be. This works with both.

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1068,7 +1068,6 @@ It has the following attributes:
 - beta
 - approved
 - live
-- avatar_src (optional)
 
 *id*, *created_at*, *updated_at*, *user_count*, and
 *classifications_count* are set by the API.
@@ -1079,7 +1078,6 @@ It has the following attributes:
   + approved (boolean) ... approved state filter
   + beta (boolean) ... beta state filter
   + live (boolean) ... live state filter
-  + avatar_src (boolean) ... include optional `avatar_src` attribute
 
 + Model
 
@@ -1130,7 +1128,6 @@ It has the following attributes:
                     }],
                     "science_case": "88mph + 1.21 GW = 1955",
                     "introduction": "asdfasdf",
-                    "avatar": "http://test.host/x02234.png",
                     "background_image": "http://test.host/12312asd.jp2",
                     "private": false,
                     "faq": "This project uses..",
--- a/apiary.apib
+++ b/apiary.apib
@@ -1068,6 +1068,7 @@ It has the following attributes:
 - beta
 - approved
 - live
+- avatar_src (optional)
 
 *id*, *created_at*, *updated_at*, *user_count*, and
 *classifications_count* are set by the API.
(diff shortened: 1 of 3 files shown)


camallen:
> Sarah and I were working on reduxifying watch.zooniverse.org and we need the avatar--but we also need classification, which isn't included in the 'card' serialization method.

Passing `?include=avatar` params would have returned the linked avatar resource in the response. The `avatar_src` was something we added for cards but I can see how it would be useful for `watch` too.

From memory, watch site will show project cards with classification counts and update from the stream right? If so it would be best to not pay the cost of serializing the full resource that won't be used. Perhaps we should just add the classification count to the cards context instead?

> I thought about taking the opportunity to switch to PanoptesRestpack on the project serializer, but it occurred to me that I don't really know what the difference is supposed to be.

It's mostly a wrapper around restpack that provides a DSL to wire up preloads and automatically preload any included resources (avoid N+1) on page routes. It also wires up correct href's for certain scopes / nested routes.

zwolf:
Adding the classification count to the card context is a good idea. The way that the panoptes-javascript-client makes the request, caches the include, and then requires another (fake) request to actually get the data makes it annoying for what we were trying to do. (<-- that sounds ridiculous so I'll let @srallen chime in if she wants to.)

srallen:
Yep, the way that the json-api-client caches requests is buggy and unintuitive especially for include requests. Changing this is close to the top of our list for fixes for the js client:
https://github.com/zooniverse/json-api-client/issues/34
Our current plan is to strip out the way it caches now and replace it with an established library like ladda as well as continuing to improve app state management using redux:
https://github.com/zooniverse/rfc/issues/16

For this, I like the idea of doing the reverse and adding the classification counts to the `cards: true` param instead.

zwolf:
Switched to include the classification count in the card context instead of exposing the avater src via query param. Would have switched to PanoptesRestpack but I was getting a failing spec that is out of scope to dig into in this PR:

```1) Api::V1::ProjectsController#index logged in  include params when the serializer model is polymorphic should include the owners in the response as linked
     Failure/Error: super(params, paging_scope(params, scope), context)

     ActiveRecord::AssociationNotFoundError:
       Association named 'owners' was not found on Project; perhaps you misspelled it?
     # ./lib/serialization/panoptes_restpack.rb:23:in `page'
     # ./app/serializers/project_serializer.rb:64:in `page'
     # ./lib/json_api_controller/indexable_resource.rb:4:in `index'
     # ./app/controllers/api/v1/projects_controller.rb:64:in `index'
     # ./spec/support/default_format.rb:15:in `process_with_default_params'
     # ./spec/controllers/api/v1/projects_controller_spec.rb:316:in `block (5 levels) in <top (required)>'```

File: apiary.apib [hunk 1]
```diff
@@ -1078,6 +1079,7 @@ It has the following attributes:
   + approved (boolean) ... approved state filter
   + beta (boolean) ... beta state filter
   + live (boolean) ... live state filter
+  + avatar_src (boolean) ... include optional `avatar_src` attribute
```
camallen: this isn't a param anymore.

PR #2373, opened 2017-06-26
Fix renamed classes in migrations

zwolf:
Migrations via `db:migrate` (rather than `db:setup`, which pulls directly from structure.rb) were broken because of the renaming of `SubjectWorkflowCount` to `SubjectWorkflowStatus`. 

Renamed!

# Review checklist


zwolf:
Must have been from when I was migrating over and over and over and they were failing. Just reverted to that file's last commit.

PR #2376, opened 2017-06-28
allow upload of audio files

camallen:
related to recent changes in the front end https://github.com/zooniverse/Panoptes-Front-End/pull/3863. Seems browsers / systems may detect them differently, e.g.

test_file.mp3: 
file cmd, terminal: audio/mpeg
chrome browser file.type: audio/mp3

test_file.m4a: 
file cmd, terminal: audio/mp4
chrome browser file.type: audio/x-m4a

# Review checklist

### DIFF
--- a/spec/models/medium_spec.rb
+++ b/spec/models/medium_spec.rb
@@ -29,10 +29,12 @@ RSpec.describe Medium, :type => :model do
       expect(m).to_not be_valid
     end
 
-    it 'should be valid with a valid content_type' do
-      Medium::ALLOWED_UPLOAD_CONTENT_TYPES.each do |content_type|
-        m = build(:medium, content_type: content_type)
-        expect(m).to be_valid
+    it 'should be valid with allowed content_types' do
+      aggregate_failures "allowed content types" do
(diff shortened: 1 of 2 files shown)


spec/models/medium_spec.rb
```diff
@@ -15,9 +15,13 @@
 
   describe "#content_type" do
 
-    it 'should not be valid without a valid content_type' do
-      m = build(:medium, content_type: "video/mp4")
-      expect(m).to_not be_valid
+    it 'should not be valid without an invalid content_type' do
```
zwolf: It should not be valid "_with_ an invalid" (or "without a valid") content type

spec/models/medium_spec.rb
```diff
@@ -26,15 +30,17 @@
     end
 
     it 'should be valid with a valid content_type' do
```
zwolf: Do you wanna aggregate these failures too, since there's about 8 tests happening in here?

camallen:
requested changes added

zwolf:
LGTM

PR #2379, opened 2017-06-30
[WIP] add code to correctly change a project owner

camallen:
Make this an explicit method on project and a rake task that can be called. 

# Todo
- [ ] Add change_owner method to project.rb
- [ ] Add rake task AND/OR expose via admin only API

# Review checklist

### DIFF
--- a/spec/support/ownable.rb
+++ b/spec/support/ownable.rb
@@ -26,4 +26,22 @@ shared_examples "is ownable" do
       expect(owned.owner?(not_the_owner)).to be_falsy
     end
   end
+
+  describe "#owner=" do
+    let(:new_owner) { create(:user) }
+    let!(:old_owner) { owned.owner }
+
+    it "should change the owner of a resource" do
+      owned.owner = new_owner
+      expect(owned.owner?(new_owner)).to be_truthy
+    end
+
+    it "should remove the old owner acl" do
+      owned.save
+      old_owner_acl = owned.owner_control_list
+      owned.owner = new_owner
+      expect(owned.owner?(@old_owner)).to be_falsy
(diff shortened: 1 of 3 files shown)


spec/support/ownable.rb
```diff
@@ -26,4 +26,22 @@
...
+      owned.owner = new_owner
+      expect(owned.owner?(new_owner)).to be_truthy
+    end
+
+    it "should remove the old owner acl" do
+      owned.save
+      old_owner_acl = owned.owner_control_list
+      owned.owner = new_owner
+      expect(owned.owner?(@old_owner)).to be_falsy
+      expect{old_owner_acl.reload}.to raise_error(ActiveRecord::RecordNotFound)
```
camallen: This currently orphans the ACL without a linked resource_id, that means we may have lots of ACL records kicking around (currently 6) from manual project owner changes.

Do we want to leave this as the default behaviour? It's that or adding after save callbacks / touching the owner= code to clean the orphan up? I'm planning on adding a `project#change_owner(new_owner)` method that can clean up the orphan as this seems to be the only place it's happening, so that's probably enough to avoid the orphans.

E.g.
```
#<AccessControlList:0x007fbb5d7a8e90
 id: 1,
 user_group_id: 1,
 roles: ["owner"],
 resource_id: nil,
 resource_type: "Project",
 created_at: Fri, 30 Jun 2017 14:10:53 UTC +00:00,
 updated_at: Fri, 30 Jun 2017 14:12:51 UTC +00:00>
```

astopy:
Could we take this a step further and add an admin-only API method for it?

camallen:
@astopy good idea, will look at adding this.

PR #2395, opened 2017-07-11
handle invalid combo task annotations in csv formatter

camallen:
some combo task annotations are malformed and do not embed as the annotation values in the expected format (see specs).

Now the code will report an error in the output file and move on instead of failing.

# Review checklist

### DIFF
--- a/lib/formatter/csv/annotation_for_csv.rb
+++ b/lib/formatter/csv/annotation_for_csv.rb
@@ -68,7 +68,7 @@ module Formatter
           new_anno['task'] = annotation['task']
           new_anno['task_label'] = nil
           new_anno['value'] ||= []
-          annotation['value'].each do |subtask|
+          Array.wrap(annotation['value']).each do |subtask|
             @current = subtask
             task_info = get_task_info(subtask)
             new_anno['value'].push case task_info['type']
@@ -181,6 +181,8 @@ module Formatter
         workflow_at_version.tasks.find do |key, task|
          key == subtask["task"]
         end.try(:last) || {}
+      rescue
+        {}
       end
 
       def task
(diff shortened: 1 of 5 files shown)


camallen:
FYI - these hound errors are from the existing specs, i had to re order them to share the shared context.

spec/lib/formatter/csv/annotation_for_csv_spec.rb
```diff
@@ -138,67 +138,95 @@
       end
 
       context "with a combo task" do
-        let(:combo_workflow) { build(:workflow, :complex_task) }
```
zwolf: Combo and complex are separate workflow/workflow content traits in the factories, but this seems to be describing a "combo" but using the "complex" factory. They may be functionally interchangeable, but this might be confusing.

camallen: complex tasks have a mix on combo and others [since this](https://github.com/zooniverse/Panoptes/commit/2c2a49ccdaffd89908da49c6281d56253a1501a3), and the existing specs require that specific workflow configuration. 

I agree that the wording is off, i'll update the let variable name to make it clearer.

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -181,6 +181,8 @@ def get_task_info(subtask)
         workflow_at_version.tasks.find do |key, task|
          key == subtask["task"]
         end.try(:last) || {}
+      rescue
```
camallen: When the subtask isn't a hash, from the specs `{ "task"=>"T3", "value"=>0 }` so:
```
> subtask = 0
 => 0 
> subtask["task"]
TypeError: no implicit conversion of String into Integer
``

camallen:
I have fixed the errors (mostly by ignoring long rspec blocks) now. It just makes the PR harder to interpret the behaviour changes vs formatting / syntax ones. 

I'm in favour of keeping hound now it's back to a working config. Also I've added the rubocop linter to my dev env now so will look at ignoring / modifying rules as we go.

PR #2396, opened 2017-07-11
Allow designator UPP config to be set by project owners

marten:


app/schemas/project_preference_update_settings_schema.rb
```diff
@@ -22,6 +22,11 @@ class ProjectPreferenceUpdateSettingsSchema < JsonSchema
         type "string", "integer"
         pattern "^[0-9]*$"
       end
+
+      property "designator" do
+        type "object"
+        additional_properties true
```
camallen: So this will be free form object changing over time, or is it mostly set already?

marten: Right now it's `{workflow_id => {subject_set_weights: {subject_set_id => weighting}, subject_set_chances: {subject_set_id => chance}`. All of those are optional. But having model IDs in there as keys makes it hard to codify the schema (while it makes lookup faster).

Issue #2406, opened 2017-07-18
Speed up subject uploads


camallen:
Avoid running user.touch [here](https://github.com/zooniverse/Panoptes/blob/a393ea3c471837963236c2ff68e8387e01f4a46c/app/controllers/api/v1/subjects_controller.rb#L38) on each subject upload. 

Possible solutions:
+ Instead look to background it with a rate limit
+ Use the (cached?) uploaded_subjects_count value when generating the cache key to invalidate the serialized response.

I'm in favour of the second option using `cache_by_resource_method` from ExtendedCacheKey with the caveat being that it could be a costly (Subject table count) solution should the cache layer be removed.

Also the subject serializer on create attempts to find any collection the newly minted subject is in. Currently due to a missing index this is a table scan and it's a significant cost that can be avoided. We need to add the missing index in and I'm thinking to just skip links (or subset them) on subjects create serializer responses.

PR #2407, opened 2017-07-19
avoid updating user on each subject upload

camallen:
Partly addresses #2406 do not touch the user model on each upload (and stress the db), instead use the uploaded_subjects_count as part of the rails cache key to invalidated the serializer.

The only down side is without caching layers this can spam the db with subjects table counts for each user request. our setup has the caching layer so won't be impacted but i wanted to get this up for discussion.

# Review checklist


marten:
I don't really mind, the user model gets hit soo often that it's pretty much always going to be a cache hit to look that up.

PR #2435, opened 2017-08-08
[WIP] Generic translations api support

camallen:
Strings translations API spike, the initial idea is to have a Translations resource linked to a translated resource (project, workflow, project_page, tutorial, field_guide, organization). Each translation resource contains the translations strings for a given language code.

The API currently looks like
`GET /translations?translated_type="project"` -> all public Projects or ones i have rights to access. 

`GET /translations?translated_type="workflow"&language=en-AU` -> all public workflow en-AU translations or ones I have access to.

`GET /translations?translated_id=1&translated_type="project"` -> all Project 1 language translations

`GET /translations/id?translated_type="tutorial"` -> the specific tutorial resource for translation

`POST /translations` -> create the translations if i have translator access to the translated resource.
`payload: { strings: { }, language: 'en-AU', translated_id: 1, translated_type: 'project' }`

`PUT /translations/id` -> update the translations if i have translator role on the translated_resource
`payload: {strings: { }, translated_id: 1, translated_type: 'project_page' }`

Authorisation derives from the required `translated_type` param to wire up the controlled resource scope for the logged in user. normally this means they own or have translate rights to the translated resource.
E.g. params = `{translated_id: 1, translated_type: 'Project'}` on the index listing will wire up the controlled resources method to call the `Project.scope_for(:index, api_user).where(id: params[:translated_id])`, thus if a project is public you'll get access, if it's private you'll need to have normal project access roles to list the translations.

# Review checklist


marten:
I like it. 

One thought I had was perhaps to expand and add an "application_id" as well, so that we don't need to reinvent the wheel for EducationAPI, Caesar and actually also just the generic bits of PFE could in theory be stored here so that everything is then editable through one and the same interface. However, that's a thought I wanted to plant now, but we shouldn't try and cram that into the first PR.

camallen:
> Caesar and actually also just the generic bits of PFE could in theory be stored here so that everything is then editable through one and the same interface

I'll have a look at switching the controller context and setting defaults via query param.

simoneduca:
The API seems reasonable, but I have a couple of questions. What does `GET /translations/id` return? Looking at the factory, it looks like it returns the equivalent of the current `project_contents` resource. Is that correct? If it is, I'd prefer it to return all resources, i.e. workflows, pages, tutorial, etc.

I love the idea of managing generic PFE strings and individual projects strings all in one place.

Re. authentication, starting just with projects seems ok to me.

eatyourgreens:
> I'd prefer it to return all resources, i.e. workflows, pages, tutorial, etc.
Would this scale for a large project like Notes from Nature? I'm concerned about performance here, during busy times like Stargazing Live. Will PFE be requesting all the project text (workflows, tutorials, field guide and pages) for every visitor to a classification page, when all that visitor wants is the task text and maybe the tutorial at first. Similarly, if I want to quickly update a tutorial in the translation app, as happened recently for Penguin Watch, I'm not entirely sure why I'd need to download the entire project first.

@simoneduca  any thoughts on how big the changes to PFE would be, to implement this API? It seems like quite a lot of work to me, but I'm maybe being too conservative and you've worked with the front-end more recently than I have.

simoneduca:
@eatyourgreens I think you're right in being concerned about performance. I might have misunderstood Cam's suggestion. For example, I don't understand if [this factory](https://github.com/camallen/Panoptes/blob/9447cc3a59ed309207736c2e0d2e4b3a9cf8ed9b/spec/factories/translations.rb#L1) will be the same across resources, especially pages, tutorials. If it will be then changes across PFE will take quite some time.

simoneduca:
Anyway, after chatting with @eatyourgreens, it seems neither of us are clear on how this would work.

eatyourgreens:
I'm assuming this API gives us a JSON dictionary, something like [this one used by Penguin Watch](/zooniverse/penguinwatch/blob/master/lib/en-us.coffee).

In that case, PFE would load the dictionary from this API, register it with counterpart, then use the `Translate` component to print out strings. At the moment, strings are rendered directly in the various PFE components.

camallen:
>The initial format is more like current translations which is a json blob on keys & translation strings for the whole project (contents, workflows, tutorials, pages, etc)

My initial thoughts were return a big json dictionary that can be keyed up to return the nested data you need for a whole project instead of requesting each resource translation, @simoneduca the factory you mention would contain a whole projects worth of translations in my current approach. Also it has not been fleshed out very much (initial work) but I was thinking to include a top key noting the resource type, e.g. `projects.name`, `workflow.name` etc. 

@eatyourgreens raises a valid point that it could be a weighty document, especially for NFN workflows (:sadpanda:). I'd discount these currently as they are already large documents so i'm not sure anything changes on that front. 

Scaling is an API concern here, hence the simplified data model. Storing the data in one representation in the database avoids complex data joining / access path sql queries so alleviates DB pressure, i.e. it's a lookup and fetch instead of a lookup, join, fetch. Also the data doesn't change often so we can cache it API side and long term use HTTP caching. 

> Similarly, if I want to quickly update a tutorial in the translation app, as happened recently for Penguin Watch, I'm not entirely sure why I'd need to download the entire project first.

Agree it most likely could be a weighty request for that use case but if there is no single representation and you wanted to do wholesale changes for the copy, you'd have to request all the disparate resources via multiple requests. Also you would have to know these resources ahead of time and would need code changes when adding new ones in PFE (maybe API too).

However if there is a single representation document it's a simple matter to add new resources (top level keys) and the request URL is still the same. 

We can look at adding translation routes for each resource, e.g. 
`GET /projects/:id/translations?language=en-au`
`GET /workflows/:id/translations?language=en-au`
`GET /field_guides/:id/translations/:id`
These nested resources would require new API code to be written / deployed. 

Alternatively we could do a flat route like:
`GET /translations?translated_type=project&language=en-au`
`GET /translations?translated_id=54&translated_type=tutorial&language=en-au`
The benefit of this mechanism would split the large JSON doc into sub resources and can be requested independently from the whole. 

However the flat route may require more contextual authorization params (previously in the request path) or we hard code the lookups(e.g. top level resource that owns the translation for authorization, workflows -> project, tutorials -> workflows -> project, and this adds complexity to the DB query. 

Happy to chat about this further. Might be best to do so in person?

spec/factories/translations.rb
```diff
@@ -0,0 +1,42 @@
...
+      researcher_quote: "This is my favorite project",
+      urls: [
+        {label: "Blog", url: "http://blog.example.com/"},
+        {label: "Twitter", url: "http://twitter.com/example"}
+      ]
+    })
+    association :translated, factory: :project
+    language "en-GB"
+
+    factory :workflow_translation do
```
marten: Is this factory meant to be nested inside?

camallen: yeah - if you use the nesting it inherits the attributes of the outer scope.

PR #2441, opened 2017-08-17
refactor media controller controlled_resources

camallen:
**Please note, this long running PR had master merged and extracted common behaviour for both media / translations controllers.**

media controller is for the Medium polymorphic resource. The controller has to interpret the nested routes to determine the correct Media scopes based on linked_id, linked_type lookups. 

This PR ensures the polymorphic (media / translations) controllers correctly link resources to the polymorphic scope (e.g. find all translations for project 1, where i can update project 1).

This PR also ensures the before_filter ordering tests the etag precondition on the correct resources not the polymorphic link resources.

Finally I added in a test to ensure we don't erroneously destroy the parent resources by mistake, i think by luck we never hit this before but it will now 500 if this happens. This is a new feature but should safegaurd against any controller attempting to delete resources types it doesn't not control.

~This will be some ground work for #2435 and I may extract this to a shared module if the pattern fits for the translations polymorphic resource authorization.~ I have extracted this to a shared controller module and included this in translations after merging from master. It cleans up the controller code quite nicely though one has to understand the included module code.

# Review checklist


marten:
I've tried understanding this, but I don't seem to be able to. Maybe you could talk me and @zwolf through it sometime?

camallen:
@marten @zwolf happy to provide a drive by of the code and glean any places I can refactor to make it more obvious. Did you want to provide some inline questions or maybe just as we go through it?

app/controllers/api/v1/media_controller.rb
```diff
@@ -122,53 +92,66 @@ def controlled_scope
...
+  # will lookup the linked attached_images media for the workflow :id
+  def media_controlled_resources
+    linked_media_scope = association_reflection.klass.where(
+      linked_id: media_parental_controlled_resources.select(:id),
+      linked_type: resource_class.name
+    )
+    if params.key?(:id)
+      linked_media_scope = linked_media_scope.where(id: params[:id])
+    end
+    @controlled_resources = linked_media_scope
```
camallen: Instead of unsetting memoized instance variables here try and provide a hook to `controlled_resources` behaviour., e.g. 
``` ruby
def controlled_resources
  find_controlled_resources(resource_class, resource_ids)
end

def find_controlled_resources(controlled_class, controlled_ids)
  @controlled_resources ||= api_user.do(controlled_scope)
     .to(resource_class, scope_context, add_active_scope: add_active_resources_scope)
     .with_ids(resource_ids)
     .scope
end
```
around here and in the `media_controller`
```
def controlled_resources
  find_controlled_resources(polymorphic_class, polymorphic_resource_ids)
end
```
https://github.com/zooniverse/Panoptes/blob/e33ae3cac8f67f646be0033aee60833458f22702/lib/role_control/roled_controller.rb#L22

thanks to @marten

camallen:
@marten I know you've reviewed / approved this PR before but i merged master and refactored common polymorphic controller behaviour building on your previous comments. The new implementation is much cleaner without hacking instance vars and carefully plotting callback timing. 

Most of this is the same as before but I would appreciate you looking at it again and I apologise for the size of this PR.

PR #2447, opened 2017-08-24
add urls to organisation serialiser

### LINKED ISSUE #2397: add urls to organization resource
add `urls` attribute to organization resource, similar to [project urls](https://github.com/zooniverse/Panoptes/blob/master/app/serializers/project_serializer.rb#L27)

camallen:
closes #2397 allow the orgs to get access to their external URL's

# Review checklist

### DIFF
--- a/spec/serializers/organization_serializer_spec.rb
+++ b/spec/serializers/organization_serializer_spec.rb
@@ -88,4 +88,22 @@ describe OrganizationSerializer do
       end
     end
   end
+
+  describe "#urls" do
+    let(:result) do
+      OrganizationSerializer.single({}, Organization.where(id: organization.id), {})
+    end
+    let(:result_urls) do
+      organization.urls.map do |url|
+        {
+          url: url["url"],
+          label: url["label"]
+        }
+      end
+    end
+
+    it "should return url attribute" do
+      expect(result[:urls]).to match_array(result_urls)
+    end
+  end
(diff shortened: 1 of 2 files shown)


spec/serializers/organization_serializer_spec.rb
```diff
@@ -88,4 +88,22 @@
       end
     end
   end
+
+  describe "#urls" do
+    let(:result) do
+      OrganizationSerializer.single({}, Organization.where(id: organization.id), {})
```
zwolf: I can't find the docs for this method in ruby/rails/AR/ActiveModelSerializers. It makes sense but I'd like to know what the empty hashes signify, can you link that for me?

camallen: it's not well documented from memory, basically the same that a page takes the params hash, the AR scope and the serializer context hash. 

From https://github.com/RestPack/restpack_serializer#exposing-an-api
> The AlbumSerializer also provides a single method which will return a serialized resource similar to as_json above.

https://github.com/RestPack/restpack_serializer/blob/8eb3096c7118968ab6f353b09b1e6976a6c36951/lib/restpack_serializer/serializable/single.rb#L5

zwolf: Ah hah, a restpack method. That explains it, thanks.

PR #2464, opened 2017-09-13
Reactivate existing membership on create

zwolf:
When a user tries to re-join a user group they had been deleted from, the existing, inactive membership causes an error on `Membership.create!`. Now, memberships are either created or reactivated on a successfully validated create request.

Primary use for this is adding/removing users from Education classrooms, which correspond directly to Panoptes user groups.

# Review checklist

### DIFF
--- a/app/operations/memberships/create.rb
+++ b/app/operations/memberships/create.rb
@@ -12,7 +12,10 @@ module Memberships
       raise Unauthorized unless user_group.verify_join_token(join_token)
       raise Unauthorized unless user.id == api_user.id
 
-      Membership.create! user: api_user.user, user_group: user_group, state: :active
+      membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group)
+      membership.state = :active
+      membership.save!
+
     end
 
     def user
--- a/app/operations/memberships/create.rb
+++ b/app/operations/memberships/create.rb
@@ -15,7 +15,6 @@ module Memberships
       membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group)
       membership.state = :active
       membership.tap(&:save!)
-
     end
 
     def user
--- a/app/operations/memberships/create.rb
+++ b/app/operations/memberships/create.rb
@@ -14,7 +14,7 @@ module Memberships
 
       membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group)
       membership.state = :active
-      membership.save!
+      membership.tap(&:save!)
 
     end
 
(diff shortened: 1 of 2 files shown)


app/operations/memberships/create.rb
```diff
@@ -12,7 +12,9 @@ def execute
       raise Unauthorized unless user_group.verify_join_token(join_token)
       raise Unauthorized unless user.id == api_user.id
 
-      Membership.create! user: api_user.user, user_group: user_group, state: :active
+      membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group)
+      membership.state = :active
+      membership.tap(&:save!)
```
camallen: what's the reason for tap here? won't it be new or existing model so save! will just work.

app/operations/memberships/create.rb
```diff
@@ -12,7 +12,9 @@ def execute
       raise Unauthorized unless user_group.verify_join_token(join_token)
       raise Unauthorized unless user.id == api_user.id
 
-      Membership.create! user: api_user.user, user_group: user_group, state: :active
+      membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group)
+      membership.state = :active
```
camallen: non-blocking - i think you can do this
```
membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group) do |m|
  m.state = :active
end
membership.save!
```

app/operations/memberships/create.rb
```diff
@@ -12,7 +12,9 @@ def execute
       raise Unauthorized unless user_group.verify_join_token(join_token)
       raise Unauthorized unless user.id == api_user.id
 
-      Membership.create! user: api_user.user, user_group: user_group, state: :active
+      membership = Membership.find_or_initialize_by(user: api_user.user, user_group: user_group)
```
camallen: should we explicitly query for deactivated memberships here to match the spec? 
```
find_or_initialize_by(
  user: api_user.user, 
  user_group: user_group, 
  state: :inactive
)
```

app/operations/memberships/create.rb
zwolf: #save returns true or false, the expected return value of this operation is the created object. Saved an extra "membership" line.

camallen: :tada:

app/operations/memberships/create.rb
zwolf: I want to get an any existing membership regardless of state with that line, then just make sure it's active with the next one.

PR #2490, opened 2017-10-12
Use Designator by default for all workflows.

marten:
* Adds an option to set a workflow explicitly to use the built-in selection.
* Changes the default selector to use Designator
* Changes the CellectController so that you can get the list of subject IDs for any workflow, not just the ones set to use Cellect (primarily so that I can update Designator to load via this API rather than connecting to the DB directly).

# Review checklist


app/models/workflow.rb
```diff
@@ -91,31 +91,15 @@ def retirement_with_defaults
   def subject_selector
     @subject_selector ||=
       case
+      when subject_selection_strategy == "builtin"
+        Subjects::BuiltInSelector.new(self)
       when subject_selection_strategy == "cellect"
         Subjects::CellectSelector.new(self)
       when subject_selection_strategy == "designator"
```
camallen: can we remove this test and just fallback to default?

app/models/workflow.rb
```diff
@@ -32,7 +32,7 @@ class Workflow < ActiveRecord::Base
   # TODO: remove this association from the cache key
   cache_by_resource_method :subjects_count, :finished?
 
-  enum subject_selection_strategy: [:default, :cellect, :designator]
+  enum subject_selection_strategy: [:default, :cellect, :designator, :builtin]
 
   scope :using_cellect, -> { where(subject_selection_strategy: subject_selection_strategies[:cellect]) }
```
camallen: do you want to modify this scope to pickup non cellect enum targets?

config/initializers/cellect.rb
```diff
@@ -6,10 +6,6 @@ def self.cellect_on
   def self.cellect_min_pool_size
     @cellect_min_pool_size ||= (ENV["CELLECT_MIN_POOL_SIZE"] || 10000).to_i
   end
-
-  def self.use_cellect?(workflow)
-    cellect_on && workflow.using_cellect?
```
camallen: we should probably clean up the feature flag `self.cellect_on` as well if we remove this

lib/subjects/selector.rb
```diff
@@ -50,21 +50,18 @@ def fallback_selection
...
-            context: {
-              workflow: workflow.id
-            }
-          )
+      sms_ids = fallback_selector.select
+      if data_available = !sms_ids.empty?
+        if Panoptes.flipper[:cellect_sync_error_reload].enabled?
+          NotifySubjectSelectorOfChangeWorker.perform_async(workflow.id)
         end
+        Honeybadger.notify(
```
camallen: I'm happy enough to remove this now, it's just noise in the HB from cellect previously. Do you want to track this for designator?

lib/subjects/strategy_selection.rb
```diff
@@ -16,31 +16,18 @@ def initialize(workflow, user, subject_set_id, limit=SubjectQueue::DEFAULT_LENGT
 
     def select
       used_strategy, selected_ids = select_sms_ids
-      eventlog.info("Selected subjects", desired_strategy: strategy, used_strategy: used_strategy, subject_ids: selected_ids)
+      eventlog.info("Selected subjects", used_strategy: used_strategy, subject_ids: selected_ids)
```
camallen: FYI - these never get logged as the eventlog is never initialised correctly.

app/models/workflow.rb
marten: The only thing that used this was the Cellect controller, and i didn't want that to return non-Cellect workflows. 

But given that you've now confirmed that nothing uses this controller yet, I can remove this completely and let the cellect controller take a `strategy` param. (And rename that controller)

lib/subjects/selector.rb
marten: Left it in for now, but did update the message.

PR #2493, opened 2017-10-16
add cache_total_count dsl to panoptes restpack

### LINKED ISSUE #2472: End points with large tables cause db load
Requesting data for the users end point returns the table count in the response meta. This is a tablescan in postgres issued from Restpack Serializer & Kaminari. 

Previously for large tables we've removed the count value completely to avoid this cost ( #2145 ). However this number is being used for volunteer counts in the home page https://github.com/zooniverse/Panoptes-Front-End/blob/66d628ca258912ce09789fd81bac2fb4fb398079/app/pages/home-not-logged-in.jsx#L54

The best solution here is to cache the count query to avoid each serializer instance requesting a table scan in the db. Perhaps just wrapping the model.count method or modifying the serializer / kaminari to count with a cache store https://github.com/edpaget/restpack_serializer/blob/9a7cdb2777006c7cd05c512859f5c9451248ca8d/lib/restpack_serializer/serializable/paging.rb#L47

camallen:
closes #2472 allow any panoptes restpack serializer to cache the total count value via a DSL option

# Review checklist

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -18,6 +18,8 @@ class UserSerializer
 
   preload :avatar
 
+  cache_total_count true
+
   def admin
     !!@model.admin
   end
@@ -57,10 +59,5 @@ class UserSerializer
 
   def add_links(model, data)
     data[:links] = {}
-  end
-
-  def self.serialize_meta(page, options)
-    page = Serialization::PageWithCachedMetadata.new(page)
-    super(page, options)
   end
 end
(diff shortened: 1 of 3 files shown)


app/serializers/user_serializer.rb
```diff
@@ -18,6 +18,8 @@ class UserSerializer
 
   preload :avatar
 
+  cache_total_count true
```
camallen: should this provide an option to set the expires in time, e.g. for users it can be default from cache store or 1 min, if we use this on other models we may want to have a shorter configurable expiry time...?

PR #2494, opened 2017-10-17
read from slave db for database dumps

camallen:
Read from slave database when dumping read-only data. The database cleaner strategy had to be set to truncation instead of transaction so the different db connections have access to the same testing data, this should be reverted if / when the read_slave requirement is removed. 

**Note** this will require new `_read_slave` db connections to be configured before merging.

# Review checklist

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -45,7 +45,9 @@ RSpec.configure do |config|
   end
 
   config.before(:each) do |example|
-    DatabaseCleaner.strategy = :transaction
+    # use truncation to handle distinct read_slave db connection
+    # revert back to :transaction (faster) when the read slave requirement is removed
+    DatabaseCleaner.strategy = :truncation
     DatabaseCleaner.start
     ActionMailer::Base.deliveries.clear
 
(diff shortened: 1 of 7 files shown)


spec/spec_helper.rb
```diff
@@ -45,7 +45,9 @@
   end
 
   config.before(:each) do |example|
-    DatabaseCleaner.strategy = :transaction
+    # use truncation to handle distinct read_slave db connection
+    # revert back to :transaction (faster) when the read slave requirement is removed
+    DatabaseCleaner.strategy = :truncation
```
marten: Can we do this with a tag on specific specs only? Tests are still running but I expect them to become at least twice as slow for the full suite (which is already taking a long time)?

camallen: yeah for sure. 

Alternatively what do you think about using this gem instead https://github.com/kenn/slavery#transactional-fixtures

It seems to have better handling of connection switching, global kill switch (flipper?) and the ability to combine read slave / write master with nested blocks. It's a small dep that i've had a read through and the code seems ok.

PR #2498, opened 2017-10-19
[WIP] - Re-engineer dump workers

camallen:
I'll rebase this against #2499 once the data model is sorted. 

~This will be a polymorphic data model to support all the dumps and the plan is to store the data in a json blob, e.g. `{ id: 1, subject_id: 2, annotations: {}, workflow_id: 1, ... }`. @marten @zwolf thoughts on the model name, data format?~ this is now just a standard model that contains the classification csv export fields. 

Use a strategy pattern to dynamically switch between the pre-calculated ClassificationRowExport instances as the dump source or the old behaviour of finding all and formatting on the go

# TODO
- [X] Create the row export model in a low priority worker from the lifecycle method
- [x] Backfill in the dump (for all the previous models)
- [ ] Include the current subject retirement state when serializing to csv
- [ ] Figure out if we can use some other hashed data for non-logged-in users.

# Review checklist


marten:
Not because my idea is better (or worse) but just because we apparently had different ideas, let me also share what I thought the approach would be. I had a `ClassificationExportRow` in mind, with columns for each of the CSV columns, plus maybe one or two for scoping queries. So not polymorphic.

I would definitely suggest going with `ExportRow` rather than `Export`. Export sounds more like the full CSV file than a small piece of it, to me.

camallen:
So we leave the subjects, workflow, etc dumps running the way they are and pay the read IOPS costs on them still (not as expensive as classifications i agree)?

ExportRow sounds good!

marten:
Oh, and keep in mind that one subject has multiple lines in the subjects.csv dump: one for each workflow it appears in.

marten:
How do we do subject retirement info that's currently in the classifications export? Have to remember to do cache invalidation on that.

camallen:
retirement cache is setup per batch processed - https://github.com/zooniverse/Panoptes/blob/76b30dbd53012b1da7d986da725db3ff9839c5a6/app/workers/classifications_dump_worker.rb#L51

marten:
Yeah but presumably that will now get stuck into the ExportRow?

camallen:
ah - good point, now I get you.

camallen:
I guess, it's still too much to ask that they join a fresh subject dump and their classification dump to determine retirement status.

app/models/classification.rb
```diff
@@ -111,6 +111,12 @@ def metadata
     read_attribute(:metadata).with_indifferent_access
   end
 
+  def export_row
```
camallen: This is a hack around the has_one requiring knowledge of the `classification.project_id` to hit the specific index (project_id, classification_id), this index is required to filter the large table into smaller dump read chunks. This can probably go as it's not being used in the current setup.
```
# this can't be used in joins as the instance attributes are not known ahead of time
has_one :classification_export_row, -> (c) { (project_id: c.project_id, classification_id: c.id) }
```

app/workers/classifications_dump_worker.rb
```diff
@@ -10,50 +10,141 @@ class ClassificationsDumpWorker
...
+      scope_filters = case resource_type
+      when "workflow"
+        { project_id: resource.project_id, workflow_id: resource.id }
+      else
+        { project_id: resource.id }
+      end
+      ClassificationExportRow.where(scope_filters)
+    end
+
+    def backfill_classification_export_rows
```
camallen: Ideally this should be extracted for re-use, perhaps in a rake task?

camallen:
Closing this as it's out of data post the refactor in # - I still think this idea has merit:

1. Use the export row as the main source of data export.
2. backfill the export row model in batches while running dumps for new work

This way we know the latest export point and dont re-compute the state every time we serialize the dumps. Also noting this doesn't have to exist in panoptes either...

PR #2499, opened 2017-10-22
Add classification row export model

camallen:
New table to support pre-computed export csv data and populate the data in the classification lifecycle event via a low data worker

# Review checklist


db/migrate/20171019115705_add_classification_export_row_model.rb
```diff
@@ -0,0 +1,27 @@
...
+    create_table :classification_export_rows do |t|
+      t.references :classification, null: false
+      t.references :project, null: false
+      t.references :workflow, null: false, index: true
+      t.references :user
+      t.string :user_name
+      t.string :user_ip
+      t.string :workflow_name
+      t.string :workflow_version
+      t.timestamp :created_at
```
camallen: this needs to be changed to the `classification.created_at`

db/migrate/20171019115705_add_classification_export_row_model.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.jsonb :annotations
+      t.jsonb :subject_data
+      t.string :subject_ids
+      t.timestamps
+    end
+
+    add_index :classification_export_rows,
+      %i(project_id classification_id),
+      name: 'index_export_rows_on_project_id_and_classification_id',
+      unique: true
```
marten: Aren't classification IDs globally unique? Why include the project in this unique index?

lib/classification_lifecycle.rb
```diff
@@ -86,6 +89,11 @@ def notify_cellect
     end
   end
 
+  def create_export_row
```
marten: Do we want a feature flag on this?

camallen: Yeah - good call.

db/migrate/20171019115705_add_classification_export_row_model.rb
camallen: they are unique, i want the project index to increase how selective this index when we run a dump for backfilling. Finding can probably just use the project_id / workflow_id index but i should check the `find_in_batches` won't use the compound index to order by classification (i think it does by default).

lib/classification_lifecycle.rb
```diff
@@ -86,6 +89,13 @@ def notify_cellect
...
+    return unless classification.complete?
+    unless Panoptes.flipper[:disable_classification_export_row_in_lifecycle].enabled?
```
marten: The negatives is a bit confusing here.

PR #2513, opened 2017-11-02
feature flag read recents from read replica

camallen:
some of the recents queries can be costly as they invert the order on created at before selecting the first page. push this to the read replica where we can

# Review checklist

### DIFF
--- a/app/controllers/concerns/recents.rb
+++ b/app/controllers/concerns/recents.rb
@@ -2,10 +2,24 @@ module Recents
   def recents
     ps = params.dup
     ps.delete "#{resource_name}_id"
-    render json_api: RecentSerializer.page(
-      ps,
-      Recent.where(:"#{resource_name}_id" => resource_ids),
-      { url_prefix: "#{resource_sym.to_s.pluralize}/#{resource_ids}" }
-    )
+    read_recents_from_database do
+      render json_api: RecentSerializer.page(
+        ps,
+        Recent.where(:"#{resource_name}_id" => resource_ids),
+        { url_prefix: "#{resource_sym.to_s.pluralize}/#{resource_ids}" }
+      )
+    end
+  end
+
+  private
+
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/recents.rb
```diff
@@ -2,10 +2,24 @@ module Recents
...
+        ps,
+        Recent.where(:"#{resource_name}_id" => resource_ids),
+        { url_prefix: "#{resource_sym.to_s.pluralize}/#{resource_ids}" }
+      )
+    end
+  end
+
+  private
+
+  def read_recents_from_database
```
marten: Wonder if we should just turn this into a globally available `Panoptes.read_from_replica(flipper_key)` method? 2/10 tho

camallen: thought about it as well, do we want granularity to have the dumps run on the replica but switch this feature back (say the replica lag is really high)?

marten: Maybe? That's why I was passing in the flipper key, but I'm happy to leave it off if you think we don't need it.

camallen: ha! bad cam, I read that completely differently. i'll extract the common code block to something like that.

app/workers/concerns/dump_worker.rb
```diff
@@ -128,13 +128,7 @@ def get_resource
...
-  def read_from_database
-    if Panoptes.flipper.enabled?("dump_data_from_read_slave")
-      Slavery.on_slave do
-        yield
-      end
-    else
-      yield
-    end
+  def read_from_database(&block)
+    DatabaseReplica.read("dump_data_from_read_slave", &block)
```
camallen: this is costlier than using a block as it converts the block to a proc but it keeps the code cleaner.

marten:
LGTM

PR #2515, opened 2017-11-03
Allow create and update actions for translations

camallen:
Properly test the custom translation route constraints to avoid 404's to unknown_route. Correctly test the ETag resource on update and fill in the polymorphic translated resource keys on the create payload using the schema validation.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/translations_controller.rb
+++ b/app/controllers/api/v1/translations_controller.rb
@@ -37,7 +37,12 @@ class Api::V1::TranslationsController < Api::ApiController
     revert_resource_name_to_controller_type
 
     translation = Translation.transaction(requires_new: true) do
-      create_params[:translated_type] = create_params[:translated_type].classify
+
+      # push these scope params into create payload
+      # to validate using the create schema
+      params[:translations][:translated_type] = params[:translated_type].classify
+      params[:translations][:translated_id] = params[:translated_id]
+
       resource = Translation.new(create_params)
       resource.save!
       resource
--- a/app/controllers/api/v1/translations_controller.rb
+++ b/app/controllers/api/v1/translations_controller.rb
@@ -37,9 +37,7 @@ class Api::V1::TranslationsController < Api::ApiController
     revert_resource_name_to_controller_type
 
     translation = Translation.transaction(requires_new: true) do
-      create_params[:translated_type] = params[:translated_type].classify
-      # TODO: should raise an error if this is missing
-      create_params[:translated_id] = params[:translated_id]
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/translations_controller.rb
```diff
@@ -37,9 +37,12 @@ def create
     revert_resource_name_to_controller_type
 
     translation = Translation.transaction(requires_new: true) do
-      create_params[:translated_type] = params[:translated_type].classify
-      # TODO: should raise an error if this is missing
-      create_params[:translated_id] = params[:translated_id]
+
+      # push these scope params into create payload
+      # to validate using the create schema
+      params[:translations][:translated_type] = params[:translated_type].classify
```
marten: Are there any security risks with calling `classify`?

lib/routes/constraints/translations.rb
```diff
@@ -6,21 +6,23 @@ class Translations
       def matches?(request)
         @params = request.params
 
-        if show_route?
+        if member_route?
            valid_show_id? && has_translated_type_param?
```
marten: Does `valid_show_id?` need to be renamed to `valid_member_id?`?

app/controllers/api/v1/translations_controller.rb
camallen: I don't think so, it converts string to string e.g. `"project"` ->  `"Project"`. 
That is then fed into the create params for Rails to interpret as a valid polymorphic class target (i.e. does that class name exist)

lib/routes/json_api_routes.rb
```diff
@@ -61,13 +61,19 @@ def media_resources(*names)
       end
     end
 
+    def json_api_default_opts
+      @json_api_default_opts ||= {
+        except: %i(new edit),
+        constraints: { id: VALID_IDS },
```
marten: Unrelated, but shouldn't https://github.com/zooniverse/Panoptes/pull/2515/files#diff-6c7cf9697e69227d67c03ea7e9bd79cdL3 be `/[0-9]+/` rather than `*`?

lib/routes/constraints/translations.rb
camallen: yeah - it probably should for consistent naming. I can update.

lib/routes/json_api_routes.rb
camallen: I believe you are correct. I'll issue a new PR to fix this in isolation.

camallen: #2517

PR #2521, opened 2017-11-09
Remove aggregation client

marten:
Removes aggregation dump worker, aggregation client, aggregation dump API endpoint, aggregation dump notification email.

# Review checklist


config/routes.rb
```diff
@@ -100,13 +100,11 @@
         media_resources :avatar, :background, :attached_images,
           classifications_export: { except: [:create] },
           subjects_export: { except: [:create] },
-          aggregations_export: { except: [:create] },
```
camallen: Do we want to response with a deprecated or just remove it and it'll 404 with missing route (as it was only used internally)?

marten: I'm 99% confident that https://github.com/zooniverse/Panoptes-Front-End/pull/4144 removes the only place where this url is ever referenced. It might be easiest if we hold off on merging this until that is merged.

PR #2522, opened 2017-11-10
user has finished a workflow if the workflow is finished

### LINKED ISSUE #2197: add workflow finished state to the selection serializer
linked to #2194 - Provide a mechanism to signal that the user has finished the workflow and/or the workflows is finished, without paying a hefty count query cost on this.

camallen:
closes #2197 - short circuit the check on seen linked counts vs linked counts. Only run this if the workflow isn't finished.

Also this flows into subject selection response context `finished_workflow` attribute. This now signals that a workflow is finished or the user has finished and can be used after each selection to determine if the client should prompt the user to choose a new worklfow and/or move to another project.

# Review checklist

### DIFF
--- a/spec/controllers/api/v1/subjects_controller_spec.rb
+++ b/spec/controllers/api/v1/subjects_controller_spec.rb
@@ -279,9 +279,9 @@ describe Api::V1::SubjectsController, type: :controller do
             get :index, request_params
           end
 
-          it 'should return user finished_workflow as false for each subject' do
+          it 'should return user finished_workflow as true for each subject' do
             seen_all = json_response["subjects"].map{ |s| s['finished_workflow']}
-            expect(seen_all).to all be(false)
+            expect(seen_all).to all be(true)
           end
 
           it 'should return all subjects as retired' do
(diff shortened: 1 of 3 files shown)


spec/controllers/api/v1/subjects_controller_spec.rb
```diff
@@ -461,9 +461,9 @@
             get :queued, request_params
           end
 
-          it 'should return user finished_workflow as false for each subject' do
+          it 'should return user finished_workflow as true for each subject', :focus do
```
marten: There's still a focus here.

camallen: urgh

marten:
This PR closes itself, rather than some other issue?

Is this done for performance reasons? Conceptually it feels like a lie, I haven't completed a workflow just because everyone else has pushed everything to retirement. If simply for messaging in the frontend, shouldn't the frontend use the workflow.finished flag instead of / in addition to this?

camallen:
Fixed the close issue. Mostly it's a way to push the data to the client without having the client re-request the workflow data. When new subject data comes down we signal the UI, yep this workflow is done. The field was already ambiguous, `finished_workflow` didn't actually mean it was finished, but the user had finished it, now it catches both events. 

I can split out the two events if you'd prefer more information being serialized to the client.

PR #2525, opened 2017-11-10
Add tags to Organizations

### LINKED ISSUE #2436: add tags to organization resource
Add `tags` attribute to organization resource, similar to project tags.

Includes organizations in filtering by discipline, for consistent UX as per [PFE Issue #3977](https://github.com/zooniverse/Panoptes-Front-End/issues/3977).

zwolf:
Adds tag resource to organizations. Extracts tag logic into concerns and support specs and adds filtering via query params on index methods.

Tags are now built (find or init) in a little service object that's used by the project and organization update and create methods/ops.

Bonus: should make it easy to add tags to other things. Lookin' at you, collections.

Fixes #2436 

# Review checklist

### DIFF
--- a/app/operations/organizations/create.rb
+++ b/app/operations/organizations/create.rb
@@ -18,8 +18,8 @@ module Organizations
         organization = build_organization
         organization.organization_contents.build(organization_contents_params)
 
-        tags = Tags::BuildTags.run!(api_user: api_user, tag_array: tags) if tags
-        organization.tags = tags unless tags.nil?
+        tag_objects = Tags::BuildTags.run!(api_user: api_user, tag_array: tags) if tags
+        organization.tags = tag_objects unless tags.nil?
 
         organization.save!
         organization
--- a/app/operations/organizations/create.rb
+++ b/app/operations/organizations/create.rb
@@ -11,11 +11,16 @@ module Organizations
     string :introduction, default: ''
     array :urls, default: []
     array :categories, default: []
+    array :tags, default: []
 
     def execute
       Organization.transaction(requires_new: true) do
         organization = build_organization
         organization.organization_contents.build(organization_contents_params)
+
+        tags = Tags::BuildTags.run!(api_user: api_user, tag_array: tags) if tags
+        organization.tags = tags unless tags.nil?
+
         organization.save!
         organization
       end
--- a/app/operations/organizations/update.rb
+++ b/app/operations/organizations/update.rb
@@ -18,6 +18,13 @@ module Organizations
             org_update.delete(field)
           end
         end
+
+        if org_update.key?(:tags)
+          tags = Tags::BuildTags.run!(api_user: api_user, tag_array: org_update[:tags])
+          organization.tags = tags unless tags.nil?
+          org_update.delete(:tags)
+        end
+
         organization.update!(org_update.symbolize_keys)
         organization.organization_contents.find_or_initialize_by(language: language).tap do |content|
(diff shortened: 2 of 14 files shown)


zwolf:
Yeah, good call, specs are always good.

app/operations/tags/build_tags.rb
```diff
@@ -0,0 +1,14 @@
+module Tags
+  class BuildTags < Operation
+
+    array :tag_array
+
+    def execute
+      tags = tag_array.try(:map) do |tag|
```
camallen: OOI - why `try` the map operation? If it's an empty array then it'll do nothing but it won't fail with missing method.

app/operations/organizations/create.rb
```diff
@@ -11,11 +11,16 @@ class Create < Operation
...
     array :urls, default: []
     array :categories, default: []
+    array :tags, default: []
 
     def execute
       Organization.transaction(requires_new: true) do
         organization = build_organization
         organization.organization_contents.build(organization_contents_params)
+
+        tag_objects = Tags::BuildTags.run!(api_user: api_user, tag_array: tags) if tags
```
camallen: What happens at the top level transaction in Organisations controller (create / update / destroy) when a `first_or_create` in the nested operation transaction fails here? Could we get inconsistent state?

http://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html

app/operations/organizations/update.rb
```diff
@@ -18,6 +18,13 @@ def execute
             org_update.delete(field)
           end
         end
+
+        if org_update.key?(:tags)
+          tags = Tags::BuildTags.run!(api_user: api_user, tag_array: org_update[:tags])
```
camallen: Same question about create operation and transactions.

app/operations/tags/build_tags.rb
zwolf: Ah, because the guts of this method was originally plucked from attempting to map over (potentially omitted) params. Seems unnecessary.

zwolf:
I didn't use Create because it will find tags that already exist, which will probably happen more often than it creates them. I realize that they were getting created when the resource was saved, but it was cumbersome to test in the spec if they were being instantiated or not without actually saving them in the operation.

app/operations/organizations/create.rb
zwolf: This is a good question. I've added a spec to make sure that an error response is returned when the tag attribute is incorrectly formatted, but that doesn't answer the question. I think it might be moot because there's nothing happening in the top level transaction except the call to the operation; that one seems superfluous. 

Inside the op, using `require_new: true` should ensure that any exceptions in any of the sub-transactions (tag creation, saving objects, etc) should bubble up and cause the operation's transaction to roll back.

zwolf:
These are good questions, but this can probably be merged as-is if it's blocking you, @mcbouslog.

app/operations/organizations/create.rb
camallen: the top level transaction will not rollback thus they could leave inconsistent state if there was a multi org update (via the array of orgs param) and one failed but the other didn't.

I'd have to double check if we support batch (array) creates, updates & deletes but that controller code looks like it does. Either way, I think we should add `require_new: true` to the top level to ensure a rollback on a nested transaction and thus keep the state consistent for the whole transaction. 

I think this PR is good to go so happy for this to come in another PR but we shouldn't allow the state to become inconsistent in the DB.

mcbouslog:
Thanks @zwolf! This is not blocking me, I have a handful of unrelated organization work before I get to the front-end pieces related to this.

PR #2528, opened 2017-11-15
Media resources scope fix

camallen:
Ensure the media resources restrict the media_type scope to the relevant route. 

E.g. get /projects/1/background should only return media resources with `type: 'project_background'`. fixes a bug reported in https://github.com/zooniverse/Panoptes-Front-End/pull/4142#issuecomment-344628072

# Review checklist

### DIFF
--- a/app/controllers/api/v1/media_controller.rb
+++ b/app/controllers/api/v1/media_controller.rb
@@ -98,4 +98,14 @@ class Api::V1::MediaController < Api::ApiController
   def has_many_assocation?
     @has_many_assocation ||= association_reflection.macro == :has_many
   end
+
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/media_controller.rb
```diff
@@ -98,4 +98,19 @@ def has_one_assocation?
...
+  def controlled_resources
+    @controlled_resources ||= if params[:media_name]
+                                super.where(type: singular_linked_media_type)
+                              else
+                                super
+                              end
+  end
+
+  # attached images have where(type: "tutorial_attached_image") } polymorphic scope
+  # so these has_many relatinos need to be singular types scopes
```
marten: typo in "relations"

Issue #2459, opened 2017-09-05
add attached_images to organization media_include


mcbouslog:
We'd like the organization landing Introduction and About sections to support images, so we'd like to give an organization collaborator the ability to upload media and utilize it in those sections, which currently support Markdown. Hopefully similar, if not the same, to how a project collaborator utilizes the Media section of the Project Builder.

Towards https://github.com/zooniverse/Panoptes-Front-End/issues/2970.

PR #2535, opened 2017-11-20
Adds attached images to organizations

zwolf:
Model, controller, serializer, router.

It doesn't look like the organization operations have specs at all. Sort of taken care of by controller specs but there's enough going on in there that they probably should. Not in this PR, though.

Resolves https://github.com/zooniverse/Panoptes/issues/2459

# Review checklist


app/models/organization.rb
```diff
@@ -15,6 +15,8 @@ class Organization < ActiveRecord::Base
   has_one :background, -> { where(type: "organization_background") }, class_name: "Medium", as: :linked
   has_many :organization_roles, -> { where.not(roles: []) }, class_name: "AccessControlList", as: :resource
   has_many :pages, class_name: "OrganizationPage", dependent: :destroy
+  has_many :attached_images, -> { where(type: "organization_attached_image") }, class_name: "Medium",
```
camallen: maybe just use "org_attached_image" or something that conveys the table link correctly to keep the index size down

zwolf: ✔️

Issue #2319, opened 2017-05-17
Slow db queries paging into large tables using offset strategy


camallen:
Linked to #136 

The classifications#project end point by default uses page numbers to offset into the result set. This is slow for large tables spaces especially projected result sets (with filtered queries project_id = x). 

Should we just turn off paging for end points on this table forcing people to use the last_id offset method instead? This is non-normal behaviour for our API though and any kind of paging on this table will degrade with higher page offset values. Perhaps we can turn of past N pages?

Thoughts?

astopy:
Would the endpoint still be useful without paging? Is there a use case for only wanting the first (few) page(s) but not the others? If so, removing/limiting paging would be a good solution. Otherwise maybe we should disable it and have the API return a message telling the user to use the last_id or CSV exports.

We could also think about keeping paging enabled but loading the data from somewhere else. This could be part of reworking the CSV exports, since we want to stop generating them from scratch every time. E.g. we could write classifications to S3 in batches (say, 1,000 lines per file), then this endpoint can use the page_size*page_number to figure out which files to read (while the exports can just cat all of them together).

camallen:
In the short term I think we could look at removing / restricting paging for this endpoint but it's not standard behaviour. I guess as long as our error messaging is good it should be ok. 

Long term i agree we should tie this into better exports so we can serialize the ordered data and avoid the lookup costs into the table.

camallen:
a per project sorted id with compound? index could work here though might be another RAM cost for just this use case. e.g. 

| id | project_id | sort_id |
| -- | --------- | ------- |
| 1  | 1 | 1 |
| 2 | 2 | 1  |
| 3 | 1 | 2 |
| 4 | 2 | 2  |

marten:
It's worth noting here that Caesar is using this endpoint to backfill old classifications. It filters to a specific subject (so it's probably not causing any slow queries), but still pages through if there are more than page_size classifications.

> We could also think about keeping paging enabled but loading the data from somewhere else. This could be part of reworking the CSV exports, since we want to stop generating them from scratch every time. E.g. we could write classifications to S3 in batches (say, 1,000 lines per file), then this endpoint can use the page_size*page_number to figure out which files to read (while the exports can just cat all of them together).

I'm not sure this would work. It's possible to call various filters on this endpoint, and then you no longer know which files to pull from.

PR #2537, opened 2017-11-21
add compound index for classifications by project and id

### LINKED ISSUE #2334: Incorrect results using last_id offset for paging
Linked to #2319

The db is returning unsorted data for the last_id offset so there is no guarantee that it'll return the expected data pages. We can fix this using small window pg queries or adding another col (&index) to do the per project sort order. Else we can look into better datastores (or postgres solutions) to support paging without using offset strategies.

camallen:
closes #2334 #2319

adds an index to filter by project_id and order by id at the same time e.g. 
`Classification.where(project_id: 1).where("id > ?", last_id).order(:id).limit(page_size)`

*Note: the rebuilt index is ~2x smaller than it was (4148 MB vs 1696 MB).* 

~I'm going to issue a new PR~ I've included a migration to concurrently rebuild and drop classifications table indexes. We should look at rebuilding the db / table indexes during downtimes or routinely rebuilding and dropping old ones to ensure we remove the index bloat. The code in the migration could be a nice target for extraction, i can do this if people think it'll be used more widely.

# Review checklist

### DIFF
--- a/db/migrate/20171121120455_rebuild_classification_indexes.rb
+++ b/db/migrate/20171121120455_rebuild_classification_indexes.rb
@@ -5,7 +5,7 @@ class RebuildClassificationIndexes < ActiveRecord::Migration
   def change
     reversible do |direction|
       direction.up do
-        rebuild_non_spare_indexes
+        rebuild_non_sparse_indexes
         rebuild_sparse_indexes
       end
     end
@@ -13,7 +13,7 @@ class RebuildClassificationIndexes < ActiveRecord::Migration
 
   private
 
-  def rebuild_non_spare_indexes
+  def rebuild_non_sparse_indexes
     index_cols.each do |column|
       rebuild_index(column) do
         add_index TABLE_NAME, column, algorithm: :concurrently
--- a/db/migrate/20171121120455_rebuild_classification_indexes.rb
+++ b/db/migrate/20171121120455_rebuild_classification_indexes.rb
@@ -3,20 +3,31 @@ class RebuildClassificationIndexes < ActiveRecord::Migration
   TABLE_NAME = :classifications
 
   def change
+    reversible do |direction|
+      direction.up do
+        rebuild_non_spare_indexes
+        rebuild_sparse_indexes
+      end
+    end
+  end
+
+  private
+
+  def rebuild_non_spare_indexes
     index_cols.each do |column|
       rebuild_index(column) do
         add_index TABLE_NAME, column, algorithm: :concurrently
       end
     end
+  end
 
+  def rebuild_sparse_indexes
     sparse_index_rebuilds.each do |column, clause|
       rebuild_index(column) do
         add_index TABLE_NAME, column, where: clause, algorithm: :concurrently
       end
     end
   end
-
-  private
 
   def rebuild_index(column)
     renamed_index_name = to_delete_index_name(column)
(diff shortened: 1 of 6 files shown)


db/migrate/20171121120455_rebuild_classification_indexes.rb
```diff
@@ -0,0 +1,57 @@
+class RebuildClassificationIndexes < ActiveRecord::Migration
```
camallen: This could be a target for extraction for a recurrent rebuild task longer term.

camallen:
@marten @zwolf can i get a review on this one too please?

spec/models/classification_spec.rb
```diff
@@ -218,16 +218,24 @@
...
         it 'returns only the classifications after last_id if provided' do
-          result = Classification.scope_for(:project, user, {last_id: Classification.first.id, project_id: project.id})
           expect(result).not_to include Classification.first
         end
 
-        it 'raise MissingParameter if project_id is not also provided' do
-          expect {
-            Classification.scope_for(:project, user, {last_id: Classification.first.id})
-          }.to raise_error(Classification::MissingParameter)
+        it 'should order by id for determinisitc result sets' do
```
zwolf: typo

spec/serializers/classification_serializer_spec.rb
```diff
@@ -33,21 +33,26 @@
...
+      end
+      let(:prefix) { "/classifications?last_id=" }
+      let(:suffix) { "&page_size=1&project_id=#{project.id}" }
 
-      it "should insert the next hightest last_id into the next_href" do
+      it "should insert the highest page set id into the next_href" do
         second = create(:classification, project: project)
         params = {project_id: project.id, last_id: last_id, page_size: 1}
-        result = ClassificationSerializer.page(params, Classification.all, {})
+        result = ClassificationSerializer.page(params, scope, {})
```
zwolf: Using the `.where().after_id()` clause in your scope but also having the `last_id` in the params seems kind of redundant. The params are what the serializer uses and is what's being tested here.

camallen: The serializer doesn't create the scope based on params, the scope is fed in via the controller action, so the project scope params here don't do anything, i'll remove them to clarify what's being tested.

camallen: Ah - both the scope setup and the params are needed. Scope is what the params will create via normal controller flows and `scope_for` so this needs to be fed into the serializer. Also the params are used to construct valid next / prev href links and they flow normally through the serializer as they are here.

db/migrate/20171121120455_rebuild_classification_indexes.rb
```diff
@@ -0,0 +1,57 @@
...
+  disable_ddl_transaction!
+  TABLE_NAME = :classifications
+
+  def change
+    reversible do |direction|
```
marten: TIL about this.

db/migrate/20171121120455_rebuild_classification_indexes.rb
```diff
@@ -0,0 +1,57 @@
...
+      direction.up do
+        rebuild_non_spare_indexes
+        rebuild_sparse_indexes
+      end
+    end
+  end
+
+  private
+
+  def rebuild_non_spare_indexes
```
marten: typo in "sparse" ;)

camallen: blargh

db/migrate/20171121120455_rebuild_classification_indexes.rb
camallen: this was new for me too :) quite useful for the index builds

PR #2538, opened 2017-11-22
Fix email dump uploads

camallen:
Handle sidekiq params symbol -> string round trip lookups and ensure we use the correct s3 signature version for the emails bucket.

# Review checklist


astopy:
I'm not really familiar enough with the code to review this, so I'll leave it to @marten

PR #2541, opened 2017-11-23
Upgrade s3 client

camallen:
bump to aws-sdk to v2 which defaults to signature_version v4 since v2.2 and update the codepaths for the new api / deprecated methods.

I manually ran the upload code in a repl to test it works instead of relying on a double to receive messages. 

https://github.com/aws/aws-sdk-ruby/blob/master/UPGRADING.md#aws-sdk-core---v220

# Review checklist


camallen:
No it uses the storage adapter to do the interaction with s3. I've got a new PR (that needs a rebase) to update it to fix the upload path and add an encryption check.

PR #2548, opened 2017-11-28
extend the default get / put presigned urls timeouts

camallen:
give the clients more time to use the presigned URL’s by default, this options can still be overridden in the calling methods.

# Review checklist

### DIFF
--- a/spec/services/media_storage/aws_adapter_spec.rb
+++ b/spec/services/media_storage/aws_adapter_spec.rb
@@ -27,18 +27,19 @@ RSpec.describe MediaStorage::AwsAdapter do
     end
 
     describe "s3 client config" do
-      before do
-        expect(Aws::S3::Client)
-        .to receive(:new)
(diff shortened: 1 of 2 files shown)


spec/services/media_storage/aws_adapter_spec.rb
```diff
@@ -18,20 +18,29 @@
...
+        .to receive(:new)
+        .with(s3_opts.except(:prefix, :bucket))
+        .and_call_original
+      end
 
-    it 'should set the aws config through the s3 client ' do
-      adapter
-    end
+      it 'should set the aws config through the s3 client ' do
+        adapter
```
marten: This style of tests continues to confuse me. I don't think I'll ever think to look for mock expectations in a before block. It makes the test examples confusing for me.

@zwolf  I need a tie-breaker on code style here.

camallen: i didn't really like it either just saved some extra typing. i'll move it.

PR #2549, opened 2017-11-29
Sync translated resource strings to translations data model

camallen:
Batch process to seed the conversion of translatable resources to new translations data model. This will be run in a rake task to initially seed the data / iron out schema bugs. 

# Review checklist

### DIFF
--- a/app/models/concerns/translation_strings.rb
+++ b/app/models/concerns/translation_strings.rb
@@ -11,6 +11,24 @@ class TranslationStrings
 
   private
 
+  # def visitor_extract_strings
+  #   if content_model_resource?
+  #     attrs
+  #   else
+  #     # this doesn't seem to be used in pandora
+  #     # looks like pandora will iterate over the list of
+  #     # translated resources and extract the attributes in visit_question
+  #     visitor = TasksVisitors::ExtractStrings.new
+  #     stripped_items = visitor.visit(attrs[*translatable_attributes])
+  #     extracted_strings = visitor.collector
+  #   end
+  # end
+  #
+  # CONTENT_MODELS = %w(project workflow).freeze
+  # def content_model_resource?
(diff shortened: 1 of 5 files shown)


app/models/concerns/translation_strings.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  def removed_nested_attributes(attributes_to_slice, key)
+    attributes[key.to_sym].map do |nested_object|
+      nested_object.slice(*attributes_to_slice)
+    end
+  end
+
+  def flatten_nested_schema(hash_schema)
+    # TODO: convert this to a visitor class that task the schema and converts
```
camallen: I'd like to get this in but i'm also happy to do this work before merging.

app/models/translation.rb
```diff
@@ -4,9 +4,12 @@ class Translation < ActiveRecord::Base
   validates_presence_of :language
 
   # TODO: add a unique validation for translated_type, id, language
+  # and a unique index
```
camallen: I'll do these model updates in another PR

app/models/concerns/translation_strings.rb
```diff
@@ -0,0 +1,113 @@
...
+  # pandora app uses the tasks key to find the tasks strings
+  def transform_workflow_attributes
+    attributes["tasks"] = attributes.delete("strings")
+  end
+
+  def transform_tutorial_attributes(nested_key="steps")
+    attributes[nested_key] = removed_nested_attributes(%i(content), nested_key)
+  end
+
+  def removed_nested_attributes(attributes_to_slice, key)
```
camallen: I don't think this method name is appropriate, perhaps it should be `extract_nested_attributes`

marten: Agree

PR #2557, opened 2017-12-04
Revert "Add Honeybadger call to get backtrace of enqueue location"

marten:
Reverts zooniverse/Panoptes#2550

Hasn't happened again in these past days. @camallen how often were these errors before?

### DIFF
--- a/app/workers/subject_set_subject_counter_worker.rb
+++ b/app/workers/subject_set_subject_counter_worker.rb
@@ -11,14 +11,6 @@ class SubjectSetSubjectCounterWorker
       }),
     unique: :until_executing
 
-  def self.perform_async(subject_set_id)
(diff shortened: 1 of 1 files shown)


camallen:
Occasional though i managed to see it a few times. Noting that I only observed this when looking at I was looking at running jobs. Let's leave it for a little bit longer as it might be a sidekiq UI issue though i somewhat doubt that.

camallen:
Also what's up with the extra revert commits?

marten:
Not sure. I just hit the revert-PR button in Github. Happy to leave this in place for a while longer, just wanted to make sure we had a PR open to remind us to take it out.

marten:
@camallen okay, well this didn't seem to help get stuff into HB. Shall we merge this?

PR #2560, opened 2017-12-07
Rails 5 prep: upgrading some gems

marten:
This branch is prep work for a future Rails 5 upgrade. There will probably be other PRs to fix up use of old deprecated gems or APIs.

# Review checklist


app/views/devise/shared/_links.erb
```diff
@@ -12,11 +12,9 @@
     <%- if devise_mapping.omniauthable? %>
       <%- resource_class.omniauth_providers.each do |provider| %>
         <%- if provider.to_s == 'facebook' %>
-          <%= link_to ("<i class='fa fa-facebook-official fa-lg'></i> Sign in").html_safe, user_omniauth_authorize_path(provider), class: "devise-shared-link #{provider.to_s}-link" %>
+          <%= link_to ("<i class='fa fa-facebook-official fa-lg'></i> Sign in").html_safe, user_facebook_omniauth_authorize_path, class: "devise-shared-link #{provider.to_s}-link" %>
         <%- elsif provider.to_s == 'google_oauth2' %>
-          <%= link_to ("<i class='fa fa-google fa-lg'></i> Sign in").html_safe, user_omniauth_authorize_path(provider), class: "devise-shared-link #{provider.to_s}-link" %>
-        <%- else %>
-          <%= link_to "Sign in with #{provider.to_s.titleize}", user_omniauth_authorize_path(provider), class: "devise-shared-link #{provider.to_s}-link" %>
```
marten: The generic helper method is gone. Rather than using `send("user_#{provider}_omniauth_authorize_path")` which seemed like a potential security issue, I opted to remove it completely. We probably want each configured provider to have an actual icon anyway.

marten:
Yeah that's why I'm conservative with gem updates going only as far as needed for Rails 5, at least in this series of PRs

PR #2561, opened 2017-12-08
Remove subject queue

marten:
# Review checklist


marten:
Not sure, @zwolf shall the three of us have a chat about preferences for this while we're all still colocated?

zwolf:
After talking with @marten, I'm fine with just pulling it out. Sending a 410 would be equivalent to breaking it anyway since we're not versioning in any meaningful way and maintaining controller code just to send a status that no one will see because this functionality is deprecated by the newer subject selection services is more trouble than it's worth. 

If we can send that status from a single line in the router, maybe. Otherwise just let it 404.

marten:
Doing the migration later was my initial take, and what @camallen originally meant. Upon reflection I'm going to go back to that, so i've removed that commit from this branch.

PR #2563, opened 2017-12-08
Convert Collection.projects to habtm association

marten:
We should have a think about how to deploy this seamlessly. I think if we migrate, run the task, and then deploy new API nodes, we should be okay, but someone else also coming to that conclusion would be very much helpful.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/collections_controller.rb
+++ b/app/controllers/api/v1/collections_controller.rb
@@ -39,7 +39,7 @@ class Api::V1::CollectionsController < Api::ApiController
   def filter_by_project_ids
     if ids_string = (params.delete(:project_ids) || params.delete(:project_id)).try(:split, ',')
       project_ids = ids_string.split(",")
-      @controlled_resources = controlled_resources.where.overlap(project_ids: project_ids)
+      @controlled_resources = controlled_resources.joins(:projects).where(projects: {id: project_ids})
     end
   end
 
--- a/app/controllers/api/v1/collections_controller.rb
+++ b/app/controllers/api/v1/collections_controller.rb
@@ -5,13 +5,15 @@ class Api::V1::CollectionsController < Api::ApiController
 
   before_action :filter_by_project_ids, only: :index
   before_action :pluralize_project_links, only: :create
+  before_action :link_up_habtm_projects, only: :create # TODO: After PR 2563, remove this
 
   require_authentication :create, :update, :destroy, scopes: [:collection]
   resource_actions :default
   schema_type :strong_params
 
+  # TODO: After 2563, remove habtm_projects
   allowed_params :create, :name, :display_name, :private, :favorite, :description,
-    links: [ :default_subject, :project, projects: [], subjects: [], owner: polymorphic ]
+    links: [ :default_subject, :project, projects: [], habtm_projects: [], subjects: [], owner: polymorphic ]
 
   allowed_params :update, :name, :display_name, :private, :description, links: [ :default_subject, subjects: [] ]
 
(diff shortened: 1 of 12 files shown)


db/migrate/20171208142645_convert_collection_belongs_to_many.rb
```diff
@@ -0,0 +1,8 @@
+class ConvertCollectionBelongsToMany < ActiveRecord::Migration
+  def change
+    create_table :collections_projects, id: false do |t|
```
camallen: not having an id makes it hard to switch to an AR model for internal use.

db/structure.sql
```diff
@@ -3449,23 +3474,23 @@ ALTER TABLE ONLY classification_subjects
 
 
 --
--- Name: fk_rails_81596e7851; Type: FK CONSTRAINT; Schema: public; Owner: -
+-- Name: tutorials fk_rails_82e4d0479b; Type: FK CONSTRAINT; Schema: public; Owner: -
 --
 
-ALTER TABLE ONLY subject_queues
```
camallen: Does this build on the subject queue removal PR, might need a rebase if so?

app/controllers/api/v1/collections_controller.rb
```diff
@@ -39,7 +39,7 @@ def check_default_subject(collection)
   def filter_by_project_ids
     if ids_string = (params.delete(:project_ids) || params.delete(:project_id)).try(:split, ',')
       project_ids = ids_string.split(",")
-      @controlled_resources = controlled_resources.where.overlap(project_ids: project_ids)
+      @controlled_resources = controlled_resources.joins(:projects).where(projects: {id: project_ids})
```
camallen: does the project_ids col still exist on collections after this PR?

db/migrate/20171208142645_convert_collection_belongs_to_many.rb
marten: Would you prefer a `has_many :collections, through: `? I was wondering, and couldn't think of any property we might stick on the join model.

camallen: I'm not sure really i find myself often trying to figure out the state of the classification -> subjects and have to write out the join statement  instead of a where on `ClassificationSubjects.where(subject_id: 1)`. If we have no need to write that `CollectionProject.where(collection_id: 1)` query then this point is moot.

marten:
Agree about the ideally, but not sure how to implement that. Where would I hook into to make sure we write to both places?

camallen:
it really depends how long this migration needs to run, if it's quick to backfill the join table maybe we can wear it? 

Technically i think we could have the new habtm relation defined under a different name then the `project_ids` method doesn't get overriden so the `belongs_to_many` should keep working. Could this relation then be setup to copy the current api project links in the controller `build_resource_for_create` with a block or setting the links to the relation name here ? https://github.com/zooniverse/Panoptes/blob/e871746789d52b71dbc26d7e8ccf87acf8763e8d/app/controllers/api/v1/collections_controller.rb#L52

marten:
Yep, I'll make it so that this just keeps working while we convert things slowly in the background.

Note to self, add habtm_projects to the serializer, and copy over data from links to also populate the habtm_projects relation on create/update

marten:
@camallen okay, wanna give this another review?

spec/factories/collections.rb
```diff
@@ -12,11 +9,7 @@
     association :owner, factory: :user
 
     project_ids do
-      if build_projects
-        [ create(:project).id ]
-      else
-        (1..10).to_a.sample(2)
-      end
+      [ create(:project).id ]
```
camallen: Collections will always build a project now (and all the cascading project resources) and if heavily used it can slow the tests down. I'm not so worried but if known maybe pass in the project when creating the factory? Also should we rename this to projects association?

marten: I considered that but there's no real way to search for locations where it does need the project to be built. Perhaps in a followup PR?

camallen: up to you, it was transient before so only those marked to build it required it by default.

PR #2578, opened 2017-12-13
Add unique db index and validation for translated resource languages

camallen:
Make sure we ever only save 1 language translation for a resource. this adds a unique compound db index that satisfies the validation query and our controller scope_for access queries. 

TODO
- [x] Add a rake task / migration step to downcase any existing language attributes.

# Review checklist

### DIFF
--- a/spec/factories/translations.rb
+++ b/spec/factories/translations.rb
@@ -22,16 +22,16 @@ FactoryGirl.define do
             type: 'drawing',
             question: "Draw a circle",
             help: "Duh?",
-            # the tool lables array might be best un-wound to avoid ambiguity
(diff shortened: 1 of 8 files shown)


spec/factories/translations.rb
```diff
@@ -22,16 +22,16 @@
             type: 'drawing',
             question: "Draw a circle",
             help: "Duh?",
-            # the tool lables array might be best un-wound to avoid ambiguity
-            # "tool_lablel.2" => "Blue",
-            tool_labels: %w(Red Green Blue)
+            "tool_lablel.0" => "Red",
```
marten: lable?

PR #2602, opened 2018-01-10
Add featured flag to projects

### LINKED ISSUE #2598: Featured project
This came up in Slack's design thoughts channel. The featured project on the home page should be managed via the API instead of us updating static content every few weeks. What this would needs is:
- a `featured` boolean
- a query param on the `/projects` endpoint

@beckyrother only wants one featured project at a time. We thought that this could be managed by a toggle on the admin page for a project. Perhaps there could be a validation on projects that only allows a single project be `featured: true` at any time?

marten:
Fixes #2598.

# TODO

- [x] Enforce that only 1 project at a time is featured

# Review checklist

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -57,12 +57,15 @@ class Project < ActiveRecord::Base
 
   validates_inclusion_of :private, :live, in: [true, false], message: "must be true or false"
 
+  validates :featured, uniqueness: {conditions: -> { featured } }
+
   ## TODO: This potential has locking issues
   validates_with UniqueForOwnerValidator
 
   after_update :send_notifications
 
   scope :launched, -> { where("launch_approved IS TRUE") }
+  scope :featured, -> { where(featured: true) }
 
   can_by_role :destroy, :update, :update_links, :destroy_links, :create_classifications_export,
     :create_subjects_export, :create_workflows_export,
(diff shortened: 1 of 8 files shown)


db/migrate/20180110133833_add_featured_to_projects.rb
```diff
@@ -0,0 +1,5 @@
+class AddFeaturedToProjects < ActiveRecord::Migration
+  def change
+    add_column :projects, :featured, :boolean, default: false, null: false, index: true
```
camallen: wouldn't this be better with a nullable type / sparse index vs this?

app/models/project.rb
```diff
@@ -57,12 +57,15 @@ class Project < ActiveRecord::Base
 
   validates_inclusion_of :private, :live, in: [true, false], message: "must be true or false"
 
+  validates :featured, uniqueness: {conditions: -> { featured } }
```
camallen: this seems like it's overly complex, couldnt the client control how many are featured with a page_size = 1 and a sort order? Also that way we get multiple featured projects and leave the client to figure out which to display?

marten: Yeah, that was my thought too, but @beckyrother was very against supporting multiple featured projects. I'm happy to not provide any API support for it if you want.

camallen: right - that is until they do ;) Up to you re leaving it and the implementation is fine.

PR #2605, opened 2018-01-11
[WIP] Stdout logging for graylog

camallen:
Staging env (and opt in dev) defaults to graylog message forwarding now by default. This may need some tweaking on the log format but it looks ok (different from the logstasher format). 

Also sidekiq will need its own logger pushing to graylog. i've left a note on how to do this. If we're happy with how this is working out we can move away from logstasher as it was writing duplicate formatted files for gelfcat too. 

We could also log to stdout and let gelfcat still work and that might help sidekiq too. that's another option available instead of using the ruby gelf gem and one i'm actually leaning towards. @astopy reasons we shouldn't use file redirection > stdout and gelfcat -> graylog?

# TODO

- [ ] Get this logging to graylog to see what the message payload looks like and see if we can get extractors working for this. 
- [ ] Test stdout logging with gelfcat docker plugin on a formatted json payload to see how that works with graylog

# Review checklist

### DIFF
--- a/lib/logging/graylog_defaults.rb
+++ b/lib/logging/graylog_defaults.rb
@@ -1,0 +1,30 @@
+module Logging
+  module GraylogDefaults
+    def self.config
+      [
+        ENV.fetch("GRAYLOG_HOST_NAME", "graylog.zooniverse.org"),
+        ENV.fetch("GRAYLOG_HOST_PORT", 12201),
+        ENV.fetch("GRAYLOG_UPD_CHUNK_CONFIG", "WAN"),
+        { host: app_source_name }
+      ]
+    end
+
+    # the source name that appears in graylog UI
+    def self.app_source_name
+      suffix = if !Rails.env.production?
+                "rails"
+               else
+                "#{Rails.env}-rails"
+               end
+      ENV.fetch("GRAYLOG_SOURCE_NAME", "panoptes-api-#{suffix}")
+    end
+
+    # TODO: add sidekiq logger for gelf / or just drop it and
+    # and keep them in stdout
+    # https://github.com/mperham/sidekiq/wiki/Logging#customize-logger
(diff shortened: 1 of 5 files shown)


lib/logging/graylog_defaults.rb
```diff
@@ -0,0 +1,30 @@
+module Logging
+  module GraylogDefaults
+    def self.config
+      [
+        ENV.fetch("GRAYLOG_HOST_NAME", "graylog.zooniverse.org"),
```
astopy: `graylog.zooniverse.org` isn't directly accessible by Panoptes. Maybe make this `panoptes-logs.zooniverse.org` instead?  That still won't be right for staging, which uses `panoptes-staging-logs.zooniverse.org`, but at least that's one less config file we need to set it in.

camallen: `panoptes-logs.zooniverse.org` is accessible from both staging and prod vpcs?

camallen:
If docker has a built in gelf forwarder i think we should use that instead of using rubyland code for this. our logging should be as simple as possible, only log to stdout from rails / sidekiq and leave it up to other services to format / push that log stream where it needs to go.

lib/logging/graylog_defaults.rb
astopy: Nope, staging can only access `panoptes-staging-logs.zooniverse.org` so we'll still need to set `GRAYLOG_HOST_NAME` there.

astopy:
Either way would be fine with me. [See designator config](https://github.com/zooniverse/operations/blob/5441969ad67bb7e99a8c28ab33ed723bf2c2bcc3/panoptes-designator/docker-compose.yml#L11) for an example.

camallen:
Closing this - i'll reopen later or start afresh.

PR #2606, opened 2018-01-11
Remove belongs_to_many AR extension

marten:
Do not merge this until #2563 is delployed and migrated. Will want to rebase this on master so that the diff doesn't include all of #2563 for review.

# Review checklist


app/controllers/api/v1/collections_controller.rb
```diff
@@ -39,7 +39,7 @@ def check_default_subject(collection)
   def filter_by_project_ids
     if ids_string = (params.delete(:project_ids) || params.delete(:project_id)).try(:split, ',')
       project_ids = ids_string.split(",")
-      @controlled_resources = controlled_resources.where.overlap(project_ids: project_ids)
+      @controlled_resources = controlled_resources.joins(:projects).where(projects: {id: project_ids})
```
camallen: can't this just be `controlled_resources.projects.where(id: project_ids)` ?

marten: That would return projects, but we need to return collections?

Issue #2600, opened 2018-01-09
add announcement to organization resource


mcbouslog:
Add `announcement` property to organization resource. Value would be a string, similar to `description` or `introduction`.

The announcement would be featured on the org landing, to announce special events, challenges, progress, etc. similar to how announcements are used on a project's landing page.

PR #2611, opened 2018-01-16
Add announcement field to organizations

zwolf:
I added this field to the organization contents. Projects store announcements in their config JSON blobs, but orgs don't/shouldn't have one of those and they should be translatable, anyway.

Resolves https://github.com/zooniverse/Panoptes/issues/2600

# Review checklist

### DIFF
--- a/app/models/organization_content.rb
+++ b/app/models/organization_content.rb
@@ -5,4 +5,5 @@ class OrganizationContent < ActiveRecord::Base
   validates_length_of :title, maximum: 255
   validates_length_of :description, maximum: 300
   validates_length_of :introduction, maximum: 5000
+  validates_length_of :announcement, maximum: 5000
 end
--- a/db/migrate/20180115214144_add_announcement_to_organization_contents.rb
+++ b/db/migrate/20180115214144_add_announcement_to_organization_contents.rb
@@ -1,5 +1,5 @@
 class AddAnnouncementToOrganizationContents < ActiveRecord::Migration
   def change
-    add_column :organization_contents, :announcement, :string, default: "", index: true
+    add_column :organization_contents, :announcement, :string, default: ""
   end
 end
--- a/db/migrate/20180115214144_add_announcement_to_organization_contents.rb
+++ b/db/migrate/20180115214144_add_announcement_to_organization_contents.rb
@@ -1,0 +1,5 @@
+class AddAnnouncementToOrganizationContents < ActiveRecord::Migration
+  def change
+    add_column :organization_contents, :announcement, :string, default: "", index: true
+  end
+end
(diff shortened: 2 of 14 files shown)


camallen:
the code looks good however I was planning on leaving project announcements from the owner / researcher in the primary language. Mostly as they are ephemeral and thus can be gone before a translator (for each language) can get to it. 

Translations as they stand are for the static content, even then they will have issues of getting out of sync. i'm not 100% certain but I think we should switch this to the way projects work and leave announcements in the primary language. Thoughts?

eatyourgreens:
+1 for adding announcements to translations. Announcements in project config was a hack, and we're now in a position where people put long-running announcement banners on their projects pages, without being able to translate them into other languages.

In practice, I've been seeing projects put banners up for weeks at a time (if not months) on the live site.

eatyourgreens:
I should add, based on experience with Penguin Watch, translators update text within 24 hours of a banner appearing on the home page, if they're given notice.

camallen:
OOI which ones have been around for weeks / months? Is this due a replacement for updating results pages? Or a way to signal paused and finished status? Automatically moving to paused / finished could alleviate some of this.

eatyourgreens:
I think it's mostly signalling paused/finished status but you'd also get announcements like the one we added to the Penguin Watch home page, I suppose.

https://www.zooniverse.org/projects/sbradford/western-montana-wildlife
https://www.zooniverse.org/projects/mrniaboc/bash-the-bug
https://www.zooniverse.org/projects/mozerm/snow-spotter
https://www.zooniverse.org/projects/zooniverse/gravity-spy

app/models/organization_content.rb
```diff
@@ -5,4 +5,5 @@ class OrganizationContent < ActiveRecord::Base
   validates_length_of :title, maximum: 255
   validates_length_of :description, maximum: 300
   validates_length_of :introduction, maximum: 5000
+  validates_length_of :announcement, maximum: 5000
```
camallen: 5K chars seems like it would take up a chunk of UI with text. maybe 2k limit?

db/migrate/20180115214144_add_announcement_to_organization_contents.rb
```diff
@@ -0,0 +1,5 @@
+class AddAnnouncementToOrganizationContents < ActiveRecord::Migration
+  def change
+    add_column :organization_contents, :announcement, :string, default: "", index: true
```
camallen: why do we need an index on this possibly large string?

app/models/organization_content.rb
zwolf: 5000 is what we gave both introduction and description. I feel like a character limit should be enforced on the front end before it hits a max and the request fails opaquely.

db/migrate/20180115214144_add_announcement_to_organization_contents.rb
zwolf: good point.

PR #2612, opened 2018-01-16
add project metadata config option to restrict self service data access

camallen:
some projects may contain sensitive data that we need to control access to. this is a first step to avoid leaking that data to the stream and associated services. If this becomes widely needed it should be a column on the project with an index.

# TODO:

- [x] Add in restrictions on classification (subjects?) dumps for this project at the API level so there are decent error messages in the UI
- [x] Add in restrictions on classification api end points for all users (admin included)
- [ ] Test this on staging / prod to ensure no data is published to the stream and classification exports cannot be requested via the api.
# Review checklist

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -77,8 +77,7 @@ class ClassificationLifecycle
   def publish_data
     return unless classification.complete?
 
-    private_data = !!project.configuration.fetch("private_data", false)
-    unless private_data
+    unless project.disabled_data_export?
       PublishClassificationWorker.perform_async(classification.id)
     end
   end
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -77,8 +77,8 @@ class ClassificationLifecycle
   def publish_data
     return unless classification.complete?
 
-    allow_in_stream = !!project.configuration.fetch("allow_in_stream", true)
-    if allow_in_stream
+    private_data = !!project.configuration.fetch("private_data", false)
+    unless private_data
       PublishClassificationWorker.perform_async(classification.id)
     end
   end
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -77,7 +77,10 @@ class ClassificationLifecycle
   def publish_data
     return unless classification.complete?
 
-    PublishClassificationWorker.perform_async(classification.id)
+    allow_in_stream = !!project.configuration.fetch("allow_in_stream", true)
+    if allow_in_stream
+      PublishClassificationWorker.perform_async(classification.id)
+    end
   end
 
   def notify_subject_selector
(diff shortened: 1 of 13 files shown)


zwolf:
Would workflow and workflow contents exports leak the kind of data this is meant to protect?

camallen:
I don't think so, though happy to block these as well. @astopy thoughts on if we need to block more exports than just classification data?

astopy:
I don't think there's anything else we need to block

lib/classification_lifecycle.rb
```diff
@@ -77,7 +77,9 @@ def update_seen_subjects
   def publish_data
     return unless classification.complete?
 
-    PublishClassificationWorker.perform_async(classification.id)
+    unless project.disabled_data_export?
```
marten: What do data exports have to do with the internal service bus we run?

marten:
I don't get it. Kinesis is a private stream, for our own internal services only. Why can't we publish data to our own services?

camallen:
Because we had a meeting saying we won't publish any of this data to be sure we don't accidentally leak it.

lib/classification_lifecycle.rb
camallen: the method can be renamed to reflect pushing data to the stream as well as csv dumps. Both are exports in my mind just different mechanisms.

marten:
Yeah but it does cause issues with the idea of how our architecture works. This little boolean is terribly named, since it disables exports, but disabling the stream also disables Caesar. I'm planning to get Designator to listen to the stream to know when to refresh caches. This would mean that Designator gets disabled too. All of these are side-effects that are really hidden, and hard to take into account (e.g. if I built that feature for designator, I'd have to remember that this flag is a thing, and then change the subject selector code to not use designator if the exports are disabled).

If the stream isn't secured properly we're in a lot of shit regardless. I'm 9/10 against this change for the time being.

camallen:
> This little boolean is terribly named

Agree the name could be better, happy to rename it and the surrounding methods to clarify intent.

Just adding, this is for a one off experimental project, it's not for wider use and was the simplest thing to do to get this out the door. We had said meeting and decided to do the bare minimum to get this out the door (including dropping stats counts). If we do this again we'll extract this work to ensure internal services have the data and re-work the public data output to ignore this data.

camallen:
I agree with your concerns regarding how this can impact our architecture and data flow. I really want to stress that this is a one off, special project and this code / functionality **will** be removed after this project has run.

chrislintott:
I think for one project, when we don;'t know how often we'll need to do this this is fine. @marten - you're not wrong in the long run, but at this point I think we should just do this and see if it creates demand for future such projects. I agree this is a stopgap solution, but sometimes stopgaps are needed.

marten:
Okay.

marten:
In that case I suggest we call the method something like `temporary_flag_to_keep_data_internal_to_panoptes`?

PR #2616, opened 2018-01-19
Remove postgres_ext gem

marten:
Compatibility with Rails 5 isn't good. There are a few forks that fix one or two issues, but nobody has merged everything, and even then I think they wouldn't combine to full compatibility.

Basically, the gem seems to be abandoned.

I advise we test this pretty well though, I'm not 100% confident everything is covered by the test suite.

# Review checklist


app/models/membership.rb
```diff
@@ -6,7 +6,7 @@ class Membership < ActiveRecord::Base
   enum state: [:active, :invited, :inactive]
 
   scope :active, -> { where(state: states[:active]) }
-  scope :identity, -> { active.where(identity: true, roles: ["group_admin"]) }
+  scope :identity, -> { active.where(identity: true).where("'group_admin' = ANY(memberships.roles)") }
```
camallen: Not a blocking comment, I think the identity group is a special one here that will only have the roles set to `["group_admin"]`. Are there any existing identity memberships that have more than just this listed role?  TBH the ANY operator looks fine, i'm only slightly concerned about relaxing the  criteria for matching an identity membership, though in theory the clauses should be the same.

marten: No, looks like all identity groups are exactly `["group_admin"]`. Changing to different SQL queries definitely wasn't intended for this PR, so nice catch.

PR #2624, opened 2018-01-23
remove old indexes from structure.sql 

camallen:
Old indexes crept into the structure from #2611 and due to whitespace formatting, the diff was large so didn't show in the PR by default. We should either leave the whitespace as is output from the `rake db:migrate` task or all strip whitespace on this file for any PR. I vote leave whitespace as it is the easiest to avoid messing up. 

Either way we need to ensure we see the diffs on this file and they do what we expect.

# Review checklist

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -58,7 +58,7 @@ SET default_tablespace = '';
 SET default_with_oids = false;
 
 --
--- Name: access_control_lists; Type: TABLE; Schema: public; Owner: -; Tablespace:
+-- Name: access_control_lists; Type: TABLE; Schema: public; Owner: -; Tablespace: 
 --
 
 CREATE TABLE access_control_lists (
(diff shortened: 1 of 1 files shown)


zwolf:
I incorrectly left those indexes in when I rebased that PR, I misread and assumed they were added and not deleted. Obviously my editor removed that whitespace for me, and it will every time I edit that file manually. Merge conflicts in this file are horrific and while re-generating after migrating it is probably the best way, try telling git that during a rebase.

camallen:
> Obviously my editor removed that whitespace for me, and it will every time I edit that file manually

Do you deal with merge conflicts in your editor? I normally use an external merge tool to deal with this. 

Yeah - completely agree about rebasing and merge on this file. I normally try my best and then clear my test db and reset the structure.sql from master (or whichever branch i'm rebasing from) and then re-apply the migrations in my PR. I'd be keen to know about better ways of dealing with this issue if peeps have them.

Issue #2586, opened 2017-12-29
org attached images not working as expected


mcbouslog:
for a project the following returns the linked attached images:
`https://panoptes-staging.zooniverse.org/api/projects/1693/attached_images?http_cache=true`

but for an org the following returns a 404:
`https://panoptes-staging.zooniverse.org/api/organizations/5/attached_images?http_cache=true`

not sure, but might be related to the org attached_image's href.
note the `href` attribute for this project linked media:
![screen shot 2017-12-27 at 6 06 07 pm](https://user-images.githubusercontent.com/12118751/34444832-8f0382d8-ec96-11e7-9854-d170fccf1b5b.png)
and for this org linked avatar (which works as expected):
![screen shot 2017-12-29 at 12 49 06 pm](https://user-images.githubusercontent.com/12118751/34444867-c0d75b72-ec96-11e7-83a8-2903f6b16813.png)

compared to the org linked attached_image:
![screen shot 2017-12-27 at 6 02 27 pm](https://user-images.githubusercontent.com/12118751/34444919-e7772e6a-ec96-11e7-87c0-1d8efb5f1915.png)

mentioning @zwolf

PR #2628, opened 2018-01-23
Org serializer link update

zwolf:
@camallen added the specs, I made "org" say "organization". This fixes the broken href in serialized orgs. 

Fixes https://github.com/zooniverse/Panoptes/issues/2586

# Review checklist

### DIFF
--- a/db/migrate/20180115214144_add_announcement_to_organization_contents.rb
+++ b/db/migrate/20180115214144_add_announcement_to_organization_contents.rb
@@ -1,5 +1,6 @@
 class AddAnnouncementToOrganizationContents < ActiveRecord::Migration
   def change
     add_column :organization_contents, :announcement, :string, default: ""
+    Medium.where(type: "org_attached_image").update_all(type: "organization_attached_image")
   end
 end
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -45,6 +45,7 @@ class Medium < ActiveRecord::Base
   end
 
   def linked_resource_details
+    # field_guide has a _ in it so manually find it vs prefix_*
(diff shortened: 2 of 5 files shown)


zwolf:
Unless you're extremely worried about a couple (tops) of lingering rows in the media table, this is new and has never been used outside of a bit of testing and no data migration is required.

camallen:
I'd prefer not to leave data out of sync with code / schema. Prod has 0, staging has 23 of these media resources. I'm not worried about staging and I'll leave it up to you, though a migration should be as easy as:
`Medium.where(type: "org_attached_image").update_all(type: "organization_attached_image")`

zwolf:
That should do it then, yeah?

db/migrate/20180115214144_add_announcement_to_organization_contents.rb
```diff
@@ -1,5 +1,6 @@
 class AddAnnouncementToOrganizationContents < ActiveRecord::Migration
   def change
     add_column :organization_contents, :announcement, :string, default: ""
+    Medium.where(type: "org_attached_image").update_all(type: "organization_attached_image")
```
camallen: Hasn't this migration already run on staging? so this change won't do anything...

zwolf: Has it? This PR hasn't merged yet, why would it have deployed to staging?

zwolf: yeah, but a whole extra migration just for this seems like overkill. I'll fix it on the console.

camallen: 👍

PR #2629, opened 2018-01-23
Asynchronous translatable resource strings updates

camallen:
Hook into the controller actions when creating/updating translatable resources to sync the related translation strings via a worker. 

# Review checklist


marten:
OOI, is there a performance concern with the syncing? Wondering why you chose a worker rather than just in-request.

camallen:
Mostly keeping response times down and keeping transactions as small as possible as well as on the requested resource / action only. If it proves to be problematic I'll look into in-lining this syncing to the request.

An aside, once this gains adoption and we iron out the bugs. I want to move the contents model back to linked resource itself as well and remove all the models & language cruft in the controllers. Once it's all single resources, in-lining should be easier as well.

PR #2636, opened 2018-01-25
Better subject classification counts used for reporting & retirement.

### LINKED ISSUE #1956: Counters don't take into account already-seens?
I think https://github.com/zooniverse/Panoptes/blob/master/app/counters/subject_workflow_counter.rb is not taking subjects seen before by a user into account. Which is probably my fault since I missed it initially in this one: https://github.com/zooniverse/Panoptes/blob/master/app/workers/reset_project_counters_worker.rb

Solving this will probably take two things:
- Different counters for "total classifications" and "unique classifications". Otherwise reclassifications won't bubble up the counters to the project homepage.
- I couldn't get SQL to do this uniquing, might be easiest to just add a `counts_towards_retirement` column on classifications and fill that with the lifecycle. Then we can add a simple where clause on it in the counter. We can then also just trigger the count worker from the lifecycle, regardless of whether it's a create or an update, since the worker can look at that field to know if an increment is needed.

### LINKED ISSUE #2394: "gold mode" classifications shouldn't count toward retirement
When research teams want to validate volunteer answers (say, to assess accuracy for publications), they need to create a "gold standard" dataset that they can compare volunteer answers to. 

When we compare the volunteer answers to the gold standard dataset, we have to strip out the gold standard data from the classification data, because otherwise you're counting your gold standard data as a volunteer classification and inflating your measure of accuracy. (If you compare the project owner's answer to itself and say you have 100% accuracy, that's not really fair...)

Note that "experts" classifying on the site is different than creating a "gold mode" classification. "gold mode" classifications are created deliberately; people with "expert" roles on the project have to deliberately select the gear icon and select "gold mode" prior to submitting every single classification. So this isn't just "experts" classifying on the site - this is deliberately creating a separate dataset that you can compare volunteer answers to.

However, gold mode classifications currently count towards retirement. This is especially problematic when you're trying to assess accuracy for projects with low retirement rules. 

**Example** 
For example, I'm assessing accuracy on Snapshots at Sea, which circulates images to 5 people; when at least 4 out of 5 people say "YES" the aggregate answer is "YES." The project owner created the gold standard dataset to compare the aggregate answers to using "gold mode." However, because this contributed to retirement, I've now only got 4 volunteer answers for these photos. Unfortunately, many of them have 3/4 people saying "YES" so that last volunteer answer is actually really important for producing the aggregated answer. Including Ted's gold mode classification - the "known answer" in the volunteer dataset would inflate our measure of accuracy.

**Question** 
Should gold mode classifications count towards retirement? How hard would it be to make them not count toward retirement? We could tell project owners to up their retirement limits if they want to create gold standard data, but I don't think that's an elegant solution.

### LINKED ISSUE #2471: Subject workflow status counter counts incomplete classifications
Linked to #2394 #1956

Currently doesn't filter incomplete classifications as it should. Check the addition of this where clause doesn't impact on the performance (this col does have an index).

camallen:
Fixes #2471, closed #1956 and resolves #2394

Ignore incomplete, gold_standard and user has seen before classifications in counts, specifically for retirement.

Making a note that  https://github.com/zooniverse/Panoptes/blob/fbb0d47e9d4a6b3e5d603d734b27bb66c69c60df/app/models/classification.rb#L28

Is currently a seq scan...it's only used with other more specific scopes
so it's a filter when walking the index but this is still row lookup (not pure index query), note: the below is a warm cache query.
>"Nested Loop  (cost=1.14..573.03 rows=2 width=4) (actual time=0.085..0.431 rows=10 loops=1)"
"  Output: classifications.id"
"  ->  Index Scan using index_classification_subjects_on_subject_id on public.classification_subjects cs  (cost=0.57..74.08 rows=58 width=4) (actual time=0.050..0.148 rows=10 loops=1)"
"        Output: cs.classification_id, cs.subject_id"
"        Index Cond: (cs.subject_id = 17083694)"
"  ->  Index Scan using classifications_pkey on public.classifications  (cost=0.57..8.59 rows=1 width=4) (actual time=0.027..0.027 rows=1 loops=10)"
"        Output: classifications.id"
"        Index Cond: (classifications.id = cs.classification_id)"
"        Filter: (classifications.completed AND (classifications.gold_standard IS NOT TRUE) AND (NOT (classifications.metadata ? 'seen_before'::text)) AND (classifications.workflow_id = 1737))"
"Planning time: 0.296 ms"
"Execution time: 0.455 ms"

Noting that the impact of this filter lookup on the counter queries is negligible in query analysis. ~Should we just bite the bullet and flip field from a sparse index to a full btree?~ Looking at the join query i don't think a non-sparse index would be used in this case anyway as the join finds the linked classification ids and walks the index applying the filter on row lookup. We'd have to have a compound id, X index to use this on the join we have. I'm not bothering about this as testing on queries look ok. We can revisit if the performance is impacted (also happy to add an feature flag experiment here if you want).

# TODO
- [x] Don't count already seens
- [x] Figure out what to do about gold standard
- [x] Decide what todo with completed sparse index

# Review checklist

### DIFF
--- a/app/counters/subject_workflow_counter.rb
+++ b/app/counters/subject_workflow_counter.rb
@@ -12,6 +12,7 @@ class SubjectWorkflowCounter
       .joins("INNER JOIN classification_subjects cs ON cs.classification_id = classifications.id")
       .where("cs.subject_id = ?", swc.subject_id)
       .where("gold_standard IS NOT TRUE")
+      .where.not("metadata ? 'seen_before'")
       .complete
 
     if launch_date = swc.project.launch_date
--- a/app/counters/subject_workflow_counter.rb
+++ b/app/counters/subject_workflow_counter.rb
@@ -11,6 +11,7 @@ class SubjectWorkflowCounter
       .where(workflow: swc.workflow_id)
       .joins("INNER JOIN classification_subjects cs ON cs.classification_id = classifications.id")
       .where("cs.subject_id = ?", swc.subject_id)
+      .complete
     if launch_date = swc.project.launch_date
       scope = scope.where("classifications.created_at >= ?", launch_date)
     end
--- a/app/counters/subject_workflow_counter.rb
+++ b/app/counters/subject_workflow_counter.rb
@@ -11,7 +11,9 @@ class SubjectWorkflowCounter
       .where(workflow: swc.workflow_id)
       .joins("INNER JOIN classification_subjects cs ON cs.classification_id = classifications.id")
       .where("cs.subject_id = ?", swc.subject_id)
+      .where("gold_standard IS NOT TRUE")
       .complete
+
     if launch_date = swc.project.launch_date
       scope = scope.where("classifications.created_at >= ?", launch_date)
     end
--- a/app/counters/subject_workflow_counter.rb
+++ b/app/counters/subject_workflow_counter.rb
@@ -7,14 +7,39 @@ class SubjectWorkflowCounter
   end
 
   def classifications
-    scope = Classification
-      .where(workflow: swc.workflow_id)
-      .joins("INNER JOIN classification_subjects cs ON cs.classification_id = classifications.id")
-      .where("cs.subject_id = ?", swc.subject_id)
-      .where("gold_standard IS NOT TRUE")
-      .where.not("metadata ? 'seen_before'")
-      .complete
+    experiment_name = "subject_workflow_counter_skip_gs_and_seen_before"
+    CodeExperiment.run(experiment_name) do |e|
+      e.run_if { Panoptes.flipper[experiment_name].enabled? }
 
+      e.use do
+        scope = Classification
+          .where(workflow: swc.workflow_id)
+          .joins("INNER JOIN classification_subjects cs ON cs.classification_id = classifications.id")
+          .where("cs.subject_id = ?", swc.subject_id)
+
+        count_classifications(scope)
+      end
+
+      e.try do
+        scope = Classification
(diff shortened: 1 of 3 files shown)


app/counters/subject_workflow_counter.rb
```diff
@@ -11,6 +11,10 @@ def classifications
       .where(workflow: swc.workflow_id)
       .joins("INNER JOIN classification_subjects cs ON cs.classification_id = classifications.id")
       .where("cs.subject_id = ?", swc.subject_id)
+      .where("gold_standard IS NOT TRUE")
+      .where.not("metadata ? 'seen_before'")
```
marten: Does this need an index or can PG reduce the scope plenty with the other ones so that a scan isn't an issue?

camallen: see the PR comment for details on the query plan, noting it's not a scan but an index with filter. For ref, there is an index on workflow_id but the query planner chooses not to use it due to the cost of combining indexes and because join (index query) returns a list classification ids that match the id index. 

It is normally cheaper to walk the selective index (join in this case) looking up rows and apply's filters to it than merging index scans, especially if one is less selective. I believe we would need a compound id, workflow_id, gs, metadata index here to support this particular query (though i'm not sure the metadata one would work, might need promotion to a col). If we had that index it would be a straight index query and defs, faster. That said in basic testing it wasn't an issue and i'm happy to do some better test query profiling on this if you want.

camallen:
I'm going to add an experiment for this to see how it impacts on the db.

PR #2643, opened 2018-01-30
Update records before triggering sidekiq job

zwolf:
When updating a classification to be complete, there is a race condition where the classification lifecycler sidekiq job is kiqed off before the update transaction completes, resulting in the job being run against the old data and not properly creating a Recent and a UserSeenSubject.

Subsequently incoming will be a new rake task to create recents for the classifications that slipped through the cracks.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -19,7 +19,8 @@ class Api::V1::ClassificationsController < Api::ApiController
   end
 
   def update
-    super { |classification| lifecycle(:update, classification) }
+    super
+    controlled_resources.each { |resource| lifecycle(:update, resource) }
   end
 
   def gold_standard


marten:
Hmm, so `JsonApiController::CreatableResource` commits and then yields if a block is given:
https://github.com/zooniverse/Panoptes/blob/master/lib/json_api_controller/creatable_resource.rb#L6-L17

I'm wondering if `UpdatableResource` shouldn't do the same, so that this kind of bug could never have happened in the first place: https://github.com/zooniverse/Panoptes/blob/master/lib/json_api_controller/updatable_resource.rb#L18-L20

For instance, I'm guessing that https://github.com/zooniverse/Panoptes/pull/2629/files#diff-f1cbc4d7220aa39ef69b11a186de46c6R13 suffers from the same type of bug if we'd merge that?

marten:
Actually, bad example. But this I think might explain occasional counter weirdness too? https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/subject_sets_controller.rb#L20

camallen:
Yep good catch about race condition between the update action transaction and the background workers and that defs explain some borked counters. Great stuff to ID this. 

The update action yields to allow custom update behaviour and we depend on that in other places, e.g. https://github.com/zooniverse/Panoptes/blob/de86fd4cb199358a5c3389d3e8c6f8772b5757e4/app/controllers/api/v1/projects_controller.rb#L58

Your example linked in the operations will still suffer from this issue as it's wrapped in a transaction in the controller. I'll have to update them to ensure they are called after the transaction is committed. 

I'm all for avoiding this kind of issue in the first place and it's a subtle bug but this shouldn't block this PR fix for the ASM issue. I think this was us mostly just misunderstanding how the yield in the transactions and the background workers interplayed with respect to transaction isolation in the db.

marten:
No I didn't mean to block this PR :) 👍

PR #2644, opened 2018-01-30
Unsubscribe all project email preferences when user unsets project_email_communication

camallen:
Linked to https://github.com/zooniverse/Panoptes-Front-End/issues/4220#issuecomment-361304385 and https://github.com/zooniverse/Panoptes-Front-End/pull/4229

When a user sets `project_email_communication` to false it should unsubscribe all their user project preferences.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -41,13 +41,20 @@ class Api::V1::UsersController < Api::ApiController
   end
 
   def update
-    super do |user|
+    [].tap do |update_email_user_ids|
 
-      unless user.project_email_communication
-        unsubscribe_all_project_emails(user)
+      super do |user|
+        unless user.project_email_communication
+          unsubscribe_all_project_emails(user)
+        end
+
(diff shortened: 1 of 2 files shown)


app/models/gold_standard_annotation.rb
```diff
@@ -1,3 +1,14 @@
+# This class contains the historical gold standard classification annotations.
```
camallen: this must have come from a stray commit on my master...will clean up

simoneduca:
I realise I've misunderstood what this PR is supposed to do. @camallen mind refreshing my senile mind please?

camallen:
Setting the `project_email_communication` field on a user to `false` will unsubscribe all their project prefs via the API since this was introduced. 

I can't remember our exact conversation in meat space but from memory we decided to add a button to  unsubscribe from all (new api end point?) and allow folks to manage the individual prefs via normal form controls as well. Should we jump on call with grant to hash this out again?

simoneduca:
Sounds good to me!

PR #2646, opened 2018-01-31
Automate pausing / unpausing projects when finishing / linking new data

### LINKED ISSUE #2365: pause a project when active workflows are finished
We should automatically pause a project when all the active workflows are finished. Like wise we should unpause it when new data is linked / new workflows are added.

camallen:
Closes #2365 - Pause a project when it's completed all the actively linked data for all active workflows. Unpause it when new data is linked and reset the completness metrics  and counts when linking / unlinking new data.

# TODO: 
- [x] Unpause a project when linking new data, via the completeness worker
- [x] Investigate better unique locking constraints in sidekiq workers
- [x] Look at chaining workers or spacing temporally (simplest) to avoid race conditions between counters

# Review checklist

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -30,14 +30,14 @@ class Api::V1::WorkflowsController < Api::ApiController
 
   def update_links
     super do |workflow|
-      UnfinishWorkflowWorker.perform_async(workflow.id)
-      WorkflowRetiredCountWorker.perform_async(workflow.id)
       post_link_actions(workflow)
     end
   end
 
   def destroy_links
-    super { |workflow| post_link_actions(workflow) }
+    super do |workflow|
+      post_link_actions(workflow)
+    end
   end
 
   def retire_subjects
@@ -82,17 +82,13 @@ class Api::V1::WorkflowsController < Api::ApiController
   # be easily located in specific controller actions
   def post_link_actions(workflow)
     if workflow.set_member_subjects.exists?
+      RefreshWorkflowStatusWorker.perform_async(workflow.id)
       case relation
-      when :retired_subjects, 'retired_subjects'
+      when /\Aretired_subjects\z/
         params[:retired_subjects].each do |subject_id|
           NotifySubjectSelectorOfRetirementWorker.perform_async(subject_id, workflow.id)
         end
-      when :subject_sets, 'subject_sets'
-        # these calls calculate the completeness worker
-        # are in a race with the other counter in update_links
-        # WorkflowRetiredCountWorker, simple effort, perahps just delay it for a bit?
-        # how long does a retired counter run for?
-        CalculateProjectCompletenessWorker.perform_async(workflow.project_id)
+      when /\Asubject_sets\z/
         NotifySubjectSelectorOfChangeWorker.perform_async(workflow.id)
       end
     end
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -82,13 +82,13 @@ def field_context
   # be easily located in specific controller actions
   def post_link_actions(workflow)
     if workflow.set_member_subjects.exists?
+      RefreshWorkflowStatusWorker.perform_async(workflow.id)
       case relation
-      when :retired_subjects, 'retired_subjects'
+      when /\Aretired_subjects\z/
```
marten: Not sure how I feel about this. Wouldn't comparing with a regex be slower? What do you think about changing the `case` to be `case relation.to_s`

camallen: Probably though benchmarks show it is negligible. I assumed reading the one regex is a little easier but the string comparison makes sense too.
```
       user     system      total        real
regex:   0.000000   0.000000   0.000000 (  0.000034)
string:  0.000000   0.000000   0.000000 (  0.000015)
       user     system      total        real
regex:   0.000000   0.000000   0.000000 (  0.000014)
string:  0.000000   0.000000   0.000000 (  0.000015)
       user     system      total        real
regex:   0.000000   0.000000   0.000000 (  0.000011)
string:  0.000000   0.000000   0.000000 (  0.000015)
.       user     system      total        real
regex:   0.000000   0.000000   0.000000 (  0.000025)
string:  0.000000   0.000000   0.000000 (  0.000009)
       user     system      total        real
regex:   0.000000   0.000000   0.000000 (  0.000013)
string:  0.000000   0.000000   0.000000 (  0.000009)
       user     system      total        real
regex:   0.000000   0.000000   0.000000 (  0.000011)
string:  0.000000   0.000000   0.000000 (  0.000011)
```

camallen: here was the code
``` ruby
3.times do
  n = 10000
  Benchmark.bm do |x|
    x.report("regex: ") do
      case relation
      when /\Asubject_sets\z/ then true
      end
    end
    x.report("string:") do
      case relation.to_s
      when "subject_sets" then true
      end
    end
  end
end
```

camallen:
This is ready to go but hasn't been approved, any takers to have a look?

zwolf:
Does `workflow.set_member_subjects.exists?` give you enough information to un-finish a workflow? If I'm reading this correctly, then since calling the refresh worker is outside the case it'll un-finish a workflow regardless of what gets newly linked by the request.

camallen:
>Does workflow.set_member_subjects.exists? give you enough information to un-finish a workflow? 

Yes, knowing that there is data linked is important however to speed up and simplify the logic here I've chosen to just call 1 worker. 

>If I'm reading this correctly, then since calling the refresh worker is outside the case it'll un-finish a workflow regardless of what gets newly linked by the request.

Yeah - the idea is to simply the controller logic and rely on 1 worker. The refresh SWS worker will run and unfinish the workflow (if finished). Then it'll recount the retired subjects and finish the workflow if needed. Finally, it'll update the project & workflow completeness metrics so the home page stats update as well.

As both retired and new subject links will impact these numbers we can rely on the order of unfinished, recount and update finish and then update metrics so this seemed a viable solution.

PR #2647, opened 2018-02-01
More robust classification lifecycling reflecting on the classification complete status

camallen:
Classification lifecycling currently relies on the background worker being available (feature flagged but on by default). When redis drops / network error, etc the classification doesn't get scheduled to lifecycle but is saved in the db. We have a job to pick up non `lifecycled_at` classifications on a [schedule](https://github.com/zooniverse/Panoptes/blob/517e397c7aff4cd563ad06273d9b0fbac9a4eb99/app/workers/requeue_classifications_worker.rb#L7). All good you say, however we are currently adding `lifecycled_at` on create actions for incomplete classifications. Thus, any update classification (`complete: true`) that fails to schedule via redis can be left in limbo without creating recents, seens, counts, etc (all the good stuff users want us to do).

This PR modifies the state changes in the lifecycle reflecting on the `complete` flag on the classification. It now will only create a new `UserProjectPreference` on create when incomplete and rely on the next lifecycle where it's update and complete set to true to run the various actions.

I also attempted to simplify the boolean logic in this class, it could probably do with a refactor and I attempted this to split between complete / incomplete lifecycle strategies but it only saved the compete boolean test in the complete strategy so i aborted (it felt like too many changes for not much gain). 

# Review checklist

### DIFF
--- a/lib/classification_lifecycle.rb
+++ b/lib/classification_lifecycle.rb
@@ -58,18 +58,21 @@ class ClassificationLifecycle
   end
 
   def process_project_preference
-    return unless should_create_project_preference?
-    UserProjectPreferences::FindOrCreate.run! user: user, project: project
+    unless classification.anonymous?
+      UserProjectPreferences::FindOrCreate.run! user: user, project: project
+    end
   end
 
   def create_recent
-    return unless should_update_seen? && subjects_are_unseen_by_user?
+    return if classification.anonymous?
+    return if subjects_are_seen_by_user?
 
     Recent.create_from_classification(classification)
   end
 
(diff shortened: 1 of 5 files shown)


lib/classification_lifecycle.rb
```diff
@@ -23,7 +23,7 @@ def initialize(classification, action)
   end
 
   def execute
-    return if create? && classification.lifecycled_at.present?
+    return if action == "create" && classification.lifecycled_at.present?
```
camallen: do we even need the action passed in now? Should we just rely on the state inspection instead of short circuiting?

PR #2648, opened 2018-02-02
remove code experiment on subject workflow counters

camallen:
switch to the proposed filtering on classification index scopes to avoid counting duplicate, incomplete and gold standard data.

Note: linked to code experiment in #2636 - If the code experiment proves to have little impact then this filtering technique can proceed as planned. Failing that we will have to look at different index access patterns to speed up this count. 

# Review checklist

### DIFF
--- a/app/counters/subject_workflow_counter.rb
+++ b/app/counters/subject_workflow_counter.rb
@@ -7,42 +7,18 @@ class SubjectWorkflowCounter
   end
 
   def classifications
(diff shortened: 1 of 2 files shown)


camallen:
I'm happy with the code experiment test run on prod workloads - the following link was running when the db was under some pretty high cpu load too. 
 https://metrics.librato.com/s/metrics/subject_workflow_counter_skip_gs_and_seen_before?duration=6699&end_time=1517957708

PR #2661, opened 2018-02-07
count things less and space them out more

camallen:
do not allow duplicate count jobs to start, drop any new ones until the current one has finished and then fallback to congestion control.

# Review checklist

### DIFF
--- a/app/workers/workflow_retired_count_worker.rb
+++ b/app/workers/workflow_retired_count_worker.rb
@@ -11,7 +11,7 @@ class WorkflowRetiredCountWorker
     key: ->(workflow_id) { "workflow_#{workflow_id}_retired_count_worker" }
   }
 
-  sidekiq_options unique: :until_executing
+  sidekiq_options unique: :until_executed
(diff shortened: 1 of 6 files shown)


app/workers/workflow_retired_count_worker.rb
```diff
@@ -11,7 +11,7 @@ class WorkflowRetiredCountWorker
     key: ->(workflow_id) { "workflow_#{workflow_id}_retired_count_worker" }
   }
 
-  sidekiq_options unique: :until_executing
+  sidekiq_options unique: :until_executed
```
camallen: this could also be handled by modifying the congestion interval but we may not know the length of the job

marten:
If we know now that it's a slow expensive query, we should probably also bump the congestion interval.

I had some concerns about missing classifications when using `until_executed` in the count depending on timing issues, but upon further reflection I think those don't apply, that was only a concern before we switched from `cancel` to `reschedule`.

PR #2662, opened 2018-02-07
Process only undeleted workflows in cls exports

zwolf:
Classification dump workers should only export classifications from workflows that have not been deleted (via the Activatable module). That is, only workflows with `activated_state: "active"` should have their classifications included.

rationale: some deleted workflows have malformed tasks that occasionally cause exports to fail due to  an un-tracked-down PFE bug. This should be happening anyway, and will clean/speed up exports for researchers.

This should not be confused with `active`, which is a flag researchers can set. Those classifications are correctly included already.

Describe your change here.

# Review checklist

### DIFF
--- a/app/models/csv_dumps/classification_scope.rb
+++ b/app/models/csv_dumps/classification_scope.rb
@@ -34,7 +34,7 @@ module CsvDumps
     def completed_resource_classifications
       resource.classifications
         .complete
-        .joins(:workflow)
+        .joins(:workflow).where(workflows: {activated_state: "active"})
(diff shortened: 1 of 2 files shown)


zwolf:
The reason for the broken exports is that PFE bug that borks the task. This doesn't fix that bug or that borked classification, but it seems that classification is on a workflow they deleted anyway. In that sense it's a workaround, but we shouldn't include these anyway and best case scenario we don't see that pop up anymore (and exports are minutely faster!).

PR #2667, opened 2018-02-09
Revert "add project metadata config option to restrict self service d…

camallen:
**DO NOT MERGE** 

this Reverts zooniverse/Panoptes#2612 and removes the "panoptes only data" work to keep specific project data in the db. Once this project is finished we can merge this to clean up the codebase and if needed revisit this idea later.


camallen:
closing for now but leaving here if we need it.

PR #2673, opened 2018-02-13
Allow admins to look up users by email

astopy:
When handling spam complaints we only have the user's email address, so I need to be able to use that to look up the user. This looks for an `email` param and splits it on commas in the same way as `login`.

# Review checklist


app/controllers/api/v1/users_controller.rb
```diff
@@ -60,7 +60,9 @@ def update
   end
 
   def index
-    if logins = params.delete(:login).try(:split, ',').try(:map, &:downcase)
+    if api_user.is_admin? && emails = params.delete(:email).try(:split, ',').try(:map, &:downcase)
```
camallen: perhaps refactor the params extraction, split, downcase map to a reusable method here?

PR #2675, opened 2018-02-14
do not modify any user project email prefs

camallen:
reverts the unsubscribe behaviour added in #2644. 

Use the project_email_communication field as a state to copy to new project prefs when users first classify. The UI will now offer a button to unset all project email prefs and allow the user to modify the state this way.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -44,10 +44,6 @@ class Api::V1::UsersController < Api::ApiController
     [].tap do |update_email_user_ids|
 
       super do |user|
-        unless user.project_email_communication
-          unsubscribe_all_project_emails(user)
-        end
-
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -44,10 +44,6 @@ def update
     [].tap do |update_email_user_ids|
```
zwolf: Is this tap to support a multi-user updates? FYI, that yield-before-save bug in updatable_resource is still outstanding, I think. I'll get a one-line PR done up now.

I realize this was in the other PR that this is sort-of reverting, but I wanted to be clear.

camallen: Yeah the tap is to collect all the updated user ids to use outside the transaction block.

PR #2676, opened 2018-02-14
[WIP] Updatable resources should yield before they save

zwolf:
Race conditions can occur if the yield is to a sidekiq worker or something that pulls from the database before the following save! can commit its transaction. In lieu of a callback, this will at least wait until the save! call finishes (successfully or not) before the yield is attempted.

# Review checklist


zwolf:
Mostly just wanted to see what would happen, tbh. I can either close this or label it WIP if it's still enough of an issue that you'd like to revisit.

camallen:
Close it if you aren't going to get back to it soon, you can always re-open it and push on. An audit of where we are using the yield before save is key, once we find all these we can figure out what we need to do.

PR #2680, opened 2018-02-20
move workers outside of destroy transactions

camallen:
add transactions to destroy actions and move the worker calls outside the transaction to ensure consistent state when they run and avoid race conditions.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subject_sets_controller.rb
+++ b/app/controllers/api/v1/subject_sets_controller.rb
@@ -32,14 +32,14 @@ class Api::V1::SubjectSetsController < Api::ApiController
   # avoid calling destroy_all on each controlled_resource
   # to optimize sets and linked relation cleanup
   def destroy
-    subject_ids = []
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -29,22 +29,28 @@ def update_links
...
-      super
-
-      subject_ids.each_with_index do |subject_id, index|
-        SubjectRemovalWorker.perform_in(index.seconds, subject_id)
+      controlled_resources.each do |subject_set|
+        smses = subject_set.set_member_subjects
+        subject_ids.concat(smses.map(&:subject_id))
+        remove_linked_set_member_subjects(smses)
+        affected_workflow_ids.concat(controlled_resource.workflows.pluck(:id))
+        controlled_resource.subject_sets_workflows.delete_all
```
marten: Should this be optimized as well?

PR #2683, opened 2018-02-21
add bundle gem cache volume for panoptes docker dev env

camallen:
avoid the cost of installing gems and rebuilding docker images, instead rely on a persisted cached volume. This should speed up dev builds. I've also updated and tested the readme instructions to get started.

# Review checklist

### DIFF
--- a/scripts/docker/start.sh
+++ b/scripts/docker/start.sh
@@ -13,7 +13,8 @@ mkdir -p tmp/pids/
 rm -f tmp/pids/*.pid
 
 if [ "$RAILS_ENV" == "development" ]; then
-  exec foreman start
+  bundle check || bundle install --binstubs="$BUNDLE_BIN"
+  exec bundle exec foreman start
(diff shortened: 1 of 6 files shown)


scripts/docker/start.sh
```diff
@@ -13,7 +13,8 @@ mkdir -p tmp/pids/
 rm -f tmp/pids/*.pid
 
 if [ "$RAILS_ENV" == "development" ]; then
-  exec foreman start
+  bundle check || bundle install --binstubs="$BUNDLE_BIN"
```
marten: Should this also be done for running the test suite?

camallen: for sure - nice catch :)

camallen: right - so the test suite won't run this entry point. The gems will have to be installed manually via a docker run command. i've updated the readme's accordingly.

PR #2686, opened 2018-02-27
extract by set cte query from SubjectWorkflowStatus

camallen:
this is heavily optimised and only used in one place, moving closer to the place of use as it requires query execution as it no longer returns AR Relations in the current form, so is not chainable.

# Review checklist

### DIFF
--- a/app/counters/workflow_counter.rb
+++ b/app/counters/workflow_counter.rb
@@ -7,21 +7,21 @@ class WorkflowCounter
   end
 
   def classifications
-    query = SubjectWorkflowStatus
-              .by_set(workflow.subject_sets.pluck(:id),
-                      "SUM(classifications_count)")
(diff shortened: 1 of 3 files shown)


app/counters/workflow_counter.rb
```diff
@@ -30,4 +30,56 @@ def retired_subjects
...
+  # WHERE subject_workflow_counts.subject_id
+  # IN (SELECT subject_id FROM sws_by_set)
+  def sws_query(select_clause)
+    cte = sws_by_set_cte(workflow.subject_sets.pluck(:id))
+    select_manager = sws_by_set_select
+    query = SubjectWorkflowStatus.where(select_manager)
+
+    query = yield(query)
+
+    query = query.select(select_clause)
```
marten: Wondering if it would be better to delegate this to the block as well?

camallen: Good point, i should have done this as well.

PR #2688, opened 2018-03-03
api controller refactor

camallen:
first attempt at trying to make the controllers more explicitly defined by reducing module use and providing hooks to customize each api authorization scheme.

# Review checklist


zwolf:
Gonna be going through this for a few, but the combination of this and the other WIP controller refactor make for some gnarly PRs. Was this just in response to the issues Adam offhandedly brought up recently? What is the impetus for the sudden large refactors? Do we have a project roadmap for panoptes?

This and #2697 seem like ones we might want to meet and go over.

camallen:
>What is the impetus for the sudden large refactors? 

A long flight and jet lag ;) but more seriously folks pointing out the high cognitive load to understand the code paths and how things like authorization scopes work in the controllers. 

We initially choose modules and inheritance as the abstraction for wiring up the API controllers. This was to provide DSLs and avoid a lot of boiler plate in the controllers when we started building panoptes. However this abstraction hid the implementation details from future developers and made it hard to understand. 

I've been in them from the start so the abstractions make enough sense to me via the DSLs and inherent knowledge i have. However, hearing folks thoughts from the "outside" shows the API controllers have a chunk of debt in them. Ideally they would be more aligned to the service objects like the organisation controllers / talk api controllers.  

I re-read the Sandi Metz article about the wrong abstraction and decided to unravel the use of modules to try and make it more obvious how the code works. This is a start on that unravelling process, I've added more in #2696 (it will need a rebase over this branch) to explicitly setup the authorization scopes using class composition instead of module inheritance. If folks think this is a useful start then we can look to progress it. 

>Do we have a project roadmap for panoptes?

Nothing formalized in stone but there are places that need work. Dumps, Controllers, resource authorization schemes (specifically polymorphic resources), serializers. That said the system works so we could address some of this via controller versioning instead of rewrites / refactors.

> This and #2697 seem like ones we might want to meet and go over.

Happy to go over this and #2696 when i'm in Chicago, if you like. I'm not sure that #2697 is linked to this as it's related to switching to the new translations API, perhaps ask questions on that PR vs this one.

All thoughts welcome on this from all folks. As i said this was me having some free time and thinking about unwinding some of the abstraction that was put in a while back.

camallen:
closing in favour of the work done in #2803

PR #2696, opened 2018-03-05
[WIP] explicit api controller authorization code

camallen:
linked and depends on to #2688

Move from the include module code injection to explicitly declaring the authorization code in each api controller. This will allow customizable api controller authorization schemes as well combined with the more explicit setup when trying to understand the code.

This is a start on migrating some controllers, if folks think this is a good way to go then i can push on and migrate the rest of the api controllers.

# Review checklist


camallen:
closing in favour of the work done in #2803

PR #2697, opened 2018-03-06
Remove model translation loading scopes

camallen:
Translations are now their own polymorphic resource, this is a start on removing the overloaded authorization scopes to load up translated content for certain models. 

# TODO
  - [ ] modify the `has_many content_association` model association to rely only the primary contents (this should probably come in another PR)

# Review checklist


camallen:
Any chance I can get a review on this?

zwolf:
LGTM, on the premise that the code that has already replaced this was properly specced and already reviewed

PR #2718, opened 2018-03-29
limit the max workflow page size to 25

camallen:
ensure large workflow responses don't tie up API resources but still allow the client to enumerate all the workflows via paging.

pinging @mcbouslog - as per our discussion in chicago, limiting this should be ok for all our sites right?

# Review checklist

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -33,6 +33,8 @@ class Workflow < ActiveRecord::Base
 
   enum subject_selection_strategy: %i{default cellect designator builtin}
 
+  max_paginates_per 25
+
   DEFAULT_RETIREMENT_OPTIONS = {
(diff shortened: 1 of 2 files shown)


mcbouslog:
- [x] searched PFE for `page_size` and confirmed limiting workflow response to 25 should be ok
- [x] searched [Zooniverse at GitHub organization level](https://github.com/search?q=org%3Azooniverse+page_size) for `page_size` and no issues apparent
- [x] offhand, I can't think of any custom front-ends or other sites that would be requesting more workflows than the API default response other than the Notes from Nature landing, which [paginates through requests based on response metadata](https://github.com/zooniverse/notes-from-nature-landing/blob/master/src/actions/workflows.js#L22)

So I think we're good to go.

PR #2728, opened 2018-04-03
Remove medium mime restrictions

camallen:
Allow users to upload any content/mime type media resources. The UI may not support these for say some subject locations, the failure mode will be in the UI. 

This still restricts the medium export types to be 'text/csv' only. 

# Review checklist


tingard:
This would enable prepping https://master.pfe-preview.zooniverse.org/projects/tingard/galaxy-builder for launch, and so would be fantastic if it could be merged by Thursday to help us meet our goal launch date.

PR #2729, opened 2018-04-04
Add user ux testing email communications field

camallen:
A new email list for ux testing, this will have to be wired up for email settings page in PFE too.

# Review checklist


camallen:
It won't i'd prefer that to come in another pull request to keep get this out and wired up so folks can subscribe to it.

astopy:
That's fine, as long as the plan is to implement that soon (I don't want us to have to manually export this a bunch of times if there are emails planned to go out soon).

PR #2736, opened 2018-04-08
move ux testing attribute backfill op to rake task

camallen:
backfilling the ux testing email comms attribute may consume all the read IOPS for the database (it did on staging db). If this is the case we should ensure we can stop (and restart) to avoid slowing down the API nodes.

# Review checklist

### DIFF
--- a/lib/tasks/user.rake
+++ b/lib/tasks/user.rake
@@ -13,7 +13,7 @@ namespace :user do
 
   desc "Backfill UX testing emails comms field in batches"
   task backfill_ux_testing_email_field: :environment do
-    User.find_in_batches do |users|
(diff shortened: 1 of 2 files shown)


lib/tasks/user.rake
```diff
@@ -11,6 +11,19 @@ namespace :user do
     end
   end
 
+  desc "Backfill UX testing emails comms field in batches"
+  task backfill_ux_testing_email_field: :environment do
+    User.find_in_batches do |users|
```
marten: Should this have a `.select(:id)` to lighten the amount of data needed?

camallen: Good shout - i'll look to add that in and get this out.

PR #2742, opened 2018-04-18
do not allow empty retirement counts

camallen:
add custom workflow retirement validatator and ensure the classification count option has an integer in it.

# Review checklist

### DIFF
--- a/app/models/concerns/retirement_validator.rb
+++ b/app/models/concerns/retirement_validator.rb
@@ -1,0 +1,49 @@
+class RetirementValidator
+  attr_reader :workflow
+
+  def initialize(workflow)
+    @workflow = workflow
+  end
+
+  def validate
+    return if empty_retirement_config?
+
+    if invalid_valid_criteria?
+      add_error(
+        :"retirement.criteria",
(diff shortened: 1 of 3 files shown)


marten:
Could this have been an actual [`ActiveModel::Validator`](http://guides.rubyonrails.org/active_record_validations.html#custom-validators). Or alternatively, since it really only cares about the `retirement` property, perhaps giving `retirement` an actual object, and then using `validates_associated` as used in Caesar:

* https://github.com/zooniverse/caesar/blob/master/app/models/reducer.rb#L42
* https://github.com/zooniverse/caesar/blob/master/app/models/reducer.rb#L96-L98
* https://github.com/zooniverse/caesar/blob/master/app/models/extract_filter.rb#L2
* https://github.com/zooniverse/caesar/blob/master/app/models/extract_filter.rb#L7-L10

camallen:
> Could this have been an actual ActiveModel::Validator

Yes i've used this pattern without the rails specific DSL - happy to rework that if it's important. 
I could wrapper the retirement config in an object and then use `validates_associated` but it seems like it's not really an association so the original pattern makes more sense.

PR #2751, opened 2018-04-25
[WIP] Allow a project to change it's primary language

### LINKED ISSUE #2749: Error when attempting to modify project primary_language
https://app.honeybadger.io/projects/40595/faults/37505651#notice-summary

```
NoMethodError: undefined method `update!' for nil:NilClass

projects_controller.rb  67 block in update(...)

[PROJECT_ROOT]/app/controllers/api/v1/projects_controller.rb:67:in `block in update'
65       content_attributes = primary_content_attributes(update_params)
66       unless content_attributes.blank?
67         resource.primary_content.update!(content_attributes)
68       end
```

I'm not even sure I'm supposed to be able to modify this attribute, but either way this is a bug (it should either modify it or return a validation error).

camallen:
fixes #2749 - If a project requests to update it's primary language code we should find any existing project contents model or build a new one, then update it's attributes and link it to the existing project resource as the primary_content model.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -59,9 +59,6 @@ class Api::V1::ProjectsController < Api::ApiController
 
   def update
     super do |resource|
-      # TODO: extract this primary project content update
-      # to a service object that sits in the project
-      # transaction.
       content_attributes = primary_content_attributes(update_params)
       unless content_attributes.blank?
         resource.primary_content.update!(content_attributes)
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/projects_controller.rb
```diff
@@ -58,13 +58,34 @@ def index
...
+    primary_language_update = update_params.delete(:primary_language)
     super do |resource|
-      # TODO: extract this primary project content update
-      # to a service object that sits in the project
-      # transaction.
       content_attributes = primary_content_attributes(update_params)
-      unless content_attributes.blank?
-        resource.primary_content.update!(content_attributes)
+
+      if primary_language_update
```
camallen: This needs an extraction to a service object or class that handles the project_contents update instead of this large conditional behaviour i have here.

camallen:
This work will have to change if the work in #2754 changes the contents association from `has_many` to `has_one`, I'll revisit after #2754

PR #2752, opened 2018-04-26
Include credited_name in all user requests

zwolf:
A user's credited name is collected for the purpose of displaying publicly and therefore should be included in public user requests. In the event that credited_name is nil OR includes an `@` (to protect against leaking an email address), the user's login is serialized as credited_name instead.

This will allow science teams (SGL, specifically) to query the API for user information and update their discoveries with appropriate names.

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -86,7 +86,7 @@ The User resource has the following attributes:
 - languages
 
 *id*, *zooniverse_id*, *created_at*, *updated_at*, and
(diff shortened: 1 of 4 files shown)


zwolf:
Before this can be merged: @trouille and @chrislintott, please confirm this behavior is correct and appropriate.

chrislintott:
Yes, this looks good.

PR #2753, opened 2018-04-26
Ensure subjects selected with Designator are returned in selection order

marten:
We were getting reports that training images were "clustering" rather than being interleaved. This was a side effect of getting the SetMemberSubjects without any explicit ordering.

# Review checklist

### DIFF
--- a/lib/subjects/designator_selector.rb
+++ b/lib/subjects/designator_selector.rb
@@ -35,6 +35,7 @@ module Subjects
 
       subject_ids = self.class.client.get_subjects(workflow.id, user.try(&:id), group_id, limit)
       return [] unless subject_ids.present?
+      raise "Hacking attempt" unless subject_ids.all? { |i| i.is_a? Integer }
 
       sms_scope = SetMemberSubject.by_subject_workflow(subject_ids, workflow.id)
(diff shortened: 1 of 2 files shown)


lib/subjects/designator_selector.rb
```diff
@@ -34,7 +34,12 @@ def get_subjects(user, group_id, limit)
       return [] unless enabled?
 
       subject_ids = self.class.client.get_subjects(workflow.id, user.try(&:id), group_id, limit)
+      return [] unless subject_ids.present?
+      raise "Hacking attempt" unless subject_ids.all? { |i| i.is_a? Integer }
+
       sms_scope = SetMemberSubject.by_subject_workflow(subject_ids, workflow.id)
+        .order("idx(array[#{subject_ids.join(',')}], set_member_subjects.subject_id)")
```
camallen: I've was thinking about this the other day, what do you say we can just return the subject ids from the selector as they? The selectors check the set linkage when loading data and keep things in sync under normal operations, why do we need this extra db query to check the set linkage? I think this is super vestigial and we can optimize the selection path by removing this extra sms scope selection. Thoughts?

marten: Yeah, I'm up for removing it. I'll look at it next week.

PR #2754, opened 2018-04-26
[WIP] Convert contents association to singular resource contents model

camallen:
related to #2751 - that PR will have to change if this makes it in.

Since we removed the load_with_languages #2697 we can now convert the (project|workflow|organization|)_contents models from has_many to has_one to allow preloading the contents model on the resource serializer scopes. Translations are now in their own resource and synced on change via the parent resource but still by the *_contents model strings so long term these *_content model strings can be folded back into the parent resource and extracted to translations model from there.

# Todo

- [x] Remove the `eager_load_projects` flipper feature once this is deployed - #2837

# Review checklist


marten:
I'm picking this back up, because publishing versions will be much simpler if we don't have the content models anymore.

camallen:
closing this in favour of contents replacement work going on in pr's like #2925

PR #2759, opened 2018-05-10
Generate structure.sql with Postgresql 9.6

marten:
It's becoming hard to actually make changes to the DB because every migration generates a huge structure.sql diff because it's so far out of date with how a modern `pg_dump` generates it. I've been manually picking chunks and committing only partial changes to work around it for the past year+, but now I hit a situation where the diff chunks weren't even possible to pick apart, and it would've meant that I'd have to manually edit structure.sql just so that I could commit it.

Instead, I'm suggesting we get with the times. This updates it to the current 9.6.8 format, without changing anything. It was the result of a blank database getting filled from `rake db:setup` and then generated by a `rake db:migrate`.

# Review checklist

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2,12 +2,18 @@
 -- PostgreSQL database dump
 --
 
+-- Dumped from database version 9.6.8
+-- Dumped by pg_dump version 9.6.8
+
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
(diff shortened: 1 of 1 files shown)


camallen:
I'm not sure we should just bump this without syncing our actual databases. Do you test/dev with 9.6 db or 9.4? Optimal solution for me is to upgrade the actual databases.

marten:
So I was devving on 9.4 as well, but for me it just generates a different format than what it was. Docker is no help because Rails generates structure.sql using your local copy of pgdump.

If you run `rake db:migrate` now, does the structure.sql file stay completely unchanged for you?

camallen:
closing this - feel free to reopen.

PR #2762, opened 2018-05-14
Allow users to edit and delete their account

### LINKED ISSUE #2410: Support sign in / register on smaller screen devices
Make the [form component](https://github.com/zooniverse/Panoptes/blob/master/app/assets/stylesheets/forms.css) respond to different screen sizes. 

Currently looks like this for simulated Nexus 5x (412 x 732)
screen shot 2017-07-21 at 14 28 28

marten:
I think that it might be easiest to keep features like these in Panoptes as HTML rather than APIs. That way we enforce consistency across frontends.

* Renamed "sign up" to "create account". The suble distinction between "sign up" and "sign in" confuses me every time I look at this screen.

image

image

image

Closes #2410 

# Review checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
--- a/app/assets/stylesheets/base.css
+++ b/app/assets/stylesheets/base.css
@@ -1,6 +1,6 @@
 html, body {
   background: #fbfbfb;
-  font: 400 10px/20px Arial, sans-serif;
+  font: 400 12px/20px Arial, sans-serif;
   -moz-osx-font-smoothing: grayscale;
   -webkit-font-smoothing: antialiased;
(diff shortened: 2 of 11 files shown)


marten:
@camallen Good catch. I've added a test for that, and also updated all of the other views such that #2410 will now be fixed with this as well. Might want to check it over again.

marten:
Oh and note that I'm renaming "sign up" to "create account" because it always confuses me (too close to "sign in")

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
```
camallen: does this file want to be in here?

marten: Yeah it's some kind of standardized way of telling editors what settings to use for common things. Mine was indenting erb with 4 spaces, and figured this might be a good way of both fixing it and documenting it, even if people don't run the plugin. http://editorconfig.org/

camallen: sneaky

PR #2765, opened 2018-05-15
upgrade structure.sql with secure schema paths

camallen:
Update the postgres structure doc from output of pgdump with a version post 9.4.17 (most of our dev envs). 

This change reflects the comments i made in https://github.com/zooniverse/Panoptes/pull/2759#issuecomment-389215000 and may supersede #2759 if we want to keep the code and database version in sync. 

# Review checklist


marten:
Okay, I've finally managed to get them running side by side. I get the same schema when I run `db:migrate` so let's merge this.

PR #2768, opened 2018-05-17
Refresh workflow worker deadlock issues

camallen:
Reduce the workflow deadlocks scope and count things less during retirement and set link events.  This will now rely on the retired count worker to call the project completeness worker to update the stats more regularly combined with more restrictive congestion rate limits on count workers to avoid concurrent workflow updates.

# Review checklist


camallen:
From this PR not really. Workflow updates are deadlocking, e.g. https://app.honeybadger.io/projects/40595/faults/37690675#notice-summary

The changes i introduced last month ran those counters and updates more frequently. The retirement events also run them more than needed and these changes should help with bulk retirement events.

PR #2773, opened 2018-05-18
Consolidate retirement worker behaviour

camallen:
Avoid 2 retirement workers (1 via the custom API & one from classification receipt) and consolidate on one for the retirement behaviour. 

The `RetireSubjectWorker` will still be called from the operation and will find or create the `subject workflow status` resource and pass this to the retirement worker. This allows us to consolidate behaviour on the `RetirementWorker` for post retirement behaviour (counting, notifications, publishing) in one worker.

# Review checklist

### DIFF
--- a/app/workers/retire_subject_worker.rb
+++ b/app/workers/retire_subject_worker.rb
@@ -1,30 +1,28 @@
 class RetireSubjectWorker
   include Sidekiq::Worker
+  attr_reader :workflow
 
   sidekiq_options queue: :high
 
   def perform(workflow_id, subject_ids, reason=nil)
-    workflow = Workflow.find(workflow_id)
-    retired_subjects = 0
-
-    Workflow.transaction do
-      subject_ids.each do |subject_id|
-        begin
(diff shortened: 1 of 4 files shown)


app/workers/retire_subject_worker.rb
```diff
@@ -1,30 +1,28 @@
...
-        begin
-          if workflow.retire_subject(subject_id, reason)
-            retired_subjects += 1
-          end
-        rescue ActiveRecord::RecordInvalid
-        end
-      end
+    Array.wrap(subject_ids).each do |subject_id|
+      count = subject_workflow_status(subject_id)
+      RetirementWorker.perform_async(count.id, reason)
```
marten: This worker has a default of "nil". I think this line will then pass in an explicit `nil` for `reason`, which will override the default value of `classification_count` defined in `RetirementWorker`?

camallen: this was the previous behaviour so i kept it (see the diff for the details). The operation validates a reason if it's passed in but does allow a nil one too.

PR #2777, opened 2018-05-21
Fix login submit

### LINKED ISSUE #2776: Cannot click "Sign In" on Sign In form
## UI Bug
Issue: users cannot click on the "Sign In" button (i.e. cannot use mouse input to trigger the form submission.)

screen shot 2018-05-21 at 12 02 05
_Clicking on Sign In does nothing. [Absolutely nothing! It's so stupid](https://youtu.be/KezvwARhBIc?t=1m25s)_

### Analysis
This appears to be a UI bug. The `nav.devise-shared-links` element actually appears _on top of the Sign In button,_ blocking any mouse/touch events.

screen shot 2018-05-21 at 12 15 53

Testing:
- Go to https://panoptes.zooniverse.org/users/sign_in
- Fill in your username and password.
- Using a mouse, click on the Sign In button
- Expected: user is taken to the sign in success/failed 
- Actual: nothing happens. Mouse click isn't registered.

Notes:
- The issue exists for touch input as well; users on touch devices cannot tap on the Sign In button.
- Users however CAN perform the Sign In operation by pressing the Enter key when

Tested on OSX+Chrome/FF/Safari, Android+Chrome.

### Solution
One way to prevent the `nav.devise-shared-links` element from blocking mouse/touch input is to add the CSS code, `pointer-events: none`.

NOTE: also make sure the devise-shared-links doesn't have a background. [Further information here.](https://stackoverflow.com/questions/3680429/click-through-a-div-to-underlying-elements)

NOTE: an alternative solution is set `.form` to `position: relative; z-index: 100;`, putting the form 'above' the blocking `nav.devise-shared-links`, but this may affect other forms, so I'm a bit iffy with it.

### Status
Fix incoming, but I'll need help testing it. Tagging @camallen @marten

marten:
Fixes #2776

image

# Review checklist

### DIFF
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -19,7 +19,9 @@
 
       <div class="submit-container"><%= f.submit "Sign in" %></div>
     <% end %>
+  </div>
 
+  <div class="panel-footer">
     <%= render "devise/shared/links" %>
   </div>
 </div>


shaunanoordin:
Nice! This is a really simple fix, and I actually like this new design. 👍

PR #2785, opened 2018-05-30
Avoid db fetch of the whole workflow in the retire operation

camallen:
When retiring subjects we can use the `params[:workflow_id]` instead of the workflow resource in the operation. 

The authorization scheme will still issue a check via `controlled_resources.exists?` which is a count query on the db to ensure we have access to retire subjects for a given workflow. As we only use the `workflow_id` attribute and don't need the whole workflow, we can avoid paying the cost to serialize and instantiate this data from the db. 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -42,7 +42,7 @@ class Api::V1::WorkflowsController < Api::ApiController
   end
 
   def retire_subjects
-    operation.with(workflow: controlled_resource).run!(params)
+    operation.run!(params)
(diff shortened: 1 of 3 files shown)


marten:
How hard would it be to get the controller action to limit to a `.select(:id)` for the auth check?

camallen:
I thought of that as well, not hard something like 
`controlled_resources.select(:id).first` instead of using `controlled_resource`. As it's a non-standard route that doesn't serialize any response object i thought this was a better way to avoid paying any more serialization / ORM costs.

marten:
Gotcha, 👍

PR #2786, opened 2018-05-31
avoid repeat loading of the workflow resource in RetireSubjectWorker

camallen:
Follow up to #2785 - we really only want to know if the workflow exists we don’t need the whole resource to do this in the RetireSubjectWorker. 

I am going to look at our workers and how they use the workflow resource as we could be serializing more data over the wire and from the db than we currently need / use. 

However littering the code with `Workflow.select(:id, :project_id, ..).find(id)` is not ideal and will probably mean we miss it in the future. It may be easier to just slice the task object to it's own model and `accept_nested_attributes` from the workflow to deal with this. That way we can safely use the workflow resource and ensure we pre/eager_load the tasks association for the serializers.

This may also be a time to think about how to better store these json task definitions and if they should be in the db at all, should we instead slice them out to a media resource in aws instead? Something to think about, in the meantime this PR is a start on avoiding lots of db fetches we don't use. 

# Review checklist

### DIFF
--- a/app/workers/retire_subject_worker.rb
+++ b/app/workers/retire_subject_worker.rb
@@ -1,28 +1,28 @@
 class RetireSubjectWorker
   include Sidekiq::Worker
-  attr_reader :workflow
+  attr_reader :workflow_id
 
   sidekiq_options queue: :high
 
   def perform(workflow_id, subject_ids, reason=nil)
-    begin
-      @workflow = Workflow.find(workflow_id)
-    rescue ActiveRecord::RecordNotFound
-      return nil
-    end
-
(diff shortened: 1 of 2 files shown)


camallen:
@marten any thoughts on the comments about slicing tasks to it's own model / remote storage location?

marten:
I can see slicing it out to a table. I assume you're thinking `WorkflowTask(workflow_id: int, key: string, config: jsonb)` and `Workflow has_many :workflow_tasks`?

marten:
Storing it in S3 is less preferable to me than getting the CDN to be able to cache the API response.

camallen:
The model would look like that yes, thanks for the comments.

PR #2792, opened 2018-06-05
ensure a finished project doesn't change state automatically

camallen:
Don't automatically change a finished project's state, allow this to be a place we archive projects manually from the paused state.

# Review checklist


zwolf:
The bug here is that finished projects were being paused on account of the worker here calculating them as complete, right? If I've got that right, :shipit:

PR #2796, opened 2018-06-08
simplify and optimize the worflow counter

camallen:
move from CTE to simple non AR join query that runs in ~20ms on sql benchmarks. 

This both optimizes the query significantly but is much simpler code to understand and maintain. I don't know why i didn't see this the first time vs using a CTE with AREL to speed up this original query. 

# Review checklist

### DIFF
--- a/app/counters/workflow_counter.rb
+++ b/app/counters/workflow_counter.rb
@@ -1,5 +1,4 @@
 class WorkflowCounter
-
   attr_reader :workflow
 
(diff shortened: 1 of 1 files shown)


zwolf:
This is a hell of an optimization! This looks rad on a first pass but I'd like to talk about what made the complexity necessary and how you got around it. If you wanna merge it ASAP I'll approve, but I'd like to go over how you came to this.

PR #2803, opened 2018-06-13
Refactor RoleControl

marten:
Turning RoleControl from a module to a class, somewhat resembling Pundit's scope objects.

Ultimate goal is to make it so that we can swap it out for Pundit on a per-controller basis.

# Review checklist


app/controllers/concerns/polymorphic_resource_scope.rb
```diff
@@ -81,10 +81,70 @@ def check_polymorphic_controller_resources
...
+    else
+      ''
+    end
   end
 
   def no_resources_error_message
     "Could not find #{polymorphic_klass_name} #{controller_name} with #{no_resources_message_ids}"
   end
+
+  def check_controller_resources
```
marten: I've basically copied over all the code from `RoleControl::Owned` to this file, and merged any method overrides (inlined `Owned` methods where this file called `super`). 

I did this because it was really difficult to see how the two interplayed, since it keeps `controlled_resources` but adds `polymorphic_*` methods alongside it. Except that some of the `controlled_*` methods *are* overridden, and it's a mess to figure out how those fit together (e.g. which methods from `Owned` are actually still needed? Do we actually call both versions sometimes?). Didn't want to think about this while also thinking about the overall refactor.

To keep PR size down, I'd like to leave it like this for now, and refactor the polymorphic stuff to policy objects in a future PR.

lib/json_api_controller.rb
```diff
@@ -56,6 +56,22 @@ def serializer
     @serializer ||= "#{ resource_name.camelize }Serializer".constantize
   end
 
+  def policy_object
```
marten: If you `include Pundit` you get a `policy` method, which is a better name, but I didn't want to confuse us with module inclusion order for now. I'll rename it at a later PR.

app/controllers/concerns/roles_controller.rb
```diff
@@ -25,6 +25,10 @@ def resource_class
     AccessControlList
   end
 
+  def policy_object
+    RoledControllerPolicy.new(api_user, resource_class, resource_name, action_name, params, scope_context: scope_context)
```
camallen: feels like scope_context could just be a param assignment vs a method call here.

app/controllers/api/api_controller.rb
```diff
@@ -5,15 +5,15 @@ module Api
 
   class ApiController < ApplicationController
     include JsonApiController
-    include RoleControl::RoledController
+    include JsonApiController::CheckResourcesExist
```
camallen: Is the plan to move this to each controller as i proposed in #2688 instead of being less visible here in the API controller?

app/controllers/api/v1/memberships_controller.rb
```diff
@@ -18,7 +18,7 @@ def create
 
   private
 
-  def add_active_resources_scope
-    false
+  def policy_object
+    RoledControllerPolicy.new(api_user, resource_class, resource_name, action_name, params, add_active_resources_scope: false)
```
camallen: did you want memoization here?

app/controllers/concerns/polymorphic_resource_scope.rb
```diff
@@ -47,10 +47,10 @@ def controlled_scope
     if action_name == "create"
       :update
     else
-      super
+      action_name.to_sym
     end
   end
-
+
```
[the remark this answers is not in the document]
camallen: agree

app/controllers/concerns/polymorphic_resource_scope.rb
camallen: Makes sense - ideally it is just a different policy object resolving to the polymorphic scopes.

app/controllers/concerns/roles_controller.rb
camallen: memoization?

app/policies/roled_controller_policy.rb
```diff
@@ -0,0 +1,54 @@
...
+                              add_active_scope: add_active_resources_scope)
+  end
+
+  def resource_ids
+    @resource_ids ||= array_id_params(_resource_ids)
+  end
+
+  private
+
+  def controlled_scope
```
camallen: minor - Should we rename this to `controlled_action` while you are at it?

app/policies/roled_controller_policy.rb
```diff
@@ -0,0 +1,54 @@
...
+    action_name.to_sym
+  end
+
+  def array_id_params(string_id_params)
```
camallen: possible extract to class target here, e.g. `ArrayParams.extract_ids(string_id_params)` and re-use in the polymorphic controllers too.

lib/json_api_controller/check_resources_exist.rb
```diff
@@ -0,0 +1,33 @@
+module JsonApiController
+  class AccessDenied < StandardError; end
+
+  module CheckResourcesExist
+    extend ActiveSupport::Concern
+
+    included do
+      before_action :check_controller_resources, except: :create
```
camallen: Long term i think these should be explicitly defined in each controller to make it obvious how the authorization schemes work. I started doing this in #2696 e.g. the ClassificationsController was a first crack at how this might look.

app/controllers/concerns/roles_controller.rb
marten: The next branch I've already got nearly wrapped up actually extracts the additional options, and inlined the scope_context. Also adds memoization :)

app/controllers/api/api_controller.rb
marten: Yes

app/controllers/api/v1/memberships_controller.rb
marten: Gets added in the `pundit-scopes` branch.

app/policies/roled_controller_policy.rb
marten: This method is refactored away in `pundit-scopes` follow-up branch.

app/controllers/concerns/roles_controller.rb
camallen: :tada:

app/policies/roled_controller_policy.rb
marten: This method is refactored away in `pundit-scopes` follow-up branch.

app/policies/roled_controller_policy.rb
marten: Aye. I've actually moved `resource_ids` out of the policy and into a new class in `pundit-scopes` followup-PR

lib/json_api_controller/check_resources_exist.rb
marten: Cool, I'll have a look

PR #2804, opened 2018-06-13
remove set_member_subject scope lookups on all subject selections

camallen:
Consolidate selectors to return only subject ids vs the extra lookup on `SetMemberSubject` as we trust the selector services correctly deal with the links between subjects and workflows. This avoids the extra query costs to convert subject_ids to sms scopes via the external selectors which were join queries and used on every selection request. 

# Review checklist

### DIFF
--- a/lib/subjects/designator_selector.rb
+++ b/lib/subjects/designator_selector.rb
@@ -33,10 +33,7 @@ module Subjects
     def get_subjects(user, group_id, limit)
       return [] unless enabled?
 
-      subject_ids = self.class.client.get_subjects(workflow.id, user.try(&:id), group_id, limit)
-      return [] unless subject_ids.present?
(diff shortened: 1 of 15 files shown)


lib/subjects/designator_selector.rb
```diff
@@ -36,11 +36,7 @@ def get_subjects(user, group_id, limit)
       subject_ids = self.class.client.get_subjects(workflow.id, user.try(&:id), group_id, limit)
       return [] unless subject_ids.present?
       raise "Hacking attempt" unless subject_ids.all? { |i| i.is_a? Integer }
```
marten: This `raise` line was in there because of the now removed raw SQL statement underneath it, to guard against a MITM-attack being able to then perform SQL injection. It can go away now.

PR #2805, opened 2018-06-13
Optimize subject selection paths

camallen:
Linked to #2804 - Avoid as much work as possible during subject selection including the custom selection context response. 

I've audited the queries to determine seens / fav, retired and left them intact as they are fast as they are.

I did find what i think is a better way of determining if the user has finished a workflow, vs counting the number of available linked subjects taking into account what the user has seen. Instead of counting we can use the selector responses to proxy this information. I.e. if designator returns data we can assume that the user has not finished the workflow, same for internal selector but if it reaches the any data fallback we assume they are finished. i have also taken into account when the external selector returns all retired/seen data then we assume that the user has finished the workflow until selector reloads data in time. 

I found an optimization to short circut the selection paths if the workflow is finished, we used to run the internal fallback and then any data, now it just runs the any data. 

Finally I refactored how the serializer constructs this selection context and update specs away from controllers. 

# Review checklist


lib/subjects/selector_context.rb
```diff
@@ -0,0 +1,77 @@
...
+      !!workflow.finished_at
+    end
+
+    def user_has_finished_workflow
+      return true if finished_workflow?
+
+      if user
+        case selector.selection_state
+        when :normal # selector service returned data
+          false
```
camallen: This could also reflect on the passed in subject_ids and compare against the seen / retired subject_ids to determine if the data is current from designator.

app/controllers/api/v1/subjects_controller.rb
```diff
@@ -20,13 +20,25 @@ def index
   end
 
   def queued
-    selector = Subjects::Selector.new(api_user.user, workflow, params)
+    subject_selector = Subjects::Selector.new(api_user.user, params)
+    selected_subject_ids = subject_selector.get_subject_ids
+
+    selected_subject_scope = Subject
```
marten: Wondering if the selector should maybe return a Selection object which would have `subject_ids`, `scope`, and `context` methods that the controller would just be able to call. What do you think?

lib/subjects/selector_context.rb
```diff
@@ -0,0 +1,79 @@
...
+    def finished_workflow?
+      !!workflow.finished_at
+    end
+
+    def user_has_finished_workflow
+      return true if finished_workflow?
+
+      if user
+        case selector.selection_state
+        when :normal # selector service returned data
```
marten: when would the normal selector return data if the user has finished the workflow? The builtin one shouldn't, right? And Designator doesn't. Is that just Cellect?

app/controllers/api/v1/subjects_controller.rb
marten: OK, happy to leave it at this

lib/subjects/selector_context.rb
camallen: I'd assume it wouldn't but it could be if the external selector service is out of sync. That branch should not run if the workflow is finished.

PR #2808, opened 2018-06-14
Pundit 1 of 3: Add Pundit with an interop layer to support RoleControl

marten:
This PR is showing a little bit of the end goal, which is to move from RoleControl to Pundit.

Any models that move over to the Pundit scopes will no longer need to have anything to do with "authorization" in the model itself (since that logic is all in the Scope). Moving authorization logic out of the model is a goal for this sequence of pull requests. With this, I mean the `can_by_role` stuff in the model. Models that move to Pundit include a `PunditInterop` module for the time being, but this is expected to go away once everything is migrated over.

On the controller side, I've moved logic into `LegacyPolicy` and `PunditPolicy` modules, which are explicitly to be included in each controller, rather than coming in implictly through JsonApiController.

`ApplicationPolicy::Scope` is built from parts of `RoleControl::Controlled` and `RoleControl::Actor`. After all, we still need to build up a query to find only accessible records, and that code was all fine. It was just a little too spread out. 

It still deals with a `scope_for` type of system. The Pundit way of doing this is to have a Scope for read actions (index, show), but for writing actions like `update`, just load the model directly and then have methods on the Policy like `update?` that return true if a user is allowed to update a specific record. For now I'm leaving the `scope_for` stuff in. I might revisit this when we have some more Policy/Scope objects.

I **really** like that the tests for `OrganizationPolicy` are now explicitly saying what records should be "in scope" for what actions. As we get more tests like this I might find some duplication that can be extracted, but I definitely want to keep the explicit "for this action and this user, expect these records" in the spec file itself. So it'll probably be more in terms of custom matchers than in terms of shared examples.

## How the Scopes work (as of this PR, future might change)

First of all, this is *only* for anything that had `include RoleControl::Controlled` and `can_by_role`. Parental controlled will come in a next PR. For each resource, typically we'll have a `ReadScope` class and a `WriteScope` class. 

* If you define a `public_scope` method on a scope, the resources returned by it will always be visible to anyone. For organizations, these are the listed ones. If there are no public resources, simply do not implement a `public_scope` method, or have it return `scope.none`.
* For the private resources, we will check if the user has an ACL onto that resource. The list of roles that is allowed is defined by `roles_for_private_scope`.
* Inside of the Policy itself, call `scope` to specify which (controller) actions should use which Scope class. This is so that you can see public resources, but not modify them. If these two are the same, it's fine to just have one Scope class, and call it Scope.
* There is no functional difference between not-logged-in and not having an ACL (aka not being an owner). 
* Admins always see everything.


spec/policies/organization_policy_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+  describe 'scopes' do
+    let(:anonymous_user) { nil }
+    let(:logged_in_user) { create(:user) }
+    let(:resource_owner) { create(:user) }
+
+    let(:listed_organization) { build(:organization, listed_at: Time.now, owner: resource_owner) }
+    let(:unlisted_organization) { build(:unlisted_organization, owner: resource_owner) }
+
+    let(:api_resource_name) { "organizations" }
+    let(:scopes) { %w(public organization) }
```
marten: This line and the line above can be deleted.

app/models/organization.rb
```diff
@@ -1,12 +1,12 @@
...
   include RoleControl::Owned
-  include RoleControl::Controlled
+  include RoleControl::PunditInterop
   include Activatable
   include Linkable
   include SluggedName
   include Translatable
 
-  scope :public_scope, -> { where(listed: true) }
+  # Still needed for HttpCacheable
```
camallen: I assume the `HttpCacheable` can move to a different kind of scope resolver / policy to determine what should be cacheable, not in this PR but wondering if that is the plan?

app/policies/application_policy.rb
```diff
@@ -0,0 +1,95 @@
...
+      @scopes_by_action[action] = with
+    end
+  end
+
+  def self.scopes_by_action
+    @scopes_by_action || {}
+  end
+
+  def scope_for(action)
+    scope = if record.is_a?(Class)
```
camallen: perhaps `scope_klass` instead of scope var name?

app/policies/application_policy.rb
```diff
@@ -0,0 +1,95 @@
+class ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  @scopes_by_action = {}
```
camallen: can't you just use the `scopes_by_action` class method to access this variable with a default instead of evaluating on load?

app/policies/application_policy.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+
+    def public_flag
+      return false unless respond_to?(:public_scope)
+      !public_scope.is_a?(ActiveRecord::NullRelation)
+    end
+
+    def user_can_access_scope(private_query, public_flag)
+      accessible = scope
```
camallen: does this mutate the `@scope` variable for the instance? I'm wondering about side effects of this method being called after setting up an initial scope.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,95 @@
...
+      user_group_memberships = user.memberships_for(action, model).select(:user_group_id)
+
+      AccessControlList
+        .where(user_group_id: user_group_memberships)
+        .where(resource_type: model.model_name.name)
+        .select(:resource_id)
+        .where("roles && ARRAY[?]::varchar[]", roles)
+    end
+
+    def model
```
camallen: perhaps `model_klass` name to indicate the intention a bit more.

lib/json_api_controller/legacy_policy.rb
```diff
@@ -0,0 +1,17 @@
+module JsonApiController
+  module LegacyPolicy
```
camallen: Not a fan of the `LegacyPolicy` name but i'm having a hard time coming up with another one, perhaps `RoledPolicy` to link the `RoledControllerPolicy` objects as well via name.

app/models/organization.rb
marten: Don't really have a plan for it yet. Maybe turn it into a `http_cacheable?` method on the model? I think right now the controller is making a separate SQL call to check if there are any private resources. But it could probably do 
```ruby
def public_resources?
  resources.all? { |resource| resource.http_cacheable? }
end
```

app/policies/application_policy.rb
marten: AFAIK all Arel stuff returns new scopes, never mutate?

camallen: You are correct, there is no mutation happening here. I tested this and the `@scope` instance variable passed in via the `ApplicationPolicy.scopes_by_action` isn't an actual AR scope, it's a class that is used to create the scopes. What confused me was the assignment `accessible = scope` as it's not needed and doesn't relay the intention either. 

Perhaps the `@scope` instance var should really be `@scope_klass` and the it can be used across the `Scope` object to resolve the underlying scopes. E.g. here could be  
```
accessible_scope = scope_klass.where(id: private_query.select(:resource_id))
accessible_scope = accessible_scope.or(public_scope) if public_flag
accessible_scope
```

lib/json_api_controller/pundit_policy.rb
```diff
@@ -0,0 +1,37 @@
...
+      scope = policy.scope_for(action_name.to_sym)
+
+      if resource_ids.present?
+        scope = scope.where(id: resource_ids)
+      end
+
+      scope
+    end
+
+    def skip_policy_scope
```
camallen: Is this used anywhere? Can't see how this is activated / used.

lib/role_control/pundit_interop.rb
```diff
@@ -0,0 +1,11 @@
+module RoleControl
+  module PunditInterop
+    extend ActiveSupport::Concern
+
+    module ClassMethods
+      def scope_for(action, api_user, opts={})
```
camallen: You may have explained this in the next PR but I'm a little confused as to why this interop exists. Why does a model using the pundit controller need the to define it's own `scope_for` method is this for link management in the `organizations_controller` or just the other linking model controllers? If so i get this then though it was confusing why we access pundit through two different channels, i.e. model -> pundit and controller->pundit.

PR #2817, opened 2018-06-22
Optimize pg selection queries to be more selective

camallen:
The SMS table has grown significantly with our adoption. As such the left outer joins no longer scale. A few of these queries  and can consume a lot of database resources to the detriment of all our projects. This work moves away from the JOINS, specifically the LEFT OUTER JOINs to detect non-retired linked subjects as the selectivity on these joins was terrible and resolved long loops :(

Instead we now compose a simpler AREL with the unnested  seen subject ids, it no longer uses exists / not exists but instead uses a left outer join the SMS table to detect seen/unseens. This LOJ query requires a very selective scan on the SMS table to accompany it at all times. I've made a note in the code and i'm going to refactor this to ensure we only create very selective queries to ensure we can't run these accidentally. 

It also moves to sub queries to detect the currently linked workflow subjects and the retired / unretired state of the data. These sub queries create the high selectivity and are backed by an index on the SMS table. Accordingly i'm seeing these count queries resolve sub ~200ms vs time outs on the current version.

So that is much faster now but there is a catch. These queries will only detect subjects that have already been counted (have a linked `SubjectWorkflowStatus (SWS)` record). Ideally our selection services will load and serve data and thus populate the the status records and then these selectors will work. I'm going to alleviate this in another PR by ensuring that each SMS record creates a linked SWS record on import / create. Currently these get created through selection pressure, instead we will do this to ensure consistent and fast internal data selection services.

The resulting query looks as such
```
--explain analyze verbose

SELECT count(*)
FROM set_member_subjects 

-- find all the seen / unseen matches
LEFT OUTER JOIN (
  SELECT UNNEST(subject_ids) as subject_id 
  FROM user_seen_subjects 
  WHERE user_seen_subjects.user_id = 6
  AND user_seen_subjects.workflow_id = 6206
) as seen_subject_ids 
ON set_member_subjects.subject_id = seen_subject_ids.subject_id 

-- select the smses that are currently linked
WHERE set_member_subjects.subject_set_id IN (
  SELECT subject_sets_workflows.subject_set_id 
  FROM subject_sets_workflows 
  WHERE subject_sets_workflows.workflow_id = 6206
) 

-- select the not retired smses for workflow 
AND set_member_subjects.subject_id IN (
  SELECT subject_workflow_counts.subject_id 
  FROM subject_workflow_counts 
  WHERE subject_workflow_counts.workflow_id = 6206
  AND subject_workflow_counts.retired_at IS NULL
)
-- finally take all the unseen ones (invert this for the seen ones)                              
AND (seen_subject_ids.subject_id IS NULL) 
```

# Review checklist

### DIFF
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -39,7 +39,7 @@ class SetMemberSubject < ActiveRecord::Base
       .where(retired_at: nil)
       .select(:subject_id)
 
-    where(subject_id: non_retired_subject_ids)
+    by_workflow(workflow_id).where(subject_id: non_retired_subject_ids)
   end
 
   def self.retired_for_workflow(workflow_id)
@@ -48,7 +48,7 @@ class SetMemberSubject < ActiveRecord::Base
       .where.not(retired_at: nil)
       .select(:subject_id)
 
-    where(subject_id: retired_subject_ids)
+    by_workflow(workflow_id).where(subject_id: retired_subject_ids)
   end
 
   # THIS can be removed after https://github.com/zooniverse/Panoptes/pull/2805
--- a/app/models/set_member_subject.rb
+++ b/app/models/set_member_subject.rb
@@ -22,12 +22,15 @@ class SetMemberSubject < ActiveRecord::Base
   before_create :set_random
 
   def self.by_subject_workflow(subject_id, workflow_id)
-    joins(subject_set: :workflows)
-      .where(subject_id: subject_id, workflows: {id: workflow_id })
+    by_workflow(workflow_id).where(subject_id: subject_id)
   end
 
-  def self.by_workflow(workflow)
-    joins(:workflows).where(workflows: {id: workflow.id})
+  def self.by_workflow(workflow_id)
+    linked_workflow_set_ids = SubjectSetsWorkflow
+      .where(workflow_id: workflow_id)
+      .select(:subject_set_id)
+
+    where(subject_set_id: linked_workflow_set_ids)
   end
 
   def self.non_retired_for_workflow(workflow)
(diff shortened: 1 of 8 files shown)


marten:
I'm a little hesitant to merge this without the accompanying update that ensures that we actually have status records for each subject. What's the thoughts on that?

camallen:
I will add a new PR to create the `SubjectWorkflowStatus` records on the optimized import and on the non-optmized sets -> subjects linking end points. I was going to add a task to backfill active projects as well and look at removing unused count records on subject deletion as well.

However in theory (and ideally practice) the selector services should take over and create the records through classifications. The first selection via async loaders on the selector services would indeed fallback to any data (as the internal may not find any linked status records) and this could create retired / seens on the first time through. However any subsequent selection requests would drive data through the system and seed the internal selector as well, more classifications should mean more linked SWS records. 

To avoid the failover scenario (seens / retired possibles) we need to have the automatic creation of the SWS records but i think we can deploy this (early tomorrow) and follow up with the auto creation / backfill PR for the linked SWS records.

marten:
i was worried that workflows might have been using the builtin selector but upon inspection it looks no workflow is actually using it? Makes me wonder how you got to find out about this issue?

camallen:
The https://www.zooniverse.org/projects/y-dot-liefting/snapshot-hoge-veluwe project selection queries were backing up and causing really high db cpu use from last Friday, that was the trigger for looking to look into the queries and how to optimize.

app/models/set_member_subject.rb
```diff
@@ -22,47 +22,57 @@ class SetMemberSubject < ActiveRecord::Base
...
+    by_workflow(workflow_id).where(subject_id: retired_subject_ids)
   end
 
-  def self.unseen_for_user_by_workflow(user, workflow)
-    seen_subjects = for_user_by_workflow_scope(user, workflow)
-    by_workflow(workflow).where(seen_subjects.exists.not)
+  # Be careful using this query as it's not selective on a large table
+  # and the LEFT OUTER JOIN can take a long time to resolve
+  #
+  # Note: push this into where it's being used to ensure it's a more selective
```
camallen: I'm 50/50 on moving this code. It's only used in 1 place `select_non_retired_unseen_for_user` but it can be a method (costly query to run) if there is a lot of sms data for the workflow. 

Keeping it as a class method provides here seems a nicer interface and this class if responsible for it's search methods. let me know if you want me to move it.

marten: I'm not too worried about it if we leave the warning in the comment.

PR #2825, opened 2018-06-27
only order subject ids if they exist

camallen:
on selection avoid trying to respect the ordering of an empty list.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -23,10 +23,13 @@ class Api::V1::SubjectsController < Api::ApiController
     subject_selector = Subjects::Selector.new(api_user.user, params)
     selected_subject_ids = subject_selector.get_subject_ids
 
-    selected_subject_scope = Subject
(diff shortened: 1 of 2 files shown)


marten:
So we're still going to ask the database to return to all the subjects matching one of an empty array of IDs, just not sort it? :)

camallen:
Yeah - the serializer needs a scope to resolve, it'll be an empty scope `SELECT"subjects.* FROM subjects WHERE subjects.activated_state = 0 AND 1=0`.

marten:
Sorry, snide remark getting in the way of being actually helpful. What i meant to suggest was using `.none`, which AR knows how to short-circuit.

```ruby
    if selected_subject_ids.empty?
      selected_subject_scope = Subject.none
    else 
      selected_subject_scope = Subject.active.where(id: selected_subject_ids)
        .order("idx(array[#{selected_subject_ids.join(',')}], id)")
    end
```

camallen:
Nice - Scope.none is a better option.

PR #2836, opened 2018-07-10
Pundit 3/3: Replace Linkable with Pundit policies

marten:
This branch converts all of `Linkable` to use Pundit policies and scopes directly instead.

It also adds documentation on how we're using Pundit to do our authorization: https://github.com/zooniverse/Panoptes/blob/canbelinked/doc/pundit.md

# Review checklist


app/policies/organization_policy.rb
```diff
@@ -18,4 +18,10 @@ class TranslateScope < Scope
   scope :index, :show, :versions, :version, with: ReadScope
   scope :update, :update_links, :destroy, :destroy_links, with: WriteScope
   scope :translate, with: TranslateScope
+
+  def linkable_projects
+    # TODO: Is this really right? This lets me stick any
+    # project in my organization. What if they don't want to?
+    policy_for(Project).scope_for(:show)
```
marten: I think this should have been using scope_for update?

camallen: Looks like it was / is `can_be_linked :project, :scope_for, :update, :user` https://github.com/zooniverse/Panoptes/blob/040204f46f7c8818e2beb37e433e5507a50b5917/app/models/organization.rb#L32

marten: Excellent, thanks for validating the entire refactor for me. You've gotten confused by the old code. 

`Organization.can_be_linked(:project)` is used by `Organization.link_to_resource(project)`, which is called by *Projects*Controller when you pass in a `links: {organization: 123}` there. 

This refactor would normally move that to `ProjectPolicy.linkable_organizations`. Except, `organization` isn't in the permitted links: 

* https://github.com/zooniverse/Panoptes/blob/master/config/routes.rb#L104
* https://github.com/zooniverse/Panoptes/blob/master/app/schemas/project_update_schema.rb#L124
* https://github.com/zooniverse/Panoptes/blob/master/app/schemas/project_create_schema.rb#L112

You *can* link projects at the organization side (https://github.com/zooniverse/Panoptes/blob/master/config/routes.rb#L97). In that case the controller would call `Project.link_to_resource(organization)` which uses `Project.scope_for(:show)` because there is no explicit `can_be_linked` defined.

marten: Oh, but please double check that I've got the above reasoning about the existing workings of JsonApiController, RelationManager and Linkable correct!

camallen: Good shout - indeed I had this backwards and always get confused by this. Indeed we did get it wrong originally when settting up orgs. The `can_be_linked` permissions should be on the other side of the relation and is so on other models. 

Would removing the `default_link_to_scope(user)` in `linkable.rb` surface the erroneous model link setups?

app/policies/workflow_policy.rb
```diff
@@ -9,4 +9,30 @@ def resolve(action)
   scope :index, :show, :update, :destroy, :update_links, :destroy_links,
         :translate, :versions, :version, :retire_subjects, :create_classifications_export,
         with: Scope
+
+  def linkable_subject_sets
+    # TODO: Any set that is not part of the same project should get duplicated, not linked.
+    # This currently gets done by the controller, and is set up such that this code here
+    # must allow it to be in scope for linking. However, this opens the door for bugs to
+    # accidentally allow linkage of foreign subject sets. It would be much better if the
```
camallen: what does foreign subject sets mean? is this about mistyping the subject_set_id when linking?

Originally the idea was any public / showable `subject_set` was available for copying.

doc/pundit.md
```diff
@@ -0,0 +1,60 @@
+# Policies
+
+Our policy objects are the heart of the authorizations in Panoptes. Roughtly
+speaking they perform three functions:
+
+1. The `Scope` objects determine which models a user is allowed to access.
+2. The `index?` methods authorize each user for a specific controller action.
+3. The `linkable_xyz` methods return a list of valid Xyzs for a specific model's association.
```
camallen: perhaps `linkable_relation_name`  vs the xyz one. Also perhaps it could use an explicit example combined with the description.

doc/pundit.md
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  scope :index, :show, with: Scope
+end
+```
+
+## Linkable
+
+In order to specify which things can be put into associations, you must
+implement a method called `linkable_xyz`. Substitute xyz for the name of
```
camallen: xyz isn't very descriptive / useful here. the examples below can be used instead and say substitute `_projects` suffix for your linking relation type.

spec/policies/project_policy_spec.rb
```diff
@@ -197,4 +197,40 @@
...
+    end
+
+    it "should allow subject_sets to link when user has update permissions" do
+      subject_set_in_project = create :subject_set, project: project, num_workflows: 0
+      subject_set_other_project = create :subject_set, num_workflows: 0
+
+      expect(policy.linkable_subject_sets).to match_array([subject_set_in_project, subject_set_other_project])
+    end
+
+    # This is not part of the create/update schema for project
```
camallen: can we clean these up?

app/policies/workflow_policy.rb
marten: My thinking here was kind of that it might be preferable for this to only return the sets that are already in the current project. Those are the ones actually allowed to be associated with the workflow, as it were.

A user calling the API with a subject set ID that's not in the project, well those aren't allowed to be associated with the workflow. The controller has a special case for those, and it makes a copy of them into the project, and then links up that copy (which is now in the project, and therefore allowed to be linked per the policy).

Note that the controller is already special casing them, by checking if anything was returned by the policy that happens not to be in the project (calling `same_project?` on them), and then copying those. My TODO is to change that so that rather than calling `same_project?`, it'd be more like (not feature-complete, and crudely doing way too many queries, bear with me):

```ruby
workflow.subject_sets = links["subject_sets"].map do |subject_set_id|
  linkable_subject_sets.find(id) ||
    policy_for(SubjectSet).scope_for(:show).find(id).copy_into(project).id
end
```

I left it as a TODO for a future me (I plan to keep making small PRs kind of as friday-afternoon projects or something) because of all the obvious flaws in the above code. Deemed as out of scope for this PR, basically. Happy to move this comment to somewhere else, like a new GH issue. Would probably make more sense.

doc/pundit-notes.org
```diff
@@ -0,0 +1,190 @@
+#+TITLE: pundit-notes
```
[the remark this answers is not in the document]
marten: Not entirely sure. It's still a useful reference for now, because I don't do any cleaning up of links in this PR, that's why I kept it in for the time being.

lib/json_api_controller/relation_manager.rb
```diff
@@ -31,16 +31,13 @@ def destroy_relation(resource, relation, value)
 
     def find_for_string_type(resource, type, id)
       relation_find(id) do
-        type.camelize
-          .singularize
-          .constantize
-          .link_to_resource(resource, api_user)
+        Pundit.policy!(api_user, resource).linkable_for(type)
```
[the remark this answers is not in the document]
marten: It doesn't generate a `PolicyScope`, generates the Policy itself. It's basically `ProjectPolicy.new(api_user, resource).linkable_workflows`. It's an instance method on the policy so that it can pass in some filter based on the resource to the `WorkflowPolicy::Scope`, e.g. scoping it to only workflows belonging to the project.

Where would your `linkable_model_resources_for_type` exist as a method?

The main thing I would see as confusing myself is that `ProjectPolicy.new`, and thus by extension `Pundit.policy!`, can take both model instances (what would be returned by `Project.find(1)`) as well as AR relation objects (i.e. `Project.where(active: true)`).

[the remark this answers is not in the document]
camallen: yeah - I used the wrong wording in my comment, i get that a Policy object runs the linkable_methods. I find the wording of `Pundit.policy!(api_user, resource).linkable_for(type)` hard to understand what it's doing, maybe splitting the policy creation helps. `linkable_for` doesn't really help me here either, `policy.linkable_for(type)`
> Where would your linkable_model_resources_for_type exist as a method?

In the `relation_manager.rb`for re-use and to capture the intention of the resolving linking AR scopes i have access to on the linking resource / type. As i said above interpreting the pundit line directly is hard.

doc/pundit-notes.org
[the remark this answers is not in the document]
camallen: up to you

lib/json_api_controller/relation_manager.rb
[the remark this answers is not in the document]
marten: In theory because both RelationManager and PunditPolicy are modules mixed into the controller, in here we could just do `policy.linkable_for`.

Suggestions (either both or just one):

1. Convert RelationManager into a service object instead of a mixin, passing in the resource and the policy object.
2. Rename `linkable_for`, perhaps `association_scope`?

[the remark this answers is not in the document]
camallen: anything to relay the intention away from how it is now is ideal. option 2 makes this clearer to me. Understanding how the linkable scopes are wired up via these methods is key to ensuring clarity long term. 

Also knowing which way the relations are wired up will help, e.g. resource vs type. i'd probably have to run the code to know which way these variables are actually passed in. if you can rename these to be clearer `resource_to_link_to` and `type_to_link` could help (if i have that the right way around).

PR #2838, opened 2018-07-10
Ensure UserSeenSubjects are unique

### LINKED ISSUE #2829: User seen subjects should be unique for user + workflows
Merge the affected rows and move to a unique index on the user_id,worklfow_id columns.

marten:
On production we found that some users have more than one USS record for
a given workflow. This breaks Designator in particular, which will fail
to load its user seens cache for these users (in that workflow only).

To deploy make sure to run the rake task for cleanup first, otherwise
the migration will fail. There are only a couple of duplicates in the
database, so I guess that the chance of a new one appearing between the
rake task and running migrations is low enough that we don't need
anything more complicated than this.

Fixes #2829

# Review checklist

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -252,4 +252,24 @@ namespace :migrate do
       end
     end
   end
+
+  namespace :user_seen_subjects do
+    desc "Merge USS for the same user/workflow"
+    task :merge_duplicates => :environment do
+      UserSeenSubject.select("user_id, workflow_id").group("user_id, workflow_id").having("COUNT(*) > 1").each do |duplicate|
+        user_seen_subjects = UserSeenSubject.where(user_id: duplicate.user_id, workflow_id: duplicate.workflow_id)
+        target = user_seen_subjects[0]
+        dupes = user_seen_subjects[1..-1]
+
+        dupes.each do |dupe|
+          UserSeenSubject.transaction do
+            target.reload
+            target.subject_ids = target.subject_ids | dupe.subject_ids
+            target.save!
+            dupe.destroy
+          end
+        end
+      end
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/migrate.rake
```diff
@@ -252,4 +252,24 @@ namespace :migrate do
...
+    task :merge_duplicates => :environment do
+      UserSeenSubject.select("user_id, workflow_id").group("user_id, workflow_id").having("COUNT(*) > 1").each do |duplicate|
+        user_seen_subjects = UserSeenSubject.where(user_id: duplicate.user_id, workflow_id: duplicate.workflow_id)
+        target = user_seen_subjects[0]
+        dupes = user_seen_subjects[1..-1]
+
+        dupes.each do |dupe|
+          UserSeenSubject.transaction do
+            target.reload
+            target.subject_ids = target.subject_ids | dupe.subject_ids
```
camallen: I'm wondering about running this rake task while someone is actively classifying and the fetched records being stale.

Should we avoid the target reload and instead do a `dupe.reload` and use something like 
```
UserSeenSubject
.where(id: target.id)
.update_all(["subject_ids = uniq(subject_ids + array[?])", dupe.reload.subject_ids])
```

marten:
Oh fair point I forgot to actually make it concurrently.

lib/tasks/migrate.rake
marten: Yeah I'll do that. Had another look, and since USS does an `update_all` when adding more subject ids, I don't think we even need to reload the dupe, since anything that was added after the original query fetched them should have been added to all of them, including the target.

camallen: Yeah i saw that too but haven't got the data to check it, in theory they should all be the same and it was a race on the first create maybe? might be a slight divergence if the sidekiq workers backed up while creating the records.

PR #2846, opened 2018-07-18
Add note about password reset token expiration

adammcmaster:


adammcmaster:
I'm guessing there's probably a way to access the relevant config rather than hard coding the number, but I'm not sure how. Could someone point me in the right direction?

https://github.com/zooniverse/Panoptes/blob/52910bb7c8cb77c7203cf7751f355c010f9d7cae/config/initializers/devise.rb#L13

marten:
I couldn't find a good way easily, so I figured we merge it for now.

PR #2854, opened 2018-07-24
Add user field to track intervention notifications

camallen:
Allow users to globally opt out of intervention notifications

# Review checklist


spec/serializers/user_serializer_spec.rb
```diff
@@ -70,4 +70,40 @@
...
+    let(:private_attrs) do
+      UserSerializer::ME_ONLY_ATTRS - ["login_prompt"]
+    end
+
+    context "when i am the permitted requester" do
+      let(:context) do
+        { requester: ApiUser.new(user) }
+      end
+
+      it 'should include the private user data' do
```
marten: Perhaps aggregate failures here and line 102?

camallen:
Good catch :+1: - i've added the strong params to allow updates

PR #2861, opened 2018-07-30
Project copying

zwolf:
Adds project copying via a `POST /projects/:project_id/copy`. Requires that a) the project is not live, b) the project's configuration object includes `template: true`, and c) the user has write privileges on the project.

Utilizes the deep_cloneable gem to keep the logic clean. This gem just provides a new `#dup` that takes a list of associations to call the same on. List of associations included (as well as excluded) associations is in the ProjectCopier lib.

# Review checklist


app/workers/project_copy_worker.rb
```diff
@@ -0,0 +1,9 @@
+class ProjectCopyWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_high
+
+  def perform(project_id, user_id)
+    ProjectCopyWorker.copy(project_id, user_id)
+  end
+end
```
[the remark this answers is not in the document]
zwolf: These newlines exist locally and git thinks the files are synced. That is, I can't commit or push them.

app/policies/project_policy.rb
```diff
@@ -21,7 +21,7 @@ class TranslateScope < Scope
         :create_subjects_export,
         :create_workflows_export,
         :create_workflow_contents_export,
-        :retire_subjects, with: WriteScope
+        :retire_subjects, :copy, with: WriteScope
```
camallen: How will users copy template projects they don't collab / own? Should this be `ReadScope` i.e. public projects or ones i can see in my lab?

app/workers/project_copy_worker.rb
```diff
@@ -0,0 +1,9 @@
+class ProjectCopyWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_high
```
camallen: Do we want to look at locking on unique args to ensure users don't trigger multiple copies by double clicking form submission?

app/workers/project_copy_worker.rb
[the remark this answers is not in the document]
camallen: Do you have a linter for your editor? Something like https://github.com/ngmy/vim-rubocop might help to run it locally as well.

lib/project_copier.rb
```diff
@@ -0,0 +1,30 @@
...
+                          :pages,
+                          :tags,
+                          :tagged_resources,
+                          :avatar,
+                          :background,
+                          { active_workflows: [ :tutorials, :attached_images, :workflow_contents ] }
+                        ].freeze
+
+  def self.copy(project_id, user_id)
+    project = Project.find(project_id)
```
camallen: what happens if i can't find the project / user by ID?

lib/project_copier.rb
```diff
@@ -0,0 +1,30 @@
+class ProjectCopier
+  EXCLUDE_ATTRIBUTES = [ :classifications_count, :launched_row_order, :beta_row_order ].freeze
+  INCLUDE_ATTRIBUTES = [  :project_contents,
```
camallen: These aren't really attributes but associations. Should the variable name reflect that?

lib/project_copier.rb
```diff
@@ -0,0 +1,30 @@
...
+                          :background,
+                          { active_workflows: [ :tutorials, :attached_images, :workflow_contents ] }
+                        ].freeze
+
+  def self.copy(project_id, user_id)
+    project = Project.find(project_id)
+    user = User.find(user_id)
+
+    copied_project = project.deep_clone include: INCLUDE_ATTRIBUTES, except: EXCLUDE_ATTRIBUTES
+    copied_project.owner = user
```
camallen: did you look at the block syntax for updating the clone? https://github.com/moiristo/deep_cloneable#optional-block

spec/lib/project_copier_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      it "renames a project when the owner is copying their own project" do
+        new_copy = described_class.copy(project.id, project.owner.id)
+        expect(new_copy.display_name).to include("(copy)")
+      end
+
+      it "has matching attributes" do
+        expect(copied_project.display_name).to eq(project.display_name)
+      end
```
camallen: You might want to test the other changes to the copied instance as well, e.g. `copied_project.update_attributes(launch_approved: false, live: false)`

app/policies/project_policy.rb
zwolf: Depends. Write makes sense to me to roll out this feature on a small number of hand selected projects (via the manually setting the configuration) and the button in PFE doesn't exist yet. For those same reasons, though, I can see not mattering very much and maybe it's better to put it in Read now and not have to change it later.

lib/project_copier.rb
zwolf: I'm not doing anything with the copy that needs any pre-exclude original attributes or anything else I can think of that requires that block. It's all happening pre-`save!`, so it's not hitting the db yet.

spec/lib/project_copier_spec.rb
zwolf: ✔️

lib/project_copier.rb
zwolf: ✔️

lib/project_copier.rb
zwolf: ✔️ Good call, added a rescue to the worker.

app/workers/project_copy_worker.rb
zwolf: ✔️

zwolf:
Thanks, updated with suggested changes. Also, somehow we both managed to miss that I was calling the ProjectCopyWorker from itself instead of the ProjectCopier. There are specs for that, now, too. :3

app/workers/subject_workflow_status_create_worker.rb
```diff
@@ -1,4 +1,4 @@
-class SubjectWorkflowStatusCreateWorker
+  class SubjectWorkflowStatusCreateWorker
```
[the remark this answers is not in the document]
zwolf: aw c'mon, The Hound, this wasn't even--aw, forget it. Fine, I'll fix it.

spec/lib/project_copier_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe ProjectCopier do
+  describe '#copy' do
+    let(:project) { create(:full_project, build_extra_contents: true)}
+    let(:copyist) { create(:user) }
+    let(:tags) { create(:tag, resource: project) }
```
camallen: none of these associated records exist as they are never called before the copy. Might want to add a `let!(:tags)` or call them before you use them. Also you should probably test these associations exist on the copied project, i wouldn't validate the copy just that the included associations are copied.

lib/project_copier.rb
```diff
@@ -0,0 +1,29 @@
+class ProjectCopier
+  EXCLUDE_ASSOCIATIONS = [ :classifications_count, :launched_row_order, :beta_row_order ].freeze
```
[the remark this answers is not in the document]
zwolf: doh! shouldn't have changed both.

lib/project_copier.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    copied_project = project.deep_clone include: INCLUDE_ASSOCIATIONS, except: EXCLUDE_ASSOCIATIONS
+    copied_project.owner = user
+    copied_project.configuration.delete(:template)
+
+    if user == project.owner
+      copied_project.display_name += " (copy)"
+    end
+
+    copied_project.update_attributes(launch_approved: false, live: false)
```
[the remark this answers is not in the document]
zwolf: absolutely, ✔️

spec/lib/project_copier_spec.rb
zwolf: yup. bang!

PR #2862, opened 2018-07-31
Move jenkins file to declarative format and use existing jobs

camallen:
The current jenkins file relies on the workspace of the operations repo to exist and it may not (see failed jobs for panoptes master branch in jenkins). This PR fixes the above failures and moves the Jenkinsfile to a declarative (https://jenkins.io/doc/book/pipeline/syntax/#declarative-directives) instead of the scripted format. 

The declarative format should provide greater consistency of the Jenkins file format across repos and allow us to use tag builds longer term (https://jenkins.io/blog/2018/05/16/pipelines-with-git-tags/), while still using a script step to build and reuse the docker image caches.

### TODO
- [x] Unlink the cascading staging jobs to ensure they run through this Jenkinsfile
- [ ] Move dockerhub named repo to original panoptes one (or in a later PR off this one)
- [ ] Add production tag image building step (or in a later PR off this one)

# Review checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,23 +46,27 @@ pipeline {
       }
     }
 
-    stage('Migrate Staging DB') {
+    stage('Migrate') {
       when {
         branch 'master'
       }
-      agent {
-        docker { image 'zooniverse/operations:latest' }
-      }
-      steps {
-        sh """#!/bin/bash -e
-          source auto_cleanup.sh
-          source deploylib.sh
-          INSTANCE_ID=\$(./launch_latest.sh -q panoptes-api-staging)
-          INSTANCE_DNS_NAME=\$(instance_dns_name \$INSTANCE_ID)
-          # Wait for instance/panoptes to come up
-          timeout_cmd "timeout 5m ssh ubuntu@\$INSTANCE_DNS_NAME docker-compose -f /opt/docker_start/docker-compose.yml -p panoptes-api-staging exec -T panoptes true"
-          ssh ubuntu@\$INSTANCE_DNS_NAME docker-compose -f /opt/docker_start/docker-compose.yml -p panoptes-api-staging exec -T panoptes ./migrate.sh
-        """
+      stages {
+        stage('Staging DB') {
+          agent {
+            docker { image 'zooniverse/operations:latest' }
+          }
+          steps {
+            sh """#!/bin/bash -e
+              source auto_cleanup.sh
+              source deploylib.sh
+              INSTANCE_ID=\$(./launch_latest.sh -q panoptes-api-staging)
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,66 +1,59 @@
...
-          ./rebuild.sh panoptes-api-staging
-        """
-      },
-      worker: {
-        sh """
-          cd "/var/jenkins_home/jobs/Zooniverse GitHub/jobs/operations/branches/master/workspace" && \
-          ./rebuild.sh panoptes-dumpworker-staging
-        """
+      when {
+        anyOf { branch 'master'; branch 'fix_jenkinsfile' }
```
camallen: Whoops - the branch `'fix_jenkinsfile'` has to go

Jenkinsfile
```diff
@@ -1,66 +1,59 @@
-#!groovy
```
adammcmaster: I'd prefer to leave this in to enable syntax highlighting in editors.

Jenkinsfile
```diff
@@ -1,66 +1,59 @@
...
-          ./rebuild.sh panoptes-dumpworker-staging
-        """
+      when {
+        anyOf { branch 'master'; branch 'fix_jenkinsfile' }
+      }
+      failFast true
+      parallel {
+        stage('Build API') {
+          steps {
+            build job: '/Build Panoptes Staging AMI'
```
adammcmaster: I want to phase out the old jobs, so I'd rather move their content into this file instead of just triggering them from here. The actual logic of building/deploying should be in this repo (if that means we need an operations Docker image or something, to make things cleaner, then we can do that).

camallen: understand - we need access to the operations repo so having a docker image would be neat or we can look into another SCM checkout to gain access like the jobs currently do.

adammcmaster: Yeah, the operations repo has a jenkins file that just does `checkout scm`, so the copy of the repo at `/var/jenkins_home/jobs/Zooniverse GitHub/jobs/operations/branches/master/workspace` is already aways up to date. Docker would obviously be cleaner than having to `cd` to there first.

adammcmaster: That's weird. I just manually re-ran the operations job and it checked it out to there. Maybe Jenkins cleans up the workspaces for jobs that haven't run in a while? In which case a docker image is definitely the way to go.

camallen:
@adammcmaster i've updated to run the deploy commands via ops docker containers but isn't as clean as i'd like (mostly from the jenkins durable task timeout issues), I'd appreciate another review when you have time.

adammcmaster:
This will do for now (to avoid delaying this any more), but we should update `timeout_cmd` to echo "." on each attempt, to save duplicating this in every job. We can come back and clean this up afterwards.

PR #2865, opened 2018-07-31
decrease SWS create worker run priority

camallen:
Ensure these run when they can.

NOTE: take care when launching large numbers of unique lock jobs as they can increase redis memory usage (possibly OOM and kill redis).

# Review checklist

### DIFF
--- a/app/workers/subject_set_statuses_create_worker.rb
+++ b/app/workers/subject_set_statuses_create_worker.rb
@@ -1,7 +1,7 @@
 class SubjectSetStatusesCreateWorker
   include Sidekiq::Worker
 
-  sidekiq_options queue: :data_medium, lock: :until_executed
+  sidekiq_options queue: :data_low, lock: :until_executed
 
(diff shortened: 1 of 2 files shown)


camallen:
Setting `SubjectSetStatusesCreateWorker` to a lower priority may not make a difference as once this job starts it will enqueue all it's create jobs while only taking 1 worker thread. Setting it lower may just delay the invocation of the start of the `SubjectSetStatusesCreateWorker` job while still increasing the queue backlog. 

This PR aims to run the queued create jobs less often to avoid swamping the database with create tasks. Perhaps these should be scheduled over a time period as well to avoid the databse issue?

Re the queue count and the redis OOM error, there should really be no problem storing a lot of jobs in the queue. I think this is related to https://github.com/mhenrixon/sidekiq-unique-jobs/issues/234 and the large hash of locked jobs that v5 of that gem creates. If that doesn't alleviate any issues then we'll have to re-assess how we queue jobs in the background for processing.

PR #2867, opened 2018-08-01
Add explicit translation scope tests for resources

camallen:
ensure we test translators can add resource translations by moving the policy scope specs to the `PunditScopeTester` with explicit scope checks across translated resources. Also removed the resource policy `*_links` action scopes that aren't wired up via routes so can't be reached and updated associated policy specs.

# Review checklist

### DIFF
--- a/app/policies/aggregation_policy.rb
+++ b/app/policies/aggregation_policy.rb
@@ -19,7 +19,7 @@ class AggregationPolicy < ApplicationPolicy
   end
 
   scope :index, :show, with: ReadScope
-  scope :update, :destroy, :update_links, :destroy_links, :versions, :version, with: WriteScope
+  scope :update, :destroy, :versions, :version, with: WriteScope
 
   def linkable_subjects
     policy_for(Subject).scope_for(:show)
--- a/app/policies/field_guide_policy.rb
+++ b/app/policies/field_guide_policy.rb
@@ -6,7 +6,7 @@ class FieldGuidePolicy < ApplicationPolicy
     end
   end
 
-  scope :index, :show, :update, :destroy, with: Scope
+  scope :index, :show, :update, :destroy, :translate, with: Scope
 
   def linkable_projects
     policy_for(Project).scope_for(:update)
--- a/app/policies/membership_policy.rb
+++ b/app/policies/membership_policy.rb
@@ -27,7 +27,7 @@ class MembershipPolicy < ApplicationPolicy
     end
   end
 
-  scope :index, :show, :update, :destroy, :update_links, :destroy_links, with: Scope
+  scope :index, :show, :update, :destroy, with: Scope
 
   def linkable_users
     User.all
--- a/app/policies/organization_page_policy.rb
+++ b/app/policies/organization_page_policy.rb
@@ -6,15 +6,8 @@ class OrganizationPagePolicy < ApplicationPolicy
     end
   end
 
-  class TranslateScope < Scope
-    def resolve(_action)
-      parent_scope = policy_for(Organization).scope_for(:translate)
-      scope.where(organization_id: parent_scope.select(:id))
-    end
-  end
-
-  scope :index, :show, with: Scope
-  scope :update, :destroy, :update_links, :destroy_links, :versions, :version, with: TranslateScope
+  scope :index, :show, :update, :destroy, :translate, :update_links,
+        :destroy_links, :versions, :version, with: Scope
 
   def linkable_organizations
     policy_for(Organization).scope_for(:update)
--- a/app/policies/organization_page_policy.rb
+++ b/app/policies/organization_page_policy.rb
@@ -6,8 +6,7 @@ class OrganizationPagePolicy < ApplicationPolicy
     end
   end
 
-  scope :index, :show, :update, :destroy, :translate, :update_links,
-        :destroy_links, :versions, :version, with: Scope
+  scope :index, :show, :update, :destroy, :translate, :versions, :version, with: Scope
 
   def linkable_organizations
     policy_for(Organization).scope_for(:update)
--- a/app/policies/project_page_policy.rb
+++ b/app/policies/project_page_policy.rb
@@ -6,9 +6,7 @@ class ProjectPagePolicy < ApplicationPolicy
     end
   end
 
-  # TODO: look into removing :update_links, :destroy_links as these route actions don't exist for this resource
-  scope :index, :show, :update, :destroy, :translate, :versions,
-        :version, :update_links, :destroy_links, with: Scope
+  scope :index, :show, :update, :destroy, :translate, :versions, :version, with: Scope
 
   def linkable_projects
     policy_for(Project).scope_for(:update)
--- a/app/policies/project_page_policy.rb
+++ b/app/policies/project_page_policy.rb
@@ -6,6 +6,7 @@ class ProjectPagePolicy < ApplicationPolicy
     end
   end
 
+  # TODO: look into removing :update_links, :destroy_links as these route actions don't exist for this resource
   scope :index, :show, :update, :destroy, :translate, :versions,
         :version, :update_links, :destroy_links, with: Scope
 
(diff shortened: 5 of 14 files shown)


marten:
Would you mind talking me through this compared to the before-pundit state? I still can't find what would indicate that these things were converted to Pundit incorrectly, or what triggers these bugs. Basically I could review the diff on it's own, but I don't understand the reason behind the diff and I'd like to.

camallen:
The translation controller manually sets the 'translate' action. This was to resolve to the correct roles on the parent resource in the old permissions model. https://github.com/zooniverse/Panoptes/blob/db02a12f073348abac5dfb76485c676babec160c/app/controllers/api/v1/translations_controller.rb#L44 

The translate role was migrated to the pundit scopes https://github.com/zooniverse/Panoptes/blob/ddcbe4edeb99e88f873d31a11404095ba604986a/app/models/project.rb#L48
https://github.com/zooniverse/Panoptes/blob/db02a12f073348abac5dfb76485c676babec160c/app/policies/project_policy.rb#L25

However the underlying issue was that the `translate` action wasn't specified in pundit for the these resources. 
https://github.com/zooniverse/Panoptes/blob/db02a12f073348abac5dfb76485c676babec160c/app/policies/project_page_policy.rb#L17
Hence the unknown action error https://app.honeybadger.io/projects/40595/faults/38913903#notice-summary

Also i noted that the custom translation policy scopes that were migrated here really just proxy up to the parent resource and they already have the correct policy actions and scopes wired up. Accordingly these have been removed and now just proxy to the parent resource on their actions. 

E.g. `OrganizationPagePolicy::Scope` with action `update` will find all the Organizations i have write access to https://github.com/zooniverse/Panoptes/blob/ca9c118f9f716d65774ed08ceefc680dcd4fd0e9/app/policies/organization_policy.rb#L19

Likewise `ProjectPagePolicy::Scope` with action `translate` will proxy to the `ProjectPolicy::TranslateScope` https://github.com/zooniverse/Panoptes/blob/ca9c118f9f716d65774ed08ceefc680dcd4fd0e9/app/policies/project_policy.rb#L25

marten:
Here's where I get confused. If I look at the old ProjectPage, there is nothing to suggest that it's translatable:

https://github.com/zooniverse/Panoptes/blob/ddcbe4edeb99e88f873d31a11404095ba604986a/app/models/project_page.rb#L10

camallen:
Yeah it is confusing (hence your refactor). see the overridden `scope_for` below that link
https://github.com/zooniverse/Panoptes/blob/ddcbe4edeb99e88f873d31a11404095ba604986a/app/models/project_page.rb#L19

marten:
But that calls it on Project, not on ProjectPage. Where is the indication that ProjectPage.scope_for will ever get called with `translate`?

camallen:
There is no indication apart from the content I linked above and double checking that won't ever get called from the translations controller. That `scope_for` would only be called when hitting `project_pages` controller.

The new translations data model uses a query param to wire up the controlled resource scope i.e. proxies to parent Project, Organization, etc for the translate actions. If you can access this you can access the translation objects. It also sets the scope action name to `translate` to access the correct scopes. 

This passed the specs as the controller spec only checks the project resource and that policy object had the correct scopes wired up. Testing more than the index action in the policy specs would still not have surfaced this error. 

I think the disconnect here is that the old translations system would return the resource with interpolated translation strings and would rely on the existing resource controller actions (what i've listed above) to test the translation permissions. E.g. this `:translate` scope action would have run https://github.com/zooniverse/Panoptes/blob/ddcbe4edeb99e88f873d31a11404095ba604986a/app/models/project_page.rb#L19

The action overrides there are vestigial and probably should have been removed / updated in the new translation system PR to avoid the confusion. Again another nod to your refactor to make it clearer and something that should be scheduled to be removed as well.

app/policies/aggregation_policy.rb
```diff
@@ -19,7 +19,7 @@ def resolve(action)
   end
 
   scope :index, :show, with: ReadScope
-  scope :update, :destroy, :update_links, :destroy_links, :versions, :version, with: WriteScope
+  scope :update, :destroy, :versions, :version, with: WriteScope
```
marten: As I understand, this is going away because it doesn't support linking in the routes. It was listed in the `can_through_parent` call before, but as I've now found out that list was meaningless and not actually referenced by anything.

app/policies/membership_policy.rb
```diff
@@ -27,7 +27,7 @@ def resolve(action)
     end
   end
 
-  scope :index, :show, :update, :destroy, :update_links, :destroy_links, with: Scope
+  scope :index, :show, :update, :destroy, with: Scope
```
marten: As I understand, this is going away because it doesn't support linking in the routes. It was listed in the can_through_parent call before, but as I've now found out that list was meaningless and not actually referenced by anything

app/policies/field_guide_policy.rb
```diff
@@ -6,7 +6,7 @@ def resolve(action)
     end
   end
 
-  scope :index, :show, :update, :destroy, with: Scope
+  scope :index, :show, :update, :destroy, :translate, with: Scope
```
marten: FieldGuides are valid targets for a Translation model to link to. This wasn't explicit anywhere in the code, and the `:translate` action wasn't listed on the call the `can_through_parent` before. This worked because ParentalControlled just passed the action along to the parent, without restricting it to the array of actions defined via `can_through_parent`. That's how this action got missed during the conversion to Pundit.

app/policies/aggregation_policy.rb
camallen: yep - those actions could have been checked against the whitelist but werent' https://github.com/zooniverse/Panoptes/blob/b4b98f9843ec759c00f5546fd4e0be3e86b633f3/lib/role_control/parental_controlled.rb#L36

app/policies/field_guide_policy.rb
camallen: The whitelist of valid actions could have been checked here but wasn't https://github.com/zooniverse/Panoptes/blob/b4b98f9843ec759c00f5546fd4e0be3e86b633f3/lib/role_control/parental_controlled.rb#L36

app/policies/organization_page_policy.rb
```diff
@@ -6,15 +6,7 @@ def resolve(action)
...
-  class TranslateScope < Scope
-    def resolve(_action)
-      parent_scope = policy_for(Organization).scope_for(:translate)
-      scope.where(organization_id: parent_scope.select(:id))
-    end
-  end
-
-  scope :index, :show, with: Scope
-  scope :update, :destroy, :update_links, :destroy_links, :versions, :version, with: TranslateScope
+  scope :index, :show, :update, :destroy, :translate, :versions, :version, with: Scope
```
marten: I'm not entirely sure about this change. The existing code is saying "you can update a translation if you can translate on the parent". The new code is saying "you can update a translation if you can update the parent" which is not the same.

I know that the translations controller will only ever pass in "translate", but then why is "update" kept in the new list?

And "destroy" was converted to "translate on parent" here in the old code, and thinking about it that was probably wrong (as in we don't want a translator to be able to delete a translation, we only want people with delete access on the parent to be able to delete a translation).

app/policies/project_page_policy.rb
```diff
@@ -6,15 +6,7 @@ def resolve(action)
...
-  class TranslateScope < Scope
-    def resolve(_action)
-      parent_scope = policy_for(Project).scope_for(:translate)
-      scope.where(project_id: parent_scope.select(:id))
-    end
-  end
-
-  scope :index, :show, with: Scope
-  scope :update, :destroy, :update_links, :destroy_links, :versions, :version, with: TranslateScope
+  scope :index, :show, :update, :destroy, :translate, :versions, :version, with: Scope
```
marten: See https://github.com/zooniverse/Panoptes/pull/2867/files#r208257077

PR #2880, opened 2018-08-08
WIP: Make translatableness explicit

marten:

### DIFF
--- a/app/models/concerns/translatable.rb
+++ b/app/models/concerns/translatable.rb
@@ -1,0 +1,9 @@
+# This module should be included in any model class that is
+# a valid target for the Translation model.
+module Translatable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :translations, as: :translated, dependent: :destroy
+  end
+end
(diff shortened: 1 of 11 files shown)


app/models/concerns/translatable.rb
```diff
@@ -1,39 +1,15 @@
...
-  # TODO: this should become the association instead of the has_many
-  def primary_content
-    @primary_content ||= if content_association.loaded?
-      content_association.to_a.find do |content|
-        content.language == primary_language
-      end
-    else
-      content_association.find_by(language: primary_language)
+    def translatable_attributes
+      %i().freeze
```
camallen: Seems like an abstract method, should this raise if it hasn't been defined in the including class?

app/models/translation.rb
```diff
@@ -9,7 +10,7 @@ class Translation < ActiveRecord::Base
     scope: %i(translated_type translated_id),
     message: "translation already exists for this resource"
 
-  # TODO: Look at adding in paper trail change tracking for laguage / strings here
+  # TODO: Look at adding in PaperTrail change tracking for language / strings here
```
camallen: Could this comment go / be updated to not use papertrail for change tracking?

marten:
Good points. Resolved.

PR #2882, opened 2018-08-08
[WIP] begin looking for inactive users

RonaCostello:
Work to pull out inactive users to send follow-up emails to

# Review checklist


app/mailers/dormant_user_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class DormantUserMailer < ApplicationMailer
+  layout false
+
+  DEFAULT_SUBJECT = "Come back to the Zooniverse"
```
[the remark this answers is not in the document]
camallen: add "String".freeze to this, see:
+ https://freelancing-gods.com/2017/07/27/an-introduction-to-frozen-string-literals.html
+ https://wyeworks.com/blog/2015/12/1/immutable-strings-in-ruby-2-dot-3

app/models/user.rb
```diff
@@ -86,6 +86,27 @@ class User < ActiveRecord::Base
     },
     ranked_by: ":tsearch + (0.25 * :trigram)"
 
+  def self.dormant(window=5)
+    DatabaseReplica.read("read_dormant_users_from_replica") do
+      # devise trackable sets the current_sign_in_at on each login
+      havent_signed_in_since = "date(now()) - date(current_sign_in_at) >= #{window}"
+      where(havent_signed_in_since).find_each do |dormant_user|
+        # Check if dormant user has classified in last 5 days, only yield block if meet warden_conditions
```
camallen: We can remove this comment, the code reads well enough that it doesn't need an explanation.

app/models/user.rb
```diff
@@ -86,6 +86,27 @@ class User < ActiveRecord::Base
...
+        latest_user_classification = Classification.where(user_id: dormant_user).order(created_at: :desc).first
```
camallen: Also i'm wondering if we can use the `Recent` class (table) for this query instead of the `Classification`, it might be a smaller index / table space search? 

I am happy to help test this out too.

app/views/dormant_user_mailer/email_dormant_user.text.erb
```diff
@@ -0,0 +1,22 @@
...
+See you out there!
+
+The Zooniverse Team
+
+UNSUBSCRIBE INSTRUCTIONS:
+You are receiving this email because you have recently registered an account with the Zooniverse.
+<%- if @user %>
+To unsubscribe to these messages instantly visit <%= unsubscribe_url(token: @user.unsubscribe_token) %>
+Please be aware that this action will unsubscribe you from ALL Zooniverse emails; if you wish to manage individual project subscriptions, please visit your settings page link below.
+Alternatively you can <a href="https://zooniverse.org/#/unsubscribe">unsubscribe here</a>.
```
camallen: We can't use the HTML tags in a text email, e.g. `<a>` etc. Construct the text link instead. If this was copied we can fix the other templates in another PR.

app/workers/dormant_user_mailer_worker.rb
```diff
@@ -0,0 +1,12 @@
+class DormantUserMailerWorker
+  include Sidekiq::Worker
+
+attr_reader :user
```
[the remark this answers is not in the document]
camallen: agree - indent this.

app/workers/dormant_user_mailer_worker.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def perform(user_id)
+    @user = User.find(user_id)
+    if @user && @user.receives_email?
```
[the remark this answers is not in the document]
camallen: change this to `if @user&.receives_email?` http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

spec/mailers/dormant_user_mailer_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    it 'should come from no-reply@zooniverse.org' do
+      expect(mail.from).to include('no-reply@zooniverse.org')
+    end
+
+    it 'should have the correct subject' do
+      expect(mail.subject).to eq("Come back to the Zooniverse")
+    end
+
+    it 'should have the user name in the body' do
+      expect(mail.body.encoded).to match("#{user.display_name}")
```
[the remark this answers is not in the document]
camallen: change to `match(user.display_name)` as it's already a string.

app/mailers/dormant_user_mailer.rb
```diff
@@ -0,0 +1,13 @@
+class DormantUserMailer < ApplicationMailer
+  layout false
+
+  DEFAULT_SUBJECT = "Come back to the Zooniverse".freeze
+
+  def email_dormant_user(user)
+    @user = user
+    @email_to = user.email
+    @last_project= UserProjectPreference.where(user_id: user.id).first
```
camallen: make sure this gets the actual project
```
UserProjectPreference
.where(user_id: dormant_user.id)
.where.not(email_communication: nil)
.order(updated_at: :desc)
.first
.project
```

app/models/user.rb
```diff
@@ -86,6 +86,39 @@ class User < ActiveRecord::Base
     },
     ranked_by: ":tsearch + (0.25 * :trigram)"
 
+  def self.dormant(window=5)
```
camallen: 5 days is really not that long, a short break away. This window really should be expanded to at least 2 weeks and probably 1 month. Thoughts?

app/views/dormant_user_mailer/email_dormant_user.text.erb
```diff
@@ -0,0 +1,21 @@
+Dear <%= @user.display_name %>,
+
+Our projects still need you!
+
+Whether it is helping to classify galaxies, map organutan nesting sites, identify antibiotic resistant bacteria or to transcribe anti-slavery manuscripts, there are many exciting research projects to get involved with in the Zooniverse.
+
+Pick up where you last left off on [project name, with embedded link]<%= @last_project %>, or check out all our projects here: (https://zooniverse.org/#/projects)
```
camallen: use the `@last_project` to get the `display_name` and slug attributes, don't print the object id.

app/views/dormant_user_mailer/email_dormant_user.text.erb
```diff
@@ -0,0 +1,21 @@
...
+
+Don't forget to log in so you can keep track of your contributions and get involved on the project discussion boards (https://zooniverse.org/#/talk).
```
camallen: remove the fragment, we don't use these in routing anymore. https://zooniverse.org/talk

app/views/dormant_user_mailer/email_dormant_user.text.erb
```diff
@@ -0,0 +1,21 @@
...
+
+The Zooniverse Team
+
+UNSUBSCRIBE INSTRUCTIONS:
+<%= yield %>
+<% if @user %>
+  To unsubscribe to these messages instantly visit <%= unsubscribe_url(token: @user.unsubscribe_token) %>
+  Please be aware that this action will unsubscribe you from ALL Zooniverse emails; if you wish to manage individual project subscriptions, please visit your settings page link below.
+<% end %>
+To manage your email subscription preferences visit https://zooniverse.org/#/settings
```
camallen: Same re fragment https://zooniverse.org/settings

app/mailers/dormant_user_mailer.rb
```diff
@@ -0,0 +1,13 @@
+class DormantUserMailer < ApplicationMailer
+  layout false
```
camallen: maybe just use the default text layout to get the unsubscribe footers.

app/models/user.rb
```diff
@@ -86,6 +86,39 @@ class User < ActiveRecord::Base
...
+    DatabaseReplica.read("read_dormant_users_from_replica") do
+      # devise trackable sets the current_sign_in_at on each login
+      havent_signed_in_since = "date(now()) - date(current_sign_in_at) >= #{window}"
+      where(havent_signed_in_since).select(:id).find_each do |dormant_user|
```
camallen: This where clause should also respect the `receives_email?` states as described in https://github.com/zooniverse/Panoptes/blob/30e33a4f28b45269d2610ddf8eea0f8436ba40c6/app/models/user.rb#L201

We should skip disabled users, users without email attributes or ones that have been marked with invalid email addresses (bouncing). Doing this in the DB will filter out more users, it shouldn't impact on query times significantly but will require a quick benchmark on real data. Check in with Marten re this one.

app/workers/dormant_user_mailer_worker.rb
```diff
@@ -0,0 +1,12 @@
+class DormantUserMailerWorker
+  include Sidekiq::Worker
+
+  attr_reader :user
+
+  def perform(user_id)
+    @user = User.find(user_id)
```
camallen: Does the mailer need the instance vars here?  Looks like it'll assign it's own instance vars, remove them if not. Also you can use the safe navigation operator here if you like:
```
@user = User.find(user_id)
if @user&.receives_email?
  DormantUserMailer.email_dormant_user(@user).deliver
end 
```
Also nice work on the check that this user can receive email, top work 👍

camallen: This should also short circuit and return straight away if we are on the staging env (non production env). If there are users on there they don't want to be getting emails that make no sense.

marten:
I'm really confused about what's up with all the commits on this branch. First thing we should look at next is probably cleaning up the commit list on this, so that after that it's clear to see what's actually new code.

app/models/user.rb
```diff
@@ -86,6 +86,46 @@ class User < ActiveRecord::Base
...
+    subset_selection = "id%10 = ?"
+    where(subset_selection, ending_in)
+  end
+
+  def self.dormant(window=14, query=User)
+    DatabaseReplica.read("read_dormant_users_from_replica") do
+      # devise trackable sets the current_sign_in_at on each login
+      havent_signed_in_since = "date(now()) - date(current_sign_in_at) >= #{window}"
+      query = query.where(valid_email: true).where(havent_signed_in_since)
+      query = query.where(activated_state: User.activated_states[:active]).select(:id)
```
marten: I assume we should also be removing users that have `global_email_communication` set to false here?

app/workers/dormant_user_mailer_worker.rb
marten: I'm not sure what cam meant, but I think he meant that the instance variables here could just be normal variables, at least in the worker. Cam's suggestion of the safe navigation operator was a nice idea, except `User.find` will raise an exception, not return nil. You could rescue `ActiveRecord::RecordNotFound` and just do nothing in that case, but with User we never ever remove user records from the database, so if you can't find them, it really is something strange going on and we probably want the default behaviour (which sends us an email about the exception, and retries automatically).

```
def perform(user_id)
  user = User.find(user_id)
  
  if user.receives_email?
    DormantUserMailer.email_dormant_user(user).deliver
  end
end
```

marten: Noting here that we have opted to not make it a scheduled job. Instead, we'll enqueue it a few times with the correct cohorts at the start, and we'll do that on production. Acceptable?

app/models/user.rb
RonaCostello: I suppose not unless we're saying a dormant user email is not a general zooniverse email update? Probably best not to start emailing people who have chosen to opt out of general emails.

app/models/user.rb
```diff
@@ -86,6 +86,46 @@ class User < ActiveRecord::Base
...
+        # A user's project preference (UPP) is updated
+        # each time they classify on a project.
+        # The last updated UPP is a proxy to a user's classification activity
+        # and the query is much less expensive than classifications or recents
+        #
+        # Ideally this should move to a last_classifcation FK attribute on the
+        # user which is updated in the classification lifecycle
+        last_classified_upp = UserProjectPreference
+          .where(user_id: dormant_user.id)
+          .where.not(email_communication: nil)
```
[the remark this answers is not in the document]
RonaCostello: Given that call within the email is based on the users last upp, it is probably best not to send them emails about it. The alternative is that if the project communication is nil, we send them the non project-specific email (if global email comms are true).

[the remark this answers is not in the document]
camallen: this is really a proxy for a user classifying on a project, if their `email_communication: nil` it means they never classified on a project.

app/workers/get_dormant_users_worker.rb
```diff
@@ -0,0 +1,10 @@
+class GetDormantUsersWorker
+  include Sidekiq::Worker
+
+  def perform(num_days_since_activity, ending_in)
+    selected_users = User.subset_selection(ending_in)
```
camallen: ` raise DoNotRunOnStagingError if Rails.env.staging?`

PR #2890, opened 2018-08-13
Speed up tests

marten:
I was sharpening up my editor--testsuite integration with some hotkeys
to run a single test, file, or rerun the last one. Because this made it
faster to run tests, it made it more apparent that we're wasting a bunch
of time in our tests.

Specifically, it turns out that in our case, deletion is faster than
truncation as a "reset the db" strategy in DatabaseCleaner. This means
we don't reset IDs back to 1 every test run, but who cares, right? This
was the cause of the delay between RSpec saying "Randomized with seed"
and it actually starting to run the first example.

A further delay was caused by Celluloid (used by Sidetiq). We've already
worked around it somewhat by setting the timeout to 1 second in test.
It's also not as bad as the truncation fix, since this delay sits at the
very end of the test run, after the results are already visible, but
still. After more than a year has passed, we still don't have a full
release of Celluloid, but at least now there is a prerelease that we can
bump to. This removes the warning and timeout delay entirely.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'active_record_union', '~> 1.3.0'
 gem 'activerecord-import', '~> 0.25'
 gem 'aws-sdk', '~> 2.10'
 gem "cellect-client", '~> 3.0.2'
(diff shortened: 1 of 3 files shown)


camallen:
Also :tada: nice work :+1:

PR #2892, opened 2018-08-14
focus the called workers for workflow link actions

camallen:
This PR adds greater targeting of the called workers when updating/deleting workflow links for subjects_sets and/or retired_subjects. E.g. when adding subject_sets the SWS Create worker should be called but not the Notify Selector of retirement worker. 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -99,7 +99,7 @@ class Api::V1::WorkflowsController < Api::ApiController
       RefreshWorkflowStatusWorker.perform_async(workflow.id)
     end
 
-    if relation_index == 0
+    if relation_index&.zero?
       NotifySubjectSelectorOfChangeWorker.perform_async(workflow.id)
     end
   end
--- a/app/controllers/api/v1/workflows_controller.rb
+++ b/app/controllers/api/v1/workflows_controller.rb
@@ -33,6 +33,17 @@ class Api::V1::WorkflowsController < Api::ApiController
   def update_links
     super do |workflow|
       post_link_actions(workflow)
+
+      case relation.to_s
+      when "retired_subjects"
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/workflows_controller.rb
```diff
@@ -79,23 +90,17 @@ def field_context
...
-        end
-      when "subject_sets"
-        Array.wrap(params[:subject_sets]).each do |set_id|
-          SubjectSetStatusesCreateWorker.perform_async(set_id, workflow.id)
-        end
-        NotifySubjectSelectorOfChangeWorker.perform_async(workflow.id)
-      end
+    end
+
+    if relation_index == 0
```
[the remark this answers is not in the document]
camallen: that fails when the `relation_index` is nil (missing), leaving this as is.

app/controllers/api/v1/workflows_controller.rb
```diff
@@ -79,23 +90,17 @@ def field_context
...
   end
 
-  # This is a very good reason to move away from the api controller work we've got
-  # it's hard to hook into the right place to run actions that in theory should
-  # be easily located in specific controller actions
+  # Allways update the state of the workflow & selector services
+  # when modifying the subject sets or retired_subjects links
   def post_link_actions(workflow)
-    if workflow.set_member_subjects.exists?
+    relation_index = %w(subject_sets retired_subjects).index(relation.to_s)
```
marten: It might be me, but this reads very strangely to me. I would much prefer just a standard `case` statement, even if that then has a duplicated line for NotifySelectorOfChange in both of the cases. By the time I get to the `.zero?` (or `== 0`) we've lost all meaning of what that zero is.

camallen: updated - i was trying to be too cute.

PR #2908, opened 2018-08-21
Also write org contents to org itself

marten:
This is part of a series of PRs to eventually remove the *Content models.

This one makes it so that we write what used to go in OrgContent also to the Org directly. It has a rake data migration to backfill existing data. A followup will then be able to use the Org directly for reading.

I'll do the same in two more PRs for Project and Workflow contents.

Once nothing is read from the content models anymore, I'll do a final PR which cleans up those models (removing the tables, we'll discuss whether to keep the API - which would then read from the main models but simulate the old resources).

# Review checklist

### DIFF
--- a/app/operations/concerns/url_labels.rb
+++ b/app/operations/concerns/url_labels.rb
@@ -1,5 +1,5 @@
 module UrlLabels
-  def extract_url_labels(urls)
+  def self.extract_url_labels(urls)
     visitor = TasksVisitors::ExtractStrings.new
     visitor.visit(urls)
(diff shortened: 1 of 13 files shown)


app/operations/concerns/url_labels.rb
```diff
@@ -1,5 +1,5 @@
 module UrlLabels
-  def extract_url_labels(urls)
+  def self.extract_url_labels(urls)
```
camallen: Remove the include (projects controller, others?) as it's directly called now in `ContentFromParams`

PR #2909, opened 2018-08-21
Write project content attributes to projects directly

marten:
Describe your change here.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -62,6 +62,12 @@ class Api::V1::ProjectsController < Api::ApiController
       # transaction.
       content_attributes = primary_content_attributes(update_params)
       unless content_attributes.blank?
+        resource.display_name = content_attributes[:title] if content_attributes.key?(:title)
+        resource.description = content_attributes[:description] if content_attributes.key?(:description)
+        resource.introduction = content_attributes[:introduction] if content_attributes.key?(:introduction)
+        resource.url_labels = content_attributes[:url_labels] if content_attributes.key?(:url_labels)
+        resource.workflow_description = content_attributes[:workflow_description] if content_attributes.key?(:workflow_description)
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/projects_controller.rb
```diff
@@ -62,6 +62,12 @@ def update
       # transaction.
       content_attributes = primary_content_attributes(update_params)
       unless content_attributes.blank?
+        resource.display_name = content_attributes[:title] if content_attributes.key?(:title)
```
marten: I'm not super happy with how this is set up. But couldn't come up with something nicer that worked. Let me hear suggestions if you have m

app/controllers/api/v1/projects_controller.rb
```diff
@@ -62,6 +62,12 @@ def update
...
+        resource.description = content_attributes[:description] if content_attributes.key?(:description)
```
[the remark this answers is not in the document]
marten: Yeah the display_name vs title was the issue causing one to be weird, and I figured a bunch of lines would be nicer than a loop with one exception. But you're right, that'll get set anyway, which makes a loop easy to do here.

marten:
@camallen is it easier to follow now?

app/controllers/api/v1/projects_controller.rb
```diff
@@ -128,14 +134,15 @@ def build_resource_for_create(create_params)
 
     content_attributes = primary_content_attributes(create_params)
     create_params[:project_contents] = [ ProjectContent.new(content_attributes) ]
+    create_params[:url_labels] = content_attributes[:url_labels]
```
camallen: It may not do but does this need the associated `urls` attribute as well?

marten: No, that's updated in place in the original params hash (which is horrible but out of scope unless you want me to clean it up)

PR #2925, opened 2018-08-31
Contents read from model

marten:
This moves the read operations away from the content models. This should save some DB queries. Still writing to both the content model and the main one directly, which leaves us the option of rolling back this PR easily.

Note that Workflow still does have to load the content model from the DB in order to determine the current version. Next up will be swinging back to the versioning system in order to decouple that (that work was what prompted me to first do some PRs to remove the use of content models. The final bit will have to look at both simultaneously at least for Workflows).

# Review checklist

### DIFF
--- a/spec/factories/workflows.rb
+++ b/spec/factories/workflows.rb
@@ -57,6 +57,23 @@ FactoryBot.define do
     retired_set_member_subjects_count 0
     subject_selection_strategy "builtin"
 
+    strings({
+            "interest.question" => "Draw a circle",
+            "interest.help" => "Duh?",
+            "interest.tools.0.label" => "Red",
+            "interest.tools.1.label" => "Green",
+            "interest.tools.2.label" => "Blue",
+            "interest.tools.3.label" => "Purple",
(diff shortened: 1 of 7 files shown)


spec/factories/workflows.rb
```diff
@@ -57,6 +57,23 @@
     retired_set_member_subjects_count 0
     subject_selection_strategy "builtin"
 
+    strings({
```
zwolf: These strings don't appear to be included in an update to the workflow contents. Other strings are scoped to a particular trait/task type. Are these meant to be the default?

marten: Yeah these were the ones that were in the default factory for workflow_content, so it used to have these via the primary content.

zwolf: So do you still need the workflow contents factory of can that file be deleted?

camallen:
This will probably need to fold in the changes from #2934

marten:
I wanted to leave the content method in there for a little bit, which references those constants.

PR #2934, opened 2018-09-11
Fixes serialization of org URLs

zwolf:
Organization URLs weren't being properly translated via the task visitor prior to serialization.

# Review checklist

### DIFF
--- a/app/serializers/organization_serializer.rb
+++ b/app/serializers/organization_serializer.rb
@@ -1,5 +1,3 @@
-require "tasks_visitors/inject_strings"
-
 class OrganizationSerializer
   include RestPack::Serializer
   include OwnerLinkSerializer
--- a/app/serializers/organization_serializer.rb
+++ b/app/serializers/organization_serializer.rb
@@ -1,10 +1,12 @@
+require "tasks_visitors/inject_strings"
+
 class OrganizationSerializer
   include RestPack::Serializer
   include OwnerLinkSerializer
   include MediaLinksSerializer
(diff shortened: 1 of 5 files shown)


app/serializers/organization_serializer.rb
```diff
@@ -1,10 +1,12 @@
+require "tasks_visitors/inject_strings"
```
camallen: Do we really need this require? Rails constant autoloading should find this..

spec/operations/organizations/update_spec.rb
```diff
@@ -18,7 +18,7 @@
 
   let(:operation) { described_class.with(api_user: api_user, id: organization.id.to_s) }
 
-  it 'sets the urls, splitting out translatable content' do
+  it 'sets the content' do
```
camallen: not sure the new spec name makes more sense. The labels should be generic, translatable names that can be replaced with translations.

zwolf: This is checking that it's setting the untranslated string (`0.label`), not the translated one. Checking that it was properly serializing the translation is what the broken `#urls` spec was supposed to be doing before this PR fixes.

So you're right, it's not accurate since the URLs field it's confirming was updated isn't in contents. But it was also wrong before, I'll update it with something else.

app/serializers/organization_serializer.rb
zwolf: All the org and project serializer specs pass without it, so I guess not.

PR #2935, opened 2018-09-12
Update the urls to contain google analytic codes

RonaCostello:
utm_source: 'dormant'

utm_campaign: depends is the email is for a user with a last user
project preference, so 'classification' or no last upp so 'no_classification'

utm_content: depends on which url was followed. 'a' is the all projects
page, 'b' is a specific project link, 'c' is the talk page link.

Describe your change here.

# Review checklist


mrniaboc:
I think the utm_source should be the source that the link is in (in this case that would be 'Newsletter').

So you should add the dormant line to the utm_campaign e.g. 'utm_campaign=dormant_class' or 'utm_campaign=dormant_no_class'

PR #2936, opened 2018-09-13
update google analytic codes for dormant user mailer

RonaCostello:
Describe your change here.

# Review checklist


RonaCostello:
I wasn't sure if it was better to expand the explanation of the campaign code or keep the url shorter. I don't know how many people would know that the link is to do with GA. But yeah a longer description is probably more transparent for those that do.

app/views/dormant_user_mailer/email_dormant_user.text.erb
```diff
@@ -19,9 +19,9 @@ Our projects still need you!
 
 Whether it is helping to classify galaxies, map orangutan nesting sites, identify antibiotic resistant bacteria or transcribe anti-slavery manuscripts, there are many exciting research projects to get involved with in the Zooniverse.
 
-Check out the complete suite of projects here: https://zooniverse.org/projects?utm_source=dormant&utm_campaign=no_classification&utm_content=a.
+Check out the complete suite of projects here: https://zooniverse.org/projects?utm_source=Newsletter&utm_campaign=dormant_no_classificationsifications&utm_content=a.
```
camallen: spelling error will break the tracking here. maybe look to extract these into mailer vars to avoid these typos and interpolate those vars in the email text. 

``` ruby
ga_prefix = "?utm_source=Newsletter&tm_campaign="
@dormant_with_classification_ga_code = "#{ga_prefix}dormant_with_classifications"
@dormant_with_classification_ga_code = "#{ga_prefix}dormant_no_classification"
```

RonaCostello: good idea

RonaCostello: also nice spot

PR #2959, opened 2018-09-25
[WIP] remove classifications created_at and gold_standard indexes

camallen:
Drop the classifications created_at & gold_standard indexes. 

Gold standard index has 0 scans as it's always combined with more restrictive index lookups and ends up as a row filter in the query planner. 

Query plans from the query in the below counter shows that the classifications_subjects join index lookup combined with the classifications id index are used with row filters on the other attributes (including created_at). This index isn't used in our query paths and can go. https://github.com/zooniverse/Panoptes/blob/1943b27116c1eb1c870a8cc2ef463d2d1edafb4a/app/counters/subject_workflow_counter.rb#L19

The only caveat with this new query path is that if the `workflow.activity_classification_id` drifts to far into the past away from the period of interest (last 24.hours by default), think paused / finished projects that don't have any new activity. These queries then become a possibly large index scan with row lookups (basically a seq scan from a later starting point). 

The only way to combat this would be to record the last classification id for a workflow to limit the search space on the id index. However this will still result in long running queries for very active projects (SGL, etc). 

## TODO
- [ ] Test if using the project_id & id composite index helps address the old projects index search. In theory this index should have a small number of rows to search for old, non-active projects and could address the less restrictive searches.

# Review checklist

### DIFF
--- a/app/workers/calculate_project_activity_worker.rb
+++ b/app/workers/calculate_project_activity_worker.rb
@@ -3,13 +3,17 @@ class CalculateProjectActivityWorker
 
   sidekiq_options queue: :data_medium
 
-  def perform(project_id)
+  def perform(project_id, period=nil)
     project = Project.find(project_id)
     Project.transaction do
       project_activity = 0
       project.workflows.each do |workflow|
-        activity_count = workflow_activity(workflow)
-        workflow.update_columns activity: activity_count
+        activity_period = WorkflowActivityPeriod.new(workflow, period)
+        activity_count = activity_period.count
+        workflow.update_columns(
+          activity: activity_count,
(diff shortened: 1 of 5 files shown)


camallen:
looks like this query still needs this index - can it be reword to use another index? classification id?
https://github.com/zooniverse/Panoptes/blob/1943b27116c1eb1c870a8cc2ef463d2d1edafb4a/app/workers/calculate_project_activity_worker.rb#L22

marten:
Are we even still querying on the gold_standard column? I thought that use-case got moved onto the GoldStandardAnnotation table+model

eatyourgreens:
Gold standard has been broken in the front-end for months. I'm not sure if anyone is using it. zooniverse/Panoptes-Front-End#4564

Issue #2708, opened 2018-03-19
Missing flipper features for a brand new install


camallen:
Some behaviour like subject uploading relies on flipper being setup and enabled `Panoptes.flipper[:subject_uploading].enabled?`. These features need to be pre-seeded or 'flipped' around to allow missing features to work.

This was initially reported in contact@ email from a user running their own system.

camallen:
closed by #2962

PR #2961, opened 2018-10-01
Track workflow versions without Papertrail

marten:
This will start keeping track of the workflow version in a way that's independent from the Papertrail calculation. It's in a new column so that it can be done while keeping the old method online for now. This will allow us to check that after a while of having this running and tracking, the version numbers are exactly the same as the papertrail ones. After that, a PR will remove the old style and get the serializer to use these new columns.

# Review checklist

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,6 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+
+
+FactoryBot.create :full_project
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -58,6 +58,21 @@ class Workflow < ActiveRecord::Base
     %i(display_name strings)
   end
 
+  before_save :update_version
+
+  def update_version
+    self.major_version ||= 0
+    self.minor_version ||= 0
+
+    if changes.include? :tasks
+      self.major_version += 1
+    end
+
+    if changes.include? :strings
+      self.minor_version += 1
+    end
+  end
+
   # select a workflow without any json attributes (some can be very large)
   # this can be used generally in most workers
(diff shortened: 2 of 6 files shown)


db/seeds.rb
```diff
@@ -5,3 +5,5 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+
+FactoryBot.create :full_project
```
camallen: Do you want to have this in here and linked to this PR?

marten: I needed some data to test the rake task. Figured I'd leave this in to spark the discussion about how we feel about building up some seed data. How do you feel about it?

marten:
Yes, I want to give those defaults and not-nulls later.

marten:
Although I might as well do that now. I was initially leaving them out because I wanted the NULLs to see which ones needed backfilling, but if someone edits a workflow before backfilling the auto-generated number is going to be wrong anyway, so the backfill just ended up needing to overwrite everything. Which means that I should just set the defaults/notnull now rather than postponing. :+1:

db/seeds.rb
camallen: seed data sounds good, fancy tackling https://github.com/zooniverse/Panoptes/issues/2708 as well.

marten:
Happy to. Were you wanting to run these on production too then, or is that also just an issue on dev setups?

db/seeds.rb
marten: Happy to. Were you wanting to run these on production too then, or is that also just an issue on dev setups?

camallen: No need on production. The seed tasks should be for smoke testing dev / new installs. Ala the configure scripts for setting up yamls / jwt keys, etc.

marten: :+1:

PR #2963, opened 2018-10-02
Add steps as a JSON property to workflows

### LINKED ISSUE #2933: Add steps to the workflow resource
For the new classifier, we have agreed to move forward with the idea of workflow steps to help a few issues we've run into with the current structure. This is documented [ADR-5](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-5.md) with some of the lead up discussion in zooniverse/front-end-monorepo#123. 

To this end, we'd like to add `steps` property to the workflow resource that can be set as an object and defaults to null. The plan is to set steps as a ES Map, but JSON spec currently doesn't directly support maps, so the workaround is to use a [JSON compliant javascript helper library](https://github.com/sonnyp/JSON8/tree/master/packages/json8#map-and-set) to serialize as an object to store it in the database and then parse it back to a map.

Even though it's not directly supported by JSON spec, it's still advantageous for us to use maps as they are ordered and keyed and are observable in MobX, so it's a natural fit for the data layer library we want to use.

marten:
Fixes #2933

Describe your change here.

### DIFF
--- a/db/migrate/20181002145749_add_workflow_steps.rb
+++ b/db/migrate/20181002145749_add_workflow_steps.rb
@@ -1,0 +1,5 @@
+class AddWorkflowSteps < ActiveRecord::Migration
+  def change
+    add_column :workflows, :steps, :jsonb
+  end
+end
--- a/db/migrate/20181002145749_add_workflow_steps.rb
+++ b/db/migrate/20181002145749_add_workflow_steps.rb
@@ -1,5 +1,14 @@
 class AddWorkflowSteps < ActiveRecord::Migration
   def change
     add_column :workflows, :steps, :jsonb
+
+
+    reversible do |dir|
+      dir.up do
+        change_column_default(:workflows, :steps, {})
+        Workflow.update_all(steps: {})
+        change_column_null :workflows, :steps, false
+      end
+    end
   end
 end
--- a/db/migrate/20181002145749_add_workflow_steps.rb
+++ b/db/migrate/20181002145749_add_workflow_steps.rb
@@ -2,7 +2,6 @@ class AddWorkflowSteps < ActiveRecord::Migration
   def change
     add_column :workflows, :steps, :jsonb
 
-
     reversible do |dir|
       dir.up do
         change_column_default(:workflows, :steps, {})
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -41,7 +41,7 @@ class Workflow < ActiveRecord::Base
     'options' => {'count' => 15}
   }.freeze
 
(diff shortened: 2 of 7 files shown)


db/migrate/20181002145749_add_workflow_steps.rb
```diff
@@ -0,0 +1,5 @@
+class AddWorkflowSteps < ActiveRecord::Migration
+  def change
+    add_column :workflows, :steps, :jsonb
```
camallen: I think this should have a default, to avoid serializing null. From braintree: https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/
```
Add column, add default as a separate command, and backfill the column with the default value
```

marten: #2933 specifically requested that it default to NULL. @srallen what was the reason for that?

marten: I think personally I'd rather also add a classifier_version attribute.

marten:
I want to hang off on merging this until I'm satisfied concerns with Map / JSON serialization are all addressed.

db/migrate/20181002145749_add_workflow_steps.rb
srallen: Ok. That's alright with me. The default value for the `classifier_version` could be `1.0` and if `steps` are defined, it is then changed to `2.0`.

srallen:
Would serializing as an array of pairs, i.e. `[['S0', { tasks: [] }]]` with a default of an empty array for `steps` work?

srallen:
@marten I've written a function to do the conversion of a JS Map to an array in this PR here: https://github.com/zooniverse/front-end-monorepo/blob/c9e89b100c32e3128507072fb7362a7f84c32a36/packages/lib-classifier/src/store/utils/convertMapToArray.js

From what I've researched, the standard is to typically convert maps to objects since a map is a kind of object in javascript. Both the library I initially looked at does this as well as a helper function built into MobX itself, however we don't want an object in the case of workflow steps since we want to preserve order when serializing to JSON. The function I wrote will convert either to a flat array of the map's values or into an array of pairs. For workflow steps we will use an array of pairs. If this is sufficient, then I'll update the ADR on the monorepo documenting this.

marten:
Cool. I'll update this to NOT NULL and add the version column and then we can maybe merge this tomorrow.

srallen:
@marten any chance this can be finished and merged soon? Thanks!

camallen:
This has been approved but needs a rebase - are we still planning to push this out?

srallen:
Looks like `classifier_version` didn't get added with this. Could that get added so we have an explicit way to tell when `steps` should be used? Thanks!

PR #2972, opened 2018-10-18
Serialize empty associations as an empty array

### LINKED ISSUE #2902: Return empty list when there are no existing links?
I'm adding a class to the Python client to handle management of links (see zooniverse/panoptes-python-client#181). This class will be used automatically for any links that are returned as a list of IDs rather than a single ID, and will allow simple code like `Project.find(1234).links.workflows.add(5678)` to link objects together.

I'm running into a problem, though, if there aren't any links already, because the Python client doesn't know anything about what attributes/links each type of object has. I'd rather not have to hard code in which classes of objects are linked to each other, since that will require keeping the client and the API code updated together.

Is it feasible for the API to just return something like this?

```json
{
  ...
  "projects": [
    {
        ...
        "links": {
          "workflows": [],
        }
    }
  ]
}
```

I know there's also a top-level `links` key in the response, which should list all the linked classes, but it doesn't give any hint about which ones are one-to-one and which are (one|many)-to-many. Another option could be to add a type field to there.

marten:
Fixes #2902 

Empty associations would be ignored by Restpack, and not serialized in the links at all. This change overrides the method in Restpack doing so, such that these will appear in the links as an empty array.
 
# Review checklist

### DIFF
--- a/lib/serialization/panoptes_restpack.rb
+++ b/lib/serialization/panoptes_restpack.rb
@@ -4,6 +4,7 @@ module Serialization
 
     included do
       include RestPack::Serializer
+      include InstanceMethodOverrides
       extend ClassMethodOverrides
     end
 
@@ -32,11 +33,8 @@ module Serialization
       end
 
       def paging_scope(params, scope)
-        if params[:include]
-          param_preloads = params[:include].split(',').map(&:to_sym) & self.can_includes
-        end
+        preload_relations = preloads | param_preloads(params)
 
-        preload_relations = preloads | Array.wrap(param_preloads)
         unless preload_relations.empty?
           scope = scope.preload(*preload_relations)
         end
@@ -45,6 +43,11 @@ module Serialization
       end
 
       private
+
+      def param_preloads(params)
+        return [] unless params[:include]
+        Array.wrap(params[:include].split(',').map(&:to_sym) & self.can_includes)
+      end
 
       def page_href(page, options)
         return nil unless page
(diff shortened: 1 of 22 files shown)


lib/serialization/panoptes_restpack.rb
```diff
@@ -80,5 +83,52 @@ def serialize_meta(page, options)
...
+                        else
+                          if model.send(association.name).loaded?
+                            model.send(association.name).collect { |associated| associated.id.to_s }
+                          else
+                            model.send(association.name).pluck(:id).map(&:to_s)
+                          end
+                        end
+
+          # PATCH: Compared to the original, deleted an `if links_value.present?`
+          # clause around this next line:
```
camallen: can this include a code link to the version if was diffed against?

marten:
>Can we leave the project serializer out of this? It needs to be reworked and tested to ensure the overridden page scopes (card context) and preloads (e.g. owner) work properly.

We can, but how could I have known about those concerns? We need to do something about discoverability of this, because we'll just run into it next time, and if you're not there to voice them we would've just merged this?

camallen:
>We can, but how could I have known about those concerns?
Apart from running the specs and reading the source and seeing the customized context / preloading I'm not sure. https://github.com/zooniverse/Panoptes/blob/22d5cb61497b2de612a7085c5ddbebb204fb3225/app/serializers/project_serializer.rb#L57

>We need to do something about discoverability of this, because we'll just run into it next time, and if you're not there to voice them we would've just merged this?

Updating the tests on the serializers was also part of the migration plan too. Seems these still work so we can have some confidence the behaviour is ok - https://github.com/zooniverse/Panoptes/blob/22d5cb61497b2de612a7085c5ddbebb204fb3225/spec/serializers/project_serializer_spec.rb#L22.
That said moving to the `Serialization::PanoptesRestpack` module means we can migrate to the preloads DSL or keep it as is if needed. I wasn't sure what the addition of `param_preloads(params)` to the Project serializer does for the owner relation and how that impacts on preloading. It should work out of the box as it was and the specs suggest it still is though the change has been introduced.

camallen:
I'm going to work on this so we can get this out.

marten:
Thanks

marten:
My gut feeling says that an empty string would be worse than a null. JS clients are far more likely to guard against null/undefined than against empty string, since there isn't the `.present?` method like Rails provides to check against both at the same time.

The Ruby client will be perfectly fine with this, and @adammcmaster specifically requested this for the Python client so I'm guessing that won't be a problem either.

eatyourgreens:
I agree with @marten. An empty string would be parsed as a linked ID in JavaScript, leading to weird bugs when that empty string is appended to endpoint URLs.

JavaScript clients should be fine as long as the null relation parses as `null` and not `'null'`.

spec/serializers/project_serializer_spec.rb
```diff
@@ -14,7 +14,7 @@
   it "should preload the serialized associations ny default" do
```
marten: oh, typo here: `ny` should be `by`

camallen:
My additions to this PR
- ensure the project serializer was still preloading all the serialized includable relations
- test the new behaviour for null / [] relation links
- cleanup the project serializer's custom behaviour

As it turns out my concerns in https://github.com/zooniverse/Panoptes/pull/2972#pullrequestreview-167354909 were already taken care of by the existing code. 

However the PR as initially proposed would not preload and `owners` link if requested by and include param, `/projects?include=owner`. The extra additions now take care of this for all serializers that using `Panoptes::Restpack` that have the owner association defined as a valid include relation.

marten:
Is there a difference between `owner` and `owners`? I see some new tests asking to include `owner` but I had to remove `owners` on the project serializer (until you fixed that).

marten:
Let's follow up on this today. I've forgotten if we still wanted changes or if you were waiting on my LGTM or whatever

PR #2973, opened 2018-10-19
Translation versions

marten:
Todo:

* optional `outdated` property in the serializer which tells the API which strings are outdated compared to the primary language
* partial updates into the string field
* extract before/after filters from model into service object?

# Review checklist


app/models/translation.rb
```diff
@@ -10,7 +13,25 @@ class Translation < ActiveRecord::Base
...
+    end
+  end
+
+  def save_version
+    if changes.include?("strings")
+      changed_keys = Translations::HashDiff.diff(*changes["strings"])
+      version = translation_versions.create!(strings: strings)
+
+      changed_keys.each do |key|
+        string_versions[key] = version.id
```
camallen: I'm not 100% sure how this version number will be used, if it's user / UI facing then using the pk sequence number from the translations table will mean version number won't be linear for the a given resource, e.g. 1, 5, 35 etc depending how many new versions get written between updates. 

An alternative could use the length of the `translation_versions` relation instead. 

Thoughts?

config/routes.rb
```diff
@@ -147,7 +147,7 @@
       json_api_resources :subject_workflow_statuses, only: [:index, :show]
 
       json_api_resources(:translations, {
-        constraints: Routes::Constraints::Translations.new,
+        constraints: Routes::Constraints::TranslationsConstraint.new,
```
camallen: Originally i felt the `Routes::Constraints::TranslationsConstraint` had the `Constraints` prefix in the module name so didn't need the extra wording as a suffix. I'm about 2/10 on this change.

app/models/translation.rb
marten: Yeah this is what I mentioned yesterday about going down the wrong path initially. This will become the version of the translated resource, set by the sync worker. The front-end will pass in the version of primary translation when it updates a foreign language translation.

config/routes.rb
marten: It confused the rails autoloader when I added the Translations::* namespace, and somehow Rails was expecting to find the `::Translations`` namespace to be defined in this file. Not sure why, but this was the simplest fix.

marten:
@camallen okay so this PR starts tracking versions for the primary language translations. It doesn't yet do any diffing or tracking for foreign languages, but i think it's probably big enough and it is in a state that we can merge. So i propose another review with a view to merging this time,

app/models/concerns/translatable.rb
```diff
@@ -12,4 +12,16 @@ def translatable_attributes
...
+
+  def translatable_language
+    if respond_to?(:primary_language)
+      primary_language
+    else
+      language
+    end
+  end
+
+  def primary_language_translation
```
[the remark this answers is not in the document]
marten: Good catch, not anymore...

marten:
Sorry, yeah every time I run into a merge conflict my editor strips out the trailing whitespace in the structure file. Should've regenerated it instead, will do

PR #2980, opened 2018-10-24
Translation update string versions

marten:
This makes the string_versions editable through the API, but checks that the referenced version numbers make sense.

# Review checklist

### DIFF
--- a/app/models/translation.rb
+++ b/app/models/translation.rb
@@ -1,14 +1,15 @@
 class Translation < ActiveRecord::Base
   belongs_to :translated, polymorphic: true, required: true
+
   validate :validate_translated_is_translatable
   validate :validate_strings
   validates_presence_of :language
-
-  before_validation :downcase_language, on: :create
-
+  validate :validate_string_versions
   validates_uniqueness_of :language,
     scope: %i(translated_type translated_id),
     message: "translation already exists for this resource"
+
+  before_validation :downcase_language, on: :create
 
   # TODO: Versioning and maintaining a live, published version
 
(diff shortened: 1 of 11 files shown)


app/models/translation.rb
```diff
@@ -54,6 +55,26 @@ def validate_translated_is_translatable
     end
   end
 
+  def validate_string_versions
+    unless string_versions.is_a?(Hash)
```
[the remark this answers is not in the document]
marten: Go home Hound, you're drunk

app/models/translation.rb
```diff
@@ -1,15 +1,16 @@
 class Translation < ActiveRecord::Base
   belongs_to :translated, polymorphic: true, required: true
+
```
marten: I reshuffled this because not all the `validate` calls were grouped together.

app/models/translation.rb
```diff
@@ -54,6 +55,26 @@ def validate_translated_is_translatable
     end
   end
 
+  def validate_string_versions
+    unless string_versions.is_a?(Hash)
+      errors.add(:string_versions, "must be present but can be empty")
+      return
```
camallen: is this return needed? If it's a hash it'll return nil as well via implicit return in ruby.

marten: If it's not a hash the rest of the validation would raise an exception when trying to use it as a hash.

marten:
> Question - Are StringVersions objects allowed to be set via the API? I assumed they would be created / updated by the internal code and read only via the API...

Yeah that's what I'd been saying previously, but it has the following issue:

* Translator opens Pandora and starts translating eg a project page.
* Project owner opens PFE and edits the project page, and saves those changes.
* Translator finishes translating the version and saves their translation

If upon saving a translation we just marked them as for the latest version in the DB, as we'd discussed IRL earlier, the translation would now be marked up-to-date. So I think it'll be better to have Pandora read the version from the primary translation for the string being edited, and send that back along with the translation strings upon saving. That way upon saving it'll be saved as up-to the version that was shown to the translator, and would still be marked as out-of-date.

Note that ETags don't help in this case, since the translator is saving the foreign translation, and only the primary translation's ETag would have changed.

PR #2991, opened 2018-10-30
Sync only when strings change

marten:
Solves failing sync translations jobs, by not enqueuing them when they would fail.

# Review checklist

### DIFF
--- a/app/controllers/concerns/sync_resource_translation_strings.rb
+++ b/app/controllers/concerns/sync_resource_translation_strings.rb
@@ -18,8 +18,11 @@ module SyncResourceTranslationStrings
   end
 
   def translatable_resources
-    if action_name == "create"
+    case action_name
+    when "create"
       created_resources
+    when "update"
+      updated_resources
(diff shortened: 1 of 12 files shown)


app/controllers/concerns/sync_resource_translation_strings.rb
```diff
@@ -9,17 +9,24 @@ module SyncResourceTranslationStrings
...
   def sync_translatable_resource_strings
     translatable_resources.each do |resource|
-      TranslationSyncWorker.perform_async(
-        resource.class.name,
-        resource.id,
-        resource.translatable_language
-      )
+      # NOTE: In Rails 5 this method is deprecated, and we should use
+      # resource.saved_changes instead.
+      if (resource.previous_changes.keys & resource.class.translatable_attributes.map(&:to_s)).present?
```
camallen: super minor comment - perhaps extract this to a method with name `translations_have_changed?` for readability?

PR #2996, opened 2018-11-08
[WIP] Backfill workflow versions

marten:
Describe your change here.

# Review checklist


marten:
How in the heck did I break that spec that Travis is failing on?

marten:
I've fixed my understanding. Reify returns the version as it was before the change that the version represents. This means that `versions[0].reify` aka the create event return nil, because before the create, there was nothing.

`versions[1].reify` returns the version as it was at creation. This was confusing to me, since `versions[1].object_changes` actually contains the changes for the update, not the create...

All this happens to work correctly with the exporter, because our version numbers start at 1 rather than 0, and we look them up directly like that. So a classification for version "1.1" will get the `versions[1].reify` data, which is the record as it was first created (which in my mind is the first version aka the 0th index, but that's how I got confused).

So now I'm just taking `versions[1..-1]` and then adding the current model itself to that list, which means we backfill them correctly now. I've also added better more specific tests.

PR #3017, opened 2018-11-29
Export using workflow versions

marten:
Do not merge until we've backfilled production.

# Review checklist

### DIFF
--- a/spec/factories/classifications.rb
+++ b/spec/factories/classifications.rb
@@ -15,7 +15,7 @@ FactoryBot.define do
     user
     project
     workflow
-    workflow_version { "#{workflow.major_version}.#{workflow.minor_version}" }
+    workflow_version { workflow ? "#{workflow.major_version}.#{workflow.minor_version}" : nil }
     subject_ids do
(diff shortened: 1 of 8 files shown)


camallen:
a first pass on this look ok - specs are failing though i haven't looked into why.

marten:
Ah a test that factories up an invalid classification without a workflow now crashes because the factory can't get the workflow version without a workflow... :P Will fix

spec/factories/classifications.rb
```diff
@@ -11,11 +11,11 @@
                  {task: "another_one",
                   value: [1, 2]}]
     user_ip "192.168.0.1"
-    workflow_version "15.15"
     completed true
     user
     project
     workflow
+    workflow_version { workflow ? "#{workflow.major_version}.#{workflow.minor_version}" : nil }
```
camallen: Does this inherit from the workflow association on the line above? for some reason i assume this had to happen in the after build hooks, nice to know not.

marten: Yep, this works fine and even you pass in a specific workflow (as evidenced by the test failing on `create :classification, workflow: nil`)

PR #3019, opened 2018-12-04
Increment major version if first_task changes

marten:
This field is tracked by the new versioning, and thus should also be in either major or minor version numbering calculation. 

It wasn't included in Papertrail, but that seems like an oversight. When adding WorkflowVersion, I fixed that without realising the implication of not adding it to the version numbering. 

# Review checklist


camallen:
OOI - do we really need to track this? Does it impact on the resulting data or just the ordering of it?

PR #3021, opened 2018-12-07
more robust dropdown annotation conversion

camallen:
handle strange dropdown select payloads found in real exports - fixes https://app.honeybadger.io/projects/40595/faults/36814777

# Review checklist

### DIFF
--- a/spec/lib/formatter/csv/annotation_for_csv_spec.rb
+++ b/spec/lib/formatter/csv/annotation_for_csv_spec.rb
@@ -264,6 +264,28 @@ RSpec.describe Formatter::Csv::AnnotationForCsv do
           formatted = described_class.new(dd_classification, dd_classification.annotations[0], dd_cache).to_h
           expect(formatted).to eq(codex)
         end
+
+        context "with a empty array value (found in production db)" do
+          let(:annotation) do
+            {"task"=>"T7", "value"=>[]}
+          end
+
+          let(:codex) do
+            {
+              "task"=>"T7",
+              "value"=> [
+                {"select_label"=>"Country"},
+                {"select_label"=>"State", "option"=>false, "value"=>nil},
+                {"select_label"=>"City", "option"=>false, "value"=>nil}
+              ]
+            }
+          end
+
+          it 'should add the correct answer labels' do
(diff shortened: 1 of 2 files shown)


spec/lib/formatter/csv/annotation_for_csv_spec.rb
```diff
@@ -233,14 +233,7 @@
...
         let(:annotation) do
-          {
-            "task"=>"T7",
-            "value"=>[
-              {"value"=>"c6e0d98477ec8", "option"=>true},
-              {"value"=>"fb39ba165bfd4", "option"=>true},
-              {"value"=>"something unlisted", "option"=>false}
-            ]
-           }
+          {"task"=>"T7", "value"=>[]}
```
marten: I don't understand this change. I guess we were missing the test for a blank dropdown annotation? But don't we also want to keep testing a filled-out dropdown annotation?

camallen: I don't really understand how this data got into our system (possibly the borked workflows bug) but that is an actual value for that workflow in our db. 

There are still tests for filled out dropdown annotations, this change is specifically to handle the bug that i was seeing with unexpected missing annotation values.

marten: I might be mistaken, but I thought that uses the annotation you're clearing on the line this comment is attached to, through:
https://github.com/zooniverse/Panoptes/blob/e9a27b9666d746705e495fb686023058c0971829/spec/lib/formatter/csv/annotation_for_csv_spec.rb#L247

marten: I think you just need to add a second test for the `it 'works with bad annotations as found in production db'` which uses the empty annotations, while leaving the existing annotations and test alone?

camallen: thanks for catching this - i completely forgot to add a new it block from my debugging. Absolutely my f*** up

marten: No probs.

PR #3024, opened 2018-12-13
Ensure sync works when translated resource doesnt have versions

marten:
Fixes https://app.honeybadger.io/projects/40595/faults/40428061

It wouldn't be able to reference the `.id` in case no WorkflowVersion existed yet (not yet backfilled). This makes it more robust against that case.

# Review checklist

### DIFF
--- a/app/models/concerns/versioning.rb
+++ b/app/models/concerns/versioning.rb
@@ -41,7 +41,8 @@ module Versioning
   end
 
   def latest_version_id
-    send(self.class.versioned_association).order(id: :desc).select(:id).first.id
+    latest_version = send(self.class.versioned_association).order(id: :desc).select(:id).first
+    latest_version&.id
   end
 
   def version_ids
--- a/spec/workers/translation_sync_worker_spec.rb
+++ b/spec/workers/translation_sync_worker_spec.rb
@@ -29,6 +29,21 @@ RSpec.describe TranslationSyncWorker do
           translation.reload.strings["title"]
         }.from(old_title).to(new_title)
       end
+
+      it 'updates strings if workflow has no versions yet' do
+        workflow = create :workflow
+        translation = create :translation, translated: workflow
+        old_string = translation.strings["tasks.shape.question"]
+        workflow.primary_content.strings["shape.question"] = "asdf"
+        workflow.save!
+        workflow.workflow_versions.delete_all
+
+        expect {
+          worker.perform(workflow.class.to_s, workflow.id, workflow.primary_language)
+        }.to change {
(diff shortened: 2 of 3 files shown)


app/models/concerns/versioning.rb
```diff
@@ -41,7 +41,8 @@ def save_version
   end
 
   def latest_version_id
-    send(self.class.versioned_association).order(id: :desc).select(:id).first.id
+    latest_version = send(self.class.versioned_association).order(id: :desc).select(:id).first
```
camallen: Will this just stick nil into the db when it's expecting a version id value?
https://github.com/zooniverse/Panoptes/blob/7c8b9f28b4a66406e196c7a3ab1548eb7bd2dd7e/app/models/translation.rb#L28

spec/workers/translation_sync_worker_spec.rb
```diff
@@ -29,6 +29,21 @@
           translation.reload.strings["title"]
         }.from(old_title).to(new_title)
       end
+
+      it 'updates strings if workflow has no versions yet' do
+        workflow = create :workflow
+        translation = create :translation, translated: workflow
+        old_string = translation.strings["tasks.shape.question"]
+        workflow.primary_content.strings["shape.question"] = "asdf"
```
camallen: Aren't workflow translations built from the workflow strings attribute now? Is this needed because the primary contents are clobbering the resource strings attribute in https://github.com/zooniverse/Panoptes/blob/2b1e23182adb5e9b19994d4c24d020cdc6121fb2/app/models/concerns/translation_strings.rb#L27

app/models/concerns/versioning.rb
marten: Yes. It might be a little more work for the frontend to take it into account. The flip-side is that using e.g. `0` instead of `nil` will mean that you can't use whatever number there to look up the actual Version object. Even though we have no plans to do so now, it seemed like a worse option.

spec/workers/translation_sync_worker_spec.rb
marten: Yep. We should probably stop doing that, but maybe not in this PR.

app/models/concerns/versioning.rb
camallen: So the translation `string_versions` attribute will be missing values for these particular non-backfilled models. Should we just backfill the data to avoid this inconsistency with the new schema?

camallen: Now the data is backfilled, what's the status of this?

PR #3031, opened 2018-12-17
Add projects to graphql query root

marten:
# Review checklist


app/graphql/types/api_user_type.rb
```diff
@@ -0,0 +1,6 @@
+module Types
+  class ApiUserType < GraphQL::Schema::Object
```
camallen: Any reason this isn't a BaseObject?
```suggestion
   class ApiUserType < BaseObject
```

marten:
Yes, and in all of the schema, so you also have to request `launchRequested` and the docs in Graphiql show that the field is called `launchRequested`. In the old code we already did this ourselves: `field :launchRequested, !types.Boolean, property: :launch_requested `. Now that can just be `field :launch_requested, Boolean, null: false`.

app/graphql/graphql_schema.rb
```diff
@@ -1,3 +1,4 @@
-GraphqlSchema = GraphQL::Schema.define do
-  query(QueryRoot)
+class GraphqlSchema < GraphQL::Schema
+  query(Types::QueryType)
+  # mutation(Types::MutationType)
```
camallen: do we need this comment?

app/graphql/types/mutation_type.rb
```diff
@@ -0,0 +1,10 @@
+module Types
+  class MutationType < Types::BaseObject
+    # TODO: remove me
```
camallen: is this still TODO?

app/graphql/types/project_type.rb
```diff
@@ -0,0 +1,46 @@
...
+    field :primary_language, String, null: true
+
+    field :href, String, null: true
+    field :tags, [String], null: true
+
+    field :completeness, Float, null: false
+    field :activity, Integer, null: false
+    field :researcher_quote, String, null: true
+    field :mobile_friendly, Boolean, null: false
+    # field :urls, [ProjectUrlType]
```
camallen: we need these?

app/policies/project_policy.rb
```diff
@@ -5,6 +5,13 @@ class ReadScope < Scope
     def public_scope
       scope.where(private: false)
     end
+
+    def resolve(action =nil)
+      puts action
```
camallen: do we want these debug statements wrapping a super?

PR #3050, opened 2019-01-28
[HELP NEEDED] Add all docs

marten:
It's be really nice if we could get some more hands on this to convert the rest over. I've added all the files here, it's all readable but would look nicer in the new format (see `_projects.md` for an example). Tag a line below with your name to claim it, so others don't start work on it.

**Running and previewing locally is as simple as `cd docs; docker-compose up`**

- [x] classifications
- [x] collection preferences
- [ ] collection roles
- [ ] collections
- [ ] memberships
- [ ] organizations
- [ ] project preferences
- [ ] project roles
- [ ] set member subjects
- [ ] subject sets
- [ ] subject workflow statuses
- [ ] subjects
- [ ] user groups
- [ ] workflows

# Review checklist

### DIFF
--- a/docs/source/includes/_classifications.md
+++ b/docs/source/includes/_classifications.md
@@ -1,0 +1,222 @@
+# Classifications
+
+```json
+{
+    "classifications": [{
+        "id": 1001,
+        "created_at": "2014-08-24T22:24:32Z",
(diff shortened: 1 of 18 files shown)


lcjohnso:
Closing in favor of https://github.com/zooniverse/panoptes/pull/4212

PR #3054, opened 2019-01-30
Updated backfilling workflow versions script

marten:
Describe your change here.

# Review checklist

### DIFF
--- a/lib/tasks/backfill_workflow_versions.rb
+++ b/lib/tasks/backfill_workflow_versions.rb
@@ -1,5 +1,3 @@
-# lib/backfill_workflow_versions.rb
-
 module Tasks
   class BackfillWorkflowVersions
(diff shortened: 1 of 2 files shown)


lib/tasks/backfill_workflow_versions.rb
```diff
@@ -2,27 +2,54 @@
...
+        puts used_version
+        workflow_index, workflow_content_index = used_version.split(".").map(&:to_i)
+
+        workflow_at_version = workflow_versions[workflow_index - 1]
+        workflow_content_at_version = workflow_content_versions[workflow_content_index - 1]
+
+        backfill_version(workflow, workflow_index - 1, workflow_content_index - 1, workflow_at_version, workflow_content_at_version)
       end
+
+      # workflow_versions.each_with_index do |workflow_at_version, w_index|
```
camallen: I assume this was the old way. Maybe note that as coment and leave it around for reference.

PR #3067, opened 2019-02-06
Primary language translations only via lab

camallen:
Only allow updates to translations that are not the translated resources primary language. I.e. If my `project.primary_language == 'en'` then i should not be able to update a translation resource for that project with the 'en' language. All other language translation updates are fine. 

This applies to all users owners included as owners should be updating the translations via the originating resources to keep the string keys in sync. 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/translations_controller.rb
+++ b/app/controllers/api/v1/translations_controller.rb
@@ -12,6 +12,7 @@ class Api::V1::TranslationsController < Api::ApiController
   schema_type :json_schema
 
   before_action :downcase_language_param, except: :create
+  before_action :non_primary_language, only: :update
 
   def create
     check_polymorphic_controller_resources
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/translations_controller.rb
```diff
@@ -12,6 +12,7 @@ class Api::V1::TranslationsController < Api::ApiController
   schema_type :json_schema
 
   before_action :downcase_language_param, except: :create
+  before_action :non_primary_language, only: :update
```
camallen: Should this apply to `create` action as well? Timing race where i could create a translation with content before the sync workers run to update it, thus maybe being a gap to inject non-approved content?

The translation sync worker will overwrite the changes eventually when they run, https://github.com/zooniverse/Panoptes/blob/a392e7e0934c835d65324ef259a5115aba8b4a2b/app/workers/translation_sync_worker.rb#L10

marten: I'd say yes. there's no reason why anyone would do a create on the primary language translation themselves.

PR #3082, opened 2019-03-07
Versioning for more models

marten:
A later PR will remove the papertrail gem completely, but for this one it's needed to convert the data over.

# Review checklist


app/models/organization_page.rb
```diff
@@ -1,11 +1,13 @@
...
   include LanguageValidation
-  include Unversioned
-
-  has_paper_trail ignore: [:language]
+  include Versioning
 
   belongs_to :organization
+  has_many :organization_page_versions, dependent: :destroy
+
+  has_paper_trail ignore: [:language]
```
camallen: Can we remove this ? I don't think anyone is using it.

db/migrate/20190222121420_create_project_page_versions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateProjectPageVersions < ActiveRecord::Migration
+  def change
+    create_table :project_page_versions do |t|
+      t.references :project_page, index: true, foreign_key: true, null: false
+      t.text :title
+      t.text :content
+      t.string :url_key
+      t.timestamps null: false
```
camallen: OOI why the null false here?

app/models/organization_page.rb
marten: Yeah I'm removing all of them in a different PR.

db/migrate/20190222121420_create_project_page_versions.rb
marten: IDK, Rails is now doing that.

marten:
No model spec, but that's a reasonable idea. I'll have a go at one.

PR #3086, opened 2019-03-19
Change WF step type from object to array

zwolf:
The steps attr of workflows isn't an object, it's an array. I mean, it's still a jsonb. And sure, it's basically an object that was blown out into arrays of two elements, a string key and a value object. 

But since you can't map over JSON in JS, this is what the monorepo team needs. This updates the create and update schemas so they can be set via the API as well as changing the field's default. There are no existing workflows that use this attribute, so the migration also updates them all to the new default.

edit: fixed structure issues, nothing to see here.

# Review checklist


zwolf:
> this needs an update to the workflows factory steps def

✔️ 

> Ideally there would be a create / update action spec that tested this attribute being set correctly - seem that was missed in the initial PR.

None of the schemas are specced independently of the controllers, so I'm not sure what you mean. I'll add the attribute to the workflow controller update and create params, but generally we're not explicitly testing every attribute for validation failure.
 
> I'm not sure what the impact of the running the update to all workflows in the transaction will be, better practice is to run this in a different migration / rake task in batches, e.g. Finally we’ll backport the default value for existing data in batches. This should be done in its own migration as well. Updating in batches allows us to lock 1000 rows at a time (or whatever batch size we prefer).
Maybe have a look to see the row counts on that table before running it.

The initial PR (https://github.com/zooniverse/Panoptes/pull/2963) to add this field did the sae thing (`Workflow.update_all`), so I assume it would be fine. There are only ~9k rows in that table. Wasn't a problem then, do you think it's a problem now?

camallen:
> but generally we're not explicitly testing every attribute for validation failure.

Validation failure on the model no - but the controllers test the schemas for creating / updating.

> Wasn't a problem then, do you think it's a problem now?

I don't know, but if it does know we had this conversation ;)

zwolf:
How about splitting the difference and using `Workflow.in_batches.update_all`? That'll take 1000 at a time by default.

db/migrate/20190319192500_change_workflow_steps_default.rb
```diff
@@ -0,0 +1,12 @@
+class ChangeWorkflowStepsDefault < ActiveRecord::Migration
+  def change
+    reversible do |dir|
+      dir.up do
+        change_column_default(:workflows, :steps, [])
+        Workflow.find_each do |wf|
```
camallen: I'd probably go with batch updating vs 1 sql update for every workflow, something like 
```
Workflow.select(:id).find_in_batches do |workflows| 
  workflows_update_scope =   Workflow.where(id: workflows.pluck(:id))
  workflows_update_scope.update_all(: steps, [])
end
```

That'll issue 1 update statement for every 1000 workflows so should be much faster.

camallen:
What's the status of this PR?

db/migrate/20190319192500_change_workflow_steps_default.rb
zwolf: `find_each` does specifically uses batch processing via `find_in_batches`: https://apidock.com/rails/ActiveRecord/Batches/find_each

camallen: It does indeed - the detail I raised was the `wf.update_column(:steps, [])` issues 1 sql update statement for each record all to set a default. If you are processing say 1000 records, that's 1000K individual SQL update statements (slow). using `find_in_batches` and `update_all` would issue 1 sql update statement.

zwolf: Oh...that one.

zwolf:
Made redundant by https://github.com/zooniverse/Panoptes/pull/3113

PR #3090, opened 2019-03-21
Workflow version number fixes

marten:
This makes it so that we can export classifications when they refer to a workflow version that we don't have in the database. Best effort guess is to use the next one-higher workflow version, but if all else fails, just use the latest.

# Review checklist

### DIFF
--- a/app/serializers/workflow_serializer.rb
+++ b/app/serializers/workflow_serializer.rb
@@ -22,7 +22,7 @@ class WorkflowSerializer
   preload :subject_sets, :attached_images
 
   def version
-    "#{@model.current_version_number}.#{content_version}"
+    "#{@model.major_version}.#{content_version}"
   end
 
   def content_language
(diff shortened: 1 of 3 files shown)


app/serializers/workflow_serializer.rb
```diff
@@ -22,7 +22,7 @@ class WorkflowSerializer
   preload :subject_sets, :attached_images
 
   def version
-    "#{@model.current_version_number}.#{content_version}"
+    "#{@model.major_version}.#{content_version}"
```
camallen: Was the `current_version_number` the source of the errors as it relied on the paper tail array index version (offset by 1)?

Should the code in `CacheModelVersion` and `ModelVersion` etc be removed (another PR)?

marten: Yes, that code is slated for removal already in a different branch. This just brings forward this line of diff since we've now found it actually fixes an issue.

marten:
With the serializer change, the classifications should now always point at correct version numbers. In this particular case, the `workflow_version` of classifications was set correctly to a number given to the client by the workflow JSON. But, that field remains something that ultimately the client sends to the backend, and we don't want to throw away classifications if the client gets it wrong (i.e. we don't want to put a validation on it to make sure it points to a known version). So I opted to also make the dump workers more lenient in the face of incorrect data (which also means we don't have to fix those classifications that are currently causing exports to fail in order to get them working again -- though I would still suggest we look into fixing them).

PR #3092, opened 2019-03-22
perform_in instead of perform_async

zwolf:
Switches SubjectWorkflowStatusCreateWorkers to use `perform_in` instead of `perform_async`. The duration is just `number of elements * 4` to pad it out a bit. Perform times for each job are a chosen randomly from within that duration.

Not a genius-at-work situation over here but it's better than locking the database. Better ideas encouraged.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/set_member_subjects_controller.rb
+++ b/app/controllers/api/v1/set_member_subjects_controller.rb
@@ -12,11 +12,9 @@ class Api::V1::SetMemberSubjectsController < Api::ApiController
     super do |sms|
       set_id = sms.subject_set_id
       update_set_counts(set_id)
-      set_ids = linked_workflow_ids(set_id)
-      duration = set_ids.length * 4 # Pad times to prevent backlogs
-      set_ids.each do |workflow_id|
+      linked_workflow_ids(set_id).each do |workflow_id|
         SubjectWorkflowStatusCreateWorker.perform_in(
-          duration.seconds*rand,
+          1.minute*rand,
           sms.subject_id,
           workflow_id
         )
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/set_member_subjects_controller.rb
```diff
@@ -12,8 +12,11 @@ def create
     super do |sms|
       set_id = sms.subject_set_id
       update_set_counts(set_id)
-      linked_workflow_ids(set_id).each do |workflow_id|
-        SubjectWorkflowStatusCreateWorker.perform_async(
+      set_ids = linked_workflow_ids(set_id)
```
camallen: this is the workflow ids - so doesn't give you an idea about how many SMS records we are creating. These are going to be pretty small, normally 1 so i think these can either be left as is or converted to use something like, make it between now and 1 minute or something?

zwolf: oops, got carried away there. Made the duration somewhere within a minute on this one.

app/workers/subject_set_statuses_create_worker.rb
```diff
@@ -16,8 +16,10 @@ def perform(subject_set_id, workflow_id)
       .where(subject_set_id: subject_set_id)
       .select(:id,:subject_id)
 
+    duration = linked_subject_select_scope.length * 4
```
camallen: use count here to count in sql, if you use length here it'll load all the object to RAM to count them. That can be very bad.

zwolf: dang. nice catch, man. always learnin' over here.

camallen: i've made that mistake too many times ;)

PR #3095, opened 2019-03-27
Remove Papertrail from models

marten:
This removes Papertrail from models, and removes all routes that let you interact with Papertrail.

The Subject and CodeExperimentConfig models will no longer be versioned after this. The others are moved over to our own Versioning module.

# Review checklist


lib/tasks/migrate.rake
```diff
@@ -328,7 +328,7 @@ namespace :migrate do
     end
 
     desc "Backfill org page versions"
-    task :backfill_organization_page_versions do
```
marten: I'll pull this into a separate PR

camallen: Did this happen?

marten: Yep

PR #3096, opened 2019-03-28
Add docs for translation and workflow publishing.

marten:
Describe your change here.

# Review checklist


docs/source/includes/_translations.md
```diff
@@ -0,0 +1,178 @@
...
+## Get a single translation
+
+## Create a translation
+
+## Update a translation
+
+## Publish the latest translation
+
+```http
+POST /api/translations/123/publish HTTP/1.1
```
eatyourgreens: Noting here that https://github.com/zooniverse/json-api-client doesn't support POST to URLs like this, which is a pain.

docs/source/includes/_translations.md
```diff
@@ -0,0 +1,178 @@
...
+Accept: application/vnd.api+json; version=1
+Content-Type: application/json
+```
+
+This API will mark the latest version of this translation as the published version.
+The published version is intended to be shown to volunteers on the website,
+allowing a translator to make changes to a translation without interrupting the
+normal classification flow. The project owner or collaborators can then review
```
eatyourgreens: Project owners probably aren't going to review individual translated resources. I expect they'll preview their project in another language eg. https://www.zooniverse.org/projects/zookeeper/galaxy-zoo/?language=de then expect to approve all the translated resources at once. The project builder definitely doesn't list all translated resources for a project, with separate controls for each.

So, either PFE needs to gather all unpublished translations and send POST requests for each or we need a mechanism to send a single request that publishes all pending changes on a project and its workflows, tutorials, field guide and pages.

docs/source/includes/_translations.md
marten: What would be your preferable alternative?

eatyourgreens: I think we're going to have to update the client. My reading of the client code is you can't POST, only PUT, to a resource that's specified by ID, but I could be looking at the wrong part of the client code.

I'll have a look at how the client handles links. From memory, they use URLs like this, I think. I might need help from someone who's more familiar with the details of the API than I am.

docs/source/includes/_translations.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Publish the latest translation
+
+```http
+POST /api/translations/123/publish HTTP/1.1
+Accept: application/vnd.api+json; version=1
+Content-Type: application/json
+```
+
+This API will mark the latest version of this translation as the published version.
```
eatyourgreens: What does 'latest version of this translation' mean here if the translation's strings have mixed version numbers eg. `{  "title": 1, "description": 2 }`?

camallen: Translation string versions relate to the translated resource, i.e. a `project` has `project_versions` 1 and 2, the title attribute hasn't changed since version 1, description was updated in version 2. 

Publishing a translation locks the set of translated strings and their resource version ids to the current known values. 

E.g. with the following translation data, `project_version` 2 contains both strings, `project_version` 1 i missing the updated description.
```
string_versions = { "title": 1, "description": 2 },
strings = { "title": "An Amazing Title", "description": "Descriptive descriptions delight" }
```

Publishing that translation locks the `translation.published_version` to the above. A translator could update the translations but they won't make it to the live site till those changes are published.

eatyourgreens: I'm more confused now about how this works. So the published version can be a partially updated translation? If the translator later updates the title to version 2, is that published to a new version since the published version 2 can't be updated?

I'm assuming that the original has `string_versions: { "title": 2, "description": 2 }` but the translation is `string_versions: { "title": 2, "description": 2 }`.

camallen: > So the published version can be a partially updated translation?

Yeah - it'll contain anything the translation strings are set to from pandora, which can be an outdated / partially translated set of strings. 

> If the translator later updates the title to version 2, is that published to a new version since the published version 2 can't be updated?

The changes to title are stored on the `translation` resource and recorded in a `translation_version` as well. Translation versions don't change they are append only and contain the updated strings between translations. Published versions just point to a different version in time. 

The new strings won't be published by default. When a project owner chooses to publish the new strings, the translation resource updates it's `published_version` link to point to the latest version. 

> I'm assuming that the original has string_versions: { "title": 2, "description": 2 } but the translation is string_versions: { "title": 1, "description": 2 }

The original being the primary language? Yeah it'll always have the latest string versions. And yep the translation itself will have the older version if it hasn't been updated via the translations app.

eatyourgreens: Ok, so given a translation of a resource how do I, as a project owner, know whether or not that translation needs to be published? I've been looking at some translations on live projects but they all have `links.published_version` set to null. I'm trying to figure out what the logic is, in the front-end, for showing a Publish button to the project owner.

The other question, I guess, is how do I, as a project owner, use the API to publish all pending changes to translations in one go? Publishing each resource individually seems prone to error, in that I might update a workflow, say, but forget that I also need to publish the tutorial.

PR #3104, opened 2019-04-02
Update Registration Form with Underage Text

wgranger:
Describe your change here.
This PR adds under age conditional strings to the registration form. Initially, I need some feedback on what I'm doing wrong. The conditions and strings are set, but I can't get the virtual attributes to work with toggling to `under_age` attribute.

# Review checklist

### DIFF
--- a/app/assets/javascripts/forms.js
+++ b/app/assets/javascripts/forms.js
@@ -4,7 +4,7 @@ $(function () {
   var DEFAULT_SUBSCRIPTION_TEXT = "It’s okay to send me email every once in a while.";
 
   function changeRegistrationFormWording() {
-    if ($("#under_age").is(':checked')) {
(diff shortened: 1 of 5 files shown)


camallen:
@wgranger i've updated the rails user model with a virtual attribute to allow the form to maintain the `under_age` checkbox state on error mode round trips back from the server. I've also updated the form to modify the wording on page (re-)load if the checkbox is set as well. 

You can test this if you have a local panoptes (via docker) install with a duplicate user, they will error and reload the form with a valid error message.

I've also modified the form error styling - removed the custom `.form` nesting which should work fine across other forms (defaults the class name to the controller action).

Mind having a look at this?

wgranger:
Looks good. I'm able to show errors on duplicate users registering and the under age flag persists. This is good to merge on my end.

PR #3105, opened 2019-04-03
optimize the latest_version_id lookup

camallen:
avoid a costly order on possibly large tables to then select a subset of filtered rows.

Instead take the max id from the optimal version_ids to find the latest known version association via the auto-incrementing PK.

# Review checklist

### DIFF
--- a/app/models/concerns/versioning.rb
+++ b/app/models/concerns/versioning.rb
@@ -40,8 +40,10 @@ module Versioning
     end
   end
 
+  # take the highest associated version_id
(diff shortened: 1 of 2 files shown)


camallen:
> lgtm. Was this just an optimization that caught your eye while working on something else, or was there a slowdown you noticed that this solved?

I was trying to sync some translation resources and noticed some workers took upwards of 7 minutes for workflow syncs, i explained the query to get a look at it and realised the order was an issue on tables with lots of versions. I didn't magically see this in code...it took me a little while to see it actually :)

PR #3106, opened 2019-04-03
Cleanup contents code left on organizations

camallen:
Cleanup the left of *contents code and update the organizations controller / serializers.

1. remove content attributes for organizations controller and serializer. 
2. Remove the `has_contents` module code
3. update translation strings lookup to not references content models
4. Ensure organization params are being schema validated on create & update

This last one could be extracted to it's own PR if folks feel this is unrelated / too big to review. 

# Review checklist


zwolf:
The contents stuff looks good. Couple questions on the schema parts. I'm going to try not to conflate controller-level JSON schema validation and operation-level validation, but it seems like both are happening and we should avoid defining the same thing in two places.

The Organizations::Update action uses the update schema by requiring the OrganizationFilter, which inherits from ActiveInteraction and loads the schema and validates against it (via that `organization :organization_params` line in the update operation). Is this now also unnecessary with the addition of the schema type in the controller? If we're validating input via the JSON schema once, we probably don't need to do it again.

Similarly, the org create op has what amounts to a schema as well. If we wanna be more consistent with the rest of Panoptes validation, I think these other bits should go as well to avoid confusion. The tests pass so it looks like this still works, but if orgs make the switch to full `:json_schema` validation, the ActiveInteraction validation should be removed.

camallen:
> The Organizations::Update action uses the update schema by requiring the OrganizationFilter, which inherits from ActiveInteraction and loads the schema and validates against it (via that `organization :organization_params` line in the update operation). Is this now also unnecessary with the addition of the schema type in the controller? If we're validating input via the JSON schema once, we probably don't need to do it again.

I've remove the OrganizationFilter in the update operation. Massive thanks for the heads up on this, i did not know that's how it was working and i wondered why the update schema wasn't being used like it normally is. Nice work. 

Ok so now the create and update operations rely on the controller to validate the org's incoming params schema and the operations use these hash payloads. 

I think the only drawback here is the lack of clarity around default values in the create schema. If the client doesn't pass these in and they aren't set as required in the json schema then it'll fallback to the table defaults that look ok but something to consider in review.

PR #3113, opened 2019-04-11
Change wf steps default, take two

zwolf:
Repeat of https://github.com/zooniverse/Panoptes/pull/3086

A while back I clobbered (then unclobbered) structure.sql, which made rebasing a pain, so this was easier. I added `steps` to the workflow resource args, so they should be being tested by the update/create specs now, too.

I recreated used my test db to load the schema directly from the up-to-date structure.sql, then loaded this migration right on top to generate the new one. That rearranging is annoying but automatic.

# Review checklist


db/migrate/20190411125709_change_workflow_steps_default.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeWorkflowStepsDefault < ActiveRecord::Migration
+  def change
+    reversible do |dir|
+      dir.up do
+        change_column_default(:workflows, :steps, [])
+        Workflow.select(:id).find_in_batches do |workflows|
+          workflows_update_scope = Workflow.where(id: workflows)
```
camallen: I kind of expect this to work but was surprised to see it without a pluck on the .id attribute. 
```suggestion
          workflows_update_scope = Workflow.where(id: workflows.pluck(:id))
```

zwolf: The `workflows` block parameter that `find_in_batches` uses is an array of workflow objects, not an AR relation, so it doesn't have a pluck method.

PR #3124, opened 2019-05-07
add education_api attribute for 'hidden' workflows

camallen:
Add an education_api attribute to workflows (default is false) to avoid serializing a project's education api only workflows. 

# TODO
- [x] Determine if migration will be a manual one - I am leaning towards yes to target just Education API workflows
- [x] Allow this attribute to be set via the education API for new workflows.
- [x] Update education API to set this attribute to true for new lab workflows

# Review checklist

### DIFF
--- a/db/migrate/20190507103007_add_education_attr_to_workflows.rb
+++ b/db/migrate/20190507103007_add_education_attr_to_workflows.rb
@@ -1,5 +1,5 @@
 class AddEducationAttrToWorkflows < ActiveRecord::Migration
   def change
-    add_column :workflows, :education_api, :boolean, default: false
+    add_column :workflows, :serialize_with_project, :boolean, default: true
   end
 end
--- a/db/migrate/20190507103007_add_education_attr_to_workflows.rb
+++ b/db/migrate/20190507103007_add_education_attr_to_workflows.rb
@@ -1,0 +1,5 @@
+class AddEducationAttrToWorkflows < ActiveRecord::Migration
+  def change
+    add_column :workflows, :education_api, :boolean, default: false
+  end
+end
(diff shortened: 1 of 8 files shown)


db/migrate/20190507103007_add_education_attr_to_workflows.rb
```diff
@@ -0,0 +1,5 @@
+class AddEducationAttrToWorkflows < ActiveRecord::Migration
+  def change
+    add_column :workflows, :education_api, :boolean, default: false
```
camallen: I'm not super happy with the attribute name, ideas welcome. `education_use`, `education_only`?

zwolf: I'd go `education_only` since it amounts to a filter. Still pretty service-specific, which we don't have a lot of elsewhere, though. `hidden` or `omit` ?

adammcmaster: Could you maybe make it something more generic so it's not tied specifically to the education API? Just thinking there might be other reasons to use this in the future and having "education" in the name doesn't make it very clear what the attribute actually does.

Maybe call it `serialize` and reverse the logic, so it defaults to true?

camallen: Thanks @adammcmaster - i like the invert paradigm and naming suggestion, I'll update.

camallen:
> The two assumptions being made here are that we'll never want to make a PB-created wf into an assignment

I made it read only after the initial create to raise this issue and avoid workflows 'disappearing'. We can add in the update paths if the need arises.

> and that we never want to serialize an edu wf without it being explicitly requested. Those things both make sense to me, so lgtm.

After discussions with @srallen the classrooms use case find the workflow by id so no need to find the workflow from the project.

PR #3128, opened 2019-05-20
[WIP] Rails 5.0 

marten:
Describe your change here.

# Review checklist


Gemfile.next
```diff
@@ -0,0 +1 @@
+Gemfile
```
marten: This is apparently what symlinks look like in Github's diffs. So note, this file is a symlink to the normal gemfile, not a file with this line in it.

.travis.yml
```diff
@@ -8,9 +8,7 @@ before_install:
   - find spec -maxdepth 1 -type d | grep spec/ > tmp/curr_spec_dirs.txt
   - if grep -Fxvc -f spec/known_dirs.txt tmp/curr_spec_dirs.txt; then echo 'Detected unkown Spec directories, check the spec/known_dirs.txt!'; exit 1; fi
   - 'echo ''gem: --no-ri --no-rdoc'' > ~/.gemrc'
-
-install:
-  - bundle install --without development production --path=${BUNDLE_PATH:-vendor/bundle}
+  - if [ "$BUNDLE_GEMFILE" = "$PWD/Gemfile.next" ]; then rm -f Gemfile.lock; fi
```
marten: If BUNDLE_GEMFILE is Gemfile.next, Bundler should also be making and using Gemfile.next.lock, so i'm confused about this change.

camallen: This was super confusing to me as well, most update guides suggest running without a *.lock file but that is just silly. Anyway, turns out that the `Gemfile.next.lock` wasn't checked in and I sank way to long figuring that out :(

camallen:
closing in favour of #3374

PR #3130, opened 2019-05-23
Copy from subject metadata to SMS attrs

zwolf:
First step in implementing sequential subject selection is allowing researchers to set the priority of each subject within a set. There's probably a clever way for the worker to not have to pull the subject and SMS and push the change all separately, but at least it's in a bg worker. Happy to entertain more efficient ways to do it.

# Review checklist


app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -105,7 +105,9 @@ def add_relation(resource, relation, value)
         [ resource.id, subject_id, rand ]
       end
 
-      SetMemberSubject.import IMPORT_COLUMNS, new_sms_values, validate: false
+      result = SetMemberSubject.import IMPORT_COLUMNS, new_sms_values, validate: false
+      SubjectPriorityWorker.perform_async(result.ids)
```
camallen: Do we have an idea about how big this array could be? https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

How can we minimize the memory pressure on redis?

app/workers/subject_priority_worker.rb
```diff
@@ -0,0 +1,13 @@
+class SubjectPriorityWorker
+  include Sidekiq::Worker
+
+  def perform(sms_ids)
+    sms_ids.each do |sms_id|
```
camallen: A lot of small updates can be bad for the db, especially if it has to update indexes though this col does not have one.

Ideally this would do batch updates in 1 sql transaction instead of issuing lots of small AR save sql statements. Sadly in our rails / pg version there isn't a great way to do this.

PG 9.5 added upserts - https://wiki.postgresql.org/wiki/UPSERT

Active record import (we use this) supports this but only for PG >= 9.5 - https://github.com/zdennis/activerecord-import/#duplicate-key-update

Rails 6 natively supports this (might be worth adding a comment explaining we should replace code in future) - https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert

https://www.oipapio.com/question-751809
I'm not sure but you might be able to use `update_all` with a select statement on the linked subjects table. If you can get that working that will be the best way, e.g.
```
SetMemberSubject.update_all('priority = (SELECT subjects.metadata->>'#priority' FROM subjects WHERE set_member_subjects.subject_id = subjects.id AND set_member_subejct.id = ?)', sms_ids)
```
I'd be happy to help you work on this query too

Otherwise the sub-optimal solution is to batch up your saves and wrap them in a transaction - https://stackoverflow.com/a/19037842/2761423

I'd be super interested to see any perf improvements between the current example and the final implementation https://blog.appsignal.com/2018/02/27/benchmarking-ruby-code.html

 FWIW - we currently spam the db when creating sws records on subject set changes, e.g. https://github.com/zooniverse/Panoptes/blob/414baf509f85a569b388bd47e16122ee253217bc/app/workers/subject_set_statuses_create_worker.rb#L21 

That needs to change as well, ideally using a batchable solution to cut down the db pressure and not just relying on time spacing the operations out.

zwolf: This is great, thanks! Exactly what I was hoping to talk about. No rush on this, I mainly wanted to get a dumb version out to start this conversation. Lots of potential pain points, this seems like something to pair on sometime? Thanks for the research, I'll go read some blog posts.

zwolf:
Making a note: During refactor, include `#training_subject` from metadata in new SMS field to enable exclusion of training subjects/sims from project completeness.

lcjohnso:
Regarding #training_subject ^^ = @zwolf and I (@lcjohnso) had a good conversation about adopting the Caesar-style metadata tag for subjects that will not retire in two contexts: 1) PH-TESS like training subjects that are served along with real-data subjects in a single workflow; 2) Practice Workflows where all subjects are for training purposes only, such as used by Earthquake Detective.  In both cases, it will be good to tag and remove these subjects from completion counters.

app/controllers/api/v1/subject_sets_controller.rb
zwolf: Had a bunch written here about whether a one job with many arguments was better than many jobs with one argument when I realized that I can just send the subject set ID and pull the SMSes out again in the worker. 🎉

app/workers/subject_metadata_worker.rb
```diff
@@ -0,0 +1,24 @@
+class SubjectMetadataWorker
+  include Sidekiq::Worker
+
+  def perform(set_id)
+
+    smses = SetMemberSubject.where(subject_set_id: set_id)
```
camallen: you can remove this entirely, the `subject_set_id` is a FK relation on the `set_member_subjects` resource

app/workers/subject_metadata_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    update_sms_priority_sql = <<-SQL
+      UPDATE set_member_subjects
+      SET    priority = CAST(subjects.metadata->>'#priority' AS INTEGER)
+      FROM   subjects
+      WHERE  subjects.id = set_member_subjects.subject_id
+      AND    set_member_subjects.id IN (?)
```
camallen:
```suggestion
      AND    set_member_subjects.subject_set_id = ?
```

camallen: I assume that this only deals with 1 subject_set_id at a time, if more you'll need the IN (id1,id2) syntax.

app/workers/subject_metadata_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+    SQL
+
+    bound_update_sms_priority_sql = ActiveRecord::Base.send(
+      :replace_bind_variables,
+      update_sms_priority_sql,
+      [smses.pluck(:id)]
```
camallen:
```suggestion
      [set_id]
```

app/workers/subject_metadata_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def perform(set_id)
+
+    smses = SetMemberSubject.where(subject_set_id: set_id)
+
+    update_sms_priority_sql = <<-SQL
+      UPDATE set_member_subjects
+      SET    priority = CAST(subjects.metadata->>'#priority' AS INTEGER)
+      FROM   subjects
+      WHERE  subjects.id = set_member_subjects.subject_id
```
camallen: I think we need a where clause to ensure we only attempt to cast existing priority values as well.

zwolf:
This is ready to go. Thanks to @camallen for finding `exec_update`. Punting on other metadata updates for now to keep this PR simple, will revisit.

app/workers/subject_metadata_worker.rb
```diff
@@ -0,0 +1,20 @@
+class SubjectMetadataWorker
+  include Sidekiq::Worker
+
+  def perform(subject_set_id)
+    update_sms_priority_sql = <<-SQL
+      UPDATE set_member_subjects
+      SET    priority = CAST(subjects.metadata->>'#priority' AS INTEGER)
```
camallen:
```suggestion
      SET    priority = CAST(subjects.metadata->>'#priority' AS NUMERIC)
```

Ensure we preserve the mantissa for folks that pass in decimal numbers

PR #3138, opened 2019-06-10
add workflow complete filter

camallen:
filter out complete and incomplete via the workflow's completness attritbue

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2092,6 +2092,7 @@ Workflows are returned as an array under the _workflows_ key.
   + sort (optional, string) ... field to sort by id by default
   + project_id (optional, integer) ... filter workflows by project id
   + include (optional, string) ... comma separated list of linked resources to load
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Looks good to me too. I'm trying to think of a practical case where I might want to query on a specific numerical value (eg. workflows where completeness > 0.9) but coming up blank so I think the boolean query param is fine.

mcbouslog:
LGTM, thanks Cam!

apiary.apib
```diff
@@ -2092,6 +2092,7 @@ Workflows are returned as an array under the _workflows_ key.
   + sort (optional, string) ... field to sort by id by default
   + project_id (optional, integer) ... filter workflows by project id
   + include (optional, string) ... comma separated list of linked resources to load
+  + complete (optional, string) ... filter workflows on the completeness attribute value should be 'true' or 'false'.
```
camallen:
```suggestion
  + complete (optional, string) ... filter complete ('true') and incomplete ('false') workflows
```

PR #3139, opened 2019-06-13
Inject bind variables in rails 4 with prepared_statements set to false

camallen:
In non local envs disable prepared statements, manual SQL work relies on testing these code paths.

Allow Rails 4 to correctly inject the bind vars for the raw SQL

# Review checklist

### DIFF
--- a/app/workers/subject_metadata_worker.rb
+++ b/app/workers/subject_metadata_worker.rb
@@ -2,7 +2,7 @@ class SubjectMetadataWorker
   include Sidekiq::Worker
 
   def perform(subject_set_id)
-    if Panoptes.rails5?
+    if ActiveRecord::VERSION::MAJOR == 5
       update_sms_priority_sql = <<-SQL
         UPDATE set_member_subjects
         SET    priority = CAST(subjects.metadata->>'#priority' AS NUMERIC)
@@ -11,7 +11,6 @@ class SubjectMetadataWorker
         AND    subjects.metadata ? '#priority'
         AND    set_member_subjects.subject_set_id = $1
       SQL
-
       ActiveRecord::Base.connection.exec_update(
         update_sms_priority_sql,
         "SQL",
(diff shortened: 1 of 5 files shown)


zwolf:
As per this Blog Post We Like (https://github.blog/2018-09-28-upgrading-github-from-rails-3-2-to-5-2/), we could leave both code paths in and switch on something like `Panoptes.rails_5?`. That would let us leave the newer syntax in so we don't have to remember to come back after the upgrade merges. But that only seems useful if the backward-compatible version of this is less performant or is otherwise compromising.

I'm for bundling the Travis updates with the SMS SQL changes if the PR name and desc are updated to reflect.

camallen:
> As per this Blog Post We Like (https://github.blog/2018-09-28-upgrading-github-from-rails-3-2-to-5-2/), we could leave both code paths in and switch on something like `Panoptes.rails_5?` (nothing seems to do anything like this right now, though). That would let us leave the newer syntax in so we don't have to remember to come back after the upgrade merges. But that only seems useful if the backward-compatible version of this is less performant or is otherwise compromising.
> 
> I'm for bundling the Travis updates with the SMS SQL changes if the PR name and desc are updated to reflect.

That sounds reasonable to put both code paths in to ensure our rails 5 code works as expected. Feel free to add the switch to this PR or i'll pick it up tomorrow.

zwolf:
A quick and dirty means of determining whether Rails 5 is running isn't pretty, but it works. This at least keeps the knowledge in the codebase, keeps it searchable, and is a decent shove toward doing the upgrade. We might want a nicer solution or maybe `ten-years-rails` has a better solution; I'm still going through it.

The best way to test the other half of this would be to merge https://github.com/zooniverse/Panoptes/pull/3128/ first, then rebase and run the test in 5. Not sure if there was more work on that to do on that PR, but that'd be a good place to start. That railsconf video in the desc of the `ten-years-rails` repo is pretty good too and I'm feeling very seen

camallen:
I've tried something different with testing the Rails 5 AR interface vs the behaviour, rails 4 AR still tests behaviour. 

I think this is good to go, we can revisit the tests in light of #3128 but this bug needs to be fixed asap.

zwolf:
Agreed it needs to be fixed, so :shipit:. Revisiting the upgrade and how we want to tackle it is worth its own discussion.

Issue #2795, opened 2018-06-07
Landing page stats update - exclude never retire subjects


aprajita:
copied from PFE on @camallen's advice.

## Expected behavior
Stats should exclude any subjects set to 'never retire'

## Current behavior
included all active sets, even if they will never be retired

## Steps to replicate
visit https://www.zooniverse.org/projects/aprajita/space-warps-hsc

![image](https://user-images.githubusercontent.com/1917970/41095606-d01f6fee-6a49-11e8-80af-c38f49166db6.png)

includes 19748 training subjects so will never be 100% complete

caused a large misunderstanding during science friday week causing many users to only get the already seen flag and multiple classifications on the same images.

thanks!

marten:
@camallen was there an idea as to how to fix this? Since the workflow is actually the one set to 'never retire', for all subjects. And then it just happens that Caesar comes along and retires things, but Panoptes doesn't really know how many subjects will or will not retire that way.

Are we adding retirement limits per set?

marten:
In this particular case, their external script outputs a field "gold" which Caesar uses to ignore gold subjects. Some other projects we read from the subject metadata directly (which I think is what the SW script is also using).

camallen:
We can adopt the caesar training data setup by using the `training_subject` subject metadata to identify training data in the Panoptes API. 

This subject metadata attribute can be copied to internal records (SMS) to allow us to filter the training subjects data for completeness counts. 

If we copy that metadata to the SetMemberSubject(SMS) attribute (index?) we can filter those SMS training subjects for both the subject_set#subjects_count field 
https://github.com/zooniverse/Panoptes/blob/051a49db038b83ed0e758bc15eaf94b55675f127/app/workers/subject_set_subject_counter_worker.rb#L16

As well as the query on retired counter 

https://github.com/zooniverse/Panoptes/blob/051a49db038b83ed0e758bc15eaf94b55675f127/app/counters/workflow_counter.rb#L19

Caesar docs on marking training subjects:
https://github.com/zooniverse/caesar/blob/75ba896d7221c39a1f1209645796fc80fc5121cc/docs/source/includes/_subject_metadata.md#training_subject

lcjohnso:
[...]

PR #3140, opened 2019-06-14
ignore training subject sets for counters and selection

camallen:
towards #2795 - Ignore all subject sets that are marked as training sets on the workflow configuration when counting the workflows retired and classification counts.  These counts propagate up to project counts via the workflow resource.

And When selecting data due internally from postgres, avoid selecting from training sets.

Thanks to @amyrebecca and @eatyourgreens for the idea to use subject sets markers to determine which records are included in the counts.

~~Benchmarking the query to filter the subject sets on metadata attribute presence / absence to determine the training sets look like a simple filter on the index scan for workflow -> subject sets. I think this is ok even though it's a `NOT` query. Thoughts welcome.~~

Modified the AR association scope query to reflect on the marked configuration object ids and filter on the `subject_sets_workflows.subject_set_id` after doing it's index scan. The number of linked sets to workflows should always be a fairly small set being filtered (page fetching) on so shouldn't impact on the query time with the combined index scan.

# Review checklist

### DIFF
--- a/spec/counters/workflow_counter_spec.rb
+++ b/spec/counters/workflow_counter_spec.rb
@@ -47,8 +47,8 @@ describe WorkflowCounter do
         let(:training_set) { workflow.subject_sets.first }
         let(:real_set) { workflow.subject_sets.last }
         before do
-          new_metadata = training_set.metadata.merge("training" => true)
-          training_set.update_column(:metadata, new_metadata)
+          real_set_ar_collection_proxy = SubjectSet.where(id: real_set)
+          allow(workflow).to receive(:non_training_subject_sets).and_return(real_set_ar_collection_proxy)
         end
 
         it "should return non training data classification count only" do
@@ -94,8 +94,8 @@ describe WorkflowCounter do
         let(:training_set) { workflow.subject_sets.first }
         let(:real_set) { workflow.subject_sets.last }
         before do
-          new_metadata = training_set.metadata.merge("training" => true)
-          training_set.update_column(:metadata, new_metadata)
+          real_set_ar_collection_proxy = SubjectSet.where(id: real_set)
+          allow(workflow).to receive(:non_training_subject_sets).and_return(real_set_ar_collection_proxy)
         end
 
         it "should return non training data retired count only" do
--- a/spec/counters/workflow_counter_spec.rb
+++ b/spec/counters/workflow_counter_spec.rb
@@ -41,6 +41,20 @@ describe WorkflowCounter do
           expect(counter.classifications).to eq(2)
         end
       end
+
+      context "with subject sets marked as training data" do
+        let(:workflow) { create(:workflow_with_subjects, num_sets: 2) }
+        let(:training_set) { workflow.subject_sets.first }
+        let(:real_set) { workflow.subject_sets.last }
+        before do
+          new_metadata = training_set.metadata.merge("training" => true)
+          training_set.update_column(:metadata, new_metadata)
+        end
+
+        it "should return non training data classification count only" do
+          expect(counter.classifications).to eq(2)
+        end
+      end
     end
   end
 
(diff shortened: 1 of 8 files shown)


spec/counters/workflow_counter_spec.rb
```diff
@@ -41,6 +41,20 @@
...
+        let(:workflow) { create(:workflow_with_subjects, num_sets: 2) }
+        let(:training_set) { workflow.subject_sets.first }
+        let(:real_set) { workflow.subject_sets.last }
+        before do
+          real_set_ar_collection_proxy = SubjectSet.where(id: real_set)
+          allow(workflow).to receive(:non_training_subject_sets).and_return(real_set_ar_collection_proxy)
+        end
+
+        it "should return non training data classification count only" do
+          expect(counter.classifications).to eq(2)
```
zwolf: Is this counter using the workflow from the top of the spec or the new one with the same name you've built in this context? I assume it's not evaluated until you call #classifications so it uses the new workflow, but could you just add the training set to the existing workflow?

spec/lib/subjects/postgresql_selection_spec.rb
```diff
@@ -35,6 +34,32 @@ def update_sms_priorities
...
+              create(:set_member_subject,
+                setup_subject_workflow_statuses: true,
+                subject: subject,
+                subject_set: real_set
+              )
+            end
+          end
+
+          it "should not include training subject sets in the results" do
+            result_ids = subject.select
```
zwolf: Why do you need the enumerator over just `[subject.id]`?

spec/counters/workflow_counter_spec.rb
camallen: > Is this counter using the workflow from the top of the spec or the new one with the same name you've built in this context?

In Rspec any nested let will override the outer let clauses, so this context `let(:workflow)` overrides the root `let(:workflow)`. When we build the counter in the root it will use the nested context let. 

> I assume it's not evaluated until you call #classifications so it uses the new workflow

As it's lazy loaded, i.e. the workflow will only get built when the counter is created via it's own let clause.

> but could you just add the training set to the existing workflow?

Yep but i feel that the workflow factory is easier to read vs manually adding a set to ensuring the correct relations and avoid building extra instances (like workflow and project)
```
let(:workflow) { create(:workflow_with_subjects, num_sets: 2) }
let(:training_set) { workflow.subject_sets.first }
let(:real_set) { workflow.subject_sets.last }
## vs
let(:training_set) do 
  create(:subject_set, project: workflow.project, workflows: [workflows]) do |set|
    workflow.subject_sets << set
  end
end
let(:real_set) { workflow.non_training_subject_sets.first }
before do
  # ensure the training set is setup on the workflow before the specs run
  training_set
  ....
end
```

spec/lib/subjects/postgresql_selection_spec.rb
camallen: I'm not sure i follow, `result_ids = subject.select` runs the selection code.

zwolf: I see. The words "select" and "subject" are used in the same code in completely different ways and I mistook the RSpec `subject` for an actual subject. Reading the same words over and over in different contexts and they soon lose all meaning...

camallen: Gotcha. Yeah - i've never really liked the rspec subject syntax especially in our domain model. I can update the spec to be `let(:selector) { Subjects::PostgresqlSelection.new(workflow, user, opts) }` if you like but it'll have a bigger diff footprint.

camallen:
@zwolf mind looking over the new diff since you approved (last commit)? I'll update the fix the failing specs but that shouldn't really impact the implementation details.

zwolf:
Might be worth including documentation for this here too. LGTM.

PR #3145, opened 2019-06-27
Upgrade ruby 2.5

camallen:
Bump next ruby version 2.5 for testing and running the system. Also test ruby 2.6 which has some failures we need to upgrade on.

# Review checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,22 +17,25 @@ before_script: &before_script
 
 matrix:
   fast_finish: true
+  allow_failures:
+    - rvm: 2.6
   include:
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS=spec/controllers/api/v1/[a-m]*.rb
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS=spec/controllers/api/v1/[n-s]*.rb
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS=spec/controllers/api/v1/[t-z]*.rb
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS="spec/controllers/**.rb spec/controllers/api/*.rb spec/models spec/operations spec/counters spec/routes"
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS="spec/lib spec/workers spec/serializers spec/services spec/requests spec/middleware spec/mailers spec/policies"
+    - rvm: 2.6
 
 services:
   - postgresql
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -17,22 +19,25 @@ before_script: &before_script
...
         - SPECS=spec/controllers/api/v1/[t-z]*.rb
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS="spec/controllers/**.rb spec/controllers/api/*.rb spec/models spec/operations spec/counters spec/routes"
-    - rvm: 2.4
+    - rvm: 2.5
       env:
         - SPECS="spec/lib spec/workers spec/serializers spec/services spec/requests spec/middleware spec/mailers spec/policies"
+    - rvm: 2.6
```
camallen: ruby 2.6 and rails 4 don't play nicely together, https://github.com/rails/rails/issues/34790#issuecomment-450190024

That is won't fix so these tests will always fail until we bump to rails 5. Probably not worth testing this now until we've done the upgrade.

.travis.yml
```diff
@@ -8,6 +8,8 @@ before_install:
   - find spec -maxdepth 1 -type d | grep spec/ > tmp/curr_spec_dirs.txt
   - if grep -Fxvc -f spec/known_dirs.txt tmp/curr_spec_dirs.txt; then echo 'Detected unkown Spec directories, check the spec/known_dirs.txt!'; exit 1; fi
   - 'echo ''gem: --no-ri --no-rdoc'' > ~/.gemrc'
+  - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
+  - gem install bundler -v '< 2'
```
zwolf: The Gemfile.lock specifies a sub-2.0 version of bundler with `bundler (>= 1.3.0, < 2.0)`, along with `BUNDLED WITH 1.17.1`. With auto version switching, that should still work with ruby 2.5. Is this necessary?

camallen: You'd think so wouldn't you, the trouble is travis CI installs the latest bundler on build and not the version that this gemfile is set to use:
> Under many configurations, Travis CI installs the Ruby runtime on the fly. This means installing the latest Bundler at run time, which may cause problems due to the unsatisfied requirements.
> If you find your builds are failing due to “bundler not installed” errors, try one of the following solutions:
https://docs.travis-ci.com/user/languages/ruby/#bundler-20

And because we are using rails 4.2 we're locked to the old v1 bundler and can't upgrade till we head to rails 5, so this is the workaround till then. Happy to add a comment to the travis yml explaining this.

zwolf: I encountered this last week doing something on Caesar and found that doc, but in the end that addition wasn't necessary because ruby 2.4 is still compatible (so long as all your gems are). If 2.5 is the issue, I'll wanna revisit this on Caesar and update ruby+bundler there, too.

camallen: the issue is on ruby 2.5 the installed version of bundler is v2 but our rails gem v4.2 requires bundler v1 only. We can't upgrade our required version of bundler to v2 sadly to avoid this.

zwolf: :sadpanda: 👍

PR #3148, opened 2019-07-02
remove the SWS create worker locks

camallen:
allow all these jobs to run as they shouldn't be enqueued multiple times.

If they are let the db unique index control state vs swamping redis with lock structures

# Review checklist

### DIFF
--- a/app/workers/subject_workflow_status_create_worker.rb
+++ b/app/workers/subject_workflow_status_create_worker.rb
@@ -1,8 +1,6 @@
 class SubjectWorkflowStatusCreateWorker
   include Sidekiq::Worker
 
-  sidekiq_options lock: :until_executed
-
   def perform(subject_id, workflow_id)
     return unless Panoptes.flipper[:subject_workflow_status_create_worker].enabled?
 


camallen:
> So this just moves any potential locking from the job (via unique args) in redis to `SubjectWorkflowStatus.create` postgres? I wasn't around for whatever problem this is solving by lgtm on principal.

If there is a large spike in the number of these enqueued in redis, i.e. large subject set linkage / upload of lots of data happens, then these add lock structures to the redis store. We had ~7million enqueued jobs yesterday morning (not all these jobs but lots were) that caused a spike in redis db size. That in turn caused errors when redis tried to flush the db to disk for persistence, eventually filling the disk. https://redis.io/topics/persistence#how-it-works

Reducing the number of locks keeps the redis db size small, fast to persist and easy to manage.

camallen:
> So this just moves any potential locking from the job (via unique args) in redis to SubjectWorkflowStatus.create postgres?

Instead we check if the SWS exists via index query (fast) and then attempt to make it via the create! method. That is backed by a uniq index so will fail if we have duplicate jobs running at the same time. 

So yeah, a little more db pressure in postgres vs redis if that duplicate case happens (and we should see those non-uniq index errors report).

PR #3150, opened 2019-07-08
extend designator client with add_seen method

camallen:
Depends on https://github.com/zooniverse/designator/pull/86

This will allow a user's seen subjects for a workflow to sync between the panoptes and designator APIs. This works by extending the designator client `add_seen` API to match the public designator API added in https://github.com/zooniverse/designator/pull/86

This will be called each time a classification is received by worker https://github.com/zooniverse/Panoptes/blob/f039669292a9709ca83a9e7e382b6ce3c8da2c6b/app/workers/notify_subject_selector_of_seen_worker.rb which in turn calls the designator selector https://github.com/zooniverse/Panoptes/blob/f039669292a9709ca83a9e7e382b6ce3c8da2c6b/lib/subjects/designator_selector.rb#L15 to call this client method to call the remote API. 

# Review checklist

### DIFF
--- a/app/services/designator_client.rb
+++ b/app/services/designator_client.rb
@@ -28,8 +28,12 @@ class DesignatorClient
   end
 
   def add_seen(workflow_id, user_id, subject_id)
-    # not needed right now
-    true
+    request(:put, "/api/users/#{user_id}/add_seen_subject") do |req|
+      req.body = {
+        workflow_id: workflow_id,
+        subject_id: subject_id
+      }.to_json
+    end
   end
 
   def load_user(workflow_id, user_id)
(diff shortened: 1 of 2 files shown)


camallen:
> This looks like the first designator API call that's going to be happening every single classification, so before this gets deployed maybe we should figure out how best to measure any performance changes to our total per-classification time

As the designator call happens in a background worker I don't think this will impact on the panoptes API at all. It may have an impact on designator but as it's built for high concurrency and the in memory ETS cache (https://github.com/sasa1977/con_cache) updates are not blocking i can't see this being a problem. That said it would be nice to have some metrics we can compare. 

> Shouldn't be much at all but it seems like this was included and then short circuited at some point (the no-op notify worker) so good to keep an eye on.

FWIW The client adapter public API was originally modeled on the cellect client that does rely on signalling seens subject ids and hence this method was a no-op for designator as it didn't care about what you'd seen.

PR #3151, opened 2019-07-12
Added strong_migrations gem and initializer

zwolf:
Initializer tells the gem to ignore all old migrations, so we're just checking from here forward and only in test/dev envs. 

For reference: https://github.com/ankane/strong_migrations

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,6 +67,7 @@ group :development, :test do
   gem 'pry-rails'
   gem 'pry-stack_explorer'
   gem 'spring'
+  gem 'strong_migrations'
 end
 
 group :test do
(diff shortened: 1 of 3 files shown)


zwolf:
My assumption was that if you include a migration in a PR, you would have already run that migration locally while writing specs and seen any potential issues it raised, then either included a circumvention (and an explanation) or fixed it before submitting. 

This isn't a benchmarking tool and any errors/suggestions it raises in CI it would raise locally first. But by the same token I don't see any harm in it. if you think there's utility in letting it go nuts, I'll put it in production, too.

camallen:
> My assumption was that if you include a migration in a PR, you would have already run that migration locally while writing specs and seen any potential issues it raised, then either included a circumvention (and an explanation) or fixed it before submitting.
> 
> This isn't a benchmarking tool and any errors/suggestions it raises in CI it would raise locally first. But by the same token I don't see any harm in it. if you think there's utility in letting it go nuts, I'll put it in production, too.

There is nothing to stop someone circumventing this (small risk) and the dev team assuming that the checks ran but really they didn't. I.e. we stop reviewing that migration code risk but rely on an automated agent that we have no guarantee it ran. This is my fear that we get lazy on this...hence I was wondering about CI tooling.

PR #3156, opened 2019-07-22
[WIP] Adds Subject#surrounding to find adjacent subjects by priority

zwolf:
This is mostly a stream of consciousness PR to basically do (likely inefficiently) the task at hand: pull priority-adjacent subjects from a whole subject set given a single subject and a couple parameters. 

Right off the bat: this loads the entire set's worth of SetMemberSubjects, which is kind of a non-starter. Another way to do this would be to pull the full possible range (`subject's priority + (window * gap)`) in SQL and then select the gap-defined ones out of that range. Either way, we're gonna need a smart new index on priority, subject id, and set id. This is also presumably doable entirely in SQL, but using the array indexes was more straightforward.

The actual #surrounding logic is pretty dumb, but I just wanted to get something working. I'd love to see how someone cleverer might handle it. (but not too clever)

The controller bit of this doesn't work yet because I was having trouble getting past a mysterious controller error: `Action :surrounding_subjects not defined for #<SubjectPolicy...`. I thought the `skip_policy_scope` helper would have handled that, just like it does for #queued, but no such luck.

Work in progress. Also, who likes "adjacent" better than "surrounding"?

# Review checklist


app/controllers/api/v1/subjects_controller.rb
```diff
@@ -71,6 +71,12 @@ def destroy
     end
   end
 
+  def surrounding
+    skip_policy_scope
+
+    subject.get_surrounding(params[:subject_set_id], params[:window], params[:gap])
```
camallen: I think the `get_surrounding` behaviour should be on the `SetMemberSubject` model as that is where the query hits.

app/models/subject.rb
```diff
@@ -71,4 +71,23 @@ def retired_for_workflow?(workflow)
       false
     end
   end
+
+  def surrounding(subject_set_id, window=5, gap=1)
+    all_smses = SetMemberSubject.where(subject_set_id: subject_set_id).order(:priority)
```
camallen: Try to avoid this `order(:priority)` as it'll order the whole set before filtering. Ideally we have a new index that covers subject_set_id and priority on SMS model i.e. `sms_subject_set_id_and_priority` and we know the current `priority` value and we can then walk forward / backward from this index location to get the gap cover without searching through the whole space. 

SO i think allow the current priority value as a param and change the current index `sms_subject_set_id` to be compound with `sms_subject_set_id_and_priority`. That should allow a faster index based find of the surrounding / adjacent priority sms records.

app/models/subject.rb
```diff
@@ -71,4 +71,23 @@ def retired_for_workflow?(workflow)
...
+    this_sms = all_smses.where(subject_id: id).first
+    this_index = all_smses.find_index(this_sms)
```
camallen: I think we pass the current SMS record / priority in as we do know where are in the stack already.

config/routes.rb
```diff
@@ -75,6 +75,7 @@
       json_api_resources :memberships
 
       json_api_resources :subjects do
+        get :surrounding
```
camallen: not sure on the naming here, i prefer `neighbours` tbh.

spec/factories/set_member_subjects.rb
```diff
@@ -2,10 +2,7 @@
   factory :set_member_subject do
     subject_set
     subject
-
-    trait :with_priorities do
-      sequence(:priority)
-    end
+    sequence(:priority)
```
camallen: Mind using the trait `with_priorities` where you need it vs adding priorities to any sms factory record? Not sure what the flow on effects are for this change and the real world data will be mostly sms records without priority.

app/controllers/api/v1/subjects_controller.rb
camallen: Also can the method name be `neighbours` ?

spec/factories/set_member_subjects.rb
zwolf: RSpec doesn’t allow one to use a trait while also passing override values (subject_set_id, in this case) to `create_list`. I’m sure there’s another way to do it, but I figured that sequence wasn’t heavy. If those values being there could cause issues elsewhere I’ll remove them but that would also be good to know.

app/controllers/api/v1/subjects_controller.rb
zwolf: a) that makes sense, send the subject_set_id and params to this method on the SMS instead

b) why? I asked in the PR desc if you might not rather it was `adjacent`, but I'm not sure `neighbours` is better, even setting aside the US/UK spellings.

app/models/subject.rb
zwolf: Yeah, I said as much in the PR desc that the new/updated index is a given. Also mentioned that crafting this query in such a way as to return only the total window of interest is necessary to avoid loading and sorting whole sets.

zwolf:
Most substantive issues raised were addressed in the PR description. This is a WIP/draft and was uploaded by request. Will refactor to the SMS model and add the index before updating further. Still fighting the controller issue, if that sounds familiar to anyone.

app/controllers/api/v1/subjects_controller.rb
camallen: no great reason - i'd like to ensure the end point naming conveys the direct line from the known subject location in the priority line to it's direct neighbours and i'd like to move away from a verb tense to a noun if we can. 

`surrounding` works, `adjacent` is good but less common. I'm trying to convey as much info and simplify the naming as we can.

spec/factories/set_member_subjects.rb
camallen: TBH i'm not sure about repercussions. 

I'd try and keep the test data as real world as we can. most subjects / SMS won't have the priority data so best to keep them off the default factories i think

zwolf:
* Moved the model logic to the SMS and added a new index. Proof:

``` [2] pry(#<SetMemberSubject>)> range_smses.explain
=> EXPLAIN for: SELECT "set_member_subjects".* FROM "set_member_subjects" WHERE "set_member_subjects"."subject_set_id" = $1 AND (priority >= 5.0 AND priority <= 17.0)  ORDER BY "set_member_subjects"."priority" ASC [["subject_set_id", 72]]
                                                             QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using index_set_member_subjects_on_subject_set_id_and_priority on set_member_subjects  (cost=0.15..8.17 rows=1 width=96)
   Index Cond: ((subject_set_id = 72) AND (priority >= 5.0) AND (priority <= 17.0))
(2 rows)
```

* It'll load the whole range, then a bit of logic I'd love to clever up a bit walks it via index and the gap value. The indexes are turned back into subjects and will (eventually) be serialized by SubjectSelectorSerializer for continuity. That is, if I can get Restpack to be okay missing the selector context and so on.

* Thought about moving the controller logic to SetMemberSubjectController, but in the end it didn't seem as logical. It makes a certain sense to keep both the subject_id and subject_set_id in the params of a call to `/set_member_subjects`...but the front end is asking about a subject, and they shouldn't have to make an extra call to get the SMS id first.

* Unfortunately, I'm still getting the following error if I don't add that exception to that CheckResourcesExist before_action: 

```Failures:

  1) Api::V1::SubjectsController#adjacent should return an array of serialized subjects
     Failure/Error: raise UnknownAction, "Action #{action.inspect} not defined for #{self}"

     ApplicationPolicy::UnknownAction:
       Action :adjacent not defined for #<SubjectPolicy:0x007f98f7695720>
     # ./app/policies/application_policy.rb:31:in `scope_klass_for'
     # ./app/policies/application_policy.rb:43:in `scope_for'
     # ./lib/json_api_controller/pundit_policy.rb:20:in `policy_scope'
     # ./lib/json_api_controller/check_resources_exist.rb:26:in `resources_exist?'
     # ./lib/json_api_controller/check_resources_exist.rb:22:in `check_controller_resources'
     # ./spec/controllers/api/v1/subjects_controller_spec.rb:802:in `block (3 levels) in <top (required)>'
     # -e:1:in `<main>' 
```

I have been using `/subjects/queued` as the model for how to allow this kind of unscoped call, but just `skip_policy_scope` isn't enough. I imagine it's got something to do with how `/queued` builds its own scope, but I haven't been able to track it down past the included workaround. Even with the workaround, trying to use the `subject` relation in the SMS class gives you the same exception:

```
ApplicationPolicy::UnknownAction: Action :adjacent not defined for #<SubjectPolicy:0x007f98db182150>
from /Users/zach/zooniverse/Panoptes/app/policies/application_policy.rb:31:in `scope_klass_for' 
```

Can't make much more progress on the controller side until I figure this out. Panoptes' controller layer is still a labyrinth after all this time (I'm not even sure I'm setting the params correctly (`params['params']['subject_set_id']`? c'mon now). Advice solicited.

camallen:
@zwolf i've added some code to fix the authorization issue of the adjacent controller action. As it's a member route with a subject id the check resources exist action was firing like you noticed. 

This code change i've just added will check the user has access to the project's subjects before allowing them to get the adjacent subjects, i.e. private projects shouldn't leak subjects...

camallen:
I'm closing this for the meantime, feel free to re-open when dev work continues on this.

PR #3157, opened 2019-07-23
remove data export restrictions added for WAMBO project

camallen:
This removes the  quick and dirty data export restrictions added in #2612 to ensure we didn't leak the the WAMBO project's PII annotation task data. 

The WAMBO classification annotation data has been removed from the database via https://gist.github.com/camallen/ca697d8d80002a8d99f7c6a7f4bd48c3. As such the export restrictions can be removed as well. 

# Review checklist


zwolf:
To be honest...is this worth removing? Is this code that could be abstracted a bit more to be able to turn off exporting problematic projects? There are a couple that I'm thinking of, one that I'm actively debugging, that probably shouldn't under almost all circumstances be exported normally.

camallen:
> To be honest...is this worth removing? 

Yes - if you see the history of the PR this was a very dirty workaround for this particular WAMBO project. It blocks downstream systems from working via kinesis stream, stats, caesar etc.

We should remove this functionality as it's not longer being used and adds overhead (code maintenance / query execution times).

> Is this code that could be abstracted a bit more to be able to turn off exporting problematic projects? 

Yes, though we should do this from a clean slate not based on this work. The plan was/is to revisit the need for this functionality but rearchitect the code to ensure we handle both internal (stream and caesar stats etc) and external (researchers) use cases.

> There are a couple that I'm thinking of, one that I'm actively debugging, that probably shouldn't under almost all circumstances be exported normally.

Research teams need their data via self service. It sounds more like a failing of the exporter code vs desired funcationality.
 
> edit: A potential point in favor of removal is Caesar's FetchClassificationWorkers include this `where (not (` clause in all their requests. We just had an awful lot of them run to backfill a project and it's probably worth looking into whether that was affecting query time.

`where.not` is almost never good, however in in this case the `not` was on the project table supplied with a `project_id` so should not have made a big impact on the query plan / lookup.

camallen:
> Fair enough. With the recent AACWC issues it seemed worth asking. LGTM.

Yep for sure, it's useful to know that we have in the past / can do this export blocking if we need to in future.

PR #3161, opened 2019-08-12
Catch NoMethodError when generating export

adammcmaster:
`workflow_at_version` can be nil. I assume there is an underlying issue
that causes that, but this does not attempt to fix it.

Regardless of the underlying problem, the export should not silently
fail.

# Review checklist


adammcmaster:
Related HB https://app.honeybadger.io/projects/40595/faults/49691900

PR #3162, opened 2019-08-16
allow selector page size to be set from workflow configuration

camallen:
Allow the selector page_size to be derived from the `workflow.configuration` object.

default is to respect the incoming `page_size` param value, then if `workflow.configuration['subject_queue_page_size']` and then a default value of 10 (current). 

The use case is from @hughdickinson, specifically some workflows need to select less than the default to ensure their expected training ratios are met.

this will allow a workflow configuration setting to set a default page size when requesting subjects.

The only issue i can see with this is more round trips to the selector for that workflow, especially once the user has seen past the point where they would receive training subjects. Perhaps we could do more introspection on the user seens but that would add more work into a critical path of the API. I'd appreciate folks thoughts on this.

# Review checklist

### DIFF
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -123,9 +123,12 @@ module Subjects
       MissingSubjects.new("No data available for selection")
     end
 
-    def subjects_page_size
-      page_size = params[:page_size] ? params[:page_size].to_i : 10
-      params.merge!(page_size: page_size)
+    def subjects_page_size(default_page_size=10)
+      page_size = (params[:page_size] || workflow.selector_page_size).to_i
+
+      # update the params as they flow through the system, i.e. to the serializer
+      params[:page_size] ||= page_size
+
       page_size
     end
 
--- a/lib/subjects/selector.rb
+++ b/lib/subjects/selector.rb
@@ -124,7 +124,7 @@ module Subjects
     end
 
     def subjects_page_size(default_page_size=10)
-      page_size = (params[:page_size] || workflow.selector_page_size).to_i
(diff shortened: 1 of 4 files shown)


srallen:
Can I suggest a slight name change? I would read `selector` as workflow selection rather than subject set selection since on the client side, we have quite a bit of logic with regard to which workflow we end up displaying in the classifier. I would suggest instead `queue_page_size` or more pedantically, `subject_queue_page_size`. If there's an important reason it should set as `selector`, then feel free to ignore my suggestion.

camallen:
> subject_queue_page_size

Happy to change it to `subject_queue_page_size`

lib/subjects/selector.rb
```diff
@@ -123,9 +123,12 @@ def missing_subjects_error
       MissingSubjects.new("No data available for selection")
     end
 
-    def subjects_page_size
-      page_size = params[:page_size] ? params[:page_size].to_i : 10
-      params.merge!(page_size: page_size)
+    def subjects_page_size(default_page_size=10)
```
zwolf: Doesn’t look like this method uses its default value since it’s been added to the workflow model.

lib/subjects/selector.rb
```diff
@@ -123,9 +123,12 @@ def missing_subjects_error
...
+      page_size = (params[:page_size] || workflow.selector_page_size).to_i
```
camallen:
```suggestion
      page_size = (params[:page_size] || workflow.selector_page_size(default_page_size)).to_i
```

lib/subjects/selector.rb
camallen: good catch - i've ensured that value can flow to the `workflow.selector_page_size` method now.

PR #3163, opened 2019-08-17
Export catch nomethoderror

marten:
Continuing on @adammcmaster's work in https://github.com/zooniverse/Panoptes/pull/3161/files in addition to letting an export proceed in case the version can't be found, this branch will create a version on the fly if the requested version is the latest, but the WorkflowVersion for it doesn't exist. It will also ensure that Honeybadger is still notified in case of issues, even though we continue with the export.

I'm also adding a rake task to make sure that every workflow has at least a version for the current state (which would otherwise not be captured anywhere if someone classifies and then a project owner makes a change). **We should be run this as soon as possible.**

# Review checklist

### DIFF
--- a/lib/classification_dump_cache.rb
+++ b/lib/classification_dump_cache.rb
@@ -1,4 +1,6 @@
 class ClassificationDumpCache
+  class MissingWorkflowVersion < StandardError; end
+
   def initialize
     @workflows = {}
     @subjects = {}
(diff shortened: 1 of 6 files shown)


lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -189,6 +189,8 @@ def task
          key == annotation["task"]
         end
         @task = task_annotation.try(:last) || {}
+      rescue NoMethodError
```
marten: There's no test covering this change.

lib/classification_dump_cache.rb
```diff
@@ -49,6 +56,13 @@ def secure_user_ip(ip_string)
   def find_workflow_at_version(workflow, major_version, minor_version)
     workflow.workflow_versions.where("major_number >= ? AND minor_number >= ?", major_version, minor_version).order("major_number ASC, minor_number ASC").first!
   rescue ActiveRecord::RecordNotFound
-    workflow.workflow_versions.order("major_number ASC, minor_number ASC").last
+    fallback = workflow.workflow_versions.order("major_number ASC, minor_number ASC").last
+    return fallback if fallback.present?
+
+    if workflow.major_version == major_version && workflow.minor_version == minor_version
```
zwolf: What's the difference between this and the `.where` above? Is it just to confirm that it doesn't exist to allow you to raise, or is there some circumstance I'm not thinking about where this would be able to build after no version (or fallback) was found?

marten: The where finds workflow_version records. If that doesn't exist, then this checks if the requested version perhaps happens to be the latest one (the workflow itself has a major/minor number equal to requested). In that case we're lucky and can make a workflow_version from the current workflow state.

PR #3175, opened 2019-09-20
ensure first_party apps can login

camallen:
ensure local apps are not set to confidential, this will allow them to convert session cookies to tokens

# Review checklist


camallen:
> lgtm, gotta keep that seed data accurate

I spent 2 hours trying to figure out why i couldn't get a token from my default first party app when testing cookie subdomains / re-use on friday. The worst part was i remembered this setting and tried it but incorrectlyset the value to false....fail.

PR #3176, opened 2019-09-20
allow cookie reuse across zooniverse subdomains

### LINKED ISSUE #3123: Request: Allow Login between Zooniverse.org Subdomains
## Feature Request

I would like to request the ability for login states to persist between Zooniverse subdomains.

TL;DR: when creating cookies for `_Panoptes_session`, please set cookie's domain to `zooniverse.org` so it works across all subdomains.

### Context and Problem

Students working on the WildCam Darien Lab educational tool first need to **login** at [classroom.zooniverse.org](https://classroom.zooniverse.org/#/wildcam-darien-lab/) so they can see their list of classroom assignments. When they start their classroom assignments, they are taken to WildCam Darien's [Classify page on Zooniverse.org](https://www.zooniverse.org/projects/wildcam/wildcam-darien/classify?workflow=8752&classroom=1&_ga=2.183173990.1927398751.1556897301-1091836498.1556897301) (opened in a new window)

Problem is, their login state is lost when moving from `classroom.zooniverse.org` to `www.zooniverse.org`

Screen Shot 2019-05-03 at 16 28 48
_Student User at WildCam Darien Lab about to start an Assignment. They are logged in. A reminder (yellow) informs them to remember to login at zooniverse.org as well. Clicking on 'Start Assignment' opens a new tab/window to www.zooniverse.org_

Screen Shot 2019-05-03 at 16 29 04
_New tab/window shows the user their Assignment at www.zooniverse.org. However, note that the user is now not logged in, since they switched subdomains._

### Analysis

When the student/user logs in at `classroom.zooniverse.org`, they do so via `panoptes.zooniverse.org`. The `_Panoptes_session` cookie created there is locked to the domain `panoptes.zooniverse.org`

Similarly, if the student/user then logs in at `www.zooniverse.org`, the cookie created there is locked to the domain `www.zooniverse.org` (note that PFE, unlike the various Custom Front Ends, doesn't login via `panoptes.zooniverse.org`)

Screen Shot 2019-05-03 at 16 40 58
_Cookie data showing the specific subdomain that's allowed to see it._

### Possible Solution

One possible solution is to specifically set the larger `zooniverse.org` domain when creating Panoptes session cookies, meaning the cookie should be visible to all subdomains of zooniverse.org.

Full solution: https://stackoverflow.com/questions/10402777/share-session-cookies-between-subdomains-in-rails

### Implications

- A "shared login session" change may have an affect on our other projects (i.e. Custom Front Ends) - such as making it impossible for a user to be logged into Account A while at `www.zooniverse.org` while simultaneously logged into Account B while at `antislaverymanuscripts.org` in the same browser...
  - ...though I'm not sure if this is a realistic use case to actually worry about.
- Further details TBD

### Status

Requested

camallen:
closes #3123 Allow cookies to be reused across subdomains as per https://github.com/plataformatec/devise/wiki/How-To:-Use-subdomains

# Review checklist

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,8 +1,14 @@
 # Be sure to restart your server when you modify this file.
 
+domain =
+  if %w(production staging).include?(Rails.env)
+    # allow zooniverse.org subdomains for staging & production
+    '.zooniverse.org'
+  else
+    # allow all subdomains for dev & test
+    :all
+  end
+
 Rails.application.config.session_store :cookie_store,
     key: '_Panoptes_session',
-    domain: {
-      production: '.zooniverse.org',
-      staging: '.zooniverse.org'
-    }.fetch(Rails.env.to_sym, :all)
+    domain: domain
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,3 +1,8 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_Panoptes_session'
+Rails.application.config.session_store :cookie_store,
+    key: '_Panoptes_session',
+    domain: {
+      production: '.zooniverse.org',
+      staging: '.zooniverse.org'
+    }.fetch(Rails.env.to_sym, :all)


camallen:
> What's the purpose of defining production and staging domains separately if they're the same? Wouldn't just `domain: '.zooniverse.org'` be sufficient?

I decided to use explicit env definitions for staging and production. I could have had brancing code here as well to decide the value. I.e. if we are in staging or production mode then setup the domain to allow *.zooniverse.org. If we are in dev mode or another, say test, then allow all domains.

The code i've got is really a compact one liner, if it doesn't make sense i can expand it. Maybe?
```
default_domain_setting = 
  if %w(production staging).include?(Rails.env)
    '.zooniverse.org',
  else 
    :all
  end

Rails.application.config.session_store :cookie_store, key: '_Panoptes_session', domain: default_domain_setting
```

PR #3195, opened 2019-10-17
Fix rubocop

camallen:
our existing ignored cops file was out of date and it looks like there wasn't a good way to migrate them (mry gem apears unmaintained).

Ideally we actually start linting all new PR code properly, i've done the same on the caesar code (https://github.com/zooniverse/caesar/pull/680) base as well. This is a clean start where we will get a lot of linter warnings that we can choose to incrementally ignore or fix as PRs come in.

I've also added some other cops (rails, rspec and performance ones). happy to remove these if they prove cumbersome or cause issues.

# Review checklist


camallen:
> I'm for this in theory, but if we haven't been linting this or Caesar outside of the hound's recs, we might need to figure a protocol for figuring out how much fixing is in scope for a given PR.
> 
> lgtm, with the caveat that we'll have to take it PR by PR if this gets noisy.

Yeah - as development work continues we can address these linter errors or add ingores for them. 

FWIW i use vscode extension https://marketplace.visualstudio.com/items?itemName=misogi.ruby-rubocop and lint as I code, this helps ensure visibility to the hound errors ahead of time.

Alternatives are to run the rubopc linter on a commit hook / git pre push hook?

PR #3205, opened 2019-11-21
cleanup docker compose

camallen:
bump the postgres version and remove the unused cellect selection system

# Review checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,5 @@
 postgres:
-  image: postgres:9.4
+  image: postgres:9.5
   environment:
     - "POSTGRES_USER=panoptes"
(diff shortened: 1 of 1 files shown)


camallen:
> Cellect is still an open question w/r/t the migration, does this mean it just doesn't need to be bundled and spun up in dev or that we're not going to need it at all?

We still use cellect for subject selection on a number of workflows (some for per set selection and historically for sequential subject selection as well). 

As we have per set selection in designator now, https://github.com/zooniverse/designator/pull/74/files we can look at changing those over. 

We should also audit the API use of cellect to determine if we need to port it as some projects may be completed / not requiring this functionality anymore.

PR #3210, opened 2019-11-29
update the default cors configs

camallen:
match the changes from in #3206 and #3199

# Review checklist


zwolf:
We're not still using Hudson at all since the move to Travis, are we? Can this file just be removed?

camallen:
These are in use but for local envs i.e. smoke testing the API locally using docker-compose etc. See point 6 in https://github.com/zooniverse/Panoptes#usage. 

Noting this will most likely change for k8 deployment.

PR #3221, opened 2019-12-03
Scope SubjectMetadataWorker updates to know subjects and sets

camallen:
This PR changes the scope of each called worker to be idempotent and targets the update queries to known set_member_subject records. It will now also avoid clobbering any previously set priority values and thus avoid update rows where possible.

Current code flow means that anytime we add subjects to a set we queue a worker to update all the SMS records for the known set. For a large set upload this can leave us with blocking updates via concurrent workers. This PR should avoid these issues via more selective worker update queries.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subject_sets_controller.rb
+++ b/app/controllers/api/v1/subject_sets_controller.rb
@@ -104,7 +104,7 @@ class Api::V1::SubjectSetsController < Api::ApiController
       end
 
       result = SetMemberSubject.import IMPORT_COLUMNS, new_sms_values, validate: false
-      SubjectMetadataWorker.perform_async(resource.id)
+      SubjectMetadataWorker.perform_async(subject_ids_to_link)
       result
     else
       super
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/subject_sets_controller.rb
```diff
@@ -104,7 +104,7 @@ def add_relation(resource, relation, value)
       end
 
       result = SetMemberSubject.import IMPORT_COLUMNS, new_sms_values, validate: false
-      SubjectMetadataWorker.perform_async(resource.id)
+      SubjectMetadataWorker.perform_async(resource.id, subject_ids_to_link)
```
camallen: Should this use the SMS ids instead of the set and subject_ids?

camallen:
> awesome. One thing: should we avoid clobbering the previous priorities? If a subject is getting relinked and the metadata has changed, shouldn't the new priority be represented in the SMS?

I don't think we need to worry about clobbering as if it's getting re-linked then the `set_member_subject` (SMS) record won't exist, as unlinking will remove the SMS record. 

FYI - The db has a unique index on the SMS table to avoid linking the same subject to a set https://github.com/zooniverse/Panoptes/blob/c6ab1e888703325db7c8dfeff4101008db012b0b/db/structure.sql#L3357

Your point about ensuring the latest subject priority value is copied is very valid. As it stands the import would fail for any existing duplicate import so clobbering is essentially impossible here. I will remove the `subject.priority IS NULL` filter from the update and just rely on the sms ID for selectivity on the update query. Thanks :)

PR #3228, opened 2019-12-16
remove set_member_subject_ids

camallen:
set_member_subject_ids is not used anymore, remove it.

# Review checklist

### DIFF
--- a/app/models/subject_workflow_status.rb
+++ b/app/models/subject_workflow_status.rb
@@ -45,8 +45,4 @@ class SubjectWorkflowStatus < ActiveRecord::Base
   def retired?
     retired_at.present?
   end
-
-  def set_member_subject_ids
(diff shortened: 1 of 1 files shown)


nciemniak:
A consideration:
Should the references to `set_member_subject_ids` in rspec files be removed, or does it not matter? It is referenced in a couple of spec files, [see here](https://github.com/zooniverse/Panoptes/search?q=set_member_subject_ids&unscoped_q=set_member_subject_ids).

camallen:
> see here

Good question - each of those specs are unrelated to this change and the db migration is old (though related). 

Basically this code was meant to be removed a long time ago but was missed - it's not on a code path (that i can tell) so i think we're good to remove (and specs passing agree).

nciemniak:
Awesome. Looks good to me 👍

PR #3259, opened 2020-02-27
[WIP] Use cellect ingress dns instead of cellect client

camallen:
Remove Cellect::Client gem and it's dependency on redis for internal node IP tracking for server / client comms. Instead we will now use the DNS (K8s ingress) to communicate with the single cellect server we will run, we do this by moving CellectClient to use faraday for direct communication with the cellect server vs finding a node IP to communicate with.

Noting that this removes the ability to run multiple cellect servers with and having the API broadcast state to these servers, e.g. reload a workflow across all servers, or remove a subject from a workflow (retire) for all servers, etc. However as we will only run one cellect node this won't be a problem till we move the remaining functionality (sequential selection) to designator and then we can remove cellect from the stack entirely

# Review checklist


app/services/cellect_client.rb
```diff
@@ -1,92 +1,83 @@
-require 'cellect/client'
-require 'subjects/cellect_session'
+# frozen_string_literal: true
 
 module CellectClient
   ConnectionError = Class.new(StandardError)
 
-  MAX_TRIES = 1
+  CELLECT_HOST = ENV.fetch('CELLECT_HOST', 'https://cellect.zooniverse.org')
```
adammcmaster: I think we should probably take account of the `RAILS_ENV` here and return either the production or staging Cellect URL as needed – otherwise we would need to make sure that `CELLECT_HOST` and `RAILS_ENV` are both updated together which introduces a risk of human error.

```suggestion
  if Rails.env.production?
    DEFAULT_CELLECT_HOST = 'https://cellect.zooniverse.org'
  else
    DEFAULT_CELLECT_HOST = 'https://cellect-staging.zooniverse.org'
  end
  CELLECT_HOST = ENV.fetch('CELLECT_HOST', DEFAULT_CELLECT_HOST)
```

camallen:
> I'm not sure how to actually test this (I guess we'll just do that in staging after merge), but the code looks OK with one suggestion from me.

I probably should have included a setup for cellect in docker-compose so this can be tested. I think i'll do that before merging to ensure this works as i expect.

camallen:
This is good to go, client is communicating nicely to the cellect server (I've manually tested). This was the test harness via `docker-compose.yml`
```
version: '3'
services:
  postgres:
    image: postgres:9.5
    environment:
      - "POSTGRES_USER=panoptes"
      - "POSTGRES_PASSWORD=panoptes"
    ports:
      - "5433:5432"

  redis:
    image: redis
    command: redis-server --appendonly yes

  cellect:
    image: zooniverse/cellect-panoptes
    environment:
      - "RACK_ENV=development"
      - "PUMA_PORT=80"
      - "PUMA_MAX_THREADS=1"
      - "DEBUG_CELLECT_START=true"
      - "DATABASE_URL=postgresql://panoptes:panoptes@pg:5432/panoptes_development?pool=16"
      - "ATTENTION_REDIS_URL=redis://redis:6379/10"
    ports:
      - "4000:80"
    links:
      - redis
      - postgres:pg

  panoptes:
    build:
      context: .
      dockerfile: Dockerfile.dev
    volumes:
      - ./:/rails_app
      - gem_cache:/usr/local/bundle
    ports:
      - "3000:3000"
    environment:
      - "RAILS_ENV=development"
      - "ATTENTION_REDIS_URL=redis://redis:6379/10"
      - "CELLECT_HOST=http://cellect"
      - "DISABLE_SPRING=true"
    links:
      - redis:redis
      - cellect:cellect
      - postgres:pg

volumes:
  gem_cache:
```

PR #3278, opened 2020-03-23
add statement_timeout to db template

camallen:
record the config we are using to timeout long sql queries at the db connection layer

I'm going to update our system database.yaml to include this config. The plan is to set the ENV variable for API nodes to just longer than the load balancer (60s). As the loadbalancer will terminate connections that surpass the timeout, we can safely cancel the db queries as the results won't make it to the client.

The 5min (300000ms) value here is for sidekiq and should handle any long running queries in this env. This default can be changed via ENV vars if needed.

# Review checklist

### DIFF
--- a/config/database.yml.hudson
+++ b/config/database.yml.hudson
@@ -8,6 +8,9 @@ default: &default
   pool: 5
   port: 5432
   prepared_statements: false
+  variables:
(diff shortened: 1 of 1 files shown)


camallen:
this needs a rebase over #3279 to build correctly as a gem was yanked :(

PR #3281, opened 2020-03-24
add custom workflows preloading project serializer

camallen:
avoid loading all the json attributes (some very large) for project workflow associations. As the links only need to lookup the association id we can use a custom preloader to select only the required data for serializing the association links

# Review checklist

### DIFF
--- a/app/serializers/project_serializer.rb
+++ b/app/serializers/project_serializer.rb
@@ -24,6 +24,8 @@ class ProjectSerializer
   can_sort_by :launch_date, :activity, :completeness, :classifiers_count,
     :updated_at, :display_name
 
+  # note: workflow(s) associations are preloaded via the
+  # self.serialize_page(page, options) below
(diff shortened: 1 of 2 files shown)


app/serializers/project_serializer.rb
```diff
@@ -54,16 +56,47 @@ def self.paging_scope(params, scope, context)
...
     end
   end
 
+  # override the serialize page method to preload the
+  # workflow and active_worklfows association data
+  # however avoid loading large json attributes
+  # instead load the data required for serializing
+  # the project and links
+  def self.serialize_page(page, options)
+    project_ids = page.pluck(:id)
```
camallen: I think i should wrap this with the existing flipper test in case it causes problems and we want to change the behaviour at runtime. 
`if Panoptes.flipper[:test_preload_workflow_associations].enabled?`

PR #3282, opened 2020-03-26
add validator to ensure only one favorite per collection

### LINKED ISSUE #2552: Only allow one favourite collection per project
Reject create requests for favourites when a favourite already exists.

nciemniak:
Closes #2552 .

New validator for `Collection` model. Ensure that only one favorite collection is created per owner for each project. Originally I wanted the validator class to live in the same file as the collection model, because it's going to be used _only_ by the collection model, but when declaring the validator class in app/models/collection.rb, I was getting errors telling me the validator class could not be found, so we're going with this. 

It looks like there are other validators that live in no-mans-land in the api folder, I've created a new folder `api/validators` to organize validators. I'd like to do an additional small PR to move existing validators floating around in the api folder into the validators folder, thoughts on this?

Note - there are a lot of changes you'll see here that are just rubocop doing it's auto-linting doowop.


app/models/collection.rb
```diff
@@ -6,20 +6,20 @@ class Collection < ActiveRecord::Base
 
   has_and_belongs_to_many :projects
   has_many :collections_subjects,
-    -> { order(id: :desc) },
-    dependent: :destroy
+           -> { order(id: :desc) },
```
camallen: Just a note: As it's a long running code base with lots of lint errors in it, I avoid fixing these kinds of lint errors in my IDE as they make the PR diff super noisy. Most of this class PR diff is lint fixes, not new behaviour.

lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -0,0 +1,16 @@
+module Validators
+  # Validator for Collection model
+  # Validate that only one favorite collection is created per owner for each project
+  class OneFavoritePerOwnerValidator < ActiveModel::Validator
+    def validate(new_record)
+      return unless new_record.owner
+
+      records = new_record.owner.collections
```
camallen: you can search for favourites here like 
```suggestion
      owners_existing_favorites = new_record.owner.collections.where(favorite: true)
```

lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+      return unless records && new_record.favorite
```
camallen: will records be a AR:Relation here? i.e. `[]` but we don't test for empty?

also the guard clause on `new_record.favorite` should be at the top not here, short circuit and avoid any db work etc if we can.

lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+
+      records.each do |record|
+        new_record.errors[:favorite] = 'Only one favorite collection can be created per owner for each project' if record.favorite
```
camallen: I think you can optimize this looping to a db call above where your find all the owner collections. instead of finding them all and iterating over them to check if any are favs we can just ask the db if there are any and mark as error if there are. Instead of this loop where you do the record fetches try something like this using https://apidock.com/rails/v4.2.9/ActiveRecord/FinderMethods/exists%3F
```
has_existing_favorite = new_record.owner.collections.where(favorite: true).exists?
# then you can mark with error
if has_existing_favorite 
  new_record.errors[:favorite] = 'Only one favorite collection can be created per owner for each project'
end
```

lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -0,0 +1,16 @@
+module Validators
+  # Validator for Collection model
+  # Validate that only one favorite collection is created per owner for each project
```
camallen:
```suggestion
  # Validate that only one favorite collection is created per owner for each project
  # This validation technique may allow a duplicate favorite record be inserted into the database
  # due to the db the read isolation level of this transaction vs concurrent transactions
  # and the fact our validations exist in ruby land and not strictly enforced at the database layer
  # it's low probablity but worth noting here if folks wonder how it can happen
```

app/models/collection.rb
nciemniak: Keeping this in mind moving forward 👍

lib/validators/one_favorite_per_owner_validator.rb
nciemniak: Ah yes, very true. Moving the guard clause 👍

lib/validators/one_favorite_per_owner_validator.rb
nciemniak: Wow yes, this is much cleaner.

lib/validators/one_favorite_per_owner_validator.rb
nciemniak: >   This validation technique may allow a duplicate favorite record be inserted into the database due to the db the read isolation level of this transaction vs concurrent transactions

Just to confirm that I'm understanding this correctly: 
the hypothetical duplicate favorite scenario you're describing here would occur if two requests to create a new favorite collection come in at the same time? Can you explain further about the read isolation level?

spec/controllers/api/v1/collections_controller_spec.rb
```diff
@@ -147,75 +147,119 @@
...
+          post :create, favorite_params
+          expect(response).to have_http_status(:bad_request)
+        end
+      end
+
+      context 'when a non-favorite collection exists for a given user/project' do
+        before { create(:collection, favorite: false, owner: owner) }
+
+        it 'successfully creates a favorite collection' do
+          post :create, favorite_params
```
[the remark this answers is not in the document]
nciemniak: Using keyword arguments does not work, when written like this,
`post :create, params: favorite_params`
params are sent as a hash with the keyword params, like so: `{ params: { collections: ... } }`, which is not correct.

app/models/collection.rb
```diff
@@ -6,20 +6,20 @@ class Collection < ActiveRecord::Base
...
   has_many :user_collection_preferences, dependent: :destroy
-  belongs_to :default_subject, :class_name => "Subject"
+  belongs_to :default_subject, class_name: 'Subject'
 
   validates :display_name, presence: true
   validates :private, inclusion: { in: [true, false], message: "can't be blank" }
-  ## TODO: This potential has locking issues
-  validates_with UniqueForOwnerValidator
+  ## TODO: UniqueForOwnerValidator potential has locking issues
+  validates_with UniqueForOwnerValidator, Validators::OneFavoritePerOwnerValidator
```
camallen, line 19: not blocking, perhaps move / refactor existing validator `UniqueForOwnerValidator` code to align with your new validator organization folder structure? for consistency in the application going forward?

lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -0,0 +1,18 @@
...
+  # this validation technique may allow a duplicate favorite record be inserted into the database
+  # due to the db the read isolation level of this transaction vs concurrent transactions
+  # and the fact our validations exist in ruby land and not strictly enforced at the database layer
+  # it's low probablity but worth noting here if folks wonder how it can happen
+  class OneFavoritePerOwnerValidator < ActiveModel::Validator
+    def validate(new_record)
+      return unless new_record.owner && new_record.favorite
+
+      owner_has_existing_favorite = new_record.owner.collections.where(favorite: true).exists?
+      return unless owner_has_existing_favorite
```
camallen, line 13:
```suggestion
      if owner_has_existing_favorite = new_record.owner.collections.where(favorite: true).exists?
```

lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      new_record.errors[:favorite] = 'Only one favorite collection can be created per owner for each project'
```
camallen, line 15:
```suggestion
      new_record.errors[:favorite] = 'Only one favorite collection can be created per owner for each project'
```
```suggestion
      new_record.errors[:favorite] = 'An owner can only have one favorite collection per project'
```

spec/controllers/api/v1/collections_controller_spec.rb
```diff
@@ -147,75 +147,119 @@
...
+          expect(response).to have_http_status(:created)
+        end
+      end
+
+      context 'when favorite collection already exists for given user/project' do
+        before { create(:collection, favorite: true, owner: owner) }
+
+        it 'raises a 400 error on trying to create a favorite collection' do
```
camallen, line 217:
```suggestion
        it 'returns a bad request error trying to create a favorite collection' do
```

spec/controllers/api/v1/collections_controller_spec.rb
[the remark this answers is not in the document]
camallen: yeah - sadly this is rails 5+ specific, new versions of the linter pickup on the rails version and ignore these defaults, https://github.com/rubocop-hq/rubocop/issues/3629

sadly houndci isn't running the later versions so we get these false errors

lib/validators/one_favorite_per_owner_validator.rb
nciemniak: Agreed 👍

app/models/collection.rb
nciemniak: Sure thing, that sounds sensible

PR #3285, opened 2020-03-27
Mark recents for deletion with hourly sweeper workers

camallen:
classification deletes are cascading down to recents which is a very slow removal and is now erroring due the the pg statement timeouts

instead of `dependent :destroy` on the classification->recents relation, move to a mark and sweep system where we use a background worker on a schedule to remove the recents. The advantage of this technique is: 
1. it's quick and will not impact the classifcation destroy
2. it allows us to have implement a simple capped collection for recents (max 50 or so) which will be updated when new classifications come in.

The removal worker will cleanup any marked recents and any old recents past 14 days by default.

The migration will remove recents older than 14 days, leave the rest and remove the bloated indexes and rebuild the only ones we use for the API.

## Todo
- [ ] Rebase / merge with schema changes in #3424 

# Review checklist

### DIFF
--- a/app/workers/recent_sweeper_worker.rb
+++ b/app/workers/recent_sweeper_worker.rb
@@ -27,7 +27,10 @@ class RecentSweeperWorker
     old_recents = Recent.where('id <= ?', old_recent.id).select(:id)
 
     old_recents.find_in_batches do |recents|
-      Recent.where(id: recents).update_all(mark_remove: true)
+      Recent.where(
+        id: recents,
+        mark_remove: false
+      ).update_all(mark_remove: true)
     end
   end
 end
--- a/app/workers/recent_sweeper_worker.rb
+++ b/app/workers/recent_sweeper_worker.rb
@@ -10,11 +10,24 @@ class RecentSweeperWorker
   recurrence { hourly }
 
   def perform
+    # mark any recents older than 14 days, keep the recents table clean
+    mark_old_recents(Recent.first_older_than(14.days))
+
     # https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches
     # move this method to in_batches when migrating past rails 4
     recent_ids_to_delete = Recent.where(mark_remove: true).select(:id)
     recent_ids_to_delete.find_in_batches do |recents|
       Recent.where(id: recents).destroy_all
+    end
+  end
+
+  def mark_old_recents(old_recent)
+    return unless old_recent
+
+    old_recents = Recent.where('id <= ?', old_recent.id).select(:id)
+
+    old_recents.find_in_batches do |recents|
+      Recent.where(id: recents).update_all(mark_remove: true)
     end
   end
 end
--- a/app/workers/recent_sweeper_worker.rb
+++ b/app/workers/recent_sweeper_worker.rb
@@ -1,0 +1,18 @@
+class RecentSweeperWorker
+  include Sidekiq::Worker
+
+  include Sidetiq::Schedulable
+
+  sidekiq_options queue: :data_low
+
+  recurrence { hourly }
+
+  def perform
+    # https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches
+    # move this method to in_batches when migrating past rails 4
+    recent_ids_to_delete = Recent.where(mark_remove: true).select(:id)
+    recent_ids_to_delete.find_in_batches do |recents|
+      Recent.where(id: recents).destroy_all
+    end
+  end
+end
--- a/app/workers/recent_sweeper_worker.rb
+++ b/app/workers/recent_sweeper_worker.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class RecentSweeperWorker
   include Sidekiq::Worker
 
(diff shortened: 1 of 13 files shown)


app/workers/recent_sweeper_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+  def mark_old_recents(old_recent)
+    return unless old_recent
+
+    old_recents = Recent.where('id <= ?', old_recent.id).select(:id)
+
+    old_recents.find_in_batches do |recents|
+      Recent.where(id: recents).update_all(mark_remove: true)
```
camallen, line 30:
```suggestion
      Recent.where(id: recents, mark_remove: false).update_all(mark_remove: true)
```

db/migrate/20200327164455_add_mark_remove_column_to_recents.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class AddMarkRemoveColumnToRecents < ActiveRecord::Migration
+  disable_ddl_transaction!
+  def up
+    add_column :recents, :mark_remove, :boolean
```
camallen, line 6: Clear the `STATEMENT_TIMEOUT=0` for running migration tasks (ENV var) or manually here to ensure none of these queries time out and break the migration.

db/migrate/20200327164455_add_mark_remove_column_to_recents.rb
```diff
@@ -0,0 +1,59 @@
...
+    change_column_default :recents, :mark_remove, false
+    remove_foreign_key :recents, :classifications
+
+    remove_old_recents
```
camallen, line 10: Should we split the DDL and data work into 2 migrations? Does this mean we can avoid `disable_ddl_transaction!` for DDL steps?

nciemniak:
@camallen , a few questions for context:
> classification deletes are cascading down to recents which is a very slow removal
1. A curiosity: what is the typical scenario for a classification being deleted? I see in the specs that there are cases where an incomplete classification is saved, in which case that classification can be deleted, though I haven't figured out how to actually do that on the front-end.
2. Why is the removal so slow? Shouldn't there only be one 'recent' entry for each classification? (and if so, why would that make deletion so slow?) Perhaps my understanding of what a `recents` record represents is incomplete, I know that you're able to view your recent classifications for a project by clicking the RECENTS tab, and Ive taken a look at the specs associated with recents.

> The migration will remove recents older than 14 days, leave the rest and remove the bloated indexes and rebuild the only ones we use for the API.

When you say 'remove bloated indexes', do you mean just narrow down which columns we have indexes set on? Also, what is the purpose of deleting an index only to rebuild it? Why not just omit them from the removal?

camallen:
> @camallen , a few questions for context:
> 
> > classification deletes are cascading down to recents which is a very slow removal
> 
>  > 1. A curiosity: what is the typical scenario for a classification being deleted? I see in the specs that there are cases where an incomplete classification is saved, in which case that classification can be deleted, though I haven't figured out how to actually do that on the front-end.

Transcription projects, i believe ASM allows users to save their work in incomplete classifications and then delete / save them as completed. https://app.honeybadger.io/projects/40595/faults/62245893

Only incomplete classifications can be deleted, https://github.com/zooniverse/Panoptes/blob/422cfe7cfdfe98e542a43d17958699ae43f76380/app/policies/classification_policy.rb#L38
 
> >    2. Why is the removal so slow? Shouldn't there only be one 'recent' entry for each classification? (and if so, why would that make deletion so slow?) 

Yep - we don't link a classification to a recent, rather a recent to a classification. So the search for a recent is on a FK index on a large table. The query is slow for some reason (see the honey badger link above and errors), these errors spurned this work which is linked to migration of db data as well.

> Perhaps my understanding of what a `recents` record represents is incomplete, I know that you're able to view your recent classifications for a project by clicking the RECENTS tab, and Ive taken a look at the specs associated with recents.

FWIW - we only show the last 20 recents in the UI but we keep all of them and they serve no purpose.

> > The migration will remove recents older than 14 days, leave the rest and remove the bloated indexes and rebuild the only ones we use for the API.
> 
> When you say 'remove bloated indexes', do you mean just narrow down which columns we have indexes set on? 

Once you delete a lot of indexed records the index generally needs to be cleaned up / rebuilt. https://wiki.postgresql.org/wiki/Index_Maintenance#Index_Bloat

> Also, what is the purpose of deleting an index only to rebuild it? Why not just omit them from the removal?

This will rebuild a fresh index with no 'bloat'. We do this concurrently to avoid blocking read / write operations on the underlying table, REINDEX will lock the table (at least in our PG version), https://www.postgresql.org/docs/9.5/sql-reindex.html
> An index build with the CONCURRENTLY option failed, leaving an "invalid" index. Such indexes are useless but it can be convenient to use REINDEX to rebuild them. Note that REINDEX will not perform a concurrent build. To build the index without interfering with production you should drop the index and reissue the CREATE INDEX CONCURRENTLY command.

I should have added more context to the reasons behind this work. The crux is:
1. we only use the last 20 recents but store them all (think every classification we receive).
0. Serving the last 20 recents is costly for user / project combos (indexes are large too - we currently serve these from the replica db)
0. classificaiton deletes for ASM were timing out and erroring (probably breaking UX)
0. we are migrating the dbs, so we don't want to pay the data transfer cost if we don't need the data
0. Keeping less data / smaller indexes is better for the db generally as the indexes use less RAM (more for others) and we shrink our storage space requirements as well.

These above is why i scheduled some time for this work before the db migration, it's been on my list for a long time but wasn't critical or at least folks weren't complaining about it.

Hope this helps, let me know if you want more info.

nciemniak:
Thanks @camallen , this is very helpful :smiley:

PR #3289, opened 2020-04-01
[WIP] Optimize classification exports

camallen:
part of April fools hackday - save the formatted version of a classification for re-use in the classifcation exports to avoid the processing costs. 

WIP PR for me to see the test suite and use for longer term work on this, If i can do 1 hour a week on this i think we can massively speed up the export system with some db storage costs (no indexes)

### TODO
- [x] Actually use the existing formatted resource after initial processing
- [x] Modify the classification to store the PK id to the cached export for fast find (avoid index searches)
- [x] Generalize the export row to polymorphic and json attributes blob (easy schema additions)
- [ ] Ensure dump processor formats the 'secret' columns like IP etc consistently to avoid leaking private data
- [x] Remove the old classification_export_row schema
- [x] Remove classification_export_row code
- [ ] Add feature flag to allow dynamic switching of behaviour (or different job schedules for admin / caching)
- [ ] Test Test Test this improves the export system

# Review checklist


camallen:
closing this, after benchmarking this work** on a db snapshot this work only increased perf by ~4%. The results were similar when not accessing subject information in the export columns as well. 

This means majority of the computational cost is not spent formatting the json data to csv, rather the majority of the time was spent loading the data from the db to the worker in batches. 

A new strategy to optimize the exports would be to split the total export over many concurrent worker processes using https://github.com/mperham/sidekiq/wiki/Batches - this would convert the serial time cost to a concurrent speed up factor (max # of batch export processes running) but with a cost of increased database read pressure.

** https://github.com/camallen/Panoptes/tree/optimize-classification-exports-bench

PR #3299, opened 2020-04-29
revoke access tokens on password change

camallen:
when a user changes a password, invalidate the access tokens for the user and the associated oauth application

Note: this will only revoke the tokens for the requesting oauth application. tokens for another oauth application will not be revoked. We may want to consider revoking all tokens for security purposes (i.e. account takeover and password was reset to fix it) but this may break some workflows like subject uploading. I wasn't really sure the best way to go with this tradeoff.

# Review checklist

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,10 @@ require:
 Style/NumericLiterals:
   Enabled: false
 
+Style/BlockDelimiters:
+  Exclude:
+    - 'spec/**/*'
+
 Layout/LineLength:
   Max: 160
 Layout/SpaceAroundEqualsInParameterDefault:
@@ -15,4 +19,9 @@ Metrics/BlockLength:
   Exclude:
     - '**/*_spec.rb'
 
+Lint/AmbiguousBlockAssociation:
+  Exclude:
+    - "spec/**/*"
 
+RSpec/ExampleLength:
+  Max: 10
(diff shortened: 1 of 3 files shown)


camallen:
> Have we considered adding a button in PFE/an API call to explicitly revoke all tokens, rather than doing it on password change? I agree that this PR is a good start, though, so maybe that's something to add to the list for later.

We can add that as a feature at the API for sure - not in the this PR though. 

This train of though actually might solve the dichotomy of which app tokens to invalidate on signing in this PR. If the user passes a flag 'invalidate all tokens' to password change we can modify the behaviour of this PR to invalidate all apps vs the requesting one. 

If the client UX makes sense and we'd add it I'd be up for modifying this PR or adding it after merge / deploy of this one.

srallen:
We had the initial idea of specifically asking the user if they'd like to sign back in rather than auto do it and we're going to experiment with that UX after this PR merges. I think the suggestion of asking if and which tokens to revoke great idea too, but I agree that any further iteration on this and its UX should be done in increments in follow up PRs. I think this UX would mean we would have to store information about the token, such as browser, OS, location authenticated sessions? (also mentioning @beckyrother about the UX suggestion)

beckyrother:
From a UX perspective I'm not sure I've seen the option to sign back in after changing a password. I'd suggest revoking the token everywhere and requiring people re-sign in – that feels the most secure on the users' end.

PR #3310, opened 2020-05-19
Switch from URL to pieces

zwolf:
Env vars can't contain `:`, as is required by a URL. A full URL, even as a quoted string, doesn't seem to make it as a set env var. 

Switching to pieces (host/username/password) bypasses this issue.

# Review checklist

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,7 +1,9 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  url: <%= ENV.fetch('DATABASE_URL', 'postgresql://panoptes:panoptes@localhost') %>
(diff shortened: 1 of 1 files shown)


adammcmaster:
Looks like the newer PRs supersede this.

zwolf:
@camallen 

>ENV vars can contain : chars, the caeasr database URL contains one

Yeah, I know, but they don't load into a secret from an env file without enclosing quotes. The problem is that the DATABASE_URL var isn't loading into the pod. 

>I don't think we should split these out, they should be left as the URL version.

Why?

>Also - you'll need a db config *_standby setup for the read replica for envs and the code that reads from replica, https://github.com/kenn/standby#install

Why? Staging has no replica and no separate readonly user.

PR #3325, opened 2020-05-28
Remove X-Forwarded-Host header

adammcmaster:
Carrying over zooniverse/operations#283

Describe your change here.

# Review checklist


adammcmaster:
I'm confused – is this broken in the current AWS deployment in that case?

camallen:
I'm pretty sure i reverted it in, https://github.com/zooniverse/operations/pull/337#issue-380814156 and closed it without merge for prod in https://github.com/zooniverse/operations/pull/284

Stupidly i left the current header in there instead of cleaning up, apologies. What is current behaviour is the nginx proxy is setting the response header from the upstream (OMFG) not setting the request header to the upstream, see https://github.com/zooniverse/operations/pull/297

camallen:
Leave this with me and I'll work on it for the staging API migration

kubernetes/deployment-staging.tmpl
```diff
@@ -47,6 +47,9 @@ data:
       }
 
       location / {
+        # avoid attacker setting the host value for links in rails_app
+        # https://github.com/rails/rails/issues/29893
+        proxy_set_header X-Forwarded-Host $server_name;
```
camallen, line 52: I believe this will set the `X-Forwarded-Host` to the server block defined in the `server_name panoptes-staging.zooniverse.org;` based on reading at http://nginx.org/en/docs/http/ngx_http_core_module.html#var_server_name

it would be great to double check this and test on staging system before deploying to production.

kubernetes/deployment-production.tmpl
```diff
@@ -47,6 +47,58 @@ data:
       }
 
       location / {
+        # avoid attacker setting the host value for links in rails_app
+        # https://github.com/rails/rails/issues/29893
+        proxy_set_header X-Forwarded-Host $server_name;
```
adammcmaster, line 52: Do we definitely want to use `$server_name` rather than `$host` here?

adammcmaster: Or maybe `$http_host`? (It's confusing that there are so many similar variables...)

camallen: I wasn't 100% tbh as my reading suggests that using `$host` or `$http_host` will take it from the request line, then the http header http://nginx.org/en/docs/http/ngx_http_core_module.html#var_server_name
>$host
    in this order of precedence: host name from the request line, or host name from the “Host” request header field, or the server name matching a request 

I beleive that means it'll match the host from the request URI as per https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html

But it may fall back to http header which is the situation we're trying to avoid. I believe using `$server_name` explicitly avoids this scenario but does mean our nginx configs have to be match our DNS entries for the target service...perhaps not ideal. 

Thoughts?

adammcmaster: Ah, ok I get it now. I was misunderstanding what we were trying to do. If we're specifically trying not to pass the user-supplied hostname then `server_name` sounds right.

adammcmaster:
I think this is ready for merge now. @camallen do you agree?

kubernetes/deployment-staging.tmpl
camallen: I've manually tested this on the staging api nginx container via 
`curl -v 'https://panoptes-staging.zooniverse.org/oauth/authorize' -H 'Host: www.test.com'`

PR #3326, opened 2020-05-28
Fix race condition when setting up assets

adammcmaster:
The "cp" command runs before the assets finish compiling.

Mount the api-assets directory directly as a volume, and precompile
assets on start if the directory is empty.

Describe your change here.

# Review checklist

### DIFF
--- a/scripts/docker/start.sh
+++ b/scripts/docker/start.sh
@@ -15,7 +15,7 @@ rm -f tmp/pids/*.pid
 if [ "$RAILS_ENV" == "development" ]; then
   exec bundle exec foreman start
 else
-  if [ ! -d public/assets ]
+  if [ ! -d public/api-assets ] || [ "$(ls -A public/api-assets)" == "" ]
   then
       bundle exec rake assets:precompile
   fi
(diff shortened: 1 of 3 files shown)


scripts/docker/start.sh
```diff
@@ -15,7 +15,7 @@ rm -f tmp/pids/*.pid
 if [ "$RAILS_ENV" == "development" ]; then
   exec bundle exec foreman start
 else
-  if [ ! -d public/assets ]
+  if [ ! -d public/api-assets ] || [ "$(ls -A public/api-assets)" == "" ]
```
camallen, line 18: I think i'd prefer we switch to building these assets in the prod / staging Dockerfile rather than on boot, we need to ensure we have short boot / deploy times before the container is active.

adammcmaster: If I remember correctly we tried that once, but it fails because Rails tries to load the secrets.

camallen: I've tried this in #3329 copied from here, so seems to work, https://github.com/zooniverse/caesar/blob/8d8b5d31d65b84a47e30cea5ccd77776ca5b4549/Dockerfile#L24

PR #3329, opened 2020-05-28
Improve container startup times 

camallen:
Improve container startup times: 
1. avoid building at startup of the container
0. setup the tmp pid dirs at build time

## TODO
- [x] test this builds and works correctly locally

# Review checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,8 @@ ADD supervisord.conf /etc/supervisor/conf.d/panoptes.conf
 ADD ./ /rails_app
 
 RUN (cd /rails_app && git log --format="%H" -n 1 > commit_id.txt)
+RUN (cd /rails_app && mkdir -p tmp/pids && rm -f tmp/pids/*.pid)
+RUN (cd /rails_app && SECRET_KEY_BASE=1 bundle exec rake assets:precompile)
 
 EXPOSE 81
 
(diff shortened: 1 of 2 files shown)


adammcmaster:
I think we'll need to put the `cp` command back on startup in Kubernetes, otherwise it'll mount an empty volume over the compiled assets.

camallen:
> I think we'll need to put the `cp` command back on startup in Kubernetes, otherwise it'll mount an empty volume over the compiled assets.

I assume we can do that on the container startup lifecycle? https://github.com/zooniverse/caesar/blob/8d8b5d31d65b84a47e30cea5ccd77776ca5b4549/kubernetes/deployment-production.tmpl#L182

adammcmaster:
Yup – actually, if we just don't merge #3326 it should be fine.

camallen:
> Yup – actually, if we just don't merge #3326 it should be fine.

Up to you, we can revert that later after this is tested properly.

adammcmaster:
I've closed it – if you and Zach could test/deploy this next week that'd be good.

adammcmaster:
I think this is actually blocking staging deploys on K8, because it takes long enough that the liveness probe kills the container.

camallen:
I've tested this locally and it's working. Much faster on boot times, I'll take responsibility for fixes in Azure staging env.

PR #3330, opened 2020-05-28
Mount doorkeeper keys from secret in a config subdir

zwolf:
Load doorkeeper keys from a kubernetes secret. There were a few different ways to do this that worked around how k8s clobbers everything in the directory on which you mount a volume.

* Mount the secret using subPaths. This will not update the container if the secret's changed, which isn't a big deal here but this method also seems to be frowned upon as micromanaging k8s volumes unnecessarily.
* Mount the secret in some other place and copy it over in a postStart lifecycle hook. This runs the hook async with the entrypoint so there's no guarantee that the app won't try to start before the copying is finished.
* Use an init container. This is a cool pattern I just learned about, but in this case I’m not sure how it could access the yet-unstarted app container’s filesystem to do the copying. Must investigate further.
* Mount the secret to an unused directory (`config/keys`) and let k8s clobber it. Then just load up the keys from where they are wherever they need to be.

I went to with the last one because it seemed the neatest, even though I didn't really want to touch app code. Keys are loaded as secrets, one per env.

This will definitely prevent proper deploying now that I'm altering paths. I'm still setting this to merge to master because I think we're beyond the point of no return on that one? I can merge to a different k8s-only branch if preferred. 

# Review checklist


zwolf:
From where I'm sitting adding five characters to a path is simpler than using an additional k8s template feature. I came to this conclusion by googling around and seeing how many bugs, blog posts, and confusion were related to `subPath` and from seeking advice elsewhere. Kubernetes intends a volume mount to be a directory, and setting a subPath for an individual file feels like a workaround, which tend to be fragile.

I evaluated the other options, so consider me the one doing the frowning. If that's not sufficient, let me know that you're withholding approval and I'll reevaluate with that in mind.

.gitignore
```diff
@@ -19,6 +19,8 @@
 /config/aws_ips.json
 /config/doorkeeper-jwt*.pem
 /config/doorkeeper-jwt*.pub
+/config/keys/doorkeeper-jwt*.pem
+/config/keys/doorkeeper-jwt*.pub
```
camallen, line 23:
```suggestion
/config/keys/doorkeeper-jwt*.pem
/config/keys/doorkeeper-jwt*.pub
```
Remove the old location for the signing files

adammcmaster:
There aren't any outstanding comments so I will merge this once the status checks complete.

PR #3335, opened 2020-06-04
Upgrade restpack kaminari

camallen:
update to the latest serializer code changes in https://github.com/zooniverse/restpack_serializer/pull/1

Fixes XSS CVE, GHSA-r5jw-62xg-j433 by bumping kaminari

# Review checklist


camallen:
> kaminari 1.0 drops rails 4 support https://github.com/kaminari/kaminari/blob/master/CHANGELOG.md#100
> 
> So I'm not sure this is compatible?

Indeed that entry in the changelog got me to, however it's only rails 4.0 support that was dropped, https://github.com/kaminari/kaminari/commit/9e97a36a5d425559a2e7dbf73bf1774575597175#diff-ac46a8b5a40536c4171a5ebee5aca2d4
and we can see that it's still being tested for rails 4.1 & 4.2 https://github.com/kaminari/kaminari/tree/cd8601cc42b67267c15a13174bc2fc9bd5de1032/gemfiles

PR #3340, opened 2020-06-11
skip dump worker rate limit for special users

camallen:
allow list of user ids to be set as ENV vars that will override the congestion rate limits for all dump exports

This is a workaround for 1715 labs projects and their use case is to envisaged to allow some special project owners to request data more frequently than the coarse rate limits apply for. 

Long term this will be combined with the work #3289 to reduce the rate limit interval for all users and this special ids check can be removed. 

# Review checklist


app/workers/concerns/rate_limit_dump_worker.rb
```diff
@@ -12,8 +14,25 @@ module RateLimitDumpWorker
...
-          requester_id.blank? || !User.find(requester_id).admin?
+          congestion_enabled?(requester_id)
         }
       }
   end
+
+  module ClassMethods
+    def congestion_enabled?(requester_id)
+      # if the user is missing, should only happen via the rails console
+      return false if requester_id.blank?
```
adammcmaster, line 25: If I'm reading this correctly this inverts the logic, is that intended? The previous code would have returned `true` if the ID was blank.

Since that only happens via the rails console, I assume this is intentional since we won't want to rate limit ourselves.

app/workers/concerns/rate_limit_dump_worker.rb
```diff
@@ -12,8 +14,25 @@ module RateLimitDumpWorker
...
+
+      if (skip_id_env = ENV.fetch('SKIP_DUMP_RATE_LIMIT_USER_IDS', nil))
```
adammcmaster, line 27: Should we parse this env var in an initialiser to keep all of them in the same place?

app/workers/concerns/rate_limit_dump_worker.rb
```diff
@@ -12,8 +14,25 @@ module RateLimitDumpWorker
...
+        skip_ids = skip_id_env.split(',').map(&:to_i)
+        skip_ids.include?(requester_id)
+        # false if the user is a special skip rate limit user
+        false
```
adammcmaster, line 31: I don't quite follow how this part works, maybe just because Ruby's implicit returns confuse me. Does execution only get as far as line 31 if line 29 evaluates to `true`? How does this return `true` if the user isn't a special user?

app/workers/concerns/rate_limit_dump_worker.rb
camallen: Yes - strangely this was never tested properly and never worked as expected. Returning true previously would have enabled the congestion middleware and thus the rate limiting.

app/workers/concerns/rate_limit_dump_worker.rb
camallen: This looked like a bug, i've modified to simplify the logic by using initializers to extract the skip user ids

app/workers/concerns/rate_limit_dump_worker.rb
camallen: I've extracted these - thanks for the suggestion 👍

PR #3342, opened 2020-06-12
Revert "allow azure zooniverse.org subdomain"

camallen:
Reverts zooniverse/panoptes#3341 - remove this addition once we switch staging to azure k8s cluster


camallen:
this addition in #3341 wasn't needed, turns out the issue was with a missing / misspelt ENV var fixed in #3343

PR #3345, opened 2020-06-12
avoid extra outer regex forward slashes

camallen:
just pass a string without the outer slashes as they are being created at https://github.com/zooniverse/panoptes/blob/d7463caac4c97e29ce058dceac9adf5da1a5014b/config/initializers/cors.rb#L23

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -178,7 +178,7 @@ spec:
                   name: panoptes-common-env-vars
                   key: AWS_SECRET_ACCESS_KEY
             - name: CORS_ORIGINS_REGEX
-              value: /^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|lab|learn)\.wildcamgorongosa\.org|www\.diagnosislondon\.org|anno\.tate\.org\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$/
+              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|lab|learn)\.wildcamgorongosa\.org|www\.diagnosislondon\.org|anno\.tate\.org\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
             - name: DESIGNATOR_API_HOST
               value: https://designator.zooniverse.org/
             - name: DESIGNATOR_API_USERNAME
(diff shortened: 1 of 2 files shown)


kubernetes/deployment-staging.tmpl
```diff
@@ -128,7 +128,7 @@ spec:
             - name: DUMP_CONGESTION_OPTS_MIN_DELAY
               value: "60"
             - name: CORS_ORIGINS_REGEX
-              value: /^https?:\/\/(127\.0\.0\.1|localhost|[a-z0-9-]+\.local|10\.[0-9]+\.[0-9]+\.[0-9]+|192\.168\.[0-9]+\.[0-9]+|[a-z0-9-]+\.zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$/
+              value: '^https?:\/\/(127\.0\.0\.1|localhost|[a-z0-9-]+\.local|10\.[0-9]+\.[0-9]+\.[0-9]+|192\.168\.[0-9]+\.[0-9]+|[a-z0-9-]+\.zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
```
camallen, line 131: I'd like to streamline this to just our subdomains and `local.zooniverse.org` for our developers...
IMO it could change to
`^https?:\/\/([a-z0-9-]+\.zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?`

kubernetes/deployment-production.tmpl
```diff
@@ -178,7 +178,7 @@ spec:
                   name: panoptes-common-env-vars
                   key: AWS_SECRET_ACCESS_KEY
             - name: CORS_ORIGINS_REGEX
-              value: /^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|lab|learn)\.wildcamgorongosa\.org|www\.diagnosislondon\.org|anno\.tate\.org\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$/
+              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|lab|learn)\.wildcamgorongosa\.org|www\.diagnosislondon\.org|anno\.tate\.org\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
```
camallen, line 181: I'd like to remove old domains here as well I believe we can remove
- all wildcamgorongosa.org domains
- www\.diagnosislondon\.org
- anno\.tate\.org\.uk

Can we remove the older old weather apps?

wgranger:
This will change how we run some frontend apps locally. I know there are good instructions on setting up `local.zooniverse.org` [here](https://stackoverflow.com/c/zooniverse/questions/109). Up for debate if we want to put these steps in a more public place, but I'll be linking to this post as new repos are created.

camallen:
> st as new

OOI - if we keep `127\.0\.0\.1|localhost` 
but remove
`[a-z0-9-]+\.local|10\.[0-9]+\.[0-9]+\.[0-9]+|192\.168\.[0-9]+\.[0-9]+`
Do folks use any of the above?

PR #3350, opened 2020-06-15
Add Redis and Sidekiq

### LINKED ISSUE #3314: How does sidekiq start in K8s env?
It appears that docker will use the default supervisor config in k8s, hence we will only have the puma app running, https://github.com/zooniverse/Panoptes/blob/5cbaf621f99799880112e7304e9b4016ff78343d/supervisord.conf

I suggest that we reconfigure how the app boots and is deployed in k8s
1. remove the dependency on supervisor
0. start the puma app by default via dockerfile CMD syntax
0. Add a deployment config for sidekiq, https://github.com/zooniverse/caesar/blob/e86609e8a7b1111f54b6283875f3cd49be8de8f7/kubernetes/deployment-staging.tmpl#L194
0. modify how docker-compose starts the dev env, add a sidekiq container instead of booting multiple processes in one container.

adammcmaster:
Starting with adding Redis. Outstanding questions before we add Sidekiq itself:

- [x] What storage service should we use for Sidekiq (discussion in zooniverse/operations#395)?
- [x] How should we handle config/environment variables? Will the big list of variables/values in the API deployment template need to be duplicated? Is there a better option?

cc @camallen @zwolf 

Closes #3314 and zooniverse/operations#395.

# Review checklist


adammcmaster:
Also related to zooniverse/operations#392

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
...
       targetPort: 80
   type: NodePort
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: panoptes-production-redis
+spec:
+  accessModes:
+    - ReadWriteMany
```
camallen, line 358: don't we want `ReadWriteOnce` - this volume would only be used by redis..

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
       port: 80
       targetPort: 80
   type: NodePort
+---
+kind: PersistentVolumeClaim
```
camallen, line 352: I think we need to add a specific reclaim that ensures this disk doesn't get deleted accidentally (upgrading the node etc) policy https://kubernetes.io/docs/concepts/storage/persistent-volumes/#retain

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
...
+apiVersion: v1
+metadata:
+  name: panoptes-production-redis
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: azurefile
```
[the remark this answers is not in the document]
camallen: Or maybe raw block store ? https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volume-using-a-raw-block-volume

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
...
+    spec:
+      containers:
+        - name: panoptes-production-redis
+          image: redis:6.0.5
+          resources:
+            requests:
+              memory: "100Mi"
+              cpu: "10m"
+            limits:
+              memory: "100Mi"
```
camallen, line 390: this will need to be much higher

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
...
+              cpu: "500m"
```
camallen, line 391: for production this will need to be at least 1 or 1.5 vcpu

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
...
+      labels:
+        app: panoptes-production-redis
+    spec:
+      containers:
+        - name: panoptes-production-redis
+          image: redis:6.0.5
+          resources:
+            requests:
+              memory: "100Mi"
+              cpu: "10m"
```
camallen, line 388: these will need to change.

kubernetes/deployment-production.tmpl
```diff
@@ -348,3 +348,64 @@ spec:
...
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: azurefile
+  resources:
+    requests:
+      storage: 1Gi
+---
+apiVersion: apps/v1
+kind: Deployment
```
camallen, line 365: Should we deploy redis as a stateful set for consistent volume / network identifiers? 

https://schoolofdevops.github.io/ultimate-kubernetes-bootcamp/13_redis_statefulset/
https://rancher.com/blog/2019/deploying-redis-cluster
https://estl.tech/deploying-redis-with-persistence-on-google-kubernetes-engine-c1d60f70a043

kubernetes/deployment-production.tmpl
adammcmaster: My understanding is that the disk would only be deleted if we explicitly delete the persistentvolumeclaim (i.e. actually running `kubectl delete`). Nodes being replaced or us removing the PVC from the deployment template wouldn't delete the PVC.

I'm happy to change the reclaim policy if we think it's necessary, but if we actually want to delete the volume in the future (e.g. if we move the data out of K8) we'd need to remember to manually clean up the data from the volume.

kubernetes/deployment-production.tmpl
adammcmaster: I'm not familiar with stateful sets, so to avoid any surprises I'd lean towards using a deployment for now and maybe revisiting this later (because of the urgency of getting this finished). A deployment with one replica will do what we need as far as I can see.

kubernetes/deployment-production.tmpl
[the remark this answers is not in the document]
adammcmaster: I think the question is how much throughput do we need? It looks like azurefile gives us 1,000 IOPS per file (standard), 5,000 IOPS per file (premium, though with a minimum volume size of 100 GB so not worth the cost). https://docs.microsoft.com/en-us/azure/storage/files/storage-files-scale-targets#file-level-limits

If I'm reading this all correctly, we'd have to provision a pretty large disk to get the same performance: https://azure.microsoft.com/en-gb/pricing/details/managed-disks/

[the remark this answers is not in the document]
adammcmaster: I could easily be missing something here though. The range of options is kind of overwhelming.

adammcmaster:
> I think we need to consider how we will backup the persistent disk as well, https://kubernetes.io/docs/concepts/storage/volume-snapshots/#volumesnapshots

This is something that will need to be set up for all of our volumes, so I've added a card here for it https://github.com/zooniverse/operations/projects/1#card-40267083

> And how other folks might be deploying HA redis? https://github.com/helm/charts/tree/master/stable/redis-ha
> 
> The staging file will need to be updated to mirror the production changes for storage class, size reclaim policy, limits etc.

I've made some updates to the staging config, but I haven't mirrored the prod config exactly. I've had a look at the two existing redis instances to judge what limits to set.

> Perhaps we should look at redis enterprise offering for managed HA redis? https://redislabs.com/redis-enterprise-cloud/pricing/

Something to consider later.

camallen:
pinging @adammcmaster - i've updated with latest from master and added some fixes for CORS, Sidekiq and redis. Mind having a look and i'll do the same.

kubernetes/deployment-staging.tmpl
```diff
@@ -288,6 +203,107 @@ spec:
...
+      containers:
+        - name: panoptes-staging-sidekiq
+          image: zooniverse/panoptes:__IMAGE_TAG__
+          resources:
+            requests:
+              memory: "700Mi"
+              cpu: "50m"
+            limits:
+              memory: "700Mi"
+              cpu: "1000m"
```
camallen:
```suggestion
              memory: "700Mi"
              cpu: "50m"
            limits:
              memory: "700Mi"
              cpu: "500m"
```

Doesn't look like this is very busy at all

kubernetes/deployment-staging.tmpl
```diff
@@ -288,6 +203,107 @@ spec:
...
+    matchLabels:
+      app: panoptes-staging-redis
+  template:
+    metadata:
+      labels:
+        app: panoptes-staging-redis
+    spec:
+      containers:
+        - name: panoptes-staging-redis
+          image: redis:6.0.5
```
camallen, line 278: https://github.com/zooniverse/operations/blob/3e30539794fd4983e6be7a0726f8b763b280ef54/panoptes-redis/docker-compose.yml#L3

will this image need to add a start script for running with persistence? https://hub.docker.com/_/redis/
```
start with persistent storage

$ docker run --name some-redis -d redis redis-server --appendonly yes

If persistence is enabled, data is stored in the VOLUME /data, which can be used with --volumes-from some-volume-container or -v /docker/host/dir:/data (see docs.docker volumes).
```
For more about Redis Persistence, see http://redis.io/topics/persistence.

And maybe a custom conf? https://github.com/zooniverse/operations/blob/3e30539794fd4983e6be7a0726f8b763b280ef54/panoptes-redis/docker-compose.yml#L8

kubernetes/deployment-production.tmpl
```diff
@@ -344,6 +253,107 @@ spec:
...
+    matchLabels:
+      app: panoptes-production-redis
+  template:
+    metadata:
+      labels:
+        app: panoptes-production-redis
+    spec:
+      containers:
+        - name: panoptes-production-redis
+          image: redis:6.0.5
```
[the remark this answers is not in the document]
adammcmaster: Persistence will work out of the box (see e.g. [lita's redis deployment](https://github.com/zooniverse/lita/blob/master/kubernetes/redis.yaml)), but good point – if we want to use an AOF we'll need to add `args: ['--appendonly yes']`

That config file only contains `appendfsync no`. I'll need to read up on that to see if we want to keep it.

kubernetes/deployment-production.tmpl
```diff
@@ -344,6 +253,109 @@ spec:
...
+          image: zooniverse/panoptes:__IMAGE_TAG__
+          resources:
+            requests:
+              memory: "1500Mi"
+              cpu: "500m"
+              ephemeral-storage: "50Gi"
+            limits:
+              memory: "1500Mi"
+              cpu: "1000m"
+              ephemeral-storage: "50Gi"
```
camallen, line 280:
```suggestion
              memory: "4000Mi"
              cpu: "1000m"
              ephemeral-storage: "50Gi"
            limits:
              memory: "4000Mi"
              cpu: "2000m"
              ephemeral-storage: "50Gi"
```
aws ec2 metrics suggest this bursts ~3.25GB, 4GB might be overkill but i'd like to avoid OOMKiller on these.

Assuming we have bigger node types CPU can burst to 2vcpus - aws metrics shows ~75% of 2vcpus most of the time.

kubernetes/deployment-production.tmpl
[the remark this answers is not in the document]
adammcmaster: I think we added that to stop it blocking on slow writes to EFS. We'll probably want that here too.

PR #3351, opened 2020-06-16
remove old CORS allow list entries

camallen:
Extracted from #3345 

Remove local dev work ip based entries in allow list, folks switch to `local.zooniverse.org` or `myapp.zooniverse.org` with local DNS (/etc/hosts) overrides.

Remove old CFE apps that are already redirecting to www.zooniverse.org

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -171,7 +171,7 @@ spec:
                   name: panoptes-common-env-vars
                   key: AWS_SECRET_ACCESS_KEY
             - name: CORS_ORIGINS_REGEX
-              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
+              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|relaunch|field-book)\.notesfromnature\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
             - name: DESIGNATOR_API_HOST
               value: https://designator.zooniverse.org/
             - name: DESIGNATOR_API_USERNAME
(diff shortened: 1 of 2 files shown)


kubernetes/deployment-production.tmpl
```diff
@@ -171,7 +171,7 @@ spec:
                   name: panoptes-common-env-vars
                   key: AWS_SECRET_ACCESS_KEY
             - name: CORS_ORIGINS_REGEX
-              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|lab|learn)\.wildcamgorongosa\.org|www\.diagnosislondon\.org|anno\.tate\.org\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
+              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|relaunch|field-book)\.notesfromnature\.org|[a-z0-9-]+\.oldweather\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
```
camallen, line 174: Can we remove `*.oldweather.org` entries? Or be more explicit in the allow list? 

Can any other entries be removed?

wgranger: It looks like oldweather.org is only an archive page. Doesn't look like it needs Panoptes.

We still need ASM and Scribes of the Cairo Geniza.

wgranger: There's also "classic.oldweather.org," which gives a ngix error, and "whaling.oldweather.org," which is also an archived page

camallen: thanks for doing the leg work will, i just tested https://classic.oldweather.org/ and it uses oauth redirection flow to login so doesn't need cors access. I'll remove it and others :)

camallen:
```suggestion
              value: '^https?:\/\/((www|staging)\.antislaverymanuscripts\.org|(www|staging)?\.?heirtagger\.ox\.ac\.uk|(www|relaunch|field-book)\.notesfromnature\.org|www\.scribesofthecairogeniza\.org|([a-z0-9-]+\.)?zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$'
```

adammcmaster:
Just noting that this will conflict with #3350 since that refactors the env vars. Rather than merging this it might be easier to just add a new commit to the other PR.

PR #3352, opened 2020-06-16
Add memcached

adammcmaster:
Adds a memcached service, renames the cache class and config env var to not mention ElastiCache.

I've set the container's memory limit to 100 MB, and the runtime memory limit in memcached to 90 MB. The current production cache seems to be consistently using less than half this amount:

Screenshot 2020-06-16 at 16 36 26

# Review checklist

### DIFF
--- a/config/cache_store.rb
+++ b/config/cache_store.rb
@@ -1,9 +1,7 @@
 module Panoptes
   module Cache
-    def self.client
-      if cluster_name = ENV["CACHE_SERVER"]
-        Dalli::Client.new("#{cluster_name}:11211")
-      end
+    def self.enabled?
+      ENV.key?('MEMCACHE_SERVERS')
     end
 
     def self.options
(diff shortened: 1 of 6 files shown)


config/cache_store.rb
```diff
@@ -1,8 +1,8 @@
 module Panoptes
-  module ElastiCache
+  module Cache
     def self.client
-      if cluster_name = ENV["ELASTIC_CACHE"]
-        Dalli::ElastiCache.new("#{cluster_name}:11211")
+      if cluster_name = ENV["CACHE_SERVER"]
+        Dalli::Client.new("#{cluster_name}:11211")
```
camallen, line 5: Remove this `self.client` method and change this to use `ENV["MEMCACHE_SERVERS"]` 
https://github.com/petergoldstein/dalli#usage-with-rails-3x-and-4x
> If your servers are specified in ENV["MEMCACHE_SERVERS"] (e.g. on Heroku when using a third-party hosted addon), simply provide nil for the servers:

PR #3353, opened 2020-06-17
Remove supervisor fix app start and logging

camallen:
Remove supervisor for k8s deployments and ensure webserver & sidekiq apps can start and log properly. It will help if this comes in before #3350 to help with logging and sidekiq app startups

in the process i had to :
1. remove all supervisor configs, foreman & procfile for dev env multiprocess
0. fix the app boot, added custom sidekiq start script (tested locally with panoptes & sidekiq docker-compose containers)
0. switch logging to std out, now the default on dev but with opt out via env var.
0. modify grace period for sidekiq (25s) to link with k8s default grace period at 30s
0. remove coming deprecations / unused sidekiq env configs or ones that match defaults
0. move staging and production to only log to stdout via `logstasher`  json format and not log to files, https://github.com/shadabahmed/logstasher

# Review checklist


config/sidekiq.yml
```diff
@@ -3,11 +3,11 @@
 # sidekiq -C config.yml
 ---
 :verbose: <%= ENV.fetch('SIDEKIQ_VERBOSE', false) %>
```
adammcmaster, line 5: Similar question to the one I asked for Caesar – does this need `.to_bool`?

camallen: no - in ruby everything is `truthy` or `falsey`, e.g 
```
irb
> ENV.fetch('SIDEKIQ_VERBOSE', false)
 => false 
2.5.3 :008 > puts "I am truthy" if ENV.fetch('SIDEKIQ_VERBOSE', false)
 => nil 
2.5.3 :002 > ENV['SIDEKIQ_VERBOSE'] = 'true'
 => "true" 
2.5.3 :003 > ENV.fetch('SIDEKIQ_VERBOSE', false)
 => "true" 
2.5.3 :005 > !!ENV.fetch('SIDEKIQ_VERBOSE', false)
 => true 
2.5.3 :006 > puts "I am truthy" if ENV.fetch('SIDEKIQ_VERBOSE', false)
I am truthy
 => nil 
```

camallen: and sidekiq cli will parse this into the correct value - that said it will make it conformant to the default.
```suggestion
:verbose: <%= !!ENV.fetch('SIDEKIQ_VERBOSE', false) %>
```

PR #3358, opened 2020-06-19
add aws env vars to storage configuration

camallen:
ensure AWS key and secret make it to the s3 client configuration options. 

# Review checklist


camallen:
closing this in favour of configuration via env variables.

PR #3365, opened 2020-06-22
account for favorite collections in other projs

nciemniak:
`OneFavoritePerOwnerValidator` was erronously checking to see if the user had any favorite collections across all projects, rather than just for the project for which the new collection was being created. Update code to fix this bug.

----

# Review checklist


lib/validators/one_favorite_per_owner_validator.rb
```diff
@@ -9,8 +9,16 @@ class OneFavoritePerOwnerValidator < ActiveModel::Validator
...
-      owner_has_existing_favorite = new_record.owner.collections.where(favorite: true).exists?
-      return unless owner_has_existing_favorite
+      owner_favorite_collections = new_record.owner.collections.includes(:projects).where(favorite: true)
+      owner_has_existing_fav_for_project = false
+
+      owner_favorite_collections.each do |collection|
+        if collection.project_ids.include?(new_record.project_ids.first)
+          owner_has_existing_fav_for_project = true
+        end
+      end
```
camallen, line 19: This query will be expensive for a user that has a lot of fav collection, it'll load them all from the db and iterate over them. FWIW best to use `.find_each` https://apidock.com/rails/ActiveRecord/Batches/find_each

I think we can modify this query to check if any of the owned fav collections have the same project ids as the one we are creating for.
```suggestion
      owner_fav_collections = new_record.owner.collections.where(favorite: true)
      owner_has_existing_fav_for_project = owner_fav_collections
        .joins(:projects)
        .where(projects: { id: new_record.project_ids })
        .exists?
```

nciemniak: Using `ActiveRecord::Base.logger = Logger.new(STDOUT)` and it is very helpful.
So here's a very cool thing! Because I used `.includes` in this line here:
```ruby
new_record.owner.collections.includes(:projects).where(favorite: true)
```
rails loads up the collections and associated projects when `owner_favorite_collections` is accessed,
```
D, [2020-06-24T16:59:44.967289 #52] DEBUG -- :   Collection Load (0.7ms)  SELECT "collections".* FROM "collections" INNER JOIN "access_control_lists" ON "collections"."id" = "access_control_lists"."resource_id" WHERE "access_control_lists"."user_group_id" = 31310 AND (roles && '{owner}') AND "access_control_lists"."resource_type" = 'Collection' AND "collections"."favorite" = 't'  [["user_group_id", 31310], ["resource_type", "Collection"], ["favorite", true]]
D, [2020-06-24T16:59:44.972837 #52] DEBUG -- :   HABTM_Projects Load (0.6ms)  SELECT "collections_projects".* FROM "collections_projects" WHERE "collections_projects"."collection_id" IN (1490, 1491)
D, [2020-06-24T16:59:44.980413 #52] DEBUG -- :   Project Load (0.7ms)  SELECT "projects".* FROM "projects" WHERE "projects"."id" IN (15785, 15787)
```
and then does not make any subsequent db calls at this line `if collection.project_ids.include?(new_record.project_ids.first)`.

That being said, I would say my solution isn't necessarily all that expensive on the db (tell me if I'm wrong), but I would say that your solution does seem cleaner to me... so I'll go ahead and switch to your suggestion.

camallen:
@nciemniak what's the status of this PR?

PR #3368, opened 2020-06-23
Add sidekiq liveness probe

camallen:
add a liveness exec cmd that uses a custom sidekiq bin (ruby) script for sidekiq process state

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command

Ruby bin script is a modified version of sidekiqctl (removed in v6), https://github.com/mperham/sidekiq/blob/74ccba6c68b1df31d615991fb2749fc19de8fbf7/bin/sidekiqctl

(removed in v6) https://github.com/mperham/sidekiq/blob/master/6.0-Upgrade.md

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -279,6 +279,11 @@ spec:
               memory: "4000Mi"
               cpu: "2000m"
               ephemeral-storage: "50Gi"
+          livenessProbe:
+            exec:
+              command:
+                - /rails_app/scripts/docker/sidekiq_status
+            initialDelaySeconds: 20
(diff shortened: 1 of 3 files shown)


camallen:
> I get this error when I try to test this locally:
> 
> ```
> $ docker-compose exec sidekiq /rails_app/scripts/docker/sidekiq_status
> ---- Processes (1) ----
> a573c7de0f14:1:0f85677aad15 [rails_app]
>   Started: 2020-06-23 12:05:24 +0000 (just now)
>   Threads: 1 (0 busy)
> Couldn't get status: can't modify frozen String
> ```

thanks - i've fixed the frozen string error now and tested via docker-compose.

PR #3373, opened 2020-06-23
graceful nginx shutdown prehook

camallen:
surprisingly nginx handles TERM (default k8s pod signal) to fast shutdown, http://nginx.org/en/docs/control.html

This can lead to dropped connections, let's avoid that and try to shutdown gracefully, https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html

# Review checklist


camallen:
> Could you copy this to the prod template too?

I'd like to test this on staging before deploying to prod, will get a follow up PR for prod in.

PR #3376, opened 2020-06-24
Add database migration jobs

adammcmaster:
And tidy up slack notifications.

The idea is that staging will always run the migration on merge, and for production we will add a new `production-migrate` tag (I'm open to suggestions if anyone would prefer a different name; `production-database`? Something else?). 

We will then use Lita to trigger a new Jenkins job similar to the existing deploy command; this command should check out the `production-release` tag and then move the `production-migrate` tag to the same commit.

# Review checklist


adammcmaster:
I think we might also need to manually delete the job once it completes successfully. See https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup. Maybe we add another Lita command to handle that too.

Jenkinsfile
```diff
@@ -93,29 +78,37 @@ pipeline {
...
-              slackSend (
-                color: '#00FF00',
-                message: "SUCCESSFUL: Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]' (${env.BUILD_URL})",
-                channel: "#ops"
-              )
-            }
-          }
+    }
+
+    stage('Migrate staging database') {
```
camallen, line 83: this will have to go before the staging deploy or combine into one staging stage with a step to migrate and a step to deploy.

kubernetes/job-db-migrate-production.tmpl
```diff
@@ -0,0 +1,20 @@
...
+        command: ["/rails_app/migrate.sh"]
+        envFrom:
+        - secretRef:
+            name: panoptes-common-env-vars
+        - secretRef:
+            name: panoptes-production-env-vars
+        - configMapRef:
+            name: panoptes-production-shared
+      restartPolicy: Never
+  backoffLimit: 4
```
camallen, line 20: I was assuming this would be 1. Are we retrying 4 times in case of transient failures? if so should we stop after 2?

kubernetes/job-db-migrate-staging.tmpl
```diff
@@ -0,0 +1,20 @@
...
+        command: ["/rails_app/migrate.sh"]
+        envFrom:
+        - secretRef:
+            name: panoptes-common-env-vars
+        - secretRef:
+            name: panoptes-staging-env-vars
+        - configMapRef:
+            name: panoptes-staging-shared
+      restartPolicy: Never
+  backoffLimit: 4
```
camallen, line 20: I was assuming this would be 1. Are we retrying 4 times in case of transient failures? if so should we stop after 2?

kubernetes/job-db-migrate-production.tmpl
adammcmaster: It uses an [exponential backoff interval](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy), so I think it makes sense to try a few times in case of temporary problems that take a bit longer to resolve themselves. Is there a down side to trying multiple times? I guess it's not ideal to retry a large migration that puts a heavy load on the DB for a while before failing.

adammcmaster:
> wait for the job to complete kubectl wait --for=condition=complete --timeout=1200s job/panoptes-migrate-db-staging # not sure about the timeout here?

Yeah, 20 minutes doesn't seem long enough. Maybe set it to a day? We basically want it to wait forever because some DB migrations could take a long time.

I think "complete" and "failed" are two separate states, so this will always wait until the timeout on failure I think.

kubernetes/job-db-migrate-production.tmpl
camallen: > I guess it's not ideal to retry a large migration that puts a heavy load on the DB for a while before failing.

I can see this happening if we're doing data changes that violate the schema (uniq index etc) in the migration but tbh that's bad practice generally and something to be avoided - looking at me https://github.com/zooniverse/panoptes/blob/d6f4fbd32761217d93a127c60e7fb7dbc5899c1a/db/migrate/20200327164455_add_mark_remove_column_to_recents.rb#L10

I'd be more inclined to report the failure so we can investigate and re-run the job (manually or via different jenkins job). If we find we need the retry behaviour for transient failures we can up the `backoffLimit` here at a later date.

camallen:
> kubectl wait

The only migraitons that should take a long time are concurrent index builds. all others should be relatively fast with

> > wait for the job to complete kubectl wait --for=condition=complete --timeout=1200s job/panoptes-migrate-db-staging # not sure about the timeout here?
> 
> Yeah, 20 minutes doesn't seem long enough. Maybe set it to a day? We basically want it to wait forever because some DB migrations could take a long time.
> 
> I think "complete" and "failed" are two separate states, so this will always wait until the timeout on failure I think.

it appears we can job TTL cleanup via https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs
```
apiVersion: batch/v1
kind: Job
metadata:
  name: pi-with-ttl
spec:
  ttlSecondsAfterFinished: 100
  template:
    spec:
      containers:
      - name: pi
        image: perl
        command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
      restartPolicy: Never
```

adammcmaster:
@camallen The docs say the TTL feature is still alpha so I've added a delete command to the Jenkinsfile on success.

Jenkinsfile
```diff
@@ -185,30 +187,50 @@ pipeline {
...
-          }
-        }
+    stage('Migrate production database') {
+      when { tag 'production-migrate' }
+      agent any
+      steps {
+        sh """
+          sed 's/__IMAGE_TAG__/${GIT_COMMIT}/g' kubernetes/job-db-migrate-production.tmpl | kubectl --context azure apply --record -f -
+
+          kubectl wait --for=condition=complete --timeout=1200s job/panoptes-migrate-db-production-${GIT_COMMIT}
```
camallen, line 199: I was going to try the job TTL instead of the timeout here but either should work, https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs

if we set a day or so for the TTL i think that'll help with the cleanup code here too.

camallen:
> @camallen The docs say the TTL feature is still alpha so I've added a delete command to the Jenkinsfile on success.

`FEATURE STATE: Kubernetes v1.12 [alpha]`
yeah - alpha in v1.12 and we're up to 1.18, pity. i'm ok for either of these strategies, we can always change it up.

PR #3378, opened 2020-06-24
Azure Adapter

nciemniak:
## Notes for reviewers
- In AWS, we have **buckets** (top-level storage resource), and within a bucket we separate things out with **prefixes** for each environment. In Azure, we have **storage accounts** (top-level storage resource), and within a storage account we separate things out with **containers** for each environment. Therefore, in the `AzureAdapter`, we have no need to use a prefix to separate things out for each environment like we do in `AwsAdapter`.
- [kubernetes/deployment-staging-azure-canary.tmpl](https://github.com/zooniverse/panoptes/blob/be0c24105483bef2a6dcb337ea143bdce39b622c/kubernetes/deployment-staging-azure-canary.tmpl) contains the instructions for configuring the **canary deployment**. It includes a deployment template, a ConfigMap with env vars that tell the canary service to use azure storage, and an Ingress. My thought is to initially have the canary-weight be set to 0, and first just test using the canary-by-cookie option. Once we confirm that this works, we can switch over to having a % of requests go to azure. See these articles for more information about canary deployments: [Canary deployment with ingress nginx](https://www.elvinefendi.com/2018/11/25/canary-deployment-with-ingress-nginx.html), [Canary Release based on Ingress-Nginx](https://kubesphere.io/docs/quick-start/ingress-canary/), [Canary annotations](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary)

## Open questions:
1. ~`get_path` vs `put_path`: 
azure gems don’t give the option to set the request method (get vs put) when making a request for a presigned URL the way AWS does. Does it even make sense to have those methods be different for the azure adapter? Or would it make more sense to just define one of the methods (say, `get_path`), and call that method from within the other method (call `get_path` in `put_path`). The only difference in how Ive set those methods up thus far is based on how options are used (looking at the existing code for `AwsAdapter`).~
2. How to handle public vs private uploads using Azure Blob Storage? [See RFC here](https://github.com/zooniverse/operations/issues/409).

## To-do:
- [ ] confirm that the presigned uri works with current PFE code as well as python client code
- [ ] deployment/testing: set up canary deployment, determine a course of action for testing on staging/prod
- [ ] determine how to set up our DNS to have panoptes-uploads.zooniverse.org point to azure
- [ ] make a decision on how to set up public vs private uploads
- [ ] determine course of action for S3 data migration to azure
- [x] update tests so that http requests to the azure blob service will not be made every time the tests are run

----

## Review checklist


nciemniak:
@camallen you are right - actually pretty helpful to look at how rails does it for active storage 👍

config/initializers/storage.rb
```diff
@@ -7,7 +7,10 @@ def self.configuration
         {
           adapter: ENV.fetch('STORAGE_ADAPTER', 'test'),
           bucket: ENV['STORAGE_BUCKET'],
-          prefix: ENV['STORAGE_PREFIX']
+          prefix: ENV['STORAGE_PREFIX'],
+          storage_account_name: ENV.fetch('AZURE_STORAGE_ACCOUNT', 'panoptes'),
```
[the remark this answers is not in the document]
nciemniak: I've prefixed all azure-specific keys with `azure_`, anything not prefixed as such is for adapters in general or aws-specific.

kubernetes/deployment-production.tmpl
```diff
@@ -133,9 +133,11 @@ data:
   PROJECT_REQUEST_BCC: i2h6y2f6a7o0x7i9@zooniverse.slack.com
   REDIS_URL: 'redis://panoptes-production-redis:6379/0'
   SIDEKIQ_CONCURRENCY: '8'
-  STORAGE_ADAPTER: aws
+  STORAGE_ADAPTER: azure
```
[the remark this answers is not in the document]
nciemniak: Makes sense, I'll switch `STORAGE_ADAPTER` back to aws, but I'll leave these azure ENV vars here because we do need to remember to include them for when we switch over, no harm having them here unless there's a good reason not to @camallen let me know if you disagree.

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    def initialize_blob_client(opts)
+      Azure::Storage::Blob::BlobService.create(
+        storage_account_name: opts[:storage_account_name],
+        storage_access_key: opts[:storage_access_key]
+      )
+    end
+
+    def initialize_signer(opts)
```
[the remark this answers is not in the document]
nciemniak: Sure, rubocop was just yelling about the `initialize` method having too many lines.

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    def put_file(path, file_path, opts={})
+      # TO DO: implement private v public uploads
+      upload_options = { content_type: opts[:content_type] }
+      upload_options[:content_encoding] = 'gzip' if opts[:compressed]
+      if opts[:content_disposition]
+        upload_options[:content_disposition] = opts[:content_disposition]
+      end
+
+      content = get_file_contents file_path
```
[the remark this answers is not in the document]
nciemniak: Ah, yes. Ive checked out the tests provided in the gem and mirrored the way they pass a `File` object.

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,92 @@
...
+      path + "#{SecureRandom.uuid}.#{extension}"
+    end
+
+    def get_path(path, opts={})
+      # TO DO: implement private v public uploads
+      expires_in = opts[:get_expires] || @get_expiration # time in minutes
+      expiry_time = get_expiry_time(expires_in)
+
+      @signer.signed_uri(
+        generate_uri(path), false,
```
camallen, line 34: ask the client for the storage URI via client `generate_uri` method vs our own possibly brittle version custom one which is brittle..
https://github.com/Azure/azure-storage-ruby/blob/e5059568b6b971f6981b454fd71cf211429d418e/common/lib/azure/core/service.rb#L39

https://github.com/rails/rails/blob/c9abc6801e0c1dd187f0f7e3ca127a29a6eba1bb/activestorage/lib/active_storage/service/azure_storage_service.rb#L128

Also please use move the `use_account_sas` boolean param to its own line.

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    private
+
+    # @param expires_in [int]: time increment in minutes
+    def get_expiry_time(expires_in)
+      Time.now.utc.advance(minutes: expires_in).iso8601 # required format is UTC time zone, ISO 8601
+    end
+
+    def generate_uri(path)
+      URI("https://#{@storage_account_name}.blob.core.windows.net/#{@container}/#{path}")
```
camallen, line 89: unless there is a really good reason, we should rely on the client to generate these for us, not our own custom one. 

https://github.com/rails/rails/blob/c9abc6801e0c1dd187f0f7e3ca127a29a6eba1bb/activestorage/lib/active_storage/service/azure_storage_service.rb#L128

https://www.rubydoc.info/github/Azure/azure-storage-ruby/Azure/Storage/Service/StorageService#generate_uri-instance_method

spec/services/media_storage/azure_adapter_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+RSpec.describe MediaStorage::AzureAdapter do
+  let(:storage_account_name) { 'tiny-watermelons' }
+  let(:container) { 'test' }
+  let(:opts) do
+    {
+      azure_storage_account: storage_account_name,
+      azure_storage_access_key: 'fake',
+      azure_storage_container: container,
+      stub_responses: true
```
camallen, line 11: Does this actually work in the azure landscape? in the aws client this avoids any communication with the remote cloud storage service.

app/services/media_storage/azure_adapter.rb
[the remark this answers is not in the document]
camallen: feel free to change the defaults / override rubocop inline https://stackoverflow.com/questions/26342570/rubocop-linelength-how-to-ignore-lines-with-comments

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,92 @@
...
+      )
+      @signer = Azure::Storage::Common::Core::Auth::SharedAccessSignature.new(
+        opts[:azure_storage_account],
+        opts[:azure_storage_access_key]
+      )
+    end
+
+    def stored_path(content_type, medium_type, *path_prefix)
+      extension = get_extension(content_type)
+      path = "#{medium_type}/"
```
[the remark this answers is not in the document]
nciemniak: Ah yes. I think I initially misunderstood the desired behavior of this method, but Im pretty sure I see it now.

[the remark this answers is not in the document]
nciemniak: Method and specs updated 👍 
I've added the env var `AZURE_PREFIX` because the azure adapter `stored_path` method will use the container name to generate the part of the URL that specifies the environment - 
https://github.com/zooniverse/panoptes/blob/9f6a1429464678d836ff1c3e1ac44a166d53418d/app/services/media_storage/azure_adapter.rb#L27
which means that what we pass in as a prefix will be different from the AWS prefix.

In addition, we might end up wanting the azure prefix to be different from the aws prefix during testing / while we have both adapters running simultaneously.

app/services/media_storage/azure_adapter.rb
nciemniak: Sure - just as a side note, the reason I put it on its own line was to mirror how rails used the method; I figured it was a convention around passing unnamed parameters vs named parameters.
https://github.com/rails/rails/blob/badba9d2fa90ae96a5d8428b0a739e18b1581607/activestorage/lib/active_storage/service/azure_storage_service.rb#L113

app/services/media_storage/azure_adapter.rb
nciemniak: You're definitely right, I didnt notice that the azure gem's `generate_uri` method actually returned the microsoft domain url, thanks for catching that 👍

spec/services/media_storage/azure_adapter_spec.rb
nciemniak: You know what, it looks like it doesn't. I definitely overlooked this, when I was first checking out the aws adapter specs I thought it was a built in rspec tool but then it slipped my mind to go back and check whether it worked as I expected.

When myself and @zwolf were working on tove, we had discussed using the [VCR](https://github.com/vcr/vcr) gem (though not sure if we actually ended up using it or not, since I'm not seeing it in the Gemfile), I'm thinking that would work well for this type of scenario, i.e. recording responses from http requests and saving them so that next time that request is made, it'll just use the saved response rather than making the request again.

Will put it on the to-do list.

camallen: I'd be in favour of using verifying doubles for the azure method calls and trust the gem does what we want (based on the rails azure adapter example we used as a template). 

I would strongly vote against having this rely on external services and adding credentials against this sever for testing etc.

zwolf: Extremely agree about keeping tests independent of external services. These specs, though, seem to just be using the gem to build and format the various URLs by feeding it fake credentials and testing the output. I do also agree though that this could be done with verifying doubles and a bit of trust that the input will generate the output without actually testing the gem.

zwolf: That is, test that the gem is receiving the method calls that you expect, with the arguments that you expect, and leave it at that rather than letting the gem do the processing and testing the output. 

VCR is useful for testing your own code that relies on a API response, but in this case we already have the API requests/responses abstracted away by the gem. We can just test that we're giving the gem the right arguments to the right method calls.

zwolf:
Had a conversation about testing and doubling and stubbing, so I'll hold off on reviewing until that's reflected here. Upshot is that we should be doubling gem classes and testing only how this repo's code uses the gem and not the gem internals, such as whether it's generating specific URLs and similar.

.rubocop.yml
```diff
@@ -16,4 +16,11 @@ Metrics/BlockLength:
     - '**/*_spec.rb'
     - 'spec/support/*.rb'
 
+  Metrics/MethodLength:
+    Description: 'Avoid methods longer than 12 lines of code.'
+    Max: 12
```
camallen, line 21: OOI - how did 12 become the magic number?

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,47 @@
+module MediaStorage
```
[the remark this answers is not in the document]
camallen: please do add this one in, https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,47 @@
...
+  class AzureAdapter < AbstractAdapter
```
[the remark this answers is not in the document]
camallen: mind disabling this cop in rubocop yaml config?

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,85 @@
...
+        opts[:azure_storage_account],
+        opts[:azure_storage_access_key]
+      )
+    end
+
+    def stored_path(content_type, medium_type, *path_prefix)
+      extension = get_extension(content_type)
+      path = prefix.to_s
+      path += '/' unless path[-1] == '/'
+      path += container + '/'
```
camallen, line 27: hmm - why are we including the container in the stored path? 

my understanding is we store the object in the container at the stored path. The container should not be included in the path.

spec/services/media_storage/azure_adapter_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+  context 'when opts are passed to the initializer' do
+    it 'uses default expiration values when no expiration values are passed' do
+      default = MediaStorage::AzureAdapter::DEFAULT_EXPIRES_IN
+      expect(adapter.instance_variable_get(:@get_expiration)).to eq(default)
+      expect(adapter.instance_variable_get(:@put_expiration)).to eq(default)
+    end
+
+    it 'defaults to current rails environment for the container name when no container is given' do
+      adapter = described_class.new(opts.except(:azure_storage_container))
+      expect(adapter.instance_variable_get(:@container)).to eq('test')
```
camallen, line 28: use the attr_readers getter methods for these instance vars
```suggestion
      expect(adapter.container)).to eq('test')
```

spec/services/media_storage/azure_adapter_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+  end
+  let(:adapter) { described_class.new(opts) }
+  let(:uri_regex) { /\A#{URI::DEFAULT_PARSER.make_regexp}\z/ }
+  let(:uuid_v4_regex) { /[a-f0-9]{8}\-[a-f0-9]{4}\-4[a-f0-9]{3}\-(8|9|a|b)[a-f0-9]{3}\-[a-f0-9]{12}/ }
+
+  context 'when opts are passed to the initializer' do
+    it 'uses default expiration values when no expiration values are passed' do
+      default = MediaStorage::AzureAdapter::DEFAULT_EXPIRES_IN
+      expect(adapter.instance_variable_get(:@get_expiration)).to eq(default)
+      expect(adapter.instance_variable_get(:@put_expiration)).to eq(default)
```
camallen, line 23:
```suggestion
      expect(adapter.get_expiration).to eq(default)
      expect(adapter.put_expiration).to eq(default)
```

spec/services/media_storage/azure_adapter_spec.rb
nciemniak: Azure client and storage objects have been replaced with verifying doubles 👍

app/services/media_storage/azure_adapter.rb
nciemniak: Yes you are right. I was thinking "stored path" more along the lines of "what is the URL at which this object can be accessed?" 

Now that Ive looked through how `stored_path` is used to generate the `src` attribute in the Media records, which then later gets passed to the `path` param of the `put_path` or `get_path` methods I see that we dont need the container in the path here.

spec/services/media_storage/azure_adapter_spec.rb
```diff
@@ -0,0 +1,211 @@
...
+
+  context 'when opts are passed to the initializer' do
+    it 'uses default expiration values when no expiration values are passed' do
+      default = MediaStorage::AzureAdapter::DEFAULT_EXPIRES_IN
+      expect(adapter.get_expiration).to eq(default)
+      expect(adapter.put_expiration).to eq(default)
+    end
+
+    it 'defaults to current rails environment for the container name when no container is given' do
+      adapter = described_class.new(opts.except(:azure_storage_container))
```
camallen, line 34: I don't think we want this as we want explicitly specify the desired container and have these pre-setup with the relevant permissions. 
rather we should raise or error if the container isn't setup correctly

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,84 @@
+module MediaStorage
+  class AzureAdapter < AbstractAdapter
+    attr_reader :prefix, :container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
+    DEFAULT_EXPIRES_IN = 3 # time in minutes, see get_expiry_time(expires_in)
+
+    def initialize(opts={})
+      @storage_account_name = opts[:azure_storage_account]
+      @container = opts[:azure_storage_container] || Rails.env
+      @prefix = opts[:azure_prefix]
```
camallen, line 9: we want the default rails env value not on the container - the container has to be explicitly setup with correct ACL etc.

The prefix - this was a means of namespacing the aws s3 bucket paths for isolation to prod / staging etc. 

https://github.com/zooniverse/panoptes/blob/d0eadcf11d61f4510d92cf73f0c1411104de3ce1/app/services/media_storage/aws_adapter.rb#L7-L8
https://github.com/zooniverse/panoptes/blob/d0eadcf11d61f4510d92cf73f0c1411104de3ce1/config/initializers/storage.rb#L10
https://github.com/zooniverse/panoptes/blob/d0eadcf11d61f4510d92cf73f0c1411104de3ce1/kubernetes/deployment-staging.tmpl#L85

spec/services/media_storage/azure_adapter_spec.rb
nciemniak: @camallen what about the prefix? I know we have the prefix set up in AWS to default to the current env when nothing is passed in; I'm not sure why we do this in AWS, want to make sure I understand before rolling out the azure adapter with the same setup.

app/workers/medium_removal_worker.rb
```diff
@@ -3,8 +3,8 @@ class MediumRemovalWorker
 
   sidekiq_options queue: :data_low
 
-  def perform(medium_src)
-    MediaStorage.delete_file(medium_src)
+  def perform(medium_src, opts={})
+    MediaStorage.delete_file(medium_src, opts)
```
camallen, line 7: Can we please update the method signature of the abstract & aws adapters? 
https://github.com/zooniverse/panoptes/blob/ff90c7ecd73c66c2af03895a997e44b369973d26/app/services/media_storage/abstract_adapter.rb#L23
https://github.com/zooniverse/panoptes/blob/ff90c7ecd73c66c2af03895a997e44b369973d26/app/services/media_storage/aws_adapter.rb#L60

app/services/media_storage/azure_adapter.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    def put_path(path, opts={})
+      container = opts[:private] ? private_container : public_container
+
+      signer.signed_uri(
+        client.generate_uri("#{container}/#{path}"),
+        false,
+        service: 'b', # blob
+        permissions: 'rcw', # read create write
```
camallen, line 56: Not blocking - Folks need the 'r' read permission to put the file?

kubernetes/deployment-production.tmpl
```diff
@@ -127,6 +127,8 @@ data:
   STORAGE_ADAPTER: aws
   STORAGE_BUCKET: zooniverse-static
   STORAGE_PREFIX: panoptes-uploads.zooniverse.org/production/
+  AZURE_STORAGE_ACCOUNT: panoptes
+  AZURE_STORAGE_CONTAINER: production
```
camallen, line 131: this needs to be update for public / private containers (e.g. staging deployment config)

app/services/media_storage/azure_adapter.rb
nciemniak: Ah true, they really don't.

app/workers/medium_removal_worker.rb
nciemniak: Sure thing. I have marked the `opts` param as unused in the aws adapter using an underscore:
https://github.com/zooniverse/panoptes/blob/8457f16bfc92e5153e725e37a70032cc75aaa26f/app/services/media_storage/aws_adapter.rb#L60

PR #3383, opened 2020-06-26
cleanup migrate.sh

camallen:
remove AWS specific config / env setup

# Review checklist

### DIFF
--- a/migrate.sh
+++ b/migrate.sh
@@ -2,11 +2,4 @@
 
 cd /rails_app
 
-ln -sf /rails_conf/* ./config/
-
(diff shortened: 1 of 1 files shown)


migrate.sh
```diff
@@ -2,11 +2,4 @@
 
 cd /rails_app
 
-ln -sf /rails_conf/* ./config/
-
-if [ -z "$RAILS_ENV" ]
-then
-    export RAILS_ENV="production"
```
camallen, line 9: I consider this dangerous to leave laying around, it will default to development to avoid someone running this and accidentally running against production db.

PR #3384, opened 2020-06-29
Increase pod count for staging

adammcmaster:
Avoid downtime when scaling/upgrading the cluster.

# Review checklist

### DIFF
--- a/kubernetes/deployment-staging.tmpl
+++ b/kubernetes/deployment-staging.tmpl
@@ -187,8 +187,8 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: panoptes-staging-app
-  minReplicas: 1
-  maxReplicas: 2
+  minReplicas: 2
+  maxReplicas: 3
   targetCPUUtilizationPercentage: 80
 ---
 apiVersion: v1
@@ -266,8 +266,8 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: panoptes-staging-sidekiq
-  minReplicas: 1
-  maxReplicas: 2
+  minReplicas: 2
+  maxReplicas: 3
   targetCPUUtilizationPercentage: 95
 ---
 kind: PersistentVolumeClaim


adammcmaster:
RollingUpdate appears to be the default:

```
$ kubectl describe deploy panoptes-production-app  | grep Strategy
StrategyType:           RollingUpdate
RollingUpdateStrategy:  25% max unavailable, 25% max surge
```

Did you specifically want to use 1 and 0 as the surge/unavailable limits instead of percentages?

camallen:
> RollingUpdate appears to be the default:
> 
> ```
> $ kubectl describe deploy panoptes-production-app  | grep Strategy
> StrategyType:           RollingUpdate
> RollingUpdateStrategy:  25% max unavailable, 25% max surge
> ```
> 
> Did you specifically want to use 1 and 0 as the surge/unavailable limits instead of percentages?

I'm fine with percentages, basically we need 0 pods unavailable during the rollout.

adammcmaster:
👍 That should be what it's already doing.

adammcmaster:
Still not sure why we were seeing downtime for staging though, in that case.

adammcmaster:
Oh, I bet draining a node doesn't trigger a rolling update. I think that's what I was thinking when I opened this PR – having an extra replica would stop the service going down if the Kubernetes node gets taken out of service. It's not really a big deal since it affects staging, but the Pingdom alerts annoy me.

camallen:
> Oh, I bet draining a node doesn't trigger a rolling update. I think that's what I was thinking when I opened this PR – having an extra replica would stop the service going down if the Kubernetes node gets taken out of service. It's not really a big deal since it only affects staging, but the Pingdom alerts annoy me.

so before draining we would scale the replica count (assuming it goes to another node) and then drain, downscale the replica?

adammcmaster:
Not exactly. Nodes get drained automatically all the time due to autoscaling, so the idea here is to just always have at least two pods running to avoid any disruption from that.

camallen:
> Not exactly. Nodes get drained automatically all the time due to autoscaling, so the idea here is to just always have at least two pods running to avoid any disruption from that.

This sounds like we need pod disruption budgets for these services, https://docs.microsoft.com/en-us/azure/aks/operator-best-practices-scheduler#plan-for-availability-using-pod-disruption-budgets

PR #3386, opened 2020-06-29
Reinstate AWS production deployments

adammcmaster:
When the `production-release` tag is pushed:
- Push `zooniverse/panoptes:production-release` to Docker Hub
- Build AMIs for panoptes-api and panoptes-dumpworker
- Deploy panoptes-api and panoptes-dumpworker

# Review checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,10 +42,149 @@ pipeline {
               newImage.push('latest')
             }
           }
+          if (env.TAG_NAME == 'production-release') {
+            stage('Update production release tag') {
+              newImage.push('production-release')
+            }
+          }
         }
       }
     }
 
+    stage('Build AMIs') {
+      when { tag 'production-release' }
+      failFast true
+      parallel {
+        stage('Production API') {
+          options {
+            skipDefaultCheckout true
+          }
+          agent {
+            docker {
+              image 'zooniverse/operations:latest'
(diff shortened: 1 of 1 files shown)


adammcmaster:
The updated environment variables file is in place, and I've tested it by configuring an EC2 instance using the Docker image currently deployed to Kubernetes.

Jenkinsfile
```diff
@@ -42,10 +42,149 @@ pipeline {
...
+                  message: "FAILED: Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]' (${env.BUILD_URL})",
+                  channel: "#ops"
+                  )
+            }
+          }
         }
       }
     }
 
+    stage('Deploy production AMIs') {
```
camallen, line 133: I don't think we want this deploy script to automatically run on `production-release` tag (migrations). We will want to run the deploy via the lita cmd as per the previous setup.

adammcmaster: My aim is to have this deploy using the newer `lita deploy` command, which works by updating the tag to trigger the deploy. I'm not trying to get the old `lita panoptes deploy` command working again.

camallen: our current issue is running migrations as they will need to run before the code deploys to ensure the db schema matches the code ORM schema. If we can finalize the migration jobs running correctly (on their own tag) then i think we can safely pursue this deploy path, i.e. 
1. merge to master deploy to staging
2. run a migration job by updating `production-migrate` tag
3. update the `production-release` tag to trigger the prod deployment

PR #3391, opened 2020-07-02
Use tmp disk for static assets

camallen:
linked to #3389 #3390 - switch to using tmp disk for static asset serving via nginx

# Review checklist

### DIFF
--- a/kubernetes/deployment-staging.tmpl
+++ b/kubernetes/deployment-staging.tmpl
@@ -170,7 +170,10 @@ spec:
               mountPath: "/etc/nginx-sites"
       volumes:
         - name: static-assets
-          emptyDir: {}
+        hostPath:
+          # directory location on host node temp disk
+          path: /mnt/panoptes-staging-app-static-assets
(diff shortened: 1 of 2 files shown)


kubernetes/deployment-staging.tmpl
```diff
@@ -170,7 +170,10 @@ spec:
               mountPath: "/etc/nginx-sites"
       volumes:
         - name: static-assets
-          emptyDir: {}
+        hostPath:
+          # directory location on host node temp disk
+          path: /mnt/panoptes-staging-app-static-assets
+          type: DirectoryOrCreate
```
adammcmaster, line 176: Oops, Jenkins flagged this:
```suggestion
          hostPath:
            # directory location on host node temp disk
            path: /mnt/panoptes-staging-app-static-assets
            type: DirectoryOrCreate
```

camallen: thank you dry run deploy jenkins 🎊

PR #3394, opened 2020-07-02
Add readiness probes for rollout traffic routing

camallen:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes

only route traffic to the new pods once they are ready to accept it. I tested boot on my dev machine and it was ~25s so going with 20 for production.

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -386,6 +386,7 @@ spec:
       containers:
         - name: panoptes-production-redis
           image: redis:6.0.5
+          args: ['--save ""', '--appendonly no']
           resources:
             requests:
               memory: "2500Mi"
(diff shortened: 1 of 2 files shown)


kubernetes/deployment-production.tmpl
```diff
@@ -386,6 +402,7 @@ spec:
       containers:
         - name: panoptes-production-redis
           image: redis:6.0.5
+          args: ['--save ""', '--appendonly no']
```
adammcmaster, line 405: Did you mean to include this here? Don't we want redis data to be persistent for Panoptes?

kubernetes/deployment-staging.tmpl
```diff
@@ -330,6 +346,7 @@ spec:
       containers:
         - name: panoptes-staging-redis
           image: redis:6.0.5
+          args: ['--save ""', '--appendonly no']
```
adammcmaster, line 349: Did you mean to include this here? Don't we want redis data to be persistent for Panoptes?

kubernetes/deployment-production.tmpl
camallen: wtf!? absolutely not. bad me

PR #3404, opened 2020-07-07
Azure flipper by user

nciemniak:
A sketch of how we might use the flipper on a per-user basis.
the `set_adapter` method would need to be called prior to using any `MediaStorage` method.

I made changes here in a way that didn't touch existing methods too much for the sake of simplicity, but if we want to move forward with this idea, I'm thinking of refactoring the code such that the initialization of `@azure_adapter` and `@aws_adapter` would be more integrated with  `load_adapter` and `load_from_included` methods, perhaps.


camallen:
I'm going to close this in favour of canary testing via k8s nginx ingress deployments - feel free to reopen.

PR #3421, opened 2020-07-14
Add nasa email communication field to users

zwolf:
Adds the field via migration along with all the assorted bits to roughly match how beta emails are handled, minus the checkbox on the Devise signup page.

# Review checklist


.travis.yml
```diff
@@ -20,6 +20,7 @@ before_script: &before_script
   - psql -c 'create database travis_ci_test;' -U postgres
   - scripts/no_focus.sh
   - bundle exec rake configure:local
+  - RAILS_ENV=test bundle exec rake db:migrate
```
camallen, line 23:
```suggestion
```
this is not required, fwiw https://github.com/zooniverse/panoptes/blob/d0eadcf11d61f4510d92cf73f0c1411104de3ce1/spec/spec_helper.rb#L9
sets up the test database schema from the current `structure.sql`. 

Please make sure any PR migrations have been run locally and the associated changes included in the comitted `structure.sql` file.

classifications_csv_dump_export.rb
```diff
@@ -0,0 +1,59 @@
+# Manual csv classifications dump
```
camallen, line 1: this file shouldn't be in this PR, please remove

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddNasaEmailFieldToUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    add_column :users, :nasa_email_communication, :boolea, default: false, algorithm: :concurrently
```
camallen, line 5: Don't add a column with a default, it'll lock the table while adding this, https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/

classifications_csv_dump_export.rb
zwolf: doh! `-a` is a dangerous flag to have committed to muscle memory

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
zwolf: Y'know, I had this same thought, and then I checked the migrations folder and saw [a couple](https://github.com/zooniverse/panoptes/blob/master/db/migrate/20190624094308_add_subjects_counter_cache_to_workflow.rb#L3) recent [migrations](https://github.com/zooniverse/panoptes/blob/master/db/migrate/20190507103007_add_education_attr_to_workflows.rb#L3) using `default:`, and not using DDL transactions, and figured I was okay. Guess not!

.travis.yml
zwolf: Running them locally causes the structure.sql diff to balloon with a bunch of noise. I think it's a local PG version issue, so I'm trying to get it done in Docker right now.

zwolf: ✔️

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
camallen: it depends on the table size. If it's big and busy it'll block which can kill the app depending on the query patterns. users is big and busy so it'll cause errors. 

new tables with no data this is ok, even smallish tables with infrequent access is ok (good to get a baseline from a snapshot) but running these migrations carries downtime risk.

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddNasaEmailFieldToUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    add_column :users, :nasa_email_communication, :boolean
+    add_index :users, :nasa_email_communication, where: "nasa_email_communication = true", algorithm: :concurrently
```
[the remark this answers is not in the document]
zwolf: It's the same as the `beta_email_communication` index. Are they supposed to be used differently?

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddNasaEmailFieldToUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    add_column :users, :nasa_email_communication, :boolean
```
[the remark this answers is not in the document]
zwolf: It's either that or put the default back here in this migration. What do you want?

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
zwolf: I guess workflows isn't huge, but it sure is frequently accessed.

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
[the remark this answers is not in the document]
camallen: i don't know - there is no code in here that actually queries the data with AR. Just make sure the index matches for the query you will write (test the sql in a local test db). If it doesn't it will use up DB RAM and not help the query.

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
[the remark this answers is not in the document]
camallen: https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value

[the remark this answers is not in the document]
zwolf: So, neither. If that link indicates the suggestion you want to make in your review, what was the purpose of the first question?

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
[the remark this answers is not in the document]
zwolf: Again, I'm making the assumption that this field will be used the same way as the other email list fields, which have the same kind of indexes: 

https://github.com/zooniverse/panoptes/blob/master/db/migrate/20150605103339_add_beta_email_to_users.rb#L5-L6

If the mailer service is filtering users on these fields then it makes sense to use that predicate, and the pressure to get this finished indicates to me that it'll be used pretty soon.

zwolf:
I'm realizing now that I didn't update a bunch of the specs that include the other email attrs as options. Gonna add those real quick.

spec/models/csv_dumps/full_email_list_spec.rb
```diff
@@ -4,6 +4,7 @@
   let!(:all_email_user) { create(:user, global_email_communication: true, beta_email_communication: true) }
   let!(:global_email_user) { create(:user, global_email_communication: true, beta_email_communication: false) }
   let!(:beta_email_user) { create(:user, global_email_communication: false, beta_email_communication: true) }
+  let!(:nasa_email_user) { create(:user, global_email_communication: false, nasa_email_communication: true) }
```
zwolf, line 7: oooh, this was actually a useful one

spec/lib/user_info_scrubber_spec.rb
```diff
@@ -7,7 +7,7 @@ def scrub_user_details(user)(user)
 describe UserInfoScrubber do
   describe '::scrub_personal_info!' do
     context "when using an active user" do
-      let(:user) { create(:user, current_sign_in_ip: '1.2.3.4', last_sign_in_ip: '1.2.3.5', tsv: 'foo', private_profile: false) }
+      let(:user) { create(:user, current_sign_in_ip: '1.2.3.4', last_sign_in_ip: '1.2.3.5', tsv: 'foo', private_profile: false, nasa_email_communication: true) }
```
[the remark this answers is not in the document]
zwolf: ah, the hound, we're just gonna have to let this one ride

db/migrate/20200714191914_add_nasa_email_field_to_users.rb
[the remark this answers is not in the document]
lcjohnso: I'm not totally familiar with the email list update + synchronization tools (which are not required immediately, as I'll manage the syncing manually to begin; I'll issue up the addition of list syncing to the mailing list worker for the future), but Zach's assumption to mimic the beta email list is the same one Mark and I have adopted when considering the PFE side of things.

And yes, this feature will get immediate use: planning to start highlighting the list and pointing people to the opt-in by the end of next week.  The PFE side of the feature is complete: see https://github.com/zooniverse/Panoptes-Front-End/pull/5747.

PR #3425, opened 2020-07-17
Add Configuration Blob to Tutorial

### LINKED ISSUE #3296: Configuration object for tutorial resource
Related to https://github.com/zooniverse/Panoptes-Front-End/issues/3611. We have a request from a couple of different projects that would like to configure the frequency of slide display for the mini-course kind of tutorial resource. It was suggested to store this configuration on the workflow resource, but I think that's only being suggested because the configuration object on the workflow resource already exist. I think a better fit would be to do a small migration on the tutorial resource to add a `configuration` property that can take any JSONB value. This way, the mini-course knows how often to display itself.

wgranger:
Describe your change here.

Closes #3296 

This PR adds the configuration JSON blob to tutorial resources. No changes here to the apiary file, as it looks like there is not a section there for the Tutorial resource.

# Review checklist

### DIFF
--- a/app/schemas/tutorial_create_schema.rb
+++ b/app/schemas/tutorial_create_schema.rb
@@ -49,5 +49,9 @@ class TutorialCreateSchema < JsonSchema
       end
 
     end
+
+    property "configuration" do
+      type "object"
+    end
   end
 end
--- a/app/schemas/tutorial_create_schema.rb
+++ b/app/schemas/tutorial_create_schema.rb
@@ -50,8 +50,8 @@ class TutorialCreateSchema < JsonSchema
 
     end
 
-    property "configuration" do
-      type "object"
+    property 'configuration' do
+      type 'object'
     end
   end
 end
--- a/app/schemas/tutorial_update_schema.rb
+++ b/app/schemas/tutorial_update_schema.rb
@@ -24,8 +24,8 @@ class TutorialUpdateSchema < JsonSchema
       end
     end
 
-    property "configuration" do
-      type "object"
+    property 'configuration' do
+      type 'object'
     end
   end
 end
--- a/app/schemas/tutorial_update_schema.rb
+++ b/app/schemas/tutorial_update_schema.rb
@@ -23,5 +23,9 @@ class TutorialUpdateSchema < JsonSchema
         end
       end
     end
+
+    property "configuration" do
+      type "object"
+    end
   end
 end
--- a/db/migrate/20200717155424_add_configuration_to_tutorials.rb
+++ b/db/migrate/20200717155424_add_configuration_to_tutorials.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class AddConfigurationToTutorials < ActiveRecord::Migration
   def change
     add_column :tutorials, :configuration, :jsonb
--- a/db/migrate/20200717155424_add_configuration_to_tutorials.rb
+++ b/db/migrate/20200717155424_add_configuration_to_tutorials.rb
@@ -1,0 +1,5 @@
+class AddConfigurationToTutorials < ActiveRecord::Migration
+  def change
+    add_column :tutorials, :configuration, :jsonb
+  end
+end
--- a/db/migrate/20200717155424_add_configuration_to_tutorials.rb
+++ b/db/migrate/20200717155424_add_configuration_to_tutorials.rb
@@ -2,6 +2,6 @@
 
 class AddConfigurationToTutorials < ActiveRecord::Migration
   def change
-    add_column :tutorials, :configuration, :jsonb
+    safety_assured { add_column :tutorials, :configuration, :jsonb, default: {} }
   end
 end
--- a/spec/controllers/api/v1/tutorials_controller_spec.rb
+++ b/spec/controllers/api/v1/tutorials_controller_spec.rb
@@ -101,6 +101,9 @@ describe Api::V1::TutorialsController, type: :controller do
           links: {
             project: project.id.to_s,
             workflows: [workflow.id]
+          },
+          configuration: {
+            an_option: "a setting"
           }
         }
       }
@@ -125,6 +128,9 @@ describe Api::V1::TutorialsController, type: :controller do
       {
         tutorials: {
           steps: [{"media" => "asdf", "content" => "asdf"}]
+        },
+        configuration: {
+          an_option: "a setting"
         }
       }
     end
--- a/spec/controllers/api/v1/tutorials_controller_spec.rb
+++ b/spec/controllers/api/v1/tutorials_controller_spec.rb
@@ -103,7 +103,7 @@ describe Api::V1::TutorialsController, type: :controller do
             workflows: [workflow.id]
           },
           configuration: {
-            an_option: "a setting"
+            an_option: 'a setting'
           }
         }
       }
@@ -130,7 +130,7 @@ describe Api::V1::TutorialsController, type: :controller do
           steps: [{"media" => "asdf", "content" => "asdf"}]
         },
         configuration: {
-          an_option: "a setting"
+          an_option: 'a setting'
         }
       }
     end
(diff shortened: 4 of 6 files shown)


camallen:
is this ready for review? Mind merging / rebasing from master to fix the merge conflicts?

wgranger:
Yep. This should be good to go. Also happy to resolve the Hound errors, but I've noticed other places in the code that are no as consistent with that rule.

db/migrate/20200717155424_add_configuration_to_tutorials.rb
```diff
@@ -0,0 +1,5 @@
+class AddConfigurationToTutorials < ActiveRecord::Migration
+  def change
+    add_column :tutorials, :configuration, :jsonb
```
camallen, line 3: This colulmn should have a default value
```suggestion
    add_column :tutorials, :configuration, :jsonb, default: {}
```

FWIW adding columns with default values via a migration can be bad as the operation will lock the table for read / writes. Read more here https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/

I'm a little surprised our strong migrations gem didn't catch this, https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value

app/schemas/tutorial_create_schema.rb
```diff
@@ -49,5 +49,9 @@ class TutorialCreateSchema < JsonSchema
       end
 
     end
+
+    property "configuration" do
```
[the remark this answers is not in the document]
camallen:
```suggestion
    property 'configuration' do
```

app/schemas/tutorial_create_schema.rb
```diff
@@ -49,5 +49,9 @@ class TutorialCreateSchema < JsonSchema
...
+      type "object"
```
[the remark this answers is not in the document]
camallen:
```suggestion
      type 'object'
```

app/schemas/tutorial_update_schema.rb
```diff
@@ -23,5 +23,9 @@ class TutorialUpdateSchema < JsonSchema
         end
       end
     end
+
+    property "configuration" do
```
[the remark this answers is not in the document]
camallen:
```suggestion
    property 'configuration' do
```

app/schemas/tutorial_update_schema.rb
```diff
@@ -23,5 +23,9 @@ class TutorialUpdateSchema < JsonSchema
...
+      type "object"
```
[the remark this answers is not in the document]
camallen:
```suggestion
      type 'object'
```

db/migrate/20200717155424_add_configuration_to_tutorials.rb
```diff
@@ -0,0 +1,5 @@
+class AddConfigurationToTutorials < ActiveRecord::Migration
```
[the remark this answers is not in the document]
camallen:
```suggestion
# frozen_string_literal: true

class AddConfigurationToTutorials < ActiveRecord::Migration
```

spec/controllers/api/v1/tutorials_controller_spec.rb
```diff
@@ -101,6 +101,9 @@
           links: {
             project: project.id.to_s,
             workflows: [workflow.id]
+          },
+          configuration: {
+            an_option: "a setting"
```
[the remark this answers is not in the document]
camallen:
```suggestion
            an_option: 'a setting'
```

spec/controllers/api/v1/tutorials_controller_spec.rb
```diff
@@ -125,6 +128,9 @@
       {
         tutorials: {
           steps: [{"media" => "asdf", "content" => "asdf"}]
+        },
+        configuration: {
+          an_option: "a setting"
```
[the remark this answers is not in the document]
camallen:
```suggestion
          an_option: 'a setting'
```

db/migrate/20200717155424_add_configuration_to_tutorials.rb
zwolf: > Instead, add the column without a default value, then change the default.

@camallen's suggestion goes against strong migration's recommendation that you should add a column without a default and then use `change_column_default` to set one. If you want to follow SM's best practices, the suggestion here should be to add a line for that after `add_column`. Are you suggesting to not bother with that because the table is so small?

zwolf: Also! My recent addition to the Users table did both the `add_column` and the `change_column_default` but didn't actually backfill the data. I recommend adding that bit in this PR.

camallen: Agree with Zach that best practice is to non add a column with a default. 

However for this case with such a small table size we can avoid the extra overhead as the table locking time will most likely be very short due to table size. 
> No need to do any fancy migration here as that table is small (~2K records) so will be milliseconds to add the column + default and no impact to the API.

As this is one of @wgranger early contributions it's up to him if he wants to invest the extra effort required to make this migration align with zero downtime / best practice strategies. Personally i would (to ensure we get into the habit of these) but as the impact risk is very low i've approved his efforts on this PR as it's good enough for the feature addition. 

@zwolf if you feel strongly about enforcing the best practice migration strategies, that's ok too. Please issue a change required review to block my approval review.

wgranger: I see what you're saying. So I think the change here might instead be:

`change_column_default :jsonb, {}`

So the idea would be to rollback, make the change, then rerun the migration?

@camallen does that make sense?

camallen: yes you will have to rollback your migrated schema to re apply the one you are creating. 

you're updated migration would then look like
``` ruby
add_column :tutorials, :configuration, :jsonb
change_column_default :tutorials, :configuration, {}
```

Noting that if you go this way - you'll also have to backfill the new column see https://github.com/zooniverse/panoptes/pull/3447 
This is due to the fact that the above won't rewrite the table when it adds in the default, rather any new rows will have the default but existing rows won't. 

The way i've suggested above (not best practice) `add_column :tutorials, :configuration, :jsonb, default: {}` would add the default on all column rows during it's rewrite (but it locks the table to do so - ok for ~2k records as it's super quick).

zwolf: Either way is fine with me, I just wanted to be clear about what you were saying here. It's not a lot more code to backfill it (in the same migration) and the strong migrations repo explains how to do it, but I agree that adding the default locking-ly wouldn't be an issue.

zwolf: To be clearer, your options are (approximately, should double check): 

`add_column :tutorials, :configuration, :jsonb, default: {}`
(easiest, not best practice but fine for this)
or

```
add_column :tutorials, :configuration, :jsonb
change_column_default :tutorials, :configuration, {}
Tutorial.in_batches do |tuts|
  tuts.update_all configuration: {}
end
```
(few more steps, but this is how it should be done for bigger tables where a lock would be impactful)

wgranger: Nice. This is pretty clear. Thanks for the descriptive feedback. I think I'll go best practices here to get into good habits.

wgranger: Ended up going the route of adding `default: {}` since this is such a small table. But! The safe migrations gem did kick in and block the migration initially, requiring a `safety_assured` wrapper to be used before migrating. 

Went this route after a chat because of the speed in comparison of how small the table was (~milliseconds of downtime). Noting that in the future, and especially for larger rows, doing this in batches is the preferred method.

PR #3436, opened 2020-07-22
Define HEARTBEAT_EMAILS for production

adammcmaster:
This adds `HEARTBEAT_EMAILS` to the production environment. There is a specific heartbeat group in gmail already. I've also added sysadmins and dev, since we'd want as many people to potentially see these alerts as possible.

# Review checklist


adammcmaster:
Hmm, good question. I don't think it's necessarily a problem if they're public, but maybe I'll move this into a secret anyway.

PR #3437, opened 2020-07-24
Canary deployment setup

nciemniak:
The kubernetes/deployment-staging-azure-canary.tmpl contains the instructions for configuring the canary deployment. It includes a deployment template, a ConfigMap with env vars that tell the canary service to use azure storage, and an Ingress. My thought is to initially have the canary-weight be set to 0, and first just test using the canary-by-cookie option. Once we confirm that this works, we can switch over to having a % of requests go to azure. See these articles for more information about canary deployments: 
- [Canary deployment with ingress nginx](https://www.elvinefendi.com/2018/11/25/canary-deployment-with-ingress-nginx.html)
- [Canary release based on Ingress-Nginx](https://kubesphere.io/docs/quick-start/ingress-canary/)
- [Canary annotations](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary)

----

# Review checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -131,6 +131,7 @@ pipeline {
       agent any
       steps {
         sh "sed 's/__IMAGE_TAG__/${GIT_COMMIT}/g' kubernetes/deployment-staging.tmpl | kubectl --context azure apply --record -f -"
+        sh "sed 's/__IMAGE_TAG__/${GIT_COMMIT}/g' kubernetes/deployment-staging-azure-canary.tmpl | kubectl --context azure apply --record -f -"
       }
     }
   }
--- a/kubernetes/deployment-staging-azure-canary.tmpl
+++ b/kubernetes/deployment-staging-azure-canary.tmpl
@@ -41,13 +41,19 @@ spec:
           env:
           - name: PG_STATEMENT_TIMEOUT
             value: '65000'
+          - name: AZURE_STORAGE_ACCOUNT
+            value: panoptes
+          - name: AZURE_STORAGE_CONTAINER_PUBLIC
+            value: public
+          - name: AZURE_STORAGE_CONTAINER_PRIVATE
+            value: private
           envFrom:
           - secretRef:
               name: panoptes-common-env-vars
           - secretRef:
               name: panoptes-staging-env-vars
           - configMapRef:
-              name: panoptes-staging-azure-canary-shared
+              name: panoptes-staging-shared
           volumeMounts:
             - name: static-assets
               mountPath: "/static-assets"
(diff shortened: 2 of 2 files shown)


kubernetes/deployment-staging-azure-canary.tmpl
```diff
@@ -0,0 +1,171 @@
...
+          configMap:
+            name: panoptes-nginx-conf-staging
+        - name: jwt-staging
+          secret:
+            secretName: panoptes-doorkeeper-jwt-staging
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: panoptes-staging-azure-canary-shared
```
camallen, line 114: We can reuse the original `panoptes-staging-shared` config map and just override the `STORAGE_ADAPTER` via the `env` directive for precendence. 
https://stackoverflow.com/questions/54398272/override-env-values-defined-in-container-spec

kubernetes/deployment-staging-azure-canary.tmpl
```diff
@@ -0,0 +1,171 @@
...
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: panoptes-staging-azure-canary-ingress
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/canary: "true"
+    nginx.ingress.kubernetes.io/canary-weight: "0"
+    nginx.ingress.kubernetes.io/canary-by-cookie: "use-azure-storage"
```
camallen, line 158: i'd prefer to have the cookie named for the behaviour we want here, which is to opt into canary traffic.
```suggestion
    nginx.ingress.kubernetes.io/canary-by-cookie: "canary-api-testing-opt-in"
```

kubernetes/deployment-staging-azure-canary.tmpl
```diff
@@ -0,0 +1,171 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: panoptes-staging-azure-canary-app
```
camallen, line 4: I'm wondering if we should use the same names as existing deployments but instead namespace the canary deployment? I believe using a namespace will help us cleanup the canary deploy...thoughts?

Jenkinsfile
```diff
@@ -131,6 +131,7 @@ pipeline {
       agent any
       steps {
         sh "sed 's/__IMAGE_TAG__/${GIT_COMMIT}/g' kubernetes/deployment-staging.tmpl | kubectl --context azure apply --record -f -"
+        sh "sed 's/__IMAGE_TAG__/${GIT_COMMIT}/g' kubernetes/deployment-staging-azure-canary.tmpl | kubectl --context azure apply --record -f -"
```
camallen, line 134: I'm wondering if we should deploy the canary resources to a custom namespace?
```suggestion
        sh "sed 's/__IMAGE_TAG__/${GIT_COMMIT}/g' kubernetes/deployment-staging-azure-canary.tmpl | kubectl --context azure apply --record -f -n panoptes-staging-canary -"
```

kubernetes/deployment-staging-azure-canary.tmpl
zwolf: We don't have very many namespaces and I'm not sure if there are any good reasons to separate them. I don't see that it would make the deployment easier to remove, since there are no other moving parts or networking or w/e.

kubernetes/deployment-staging-azure-canary.tmpl
zwolf: I guess order-of-operations wise so long as the this is parsed second this config map'll be there. Might as well keep this short.

PR #3439, opened 2020-08-03
add www.zooniverse.org nginx server block

camallen:
allow serving requests proxied from the www.zooniverse.org domain without incorrect host name redirection. 
Extract the common nginx config blocks to it's own config map and reuse these across different server name blocks.

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -54,19 +54,19 @@ data:
     server {
       server_name www.zooniverse.org;
       include /etc/nginx/ssl.default.conf;
-      include /etc/nginx-sites/panoptes-nginx-common.conf;
+      include /etc/nginx/panoptes-nginx-common.conf;
     }
 
     server {
       server_name panoptes.zooniverse.org;
       include /etc/nginx/ssl.default.conf;
-      include /etc/nginx-sites/panoptes-nginx-common.conf;
+      include /etc/nginx/panoptes-nginx-common.conf;
     }
 
     server {
       server_name signin.zooniverse.org;
       include /etc/nginx/ssl.default.conf;
-      include /etc/nginx-sites/panoptes-nginx-common.conf;
+      include /etc/nginx/panoptes-nginx-common.conf;
     }
 kind: ConfigMap
 metadata:
@@ -207,7 +207,7 @@ spec:
             - name: panoptes-nginx-config
               mountPath: "/etc/nginx-sites"
             - name: panoptes-nginx-common.conf
-              mountPath: "/etc/nginx-sites"
+              mountPath: "/etc/nginx"
       volumes:
         - name: static-assets
           hostPath:
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -1,96 +1,72 @@
 apiVersion: v1
 data:
   nginx.conf: |+
+    gzip_types *;
+
+    proxy_buffer_size 8k;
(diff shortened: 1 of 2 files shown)


kubernetes/deployment-production.tmpl
```diff
@@ -1,96 +1,72 @@
...
+---
+apiVersion: v1
+data:
+  nginx.conf: |+
+    upstream docker-panoptes {
+      server localhost:81;
+    }
+    
+    server {
+      server_name www.zooniverse.org;
```
adammcmaster, line 55: If these server blocks are all the same you don't really need more than one. You can group multiple names together:

```
      server_name www.zooniverse.org panoptes.zooniverse.org signin.zooniverse.org;
```

kubernetes/deployment-production.tmpl
```diff
@@ -1,96 +1,72 @@
...
+      include /etc/nginx/ssl.default.conf;
+      include /etc/nginx-sites/panoptes-nginx-common.conf;
```
adammcmaster, line 57: And if you group them together you probably don't need to split this out into a separate file.

kubernetes/deployment-production.tmpl
```diff
@@ -230,6 +206,8 @@ spec:
               mountPath: "/static-assets"
             - name: panoptes-nginx-config
               mountPath: "/etc/nginx-sites"
+            - name: panoptes-nginx-common.conf
+              mountPath: "/etc/nginx-sites"
```
adammcmaster, line 210: If you're going to keep this in a separate file you'll want to mount it to `/etc/nginx/` instead. `/etc/nginx-sites/*` is included by the main nginx config and including this file there will cause an error.

kubernetes/deployment-production.tmpl
camallen: my reading suggests that this will use the `www.zooniverser.org` as the $server_name variable which will then be passed to the upstream via host header, https://github.com/zooniverse/panoptes/blob/325d71760298c701efa5db1202e8b5b4232112b1/kubernetes/deployment-production.tmpl#L40

this will break redirects as per the issue outlined in https://github.com/zooniverse/operations/issues/459

If we can verify how the $server_name variable is filled when using one server block (wildcard subdomain / regex are options here) then i'm happy to collapse them - i did look into collapsing them but couldn't ensure the $server_name var will be correctly set.

adammcmaster: In my testing it seems to set `$server_name` to the first hostname value:

```nginx
server {
	server_name test testing;
	rewrite ^.*$ http://$server_name redirect;
}
```

```
$ curl -v -H "Host: test" localhost:8080/ | grep Location
< Location: http://test
$ curl -v -H "Host: testing" localhost:8080/ | grep Location
< Location: http://test
```

So I guess that's not what we want.

We'll need to keep them separated as you have done then in that case. (Another option might be to use `$http_host` but that could be dangerous if we accidentally deploy with no default server block, as that would open it to every host name.)

camallen: thanks for doing the actual testing Adam - appreciated. FWIW the security issue is fixed in rails 6+ so we can remove this nginx header config if / when we get there.

camallen:
> This looks OK now, but I would also recommend doing a similar refactoring to the staging config. I know staging doesn't need the multiple hostnames but it would be good practice to keep the two configs as similar as possible.

Will do before merging - in fact i might get a staging PR out to ensure this works as expected.

PR #3443, opened 2020-08-10
Revert "Remove the canary weight"

camallen:
Reverts zooniverse/panoptes#3442

We want the canary deploy to only be reachable via the cookie presence. Removing the weight will modify how the traffic is routed. 

We want the canary to only be reachable while the header is set, if it's missing we don't want traffic to hit the canary inadvertently.

### DIFF
--- a/kubernetes/deployment-staging-azure-canary.tmpl
+++ b/kubernetes/deployment-staging-azure-canary.tmpl
@@ -121,6 +121,7 @@ metadata:
   annotations:
     kubernetes.io/ingress.class: nginx
     nginx.ingress.kubernetes.io/canary: "true"
+    nginx.ingress.kubernetes.io/canary-weight: "0"
     nginx.ingress.kubernetes.io/canary-by-cookie: "canary-api-testing-opt-in"
 spec:
   tls:


zwolf:
Why is the weight necessary to prevent normal, cookie-less traffic from being correctly routed?

camallen:
> Why is the weight necessary to prevent normal, cookie-less traffic from being correctly routed?

Good question - I haven't tested but my memory is that without a weight it does round robin between the canary and the other upstreams. This is most likely incorrect as docs don't mention it,  and tbh without any setting it's most likely ignoring the canary for traffic distribution. It would be good to test this. 

That said, explicitly setting the weight to 0 also ensures we never use the weight preference to distribute traffic to the canary. Only the cookie / header will ever force the canary to be used which is what we want for testing this via cookie opt in.

zwolf:
> nginx.ingress.kubernetes.io/canary-by-cookie: The cookie to use for notifying the Ingress to route the request to the service specified in the Canary Ingress. When the cookie value is set to always, it will be routed to the canary. When the cookie is set to never, it will never be routed to the canary. For any other value, the cookie will be ignored and the request compared against the other canary rules by precedence.

> Canary rules are evaluated in order of precedence. Precedence is as follows: canary-by-header -> canary-by-cookie -> canary-weight

(https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary)

Doesn't sound necessary to include a weight and seems to be working as intended on staging, but I'll test it bit more this afternoon.

camallen:
> > nginx.ingress.kubernetes.io/canary-by-cookie: The cookie to use for notifying the Ingress to route the request to the service specified in the Canary Ingress. When the cookie value is set to always, it will be routed to the canary. When the cookie is set to never, it will never be routed to the canary. For any other value, the cookie will be ignored and the request compared against the other canary rules by precedence.
> 
> > Canary rules are evaluated in order of precedence. Precedence is as follows: canary-by-header -> canary-by-cookie -> canary-weight
> 
> (https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary)
> 
> Doesn't sound necessary to include a weight and seems to be working as intended on staging, but I'll test it bit more this afternoon.

Excellent - if we can verify the canary is only hit when the cookie is present and never without a cookie present, then this can close. It's really important for production testing we don't allow the canary to be hit by normal users. Let us know what your tests indicate.

Issue #2308, opened 2017-04-26
aws adapter get_path should work without CF distribution or NGINX proxy


camallen:
The aws storage adapter uses a domain prefix that is meant to  resolves to an S3 bucket path. In our production / staging setup we use a CloudFront distribution to proxy the prefix CNAME to the s3 bucket path URL. 

An out of the box installation shouldn't require another service (nginx / CF) to resolve the `get_path` to the correct s3 bucket URL

e.g. storage.yml 
``` yaml
development:
  adapter: aws
  prefix: development
  bucket: labelmars-image-data
  access_key_id: <key>
  secret_access_key: <secret>
```

sets an avatar media resource src as `development/subject_location/12d7141b25-0353-4a7e-aee8-480c351521d3.jpeg` and thus the aws adapter [get_path](https://github.com/zooniverse/Panoptes/blob/20483e32c5547300c4e7c2bb630f6c18e6b9f356/app/services/media_storage/aws_adapter.rb#L40) code returns 
`https://development/subject_location/12d7141b25-0353-4a7e-aee8-480c351521d3.jpeg`

We should provide a configuration directive to modify the default behaviour to not use a CDN or maybe find the bucket URL programatically or even allow the dev to add a `domain_prefix` yaml option to ensure get_paths return the s3 bucket URL's. 
``` yaml
  domain_prefix: labelmars-image-data.s3.amazonaws.com
```
that results in `https://labelmars-image-data.s3.amazonaws.com/development/subject_location/12d7141b25-0353-4a7e-aee8-480c351521d3.jpeg`

Thanks to @squiddon for bringing this up.

astopy:
See also #1763

camallen:
keeping this open

lcjohnso:
closed by #3453

Issue #480, opened 2015-02-02
Addition to the docs - adding mysql as a requirement


rogerhutchings:
On OSX, `bundle install` will fail if you don't already have MySQL installed. It might be worth adding a Requirements section to the README for pre-install requirements

edpaget:
@camallen I don't think we're using the zooniverse_user import rake task I wrote. Should we just go ahead and remove it and the mysql dependency?

camallen:
Alas i wish we could. The plan for stargazing is to use this to sync with zoo home.

camallen:
@rogerhutchings are you going to add the doc updates? It's be super sweet taking into account this issue too https://github.com/zooniverse/Panoptes/issues/477

rogerhutchings:
Yeah, I was going to - I'll have a look at this one as well.

PR #3453, opened 2020-08-19
Extract domain & path prefix in azmedia adapter

### LINKED ISSUE #1763: Updating media URLs/prefixes
Currently we configure storage with something like:

```yaml
production:
  adapter: aws
  bucket: zooniverse-static
  prefix: panoptes-uploads.zooniverse.org/production/
```

And we assume the path starts with an FQDN here: https://github.com/zooniverse/Panoptes/blob/master/app/services/media_storage/aws_adapter.rb#L46

To make this more flexible (for both ourselves and third parties) we should instead specify the domain name (and possibly the prefix) separately:

```yaml
production:
  adapter: aws
  bucket: zooniverse-static
  prefix: panoptes-uploads.zooniverse.org/production/
  http_prefix: https://panoptes-uploads.zooniverse.org/production/
```

For our current config this does look more redundant, but when we move the media to a different bucket as planned the config becomes:

```yaml
production:
  adapter: aws
  bucket: panoptes-storage
  prefix: production/
  http_prefix: https://panoptes-uploads.zooniverse.org/production/
```

camallen:
closes #1763 #2308 

Based on the work in #3378 (i'll rebase / merge with master once that is in)

This will allow us to set the hosting domain manually via an env var to easily specify the target domain our media assets are hosted on. This PR will allow explicitly modify the AWS stored media src paths to fit into our azure storage path mappings, e.g. in aws a resource like in the static bucket with path
`panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg`
will map to azure public container path location 
`user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg`

This conversion will only apply for AWS resources with the prefix matching `$domain/env/`so new resource created in azure won't be transformed by this change.

These changes will allow us to easily test our azure media adapter by:
1. providing a custom blob store URL vs our CDN url.
2. our desired path mappings to ensure we transfer our resources between s3 and azure blob store correctly so all migrated and new resources will be available as they are in AWS media adatper.

# Review checklist


kubernetes/deployment-staging-azure-canary.tmpl
```diff
@@ -43,6 +43,10 @@ spec:
             value: '65000'
           - name: STORAGE_ADAPTER
             value: azure
+          - name:  STORAGE_URL
+            value: 'https://panoptes.blob.core.windows.net/staging'
```
nciemniak, line 47:
```suggestion
            value: 'https://panoptes.blob.core.windows.net'
```

camallen: the actual storage account container URL in the azure portal is `https://panoptes.blob.core.windows.net/public` so it's the `storage-account.domain/container-name-path`

Not sure where I got staging from tbh. I think it needs to change to `https://panoptes.blob.core.windows.net/public` instead. What do you think?

camallen:
> Based on our conversations in slack and [this PR](https://github.com/zooniverse/operations/pull/480/files), the environment will be set at the storage-account level (i.e. part of the domain, like https://panoptes-staging.blob.core.windows.net), so we wont want to be specifying the environment on the path level. That means we will need to remove the environment from the path if its present (though, we've noted before that for some AWS resources no env is specified), and also not include it in the `STORAGE_URL` env variable.

I may not be understanding properly how the URL will work from the details based in #480. To me (and what i see in the portal) is we have to provide the storage account URL, then the container path prefix and then the actual data path to create the mapped URL. 

Tha said, we won't need this full path for the CDN config for the platform, i.e. 
`https://panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg` URLs they will be mapped by the CDN to the correct storage account container path, e.g. `https://panoptes.blob.core.windows.net/public/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg`

However for manual testing without a CDN we will have to supply the whole URL as per above, e.g. `https://panoptes.blob.core.windows.net/public/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg` to ensure the URL maps to the container path data correctly.

What do you think? Either way it's a simple change on this PR / the ENV variables (url & prefix) to ensure we have the correct path mappings and we can try them out in canary pretty quickly.

app/services/media_storage/stored_path.rb
```diff
@@ -0,0 +1,22 @@
...
+  module StoredPath
+    # Construct the media access URL
+    # and ensure we remove the old aws s3 domain prefix in the stored paths
+    # but leave intact any non old s3 URL, i.e. new azure ones
+    def self.media_url(url, stored_path)
+      uri = URI("https://#{stored_path}")
+      PublicSuffix.parse(uri.host)
+      # we have a valid domain prefix here so remove it
+      # to allow us to construct the URL correctly
+      File.join(url, uri.path)
```
nciemniak, line 13: We will need to remove the environment from the path and add `public`, which will specify the  public container. Something like this-
```suggestion
      uri_path = uri.path.delete_prefix('/'+ENV['RAILS_ENV'])
      uri_path = '/public' + uri_path
      File.join(url, uri_path)
```

app/services/media_storage/azure_adapter.rb
```diff
@@ -42,7 +43,7 @@ def get_path(path, opts={})
           expiry: get_expiry_time(opts[:get_expires] || get_expiration)
         ).to_s
       else
-        "https://#{path}"
+        StoredPath.media_url(url, path)
```
nciemniak, line 46: Rather than storing/hardcoding the url in a env var, we could use the Blob Client's method `@client.generate_uri`, which will return the microsoft domain based off of the storage account name it was initialized with as a URI, e.g. `#<URI::HTTPS https://panoptes.blob.core.windows.net/>
`. That way, if we change the `AZURE_STORAGE_ACCOUNT` env var, we wont also have to change the `STORAGE_URL` env var. In this case, we wouldn't even need THE `STORAGE_URL` env var.

app/services/media_storage/stored_path.rb
camallen: We will need to keep these in for the single `panoptes-uploads.zooniverse.org` CDN all our data sits behind (in production & staging) to distinguish between staging / production target paths that it proxies to.

app/services/media_storage/azure_adapter.rb
camallen: We want to be able to specify this to be different from the blob, i.e. our CDN URL for production deployment. However we want to be able to test out canary / staging etc via the direct path to the storage account container data path. 

For our production deployment we can't break existing URLs so we plan to use `panoptes-uploads.zooniverse.org` as a singular CDN target (exactly how the production & staging systems currently work). We wont' expose the windows blob URLs to our consuming clients (only in canary testing etc)

nciemniak:
> That said, we won't need this full path for the CDN config for the platform, i.e.
https://panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg URLs they will be mapped by the CDN to the correct storage account container path, e.g. https://panoptes.blob.core.windows.net/public/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg

Ohhh right right. this is true... I forgot that we ended up deciding that we'll be mirroring the entire folder structure of what's currently stored in AWS, including the environment folder.

The direction you were going makes sense now 👍 
After spinning around on this for a bit I also thought about how we could just contstruct this URL in the code using the blob client's generate_uri method rather than hardcoding - see the last comment I added!

camallen:
> > That said, we won't need this full path for the CDN config for the platform, i.e.
> > https://panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg URLs they will be mapped by the CDN to the correct storage account container path, e.g. https://panoptes.blob.core.windows.net/public/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
> 
> Ohhh right right. this is true... I forgot that we ended up deciding that we'll be mirroring the entire folder structure of what's currently stored in AWS, including the environment folder.
> 
> The direction you were going makes sense now 👍
> After spinning around on this for a bit I also thought about how we could just contstruct this URL in the code using the blob client's generate_uri method rather than hardcoding - see the last comment I added!

https://github.com/zooniverse/panoptes/pull/3453#discussion_r474249407

Sadly we have to keep our existing URLs as they are bookmarked, linked in the wild. Breaking them / moving away from the the existing URLs would be very bad for our projects, we would break a lot of data in them, subjects, html pages, tutorials a lot. 

Using blob store URLs instead of a domain we own ties us into the cloud environment, i.e. we can't control the domain so can't rewrite, migrate from it etc, i.e. we have to keep the data in azure. So we want to ensure our media URLs still map to `panoptes-uploads.zooniverse.org` domain to keep our systems working.

zwolf:
Please see https://github.com/zooniverse/operations/pull/480 and ensure that those are the names you want. The CDN will sit in front and do a bunch of proxying but that PR outlines what the storage account and container names are gonna be.

camallen:
> Please see [zooniverse/operations#480](https://github.com/zooniverse/operations/pull/480) and ensure that those are the names you want. The CDN will sit in front and do a bunch of proxying but that PR outlines what the storage account and container names are gonna be.

Sure will do - i was using the ones that currently exist (and are configured) for staging canary testing but can change to match the ones that will be created by TF. https://github.com/zooniverse/panoptes/blob/64b4207c057708fdb1d7198a8b1f82b8f71d153b/kubernetes/deployment-staging.tmpl#L94-L96

PR #3455, opened 2020-08-21
Azure adapter: method to get required headers

nciemniak:
The blob service will return a 400 error notifying you of a missing header if you do not include the `x-ms-blob-type` header when making a PUT request to the blob path. Here, I add a method that PFE will need to call in order to see what headers will need to be added to the request. We will also need to make changes to PFE in order for this to work.

I have named the method after [Rails Active Storage](https://github.com/rails/rails/blob/357e15290ab06af8b4f32dd480b48a8c6e31e9e7/activestorage/lib/active_storage/service/azure_storage_service.rb#L104), but I was also thinking of naming it `headers_for_put_path` to better line up with our `put_path` method name (the method will need to be used in conjunction with the `put_path` method). Any preferences?

Also curious if other have thoughts on different ways to fix the current missing header issue.

-----

# Review checklist

### DIFF
--- a/app/services/media_storage/abstract_adapter.rb
+++ b/app/services/media_storage/abstract_adapter.rb
@@ -12,11 +12,18 @@ module MediaStorage
       raise NotImplementedError
     end
 
+    # Returns URL/path to be used to make a GET request to the storage service
     def get_path(path, opts={})
       raise NotImplementedError
     end
 
+    # Returns URL/path to be used to make a PUT request to the storage service
     def put_path(path, opts={})
+      raise NotImplementedError
+    end
+
+    # Returns required headers for making a put request to the storage service
+    def headers_for_direct_upload
       raise NotImplementedError
     end
 
--- a/app/services/media_storage/azure_adapter.rb
+++ b/app/services/media_storage/azure_adapter.rb
@@ -24,6 +24,7 @@ module MediaStorage
       )
     end
 
+    # Returns the path at which the medium lives in Blob Storage
     def stored_path(content_type, medium_type, *path_prefix)
       extension = get_extension(content_type)
       path = prefix.to_s
@@ -33,6 +34,7 @@ module MediaStorage
       path + "#{SecureRandom.uuid}.#{extension}"
     end
 
+    # Returns URL/path to be used to make a GET request to the Blob Service
     def get_path(path, opts={})
       if opts[:private]
         signer.signed_uri(
@@ -47,6 +49,7 @@ module MediaStorage
       end
     end
 
+    # Returns URL/path to be used to make a PUT request to the Blob Service
     def put_path(path, opts={})
       container = opts[:private] ? private_container : public_container
 
@@ -58,6 +61,11 @@ module MediaStorage
         expiry: get_expiry_time(opts[:put_expires] || put_expiration),
         content_type: opts[:content_type]
       ).to_s
+    end
+
+    # Returns required headers for making a put request to the Blob Service
+    def headers_for_direct_upload
+      { "x-ms-blob-type" => "BlockBlob" }
     end
 
     def put_file(path, file_path, opts={})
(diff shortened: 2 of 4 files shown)


app/services/media_storage/abstract_adapter.rb
```diff
@@ -12,14 +12,21 @@ def stored_path(content_type, media_type, *path_prefix)
...
       raise NotImplementedError
     end
 
+    # Returns URL/path to be used to make a PUT request to the storage service
     def put_path(path, opts={})
       raise NotImplementedError
     end
 
+    # Returns required headers for making a put request to the storage service
+    def headers_for_direct_upload
```
camallen, line 26: is this abstract across all adapter implementations? If not it's probably a private? method on the azure adapter only

app/services/media_storage/azure_adapter.rb
```diff
@@ -60,6 +63,11 @@ def put_path(path, opts={})
       ).to_s
     end
 
+    # Returns required headers for making a put request to the Blob Service
+    def headers_for_direct_upload
```
camallen, line 67: where will this be called as a public method? 

I assume these are required for by a client when using a signed PUT URL. If so perhaps they should be returned as headers on the put URL response? 

Alternatively if these aren't dynamic we can add these in as code as well on the uploading clients themselves (PFE, python client etc). That way we provide the client with the info it needs via response headers but also ensure we have these in use on our trusted clients.

nciemniak: > I assume these are required for by a client when using a signed PUT URL. If so perhaps they should be returned as headers on the put URL response?

@camallen yep the header is required when using a PUT URL whether it is signed or not (i.e. required for both public and private uploads). What do you mean when you say 'they should be returned as headers on the PUT URL response'? Currently, we send over the PUT URL in the `src` attribute of the serialized medium object- my thought was that we could add a `headers` attribute on the medium serializer that the client could access. Is there a different way to do it than that?
https://github.com/zooniverse/panoptes/blob/85f26dcaaf4b8788a802874d28521c9eed0b162f/app/serializers/medium_serializer.rb#L19

>Alternatively if these aren't dynamic we can add these in as code as well on the uploading clients themselves (PFE, python client etc).

Indeed, the headers are not dynamic. So you're saying we could just add code for the header _just_ in our uploading clients (PFE, python client etc) and not make any changes to the panoptes backend? My only concern with this is that we only want to be adding the header if we're talking to azure storage, and technically we dont want the client to need to worry about whether it is uploading to azure or to aws.

camallen: > What do you mean when you say 'they should be returned as headers on the PUT URL response'?
As literal response headers, not in the response object. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

I like the fact you've copied the ActiveStorage code here, checkout how it's used in the HTTP response object as headers, https://github.com/rails/rails/blob/855de1a7004a6cb87b7f637b2e94edda5f3e4e56/activestorage/app/controllers/active_storage/direct_uploads_controller.rb#L20

I believe our clients do most of this work themselves, e.g. https://github.com/zooniverse/panoptes-python-client/blob/7f3f4833adbd89bb60e9965b4a747ef7d552283a/panoptes_client/subject.py#L185

Providing support for suggested client upload headers from the API repsonse is a good idea 👍

app/services/media_storage/abstract_adapter.rb
camallen: If we provide content-type, content-disposition etc then this method would exist in each concrete implementation class, so this would be an abstract method. e.g https://github.com/rails/rails/blob/f16431e0c04a72e5a42fe1f7c5b5aefe76220ddc/activestorage/lib/active_storage/service.rb#L136-L138

camallen:
I'm closing for now, we can re-open if we get time to work on this again.

PR #3467, opened 2020-09-08
fetch serialised classification data from read replica

camallen:
move this read query load off the primary db to the replica

# Review checklist

### DIFF
--- a/app/controllers/api/v1/classifications_controller.rb
+++ b/app/controllers/api/v1/classifications_controller.rb
@@ -53,9 +53,7 @@ class Api::V1::ClassificationsController < Api::ApiController
   end
 
   def incomplete
-    DatabaseReplica.read('classification_serializer_data_from_replica') do
-      index
-    end
+    index
   end
 
   def project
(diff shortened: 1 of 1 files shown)


nciemniak:
Just an initial thought, I'm thinking it'd be good to add the `classification_serializer_data_from_replica` flag to this wiki page for future reference: https://github.com/zooniverse/panoptes/wiki/Feature-Flags

nciemniak (approved):
Code changes look good to me 👍 would just ask for documentation of the new feature flag in the wiki like I mentioned in previous comment.

Just one question - why is classification data a good candidate when it comes to retrieving data from the read replica? I vaguely agree with the choice, but off the top of my head can't think of the places where the api/classifications endpoint is hit in PFE and whether we need the data retrieved at that point to be *the most-up-to-date* data.

camallen:
Added the flag docs - https://github.com/zooniverse/panoptes/wiki/Feature-Flags#classification_serializer_data_from_replica-on

> Just one question - why is classification data a good candidate when it comes to retrieving data from the read replica? I vaguely agree with the choice, but off the top of my head can't think of the places where the api/classifications endpoint is hit in PFE and whether we need the data retrieved at that point to be _the most-up-to-date_ data.

These end points aren't really used in PFE, rather they are used by some research teams to ingest their project classification data to their research databases for their own analysis. 

That said the `/incomplete` route is used by ASM (i believe) to save a users partially completed classification data so they can access it at a later point. The ASM project is complete so not using this feature any more.

nciemniak:
Awesome, thanks for the info!

PR #3470, opened 2020-09-11
Azure adapter: remove env prefix from stored path

nciemniak:
Azure Adapter: 
Construct `stored_path` without env prefix.

`media.src` attributes (in the database) will now look something like this:
`subject_location/7c669787-2677-40ce-90bc-5d3ef185863d.jpeg`
`project_avatar/1adcb95b-06bf-4f84-ab83-530c1df7c086.jpeg`

In leu of using the env prefix, we will have two CDN domains, one for prod and one for staging:
staging: `panoptes-uploads-staging.zooniverse.org`
production: `panoptes-uploads.zooniverse.org`
We will store the respective domain URLs in the `STORAGE_URL` env var. We'll also include the container name in the `STORAGE_URL` env var, so it'll look like `panoptes-uploads-staging.zooniverse.org/public`, etc.

The azure adapter `get_url` method will call the `StoredPath` class, which will use the correct domain to construct the URL at which the public media item can be accessed. The final URL returned to the client will look something like this:
`panoptes-uploads-staging.zooniverse.org/public/subject_location/7c669787-2677-40ce-90bc-5d3ef185863d.jpeg`

-----

# Review checklist

### DIFF
--- a/app/services/media_storage/azure_adapter.rb
+++ b/app/services/media_storage/azure_adapter.rb
@@ -3,6 +3,7 @@
 module MediaStorage
   class AzureAdapter < AbstractAdapter
     attr_reader :url, :public_container, :private_container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
+
     DEFAULT_EXPIRES_IN = 3 # time in minutes, see get_expiry_time(expires_in)
 
     def initialize(opts={})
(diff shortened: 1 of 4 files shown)


camallen (approved):
LGTM - I was wondering if we could have left the prefix behaviour and removed the default prefix value being set, e.g. `@prefix = opts[:prefix]` instead of removing completely. That said we probably won't need this and can add it back in if we do.

nciemniak:
My reasoning on why I removed it was that I wanted the code to reflect the actual behavior (so that when we go back and look at it we dont end up spending time thinking about whether that `@prefix` attribute ever gets used or not), if our current plan does not use the prefix, might as well remove it from the code. 

If it turns out that this way of proceeding doesn't work, we'll probably need to make other code changes as well if we need to add the prefix back in.

PR #3476, opened 2020-09-17
ignore PG::QueryCanceled statement timeout errors

camallen:
Ignore the raised errors when db connections hit their timeout limit. 

Since we set explicit DB connection timeouts in #3369 and #3278 to avoid long running queries soaking up db resources (API is 65s as the loadbalancer times out at 60s, workers are 5 mins), we can safely ignore these errors and there is no need to send these to HB for reporting either as they are expected errors for slow queries.

We could turn this reporting to HB on with a feature flag in case we want to monitor these errors via HB but i'm not sure that's really useful. For ~3-4 months now we've seen these come in via HB and basically ignored them as they can happen across our app queries as are related to how busy the app /db is. 

Noting we will still see these as response status 500 errors on our logs, loadbalancers / CDN etc we just won't get notified / track these in honeybadger.

# Review checklist


zwolf (changes requested):
I think the `get_request` let in the test needs to actually be called somewhere. The `let` is w/o a bang and outside of a before block, while the expect block below only flushes HB and doesn't call it or the `get` directly. I don't think the request to the test controller is actually being made, but the test still passes because nothing is changing.

camallen:
> I think the `get_request` let in the test needs to actually be called somewhere. The `let` is w/o a bang and outside of a before block, while the expect block below only flushes HB and doesn't call it or the `get` directly. I don't think the request to the test controller is actually being made, but the test still passes because nothing is changing.

Thanks @zwolf nice catch and fail on my part :(

I've updated the specs to ensure we do call the actual test this time. Also fwiw this test is complex and modifies the routing rules / adds a custom anonymous controller for integration style specs, it's not nice. I really wanted to make sure i new how the error handler worked in HB and be able to actively debug it should i need. 

I'd be happy to drop the test and add a note on the PR to doc how to test the HB notify behaviour, it's quite well documented in HB and we're using public API here so i don't really think we need a test. Thoughts?

zwolf:
I had to check the docs to review this PR, so maybe including a link to the docs as a comment would be a good idea. Those docs are pretty comprehensive so I'm fine with the complexity in the test but if you don't figure it's really necessary, that's fine. Maybe it's worth a test ensuring that HB is being informed when you want it to be, 

That said, another quick non-blocking question: why match the full text of an error instead of the class? Your module could store an array of classes and you could use `includes?`.

zwolf (approved):

config/initializers/honeybadger.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module IngoredErrorRegexes
+  PG_QUERY_TIMEOUT = /^ActiveRecord::StatementInvalid: PG::QueryCanceled: ERROR:  canceling statement due to statement timeout/.freeze
+end
+
+Honeybadger.configure do |config|
+  # ignore the active record PG statement timeout errors
```
camallen, line 8:
```suggestion
  # ignore the active record PG statement timeout errors
  # https://docs.honeybadger.io/lib/ruby/getting-started/ignoring-errors.html#ignore-programmatically
```

camallen:
>  had to check the docs to review this PR, so maybe including a link to the docs as a comment would be a good idea. Those docs are pretty comprehensive so I'm fine with the complexity in the test but if you don't figure it's really necessary, that's fine. Maybe it's worth a test ensuring that HB is being informed when you want it to be,

I've added the docs. I'm going to leave the specific ignore by error message behaviour test in for now. I'd rather test the explicit behaviour we have modified in code vs the normal gem behaviour (that's well tested).

> That said, another quick non-blocking question: why match the full text of an error instead of the class? Your module could store an array of classes and you could use includes?.

The error class of all invalid DB statements is `ActiveRecord::StatementInvalid` this class encompasses the `PG::QueryCanceled` statement timeout as well as other error types like `PG::ConnectionBad: PQconsumeInput() server closed the connection`. Ignoring based solely on error class will not report those errors as well and we want to see errors like that.

PR #3485, opened 2020-10-15
add sidekiq pod disruption budget

camallen:
ensure we have at least 1 sidekiq pod running in the k8s cluster

when i was upgrading the nodepool today i managed to evict these pods and they couldn't reschedule due to limits on the available node pool (autoscaler didn't kick in) so sidekiq was down for 30 mins

this PDB should alleviate this issue as when draining nodes, the pods won't evict till rescheduled properly all without service downtime

# Review checklist


zwolf (approved):
LGTM

PR #3486, opened 2020-10-16
control ratio of real and training subjects in fallback selection

camallen:
by default provide at least 1 training subject for every 10 real subjects in fallback selection

avoid showing the user mostly training subjects

# Review checklist

### DIFF
--- a/lib/subjects/fallback_selection.rb
+++ b/lib/subjects/fallback_selection.rb
@@ -2,30 +2,58 @@ module Subjects
   class FallbackSelection
     attr_reader :workflow, :limit, :options
 
+    def self.num_training_subjects_limit
+      ENV.fetch('FALLBACK_NUM_TRAINING_SUBJECTS_LIMIT', 10)
+    end
+
     def initialize(workflow, limit, options = {})
       @workflow, @limit, @options = workflow, limit, options
     end
 
     def any_workflow_data
-      any_workflow_data_scope
-        .limit(limit)
-        .pluck("set_member_subjects.subject_id")
-        .shuffle
+      if workflow.grouped
+        fallback_grouped_selection.shuffle
+      else
+        fallback_selection.shuffle
+      end
     end
 
     private
 
-    def any_workflow_data_scope
-      scope = workflow.set_member_subjects
-      if workflow.grouped
-        if subject_set_id = options[:subject_set_id]
-          scope = scope.where(subject_set_id: subject_set_id)
-        else
-          msg = "subject_set_id parameter missing for grouped workflow"
-          raise Subjects::Selector::MissingParameter.new(msg)
-        end
+    def fallback_grouped_selection
+      unless (subject_set_id = options[:subject_set_id])
+        raise Subjects::Selector::MissingParameter, 'subject_set_id parameter missing for grouped workflow'
       end
-      scope
+
+      workflow
+        .set_member_subjects
+        .where(subject_set_id: subject_set_id)
+        .limit(limit)
+        .pluck('set_member_subjects.subject_id')
+    end
+
+    def fallback_selection
+      (non_training_subject_ids | training_subject_ids)
+    end
+
+    def non_training_subject_ids
+      SetMemberSubject
+        .where(subject_set_id: workflow.non_training_subject_set_ids)
+        .limit(limit)
+        .pluck('set_member_subjects.subject_id')
+    end
+
+    def training_subject_ids
+      return [] if workflow.training_set_ids.empty?
+
+      # by default a ~10:1 ratio of real subjects to training
+      # modify this ration by num_training_subjects_limit class method
+      training_limit = (limit.to_f / self.class.num_training_subjects_limit).round
+
+      SetMemberSubject
+        .where(subject_set_id: workflow.training_set_ids)
+        .limit(training_limit)
+        .pluck('set_member_subjects.subject_id')
     end
   end
 end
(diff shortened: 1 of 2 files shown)


nciemniak:
@camallen for context, what is fallback selection? when does it get used?

camallen:
Fallback selection is used as a last resort to keep the interface working / show the user some data. 

When a workflow is completed OR when a volunteer has seen all the available (non-retired) data, we trigger this short circuit to save on compute resources. 

So this PR will ensure we aren't returning only training data to the requesting volunteer, most likely a power user. This ensures we see both 'real' and 'training' subjects in the interface.

lib/subjects/fallback_selection.rb
```diff
@@ -2,30 +2,58 @@ module Subjects
...
+        .limit(limit)
+        .pluck('set_member_subjects.subject_id')
+    end
+
+    def fallback_selection
+      (non_training_subject_ids | training_subject_ids)
+    end
+
+    def non_training_subject_ids
+      SetMemberSubject
```
nciemniak, line 40: It looks like the original fallback selection code was filtering based on non retired and unseen for user. See [workflow.set_member_subjects](https://github.com/zooniverse/panoptes/blob/master/lib/subjects/fallback_selection.rb#L19) , if you click into the def'n you'll see that it returns either `select_non_retired_unseen_for_user` or `select_non_retired`. Is that something we want to include in this query here?

nciemniak (changes requested):
A few issues/questions - see the comments!

nciemniak:
Also, just one more big-picture question in regards to the use of the fallback selector code - you said the common cases where fallback selector is used is when workflow is completed or volunteer has seen all non-retired data. I'm wondering why don’t we instead tell the user that there are no more subjects to classify in these cases? Wouldn't that make more sense?

nciemniak (approved):
Approving / requesting minor changes listed previously 👍

lib/subjects/fallback_selection.rb
```diff
@@ -2,30 +2,58 @@ module Subjects
   class FallbackSelection
     attr_reader :workflow, :limit, :options
 
+    def self.num_training_subjects_limit
+      ENV.fetch('FALLBACK_NUM_TRAINING_SUBJECTS_LIMIT', 10)
```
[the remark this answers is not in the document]
camallen: i've renamed to `FALLBACK_NUM_TRAINING_SUBJECTS_DENOMINATOR` and `num_training_subjects_denominator` to reflect their use in creating the underlying ratio limits.

spec/lib/subjects/fallback_selection_spec.rb
```diff
@@ -27,6 +33,22 @@
...
+      end
+
+      before do
+        allow(workflow).to receive(:training_set_ids).and_return(training_set_id)
+      end
+
+      it 'includes some training subject ids from the workflow' do
+        selected_training_subject_ids = training_subject_ids & subject_ids
+        expect(selected_training_subject_ids).not_to be_empty
+      end
```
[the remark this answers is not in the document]
camallen: Nice catch - I've modified the code to respect the original request limit when there is training and no training linked subjects.

lib/subjects/fallback_selection.rb
camallen: I'm not sure this is the case, the `workflow.set_member_subjects` is a defined association on the workflow and doesn't filter those records on their retirement status at all, https://github.com/zooniverse/panoptes/blob/1fabe3c3ae29955d07c8fdb82b273381660bbcaa/app/models/workflow.rb#L19

camallen:
> Also, just one more big-picture question in regards to the use of the fallback selector code - you said the common cases where fallback selector is used is when workflow is completed or volunteer has seen all non-retired data. I'm wondering why don’t we instead tell the user that there are no more subjects to classify in these cases? Wouldn't that make more sense?

One of the original use cases for this code path is to keep the interface working for data provenance / historical reasoning about the UX. As such the selector had to return data to meet this use case. FWIW the classifier UI fails if this end point doesn't return any data (thought that can change).

We do supply a lot of contextual information during subject selection which the classifier can reflect on to provider improved UX to the volunteers, indeed the new classifier in the monorepo does provide improved UX on this front.

See https://github.com/zooniverse/panoptes/blob/1fabe3c3ae29955d07c8fdb82b273381660bbcaa/lib/subjects/selector_context.rb#L17-L25
which is used in subject selection via https://github.com/zooniverse/panoptes/blob/708167b59adef74e7545a8a38aab9e53ab1bad96/app/controllers/api/v1/subjects_controller.rb#L39-L42
and then used in the serializer to provide some context on why each subject was selected via https://github.com/zooniverse/panoptes/blob/708167b59adef74e7545a8a38aab9e53ab1bad96/app/serializers/subject_selector_serializer.rb#L25-L47

lib/subjects/fallback_selection.rb
nciemniak: Ah, you're totally right. My bad, when I clicked on `workflow.set_member_subjects` in github it brought me to [SetMemberSubjectSelector#set_member_subjects](https://github.com/zooniverse/panoptes/blob/1fabe3c3ae29955d07c8fdb82b273381660bbcaa/lib/subjects/set_member_subject_selector.rb#L11) and I mistakenly assumed that was the method that was being called. Thanks for clarifying that 👍

nciemniak:
Makes sense, I appreciate the extra context!

PR #3488, opened 2020-10-19
Remove env prefix from stored path if present

nciemniak:
We are no longer specifying the environment via the URI path; instead, environment is specified via the domain. Update the code to remove the environment prefix if it is present in the URI path.

------
# Review checklist

### DIFF
--- a/app/services/media_storage/stored_path.rb
+++ b/app/services/media_storage/stored_path.rb
@@ -10,11 +10,15 @@ module MediaStorage
       PublicSuffix.parse(uri.host)
       # we have a valid domain prefix here so remove it
       # to allow us to construct the URL correctly
+      env_prefix = '/' + Rails.env
+      if uri.path.start_with? env_prefix # remove env prefix if present
+        uri.path.slice! env_prefix
+      end
       File.join(url, uri.path)
     rescue PublicSuffix::DomainNotAllowed
       # failure here indicates we do not have
       # a valid domain prefix in the stored path
-      # e.g. production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
+      # e.g. /user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
       # so we do not need to rewrite the URL
       File.join(url, stored_path)
     end
--- a/app/services/media_storage/stored_path.rb
+++ b/app/services/media_storage/stored_path.rb
@@ -7,13 +7,17 @@ module MediaStorage
     # but leave intact any non old s3 URL, i.e. new azure ones
     def self.media_url(url, stored_path)
       uri = URI("https://#{stored_path}")
+      # if the parse succeeds, we know we have a valid domain prefix,
+      # i.e. it is an old s3 stored path - hence, we need to rewrite the url
       PublicSuffix.parse(uri.host)
-      # we have a valid domain prefix here so remove it
-      # to allow us to construct the URL correctly
-      env_prefix = '/' + Rails.env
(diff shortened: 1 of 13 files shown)


app/services/media_storage/stored_path.rb
```diff
@@ -10,11 +10,14 @@ def self.media_url(url, stored_path)
       PublicSuffix.parse(uri.host)
       # we have a valid domain prefix here so remove it
       # to allow us to construct the URL correctly
+      env_prefix = '/' + Rails.env
```
camallen, line 13: I assume we won't ever have a `stored_path` value of `production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg` (no leading `/`) ?

If we do this env_prefix won't strip the value correctly due to missing the `/`

app/services/media_storage/stored_path.rb
```diff
@@ -10,11 +10,14 @@ def self.media_url(url, stored_path)
...
+      uri.path.slice! env_prefix if uri.path.start_with? env_prefix # remove env prefix if present
```
camallen, line 14: I prefer a regex with capture group for this use case rather than testing if the path starts with and if so slicing the substring out. A regex could handle a missing leading `/` too.

I think the code is simpler this way but i'll defer to you, e.g. regex like 
`\A(?:\/?#{Rails.env})?(.+)\z`

where `\A` is start of string, `\z` is end of string
`(?:\/?production)?` is a non-capture group `(?:)`, it may have a leading `/` (zero or more) followed by the rails env we want to ignore, note this non-capture group doesn't have to be present but will match if it is.
`(.+)` the greedy capture group for the resource path we want to serve
try it at https://rubular.com/ with input like
```
/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
/test/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
/staging//user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
```
e.g. 
```
# micro op make this regex a class constant or class method to avoid instantiating it each time this method is called
RAILS_ENV_REMOVAL_REGEX = /\A(?:\/?production)?(.+)\z/

# get the 1st capture group (0 index is the original string) which should be the trailing media type / UUID regex etc.
uri_path = uri.match(RAILS_ENV_REMOVAL_REGEX)[1]
```

camallen (approved):
I left some non-blocking comments, something about the slice! method way bugs me. I think it's because it modifies the input string `!` and the `starts_with` test... is i normally use a regex for this type of sub-string matching work.

app/services/media_storage/stored_path.rb
nciemniak: The value we're checking in this case will always be the `.path` value of the URI object. `.path` will reliably return the path with a leading `/`, so we will never end up with a value without a leading `/`.

nciemniak:
@camallen because we can reliably expect the `.path` method of the URI object to return a string with a leading `/`, I'd prefer to use `.start_with?` - to me this is simple, clean and straightforward.

As for the use of `slice!`, maybe could use `sub` instead and make the update more explicit? Like so
```ruby
      uri_path = uri.path

      env_prefix = '/' + Rails.env
      if uri_path.start_with? env_prefix # remove env prefix if present
        uri_path = uri_path.sub(env_prefix, '')
      end
```

nciemniak:
I'm going to go ahead with the merge since the discussion over `slice!` vs `sub` isn't really worth holding up this PR over 😎

PR #3489, opened 2020-10-23
Modify the production & staging pod disruption budgets (pdb)

camallen:
#### switch pdb to percent of available replicas
avoid hard coding the number of PDB replicas, instead use a percent of the available replicas https://kubernetes.io/docs/tasks/run-application/configure-pdb/

This allows more lenient eviction rules for the relevant pods. I came across pods that wouldn't evict due to the PDB settings when doing the cluster / node pool upgrades the other day. Specifically from https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget
> If you set maxUnavailable to 0% or 0, or you set minAvailable to 100% or the number of replicas, you are requiring zero voluntary evictions. When you set zero voluntary evictions for a workload object such as ReplicaSet, then you cannot successfully drain a Node running one of those Pods. If you try to drain a Node where an unevictable Pod is running, the drain never completes. This is permitted as per the semantics of PodDisruptionBudget.

#### add pdb to production redis node
ensure we have the panoptes redis pod available at all time. 
Note that using the same value as our replica count will mean this pod can't voluntarily evict which is good as we want this redis node up all the time. That said, this setting will have consequences for node pool upgrades, these can be alleviated by redeploying e.g. `kubectl rollout restart deployment panoptes-production-redis`

#### remove the staging app pdb
Allow the scheduler to keep this node up, we can tolerate some downtime on the staging app and if it's down we can manually investigate / intervene (most likely due to node issues)

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -245,7 +245,7 @@ kind: PodDisruptionBudget
 metadata:
    name: panoptes-production-app-pdb
 spec:
-  minAvailable: 2
+  minAvailable: 50%
   selector:
     matchLabels:
       app: panoptes-production-app
@@ -337,7 +337,7 @@ kind: PodDisruptionBudget
 metadata:
    name: panoptes-production-sidekiq-pdb
 spec:
-  minAvailable: 1
+  minAvailable: 50%
   selector:
     matchLabels:
       app: panoptes-production-sidekiq
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -428,6 +428,16 @@ spec:
       targetPort: 6379
   type: NodePort
 ---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+   name: panoptes-production-redis-pdb
+spec:
+  minAvailable: 1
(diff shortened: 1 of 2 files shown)


zwolf (changes requested):
Was this linted/dry run? Those percentages might need to be represented as strings ("50%")

>When you specify a percentage by setting the value to a string representation of a percentage (eg. "50%"), it represents a percentage of total Pods. For instance, if you set maxUnavailable to "50%", then only 50% of the Pods can be unavailable during a disruption.

camallen:
> Was this linted/dry run? Those percentages might need to be represented as strings ("50%")
> 
> > When you specify a percentage by setting the value to a string representation of a percentage (eg. "50%"), it represents a percentage of total Pods. For instance, if you set maxUnavailable to "50%", then only 50% of the Pods can be unavailable during a disruption.

Yes - it's linted through the Jenkinsfile https://github.com/zooniverse/panoptes/blob/43921bab5a65c530f1c5cc24774e0d4d4d0b205e/Jenkinsfile#L49-L57

This was tested in caesar deployment, see https://github.com/zooniverse/caesar/pull/1223#issuecomment-717272845

PR #3496, opened 2020-10-28
extract azure path from s3 url for private media

nciemniak:
Add method for getting just the media path to the StoredPath module. For use by the azure adapter. This will fix requests for private media items stored in azure.

-----

# Review checklist

### DIFF
--- a/app/services/media_storage/azure_adapter.rb
+++ b/app/services/media_storage/azure_adapter.rb
@@ -2,7 +2,7 @@
 
 module MediaStorage
   class AzureAdapter < AbstractAdapter
-    attr_reader :url, :public_container, :private_container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
+    attr_reader :domain_prefix, :public_container, :private_container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
 
     DEFAULT_EXPIRES_IN = 3 # time in minutes, see get_expiry_time(expires_in)
 
@@ -10,7 +10,7 @@ module MediaStorage
       @storage_account_name = opts[:azure_storage_account]
       @public_container = opts[:azure_storage_container_public]
       @private_container = opts[:azure_storage_container_private]
-      @url = opts[:url]
+      @domain_prefix = opts[:domain_prefix]
       @get_expiration = opts.dig(:expiration, :get) || DEFAULT_EXPIRES_IN
       @put_expiration = opts.dig(:expiration, :put) || DEFAULT_EXPIRES_IN
 
@@ -42,7 +42,7 @@ module MediaStorage
           expiry: get_expiry_time(opts[:get_expires] || get_expiration)
         ).to_s
       else
-        StoredPath.media_url(url, path)
+        StoredPath.media_url(domain_prefix, path)
       end
     end
 
--- a/app/services/media_storage/azure_adapter.rb
+++ b/app/services/media_storage/azure_adapter.rb
@@ -33,8 +33,9 @@ module MediaStorage
 
     def get_path(path, opts={})
       if opts[:private]
+        azure_path = StoredPath.media_path(path)
         signer.signed_uri(
-          client.generate_uri("#{private_container}/#{path}"),
+          client.generate_uri("#{private_container}/#{azure_path}"),
           false,
           service: 'b', # blob
           permissions: 'r', # read
--- a/app/services/media_storage/azure_adapter.rb
+++ b/app/services/media_storage/azure_adapter.rb
@@ -2,7 +2,7 @@
 
 module MediaStorage
   class AzureAdapter < AbstractAdapter
-    attr_reader :domain_prefix, :public_container, :private_container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
+    attr_reader :url_prefix, :public_container, :private_container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
 
     DEFAULT_EXPIRES_IN = 3 # time in minutes, see get_expiry_time(expires_in)
 
@@ -10,7 +10,7 @@ module MediaStorage
       @storage_account_name = opts[:azure_storage_account]
       @public_container = opts[:azure_storage_container_public]
       @private_container = opts[:azure_storage_container_private]
-      @domain_prefix = opts[:domain_prefix]
+      @url_prefix = opts[:url_prefix]
       @get_expiration = opts.dig(:expiration, :get) || DEFAULT_EXPIRES_IN
       @put_expiration = opts.dig(:expiration, :put) || DEFAULT_EXPIRES_IN
 
@@ -42,7 +42,7 @@ module MediaStorage
           expiry: get_expiry_time(opts[:get_expires] || get_expiration)
         ).to_s
       else
-        StoredPath.media_url(domain_prefix, path)
+        StoredPath.media_url(url_prefix, path)
       end
     end
 
--- a/app/services/media_storage/stored_path.rb
+++ b/app/services/media_storage/stored_path.rb
@@ -10,36 +10,43 @@ module MediaStorage
   # These URLs need to be rewritten such that domain prefix and env get removed
   # Azure native paths do not need to be rewritten
   module StoredPath
-    # Construct path and join to the passed in URL
-    def self.media_url(url, stored_path)
-      uri = URI("https://#{stored_path}")
-      # if the parse succeeds, we know we have a valid domain prefix,
-      # i.e. it is an old s3 stored path and we need to rewrite it
-      PublicSuffix.parse(uri.host)
+    class << self
+      ENV_PATH_PREFIX = '/' + Rails.env
 
-      uri_path = path_without_env_prefix(uri.path)
-      File.join(url, uri_path)
-    rescue PublicSuffix::DomainNotAllowed
-      # if parse fails, we do not have a valid domain prefix in the stored path
-      # e.g. /user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
-      # so we do not need to rewrite the URL
-      File.join(url, stored_path)
-    end
+      def media_path(stored_path)
+        rewrite_stored_path(stored_path)
+      rescue PublicSuffix::DomainNotAllowed
+        # a valid stored path without a TLD prefix
+        stored_path
+      end
 
-    def self.media_path(stored_path)
-      uri = URI("https://#{stored_path}")
-      PublicSuffix.parse(uri.host)
-      # S3 path
-      path_without_env_prefix(uri.path)
-    rescue PublicSuffix::DomainNotAllowed
-      # azure path
-      stored_path
-    end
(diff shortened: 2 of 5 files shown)


camallen:
> I am thinking of extracting some of the comments in the `StoredPath` module into an explanatory Wiki page since the comments feel a bit cluttery in this module, keeping a brief description in the code as well as a reference to the wiki page? would feel cleaner to me.

by all means cleanup the code / comments, perhaps the code isn't as clear as it could be written? However I would avoid adding docs outside of the code. In my experience these quickly get out of sync with the actual code and are not easily found when looking at the code.
 
> Rather than relying on `PublicSuffix.parse(uri.host)` to error and needing to rescue. Any reason to believe that using `PublicSuffix.parse` is preferable?

The reason I used the`PublicSuffix` module was to distinguish between valid TLD (top level domain) prefixed paths, like we did in AWS land. The `PublicSuffix.parse` method will allow real URI host(s) (really just the first part of the `/` delimited `stored_path` string) but not allow non valid TLD, i.e. our new azure `stored_path` strings `subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg`

> Also, I would propose doing the domain/host check with something like this:
> 
> ```ruby
> if uri.host
>   # we have a valid domain
> else
>   # we dont have a valid domain
> end
> ```
In your example if we convert the `subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg` string to a URI then the host will be `subject_location` - so we need a way to detect and remove the TLD in the old stored paths, i.e. the `panoptes-uploads.zooniverse.org` part.

E.G.
``` bash
# with a non domain prefixed stored path (new azure paths)
[1] pry(main)> stored_path = 'production/subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg'
=> "production/subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg"
[2] pry(main)> uri = URI("https://#{stored_path}")
=> #<URI::HTTPS https://production/subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg>
[3] pry(main)> uri.host
=> "production"
[4] pry(main)> PublicSuffix.parse(uri.host)
PublicSuffix::DomainNotAllowed: `production` is not allowed according to Registry policy

# with an old domain prefixed stored path (new azure paths)
[8] pry(main)> stored_path = 'panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg'
=> "panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg"
[9] pry(main)> uri = URI("https://#{stored_path}")
=> #<URI::HTTPS https://panoptes-uploads.zooniverse.org/production/user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg>
[10] pry(main)> uri.host
=> "panoptes-uploads.zooniverse.org"
[11] pry(main)> PublicSuffix.parse(uri.host)
=> #<PublicSuffix::Domain:0x00007f81a87bccb8 @sld="zooniverse", @tld="org", @trd="panoptes-uploads">
end
```

nciemniak:
> by all means cleanup the code / comments, perhaps the code isn't as clear as it could be written? However I would avoid adding docs outside of the code. In my experience these quickly get out of sync with the actual code and are not easily found when looking at the code.

Valid point. I think the reason we need comments/documentation here in the first place is because we have this funky situation of having two different URL formats (s3 vs azure), it's one of those situations where just the code by itself won't explain why we need to do what we're doing. I'm going to rewrite comments a little to see if I can make it more concise/direct.

> In your example if we convert the `subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg `string to a URI then the host will be `subject_location`

Ahh, you are completely right. I was testing it out in IRB by passing in the string without `https://` in front of it. In that case, it seems like it'll be best to stick with using PublicSuffix.

app/services/media_storage/stored_path.rb
```diff
@@ -1,28 +1,45 @@
...
+  # Old S3 paths will include the panoptes-uploads domain prefix
+  # and the environment in their path, for example:
+  # panoptes-uploads.zooniverse.org/staging/subject_location/49f02f969a5.jpeg
+  # These URLs need to be rewritten such that domain prefix and env get removed
+  # Azure native paths do not need to be rewritten
   module StoredPath
-    # Construct the media access URL
-    # and ensure we remove the old aws s3 domain prefix in the stored paths
-    # but leave intact any non old s3 URL, i.e. new azure ones
+    # Construct path and join to the passed in URL
```
camallen, line 13: URL here isn't really a good descriptor (or param name tbh). Really this is a domain prefix / hostname rather than a URL, https://developer.mozilla.org/en-US/docs/Web/API/URL

Perhaps we should rename here (and other places) while we're thinking about this?  

Perhaps update this as well as i don't think this is a URL either, perhaps a URL prefix? https://github.com/zooniverse/panoptes/blob/6287de9674f6c684c522e432a0016ff36ad27532/app/services/media_storage/azure_adapter.rb#L13

camallen (commented):
I think this could do with a refactor to reuse common logic
Perhaps the `MediaStorage::StoredPath` logic could change to something like the following.
``` ruby
ENV_PATH_PREFIX = env_prefix = '/' + Rails.env

def self.media_path(stored_path)
  rewrite_stored_path(stored_path)
rescue PublicSuffix::DomainNotAllowed
  # a valid stored path without a TLD prefix
  stored_path
end

def self.media_url(domain_prefix, stored_path)
  rewritten_path = rewrite_stored_path(stored_path)
rescue PublicSuffix::DomainNotAllowed
  rewritten_path = stored_path
ensure
  # a valid stored path without a TLD prefix, combine with the supplied domain prefix
  File.join(domain_prefix, rewritten_path)
end

def rewrite_stored_path(stored_path)
  uri = URI("https://#{stored_path}")
  
  # throw PublicSuffix::DomainNotAllowed if uri.host parse fails
  # this indicates no valid top level domain (TLD) prefix in the stored path
  # e.g. /user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
  # this means the stored_path is valid and we do not need to rewrite the URL at all
  PublicSuffix.parse(uri.host)
   
  # remove Rails env prefix if present (remnant path prefix from s3 land)
  if uri.path.start_with?(ENV_PATH_PREFIX)
    uri.path.sub(env_prefix, '')
  else
    uri.path
  end
end
```

I wouldn't be too concerned with private class methods, perhaps this is an indicator we could use a real class and public / private methods vs `static` class methods for this state manipulation?

nciemniak:
Big fan of the suggestion, definitely dries up the code 👍 Note that in the `media_url` method I was not able to to use `ensure` because ensure does not alter the return value. Also could not do a return within the ensure because it silently rescues (i.e. it would do a rescue even if the error was something other than `DomainNotAllowed`).

> I wouldn't be too concerned with private class methods, perhaps this is an indicator we could use a real class and public / private methods vs static class methods for this state manipulation?

I went ahead and used `class << self`, that way we can easily stick with using class methods while also using private methods like normal. 

Also renamed the `url` variable in AzureAdapter to be `url_prefix`.

app/services/media_storage/azure_adapter.rb
```diff
@@ -2,15 +2,15 @@
...
+    attr_reader :url_prefix, :public_container, :private_container, :storage_account_name, :get_expiration, :put_expiration, :client, :signer
 
     DEFAULT_EXPIRES_IN = 3 # time in minutes, see get_expiry_time(expires_in)
 
     def initialize(opts={})
       @storage_account_name = opts[:azure_storage_account]
       @public_container = opts[:azure_storage_container_public]
       @private_container = opts[:azure_storage_container_private]
-      @url = opts[:url]
+      @url_prefix = opts[:url_prefix]
```
camallen, line 13: This change will need some updates to the failing specs, https://travis-ci.org/github/zooniverse/panoptes/jobs/741473601
```
1) MediaStorage::AzureAdapter#get_path when medium is public returns the path as a https link
     # ./spec/services/media_storage/azure_adapter_spec.rb:125:in `block (4 levels) in 

  2) MediaStorage::AzureAdapter#get_path when medium is public uses the StoredPath.media_url method
     # ./spec/services/media_storage/azure_adapter_spec.rb:120:in `block (4 levels) in 
```

As well as the storage initializer https://github.com/zooniverse/panoptes/blob/8f9e1c06cad27829bb931f69bda36dddb51b3da7/config/initializers/storage.rb#L9

app/services/media_storage/stored_path.rb
```diff
@@ -1,28 +1,51 @@
...
-      # a valid domain prefix in the stored path
-      # e.g. /user_avatar/1e5fc9b5-86f1-4df3-986f-549f02f969a5.jpeg
-      # so we do not need to rewrite the URL
-      File.join(url, stored_path)
+      def media_url(domain_prefix, stored_path)
+        azure_path = rewrite_stored_path(stored_path)
+        File.join(domain_prefix, azure_path)
+      rescue PublicSuffix::DomainNotAllowed
+        azure_path = stored_path
+        File.join(domain_prefix, azure_path)
```
camallen, line 28: Not blocking at all, can this be written like this?
```suggestion
        begin
          azure_path = rewrite_stored_path(stored_path)
        rescue PublicSuffix::DomainNotAllowed
          azure_path = stored_path
        end
        File.join(domain_prefix, azure_path)
```

spec/services/media_storage/azure_adapter_spec.rb
```diff
@@ -8,7 +8,7 @@
   let(:private_container) { 'secret-magic-container' }
   let(:opts) do
     {
-      url: 'https://test-uploads.zooniverse.org/container_name',
+      domain_prefix: 'https://test-uploads.zooniverse.org/container_name',
```
camallen, line 11: i don't think `domain_prefix` matches `url_prefix`
```suggestion
      url_prefix: 'https://test-uploads.zooniverse.org/container_name',
```

spec/services/media_storage/stored_path_spec.rb
```diff
@@ -41,14 +36,46 @@
...
+      end
+
+      it 'returns the expected path' do
+        expect(result).to eq(expected_result)
+      end
+    end
+
+    context 'with with an azure native stored path (no old aws prefixes)' do
+      let(:stored_path) { 'subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg' }
+      let(:expected_result) { 'subject_location/f2eb4dbc-1353-4598-b1f6-3bf9e9a14169.jpeg' }
```
camallen, line 74: this might be useful if we update `stored_path` let definition
```suggestion
      let(:expected_result) { stored_path }
```

camallen (approved):
this is super close, it needs the failing specs fixed but nothing blocking from me. Once specs are fixed let's get this deployed for more testing.

PR #3497, opened 2020-10-29
ensure training_set_ids respect the linked set ids

camallen:
related to #3486 - @mwalmsley found that he could enable fallback selection for subject sets that were no longer actively linked to the workflow.

This PR fixes the found bug by ensuring the `workflow.training_set_ids` method returns the intersection of workflow config `training_set_ids` and the currently linked `subject_set_ids`. This ensures the configuration directive doesn't allow selection of non linked sets.

# Review checklist

### DIFF
--- a/app/models/workflow.rb
+++ b/app/models/workflow.rb
@@ -161,8 +161,10 @@ class Workflow < ActiveRecord::Base
   end
 
   def training_set_ids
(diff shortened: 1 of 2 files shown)


nciemniak (approved):
Looks great 👍

PR #3505, opened 2020-11-12
add production canary deploy template

nciemniak:
Like in the staging canary template, the k8s template for the production canary deployment includes definitions for the deployment, the service and the ingress.

I'm thinking that giving it the same resource allocation (memory/CPU) as the staging canary should suffice, since the canary deploy will only be touched by our team. Let me know if you disagree.

Once this PR is deployed, a CDN record and Front Door for the canary production URL needs to be created (akin to https://canary.pfe-preview.zooniverse.org/). URL will be https://canary.zooniverse.org.

-----

# Review checklist

### DIFF
--- a/kubernetes/deployment-production-azure-canary.tmpl
+++ b/kubernetes/deployment-production-azure-canary.tmpl
@@ -22,7 +22,7 @@ spec:
               cpu: "50m"
             limits:
               memory: "700Mi"
-              cpu: "1000m"
+              cpu: "500m"
           livenessProbe:
             httpGet:
               path: /
--- a/kubernetes/deployment-production-azure-canary.tmpl
+++ b/kubernetes/deployment-production-azure-canary.tmpl
@@ -1,20 +1,20 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: panoptes-production-azure-canary-app
+  name: panoptes-production-canary-app
   labels:
-    app: panoptes-production-azure-canary-app
+    app: panoptes-production-canary-app
 spec:
   selector:
     matchLabels:
-      app: panoptes-production-azure-canary-app
+      app: panoptes-production-canary-app
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


kubernetes/deployment-production-azure-canary.tmpl
```diff
@@ -0,0 +1,154 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: panoptes-production-azure-canary-app
```
camallen, line 4: Not blocking - I don't think we need the azure in the name here tbh
```suggestion
  name: panoptes-production-canary-app
```

kubernetes/deployment-production-azure-canary.tmpl
```diff
@@ -0,0 +1,154 @@
...
+      containers:
+        - name: panoptes-production-azure-canary-app
+          image: zooniverse/panoptes:__IMAGE_TAG__
+          resources:
+            requests:
+              memory: "700Mi"
+              cpu: "50m"
+            limits:
+              memory: "700Mi"
+              cpu: "1000m"
```
camallen, line 25: If this is just for internal team testing the i think we can get away with half a vcpu. If response times are impacted it's not user facing. 
```suggestion
              cpu: "500m"
```

camallen (approved):
LGTM

kubernetes/deployment-production-azure-canary.tmpl
nciemniak: Makes sense 👍

kubernetes/deployment-production-azure-canary.tmpl
nciemniak: Valid - the reasoning behind putting azure in the name was so that if we set up another canary for a different purpose in the future, there wouldn't be confusion. But most likely, at that point we will no longer be needing the azure canary for testing.

PR #3507, opened 2020-11-13
Only allow POST requests to omniauth

nciemniak:
Only allow POST requests to omniauth to guard against CSRF vulnerability in omniauth gem - as recommended in omniauth's [mitigation instructions](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284).

I'm thinking we should also take the next step of removing the gem and all associated code (if it's not being used), unless people think it's not worth our time to do so. Thoughts?

-----

# Review checklist

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,0 +1,1 @@
+OmniAuth.config.allowed_request_methods = [:post]


camallen (commented):
I'd rather we removed this dependency and dependent gems completely

camallen:
closing in favour of #3512 - feel free to reopen

PR #3511, opened 2020-11-20
Add subject_group data model

camallen:
finishes the work started in #3271 and linked to the work in https://github.com/zooniverse/front-end-monorepo/issues/1484

Adds the `SubjecGroup` data model to dynamically link existing subjects in newly created groups. These `subject_groups` will be dynamically created via subject selection

The SubjectGroup is a placeholder for the linked member subjects. Each `SubjectGroup`
+ `has_many members` the join resource for the subjects
+ `has_many subjects` through the join table above
+ `has_one group_subject` a new aggregate subject for use in talk and to collect the classification for the `subject_group`
    + See https://github.com/zooniverse/front-end-monorepo/issues/1487 for thoughts on keeping the existing talk interface working

Of note, the `SubjectGroup` resource will have a unique identifier key which is comprised of the group's subject_ids joined into a string. This key is dependent on the subject_id order and has a unique db index as well.  This will allow the group (and group_subject) to be reused if the selector returns the unique 'group' of subjects to classify. 

The SubjectGroupMember resource is a join resource between `Subject` and `SubjectGroup` but also encodes the display order of the subject in the group. 

I've used the service object pattern to add a `SubjetGroup::Create` operation to correctly construct the `SubjectGroup` and it's associated resources. This avoids the need for complex model callbacks to create the linked resources and keeps the model logic focussed on persistence. I plan to use this operation in the subjects (or `SubjectGroup`) controller when requesting the `SubjectGroup` resource.
 

# Review checklist


app/operations/subject_groups/create.rb
```diff
@@ -0,0 +1,111 @@
...
+    def project
+      @project ||= Project.find(project_id)
+    end
+
+    def uploader
+      @uploader ||= User.find(uploader_id)
+    end
+
+    def find_subjects_in_order(subject_ids)
+      @find_subjects_in_order ||=
```
nciemniak, line 59: Should we also be ensuring that all subjects belong to the correct project?

app/models/subject_group.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  has_many :members, class_name: 'SubjectGroupMember', dependent: :destroy
+  has_many :subjects, through: :members
+  # a 'group' subject to record the grouped view of linked subject locations
+  # for use in talk and to collect retirement information about the group
+  belongs_to :group_subject, class_name: 'Subject'
+
+  validates :project, presence: true
+  validates :members, presence: true
+  validates :key, presence: true
```
nciemniak, line 14: If it was me I'd probably add a uniqueness validation as well on `key` to prevent us hitting the error on the db.

nciemniak (approved):
Nice. This piece is looking great, LGTM 😎 just a couple thoughts/comments below, not blocking.

One question. Wanted to make sure I'm following here: did you add the `key` attribute to the SubjectGroup primarily for the purpose of making subject id lookup more efficient and ensuring uniqueness of subjects? I was just thinking about how we could retrieve that same information via an ordered query on SubjectGroupMembers, but that would put the job of making sure a group is unique on the developer (rather than letting the db and built in validation handle it).

app/operations/subject_groups/create.rb
camallen: interesting - i don't think we should restrict a subject group this way as groups may be comprised of subjects that are re-used across projects (a feature we have in the API). 

I'd rather leave this to the selector systems / input to the group to handle and leave this check out of the group.

app/models/subject_group.rb
camallen: I want to avoid extra db queries on the model save / validation tbh - my plan was to use a find by key to ensure we reuse existing subject groups outside the model / create operation. 

If we need this extra check on validation, due to race conditions when creating subject groups, i'll add the key uniq validation step here.

spec/factories/subject_groups.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :subject_group do
+    project { build(:project) }
+    group_subject { build(:subject) }
+    sequence(:key, &:to_s)
```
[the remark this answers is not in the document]
camallen: could do but i don't think it really matters for the factory as this is used to test the model validations. We test the correct ids are set via the create operation as its it's external to the model responsibility (persist and retrieve the model data).

Also, if we do this we will have to create the `subject_group_member` in the build step of this factory to ensure they have ids which increases db operations in our specs (slower specs), i'd rather leave this right now and revisit if needed.

camallen:
> One question. Wanted to make sure I'm following here: did you add the `key` attribute to the SubjectGroup primarily for the purpose of making subject id lookup more efficient and ensuring uniqueness of subjects? 

Yes, that and the ability to have the tracked subject ids as data on the model without loading an association from the db. Also this should be a smaller table / uniq index to search rather than the `subject_group_members` join table and allows me to write a simpler query `SubjectGroup.find(key: '1-2-3')` vs the join query to find the subject_group when i have the subject ids. 

> I was just thinking about how we could retrieve that same information via an ordered query on SubjectGroupMembers, but that would put the job of making sure a group is unique on the developer (rather than letting the db and built in validation handle it).

yes - and that query is across all groups so will be a bigger table / index to scan and a slightly more complex query. Plus we have to load the association to understand the state of linked subject ids. 

Finally - this can change if we want to ensure consistency with the linked associations. I can't see these groups changing once created so the `key` attribute should be static once set for the lifetime of the group. This can evolve / change though :)

PR #3512, opened 2020-11-20
Remove omniauth gem

nciemniak:
Remove omniauth gem and related code - no longer using omniauth.

For reference, here are a few of the commits where the omniauth functionality was initially added in:
- [Add support for facebook logins](https://github.com/zooniverse/panoptes/commit/a6e0dc8e1fb1a2691219c35d6b55dcda34d73977#diff-0f99d81511d17dcd32f5857adf7f1076108302cd0a91ee61d66727ff4f1a1284)
- [Add Google+ login](https://github.com/zooniverse/panoptes/commit/55a6439cca6c737b1d1f1ad23fa5fa8751605c14#diff-cb6e8126296dbc007e7625eee863de5c3213ed949b9999ea3418775f65826531)
- [Move authorizations to separate model](https://github.com/zooniverse/panoptes/commit/d4ea244072e2fdc588abef04fabdfff1e40541b6#diff-5b336c7be8c98e9e6f861d418751319f4a77d244af3d2cc22a004754beb01ebd)

Remaining TO DO:
- remove authorizations table and references to it

------

# Review checklist


camallen:
@nciemniak what's the status of this PR? can this be moved out of draft status?

yuenmichelle1:
Hey @Tooyosi , 
Apologies for the delay. I forgot that we discussed this PR review in our last Thursday call.
I think PR is fine. Personally, I would split the deletion of code and the migration into two separate PRs, but I think this way is fine as well. I think if you have a good sense of order of operations to push this change to production then I'm ok. 👍🏼 

I have a couple small questions/things:
- Probably best to remove gem from Gemfile-next.lock file as well. 
- I think we can also remove the authorizations.rb factory (spec/factories/authorizations.rb)
- Can we remove authorization.rb file within models altogether since we are removing the authorizations table?

Tooyosi:
> Hey @Tooyosi , Apologies for the delay. I forgot that we discussed this PR review in our last Thursday call. I think PR is fine. Personally, I would split the deletion of code and the migration into two separate PRs, but I think this way is fine as well. I think if you have a good sense of order of operations to push this change to production then I'm ok. 👍🏼
> 
> I have a couple small questions/things:
> 
> * Probably best to remove gem from Gemfile-next.lock file as well.
> * I think we can also remove the authorizations.rb factory (spec/factories/authorizations.rb)
> * Can we remove authorization.rb file within models altogether since we are removing the authorizations table?

Thanks @yuenmichelle1 , i agree and have implemented the suggestions

yuenmichelle1 (approved):
Hey @Tooyosi  changes look fine. 

Just to be safe,  since this change also has a migration, make sure you sync up with the rest of the team in terms of your game plan for the order of operations to push this change to production. 

cc @lcjohnso  @zwolf

PR #3519, opened 2020-11-30
cleanup left over tmp files every 6h

### LINKED ISSUE #3516: Classifications not in data export
We have a workflow (https://www.zooniverse.org/lab/12335/workflows/16579) that completed on 2020-11-22.
The workflow classification export appears to only contains data up until 2020-10-27. The stats show 3,532 / 3,532 subjects retired and the workflow was auto-removed from the project landing page. 

I don't understand why there is only classification data up until 2020-10-27. As an additional test, I classified and submitted a subject (marked as "FINISHED!") in the workflow myself on 2020-11-25 and generated a new export 24 hours later. My classification isn't recorded.

Can someone help me understand what might be going on?

camallen:
closes #3516 

Use a more aggressive exporter artifacts cleanup in /tmp. avoid these exports filling up our tmp disks and causing more errors

Linked to https://app.honeybadger.io/projects/40595/faults/70494224

# Review checklist

### DIFF
--- a/scripts/docker/start.sh
+++ b/scripts/docker/start.sh
@@ -3,7 +3,11 @@
 cd /rails_app
 
 # cleanup the dump worker temp files
-tmpreaper 7d /tmp/
+# that have not changed (had data added to them) in the last 6 hours
+# use the mtime (default is atime)
+# reason: handle the case where the exporter died mid export
+#         e.g the disk filled up OR OOM error on container process
+tmpreaper 6h --mtime /tmp/
 
 mkdir -p tmp/pids/
 rm -f tmp/pids/*.pid


nciemniak (approved):
Makes sense 👍 
So it looks like we write exports to the tmp file when we first generate them? Just want to confirm for my understanding here.

camallen:
> Makes sense 👍
> So it looks like we write exports to the tmp file when we first generate them? Just want to confirm for my understanding here.

Yes - the export code uses tmp files to write the csv data, then compress and upload. Once the export is done the tmp file cleans itself up. 
https://github.com/zooniverse/panoptes/blob/29896b1f62878aa310a166a8683cf16391738b0a/lib/csv_dump.rb#L5
https://github.com/zooniverse/panoptes/blob/29896b1f62878aa310a166a8683cf16391738b0a/app/models/csv_dumps/dump_processor.rb#L26-L30

PR #3529, opened 2020-12-14
Add rake task to rewrite supernova file locations

nciemniak:
Media resources id 1 - 457799 are all subjects for project Snapshot Supernova (project id 1). They currently exist outside of our media folder organizational structure. All files have already been copied over to `/subject_location`, this task must be run to update the media record `src` attribute (which points to where the file is located).

------


lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+      # regex to check whether current src is an old path
+      OLD_PATH_REGEX = /\Apanoptes-uploads.zooniverse.org\/1\/0\//.freeze
+      # this will return match data that will give us just the file name,
+      # which is the part of the string following the old path
+      CAPTURE_GROUP_REGEX = /\A(?:panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/.freeze
+
+      puts 'starting supernova src location rewrite'
+      Media.where('id < 457800').find_each do |medium|
+        if OLD_PATH_REGEX.match(medium.src)
+          file_name = medium.src.match(CAPTURE_GROUP_REGEX)[1]
```
camallen, line 280:
```suggestion
        if matches = CAPTURE_GROUP_REGEX.match(medium.src)
          file_name = matches[1]
```
```
test = 'panoptes-uploads.zooniverse.org/1/0/file-name.jpeg'
 => "panoptes-uploads.zooniverse.org/1/0/file-name.jpeg" 
2.5.3 :030 > regex = /\A(?:panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/
 => /\A(?:panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/ 
2.5.3 :031 > if regex_capture_groups = regex.match(test)
2.5.3 :032?>     regex_capture_groups[1]
2.5.3 :033?>   else
2.5.3 :034?>     puts 'why you no match'
2.5.3 :035?>   end
 => "file-name.jpeg" 
```

lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+          new_path = 'subject_location/' + file_name
```
camallen, line 281: I think this matches the new public container path format...right?  Resolve this as ok if it matches.

lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+          medium.src = new_path
+          medium.save!
```
camallen, line 283: This will be slow as it'll run validations, it might be better / quicker (and less db / cpu pressure) to  use `.update_column(src, new_path)` instead. `update_column` will run the sql to just update the column and skip model validations etc.

lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+    desc 'Update src to subject_locations for Snapshot Supernova media'
+    task rewrite_supernova_file_locations: :environment do
+      # regex to check whether current src is an old path
+      OLD_PATH_REGEX = /\Apanoptes-uploads.zooniverse.org\/1\/0\//.freeze
+      # this will return match data that will give us just the file name,
+      # which is the part of the string following the old path
+      CAPTURE_GROUP_REGEX = /\A(?:panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/.freeze
+
+      puts 'starting supernova src location rewrite'
+      Media.where('id < 457800').find_each do |medium|
```
camallen, line 278: super minor - OOI is 457800 a supernova project subject? If so the `where` clause should match it as well `'id <= ?'`
```suggestion
      last_known_supernova_project_subject_id = 457800
      Media.where('id < ?', last_known_supernova_project_subject_id').find_each do |medium|
```

nciemniak: Nope it is not. I'll make it more explicit with `id <= 457799`

lib/tasks/migrate.rake
nciemniak: Good point.

lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+    # They currently exist outside of our media folder organizational structure.
+    # All files have already been copied over to /subject_location, this task must be run
+    # to update the media record src (which points to where the file is located)
+    desc 'Update src to subject_locations for Snapshot Supernova media'
+    task rewrite_supernova_file_locations: :environment do
+      # regex to check whether current src is an old path
+      OLD_PATH_REGEX = /\Apanoptes-uploads.zooniverse.org\/1\/0\//.freeze
+      # this will return match data that will give us just the file name,
+      # which is the part of the string following the old path
+      CAPTURE_GROUP_REGEX = /\A(?:panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/.freeze
```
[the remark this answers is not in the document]
nciemniak: sure thing 👍

lib/tasks/migrate.rake
nciemniak: Original regex `/\A(?:panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/` would not have filtered out strings that did not start with `panoptes-uploads.zooniverse.org/1/0/` (due to that capture group starting with `?:`) but Ive updated it as needed so we are good.

lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+      OLD_PATH_REGEX = /\A(panoptes-uploads.zooniverse.org\/1\/0\/)?(.+)\z/.freeze
+
+      puts 'starting supernova src location rewrite'
+      Media.where('id <= 457799').find_each do |medium|
+        matches = OLD_PATH_REGEX.match(medium.src)
+        # array index 1 will be nil if src does not start with panoptes-uploads.../0/
+        if matches[1]
+          # array index 2 is the second capture group (anything after panoptes-uploads.../0/)
+          file_name = matches[2]
+          new_path = 'subject_location/' + file_name
```
camallen, line 282: TBH i think you can simplify this setup. Ditch the regex code and use `File.basename` to get the path filename from the `src` attribute directly. Up to you.
```suggestion
          new_path = "subject_location/#{File.basename(medium.src)}"
```

```
2.5.3 :021 > File.basename('panoptes-uploads.zooniverse.org/1/0/uuid.jpeg')
 => "uuid.jpeg" 
```

lib/tasks/migrate.rake
```diff
@@ -261,6 +261,32 @@ namespace :migrate do
...
+          medium.update_column(:src, new_path)
+        end
+      end
```
camallen, line 285: I'd put some logging code in here on each 1000 records to get some feedback on the rake task.
```suggestion
        puts "Processed 1000 records" if index % 1000 == 0 
      end
```

Adding the above will require an addition to the `find_each`
```
Media.where('id <= 457799').find_each.with_index do |medium, index|
```

camallen (approved):
A couple of minor comments nothing blocking.

PR #3532, opened 2020-12-17
use the medium get_path url for avatars

camallen:
ensure the special card routes have the updated avatar src urls (rewritten) via azure storage adapter.

Since we've modified how the panoptes-upload URLs work we need to avoid serializing the linked avatar URL path directly from the `model.src` attribute. Instead use custom Media storage adapter (Azure in this case) for rewriting the staging `src` attributes to `panoptes-uploads-staging` domain and stripping the `/production` paths on the `panoptes-uploads` domain.

# Review checklist


nciemniak (commented):
Great 👍

PR #3533, opened 2020-12-18
feature flag the subject group selection

camallen:
Addresses the comments in https://github.com/zooniverse/panoptes/pull/3530#pullrequestreview-554897297

- fixes the incorrect env var spelling 🤦 
- adds the production env vars to avoid 500 errors on this feature end point
- adds a flipper feature flag to disable / enable the feature dynamically without a deploy should we need it

I've added the feature flags in both envs APIs, production == disabled, staging == enabled

# Review checklist


nciemniak (approved):
LGTM 👍😎

PR #3537, opened 2020-12-22
Consolidate direct to object store medium uploader

camallen:
Longer term replacement for the aws upload class for email exports in #3535 

this PR introduces a duck type `DirectToObjectStorage` class for interface of the 'Medium' class. This interface allows the dump processor to call `.put_file` to put a file to an object store independent of how the default MediaStorage adapter is configured, i.e. to a different cloud service (Aws s3 vs azure blob) and do a different storage account / bucket location etc.

This `DirectToObjectStorage` class removes the need for a custom `DirectToS3` class type as it allows the adapter to be chosen and configured at runtime via ENV vars currently. Safe default for Email exports right now is to Aws service as that is how our email lists are synced currently. 

Before we can switch to azure adapter object store via this PR we will have to: 
1. configure the `zooniverse-exports` storage account and private container (while locking the acl down)
2. update the mail list syncing code to download the exported email list files from azure storage account not aws

# Review checklist


app/models/csv_dumps/direct_to_object_storage.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def put_file_opts
+      {
+        private: true,
+        compressed: true,
+        content_disposition: "attachment; filename=\"#{export_file_name}.csv\""
+      }
+    end
+
+    def object_storage_emails_export_path(path='email_exports')
```
nciemniak, line 63: I know that this was just modeled off of the `DirectToS3` code, but I'm wondering what the purpose is of constructing the path via `storage_adapter.stored_path`? The path for the email exports file is static, and lives in a separate bucket (and will live in a separate container) than other media uploads. Can't we construct the path without the use of the `stored_path` method? This just seems to be overcomplicating things in my mind, makes the code hard to read.

If we really do want to use the `stored_path` method, I would suggest potentially adding an optional `file_name` param, such that we could construct the `storage_path` in a single line like:
```ruby
storage_path = storage_adapter.stored_path('text/csv', path, export_file_name)
```

nciemniak (approved):
Looks good, would just consider rewriting the `object_storage_emails_export_path` method 😎

app/models/csv_dumps/direct_to_object_storage.rb
camallen: Right - so instead i could just inline the modified functionality of `stored_path` method where needed. 

Perhaps something like this or similar?
``` ruby
# TODO figure out how to get the correct file extension from the mime type
# or just hard code it if we're only concerned with 'text/csv'
extension = extension_from_mime_type('text/csv')
"#{path}/#{export_file_name}.#{extension}"
```
private method to get the mime type
https://github.com/zooniverse/panoptes/blob/4cbea3e186595be93e2642dd0aa1442c37a8067d/app/services/media_storage/abstract_adapter.rb#L48

PR #3538, opened 2020-12-22
rewrite the azure storage paths when cleaning up

camallen:
fixes the errors reported in https://app.honeybadger.io/projects/40595/faults/72161201

as our older s3 medium objects have different paths as what results in azure, we have to ensure we modify these paths to cleanup in azure correctly. 

Use the same path rewriter code as we do for the `adapter.get_paths` to determine which azure object store path we should attempt to delete.

# Review checklist

### DIFF
--- a/app/workers/medium_removal_worker.rb
+++ b/app/workers/medium_removal_worker.rb
@@ -7,7 +7,7 @@ class MediumRemovalWorker
 
   def perform(medium_src, opts={})
     MediaStorage.delete_file(object_store_path(medium_src), opts)
-  rescue Aws::S3::Errors::AccessDenied, Azure::Core::Http::HTTPError
+  rescue Aws::S3::Errors::AccessDenied
     # do nothing and don't retry this worker
   end
 
--- a/app/workers/medium_removal_worker.rb
+++ b/app/workers/medium_removal_worker.rb
@@ -1,10 +1,30 @@
+# frozen_string_literal: true
+
 class MediumRemovalWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: :data_low
 
   def perform(medium_src, opts={})
-    MediaStorage.delete_file(medium_src, opts)
-  rescue Aws::S3::Errors::AccessDenied
+    MediaStorage.delete_file(object_store_path(medium_src), opts)
+  rescue Aws::S3::Errors::AccessDenied, Azure::Core::Http::HTTPError
+    # do nothing and don't retry this worker
+  end
+
+  private
+
+  def object_store_path(path)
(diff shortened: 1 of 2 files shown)


app/workers/medium_removal_worker.rb
```diff
@@ -1,10 +1,30 @@
...
 class MediumRemovalWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: :data_low
 
   def perform(medium_src, opts={})
-    MediaStorage.delete_file(medium_src, opts)
-  rescue Aws::S3::Errors::AccessDenied
+    MediaStorage.delete_file(object_store_path(medium_src), opts)
+  rescue Aws::S3::Errors::AccessDenied, Azure::Core::Http::HTTPError
```
camallen, line 10: I'm a bit worried that swallowing this `Azure::Core::Http::HTTPError` right now will silently ignore us attempting to delete the incorrect azure paths and leave us storing (and paying) for file in the az blob store that we don't need.

perhaps changing it back to how it was and let these errors surface in hb for us to check on would be best while we figure out how this is working with the paths post s3 migration? Thoughts? 
```
rescue Aws::S3::Errors::AccessDenied
```

nciemniak: Yes, agreed - until we are aware of the cases where this type of error is encountered with Azure, I think its best _not_ to catch the error. I'd rather have a chance to monitor before turning on this error catching.

spec/workers/medium_removal_worker_spec.rb
```diff
@@ -1,15 +1,50 @@
...
+  end
+
+  it 'ignores any missing azure (BlobNotFound (404)) media paths' do
+    require 'azure/core/http/http_error'
+    response = instance_double('Azure::Core::Http::HTTPResponse', uri: 'fake-uri', status_code: 500, body: '', reason_phrase: '')
+    allow(MediaStorage).to receive(:delete_file).and_raise(Azure::Core::Http::HTTPError, response)
+    expect { worker.perform(medium_src) }.not_to raise_error
+  end
+
+  it 'does not modify the storage path for the object store' do
```
nciemniak, line 30: Would be helpful to specify that storage path is not modified specifically for the case where we are not using the azure adapter. Otherwise might be confusing for us when we're looking back on this to read this as default behavior.
```suggestion
  it 'does not modify the storage path for the object store when not using azure adapter' do
```

nciemniak (approved):
Agree with not rescuing Azure error until we can monitor it a bit and know that it won't be an issue - small code change will be needed for that, otherwise we're good!

Issue #3539, opened 2020-12-22
Data exports being written to old src path in Azure Blob Storage


lcjohnso:
**UPDATE: Edits in bold to bring up-to-date with current understanding of issue.**

TL;DR -- Data exports that are requested AND previously existed are **not being uploaded into the correct location**: the CSV media files **at the expected src path remain not updated** but the worker is not throwing any errors and the media entry in the DB is updated such that it appears everything is working correctly.

**Desired Behavior:** When a new data export is requested through the lab (e.g., a workflow classification export) and the export has previously been requested (i.e., there is an existing media item entry in DB + CSV in Azure Blob storage), the dump worker executes the following: update the CSV in blob store with the new dump, the media entry in the DB is updated with a new `updated_at` timestamp + metadata is updated with `"state": "ready"`, and mailer is set to requester to notify them that the dump is complete.

**Current Behavior:** The dump workers seems to exit without errors (no entries show in Honeybadger), and the media entry in the DB is correctly updated, but **the new CSV export file is being stored at the old AWS src path**.

Example: I requested a workflow classification export for workflow_id=7924, which is linked to the following media item: id=63620800, type='workflow_classifications_export', and src='panoptes-uploads.zooniverse.org/production/workflow_classifications_export/7859c7a6-1e6a-46ce-945d-13d3db461f4c.csv'.  The `updated_at` tag is now 2020-12-21, but the file in the blob store is last updated on 2020-11-30 (when the file was migrated from S3) and was not updated.  As a result, the file is missing classifications after 2020-11-11 (when it was last requested; the updated_at value before the most recent unsuccessful request).

If the export has not been previously requested, and the media resource does not exist, then the data export **_does_** seems to work OK.  Example: workflow_id=17025 has export with media id=96429345, type='workflow_classifications_export', and src='workflow_classifications_export/0aae2628-b4d2-45fd-b8f6-0a3666904584.csv' (**NOTE: new simple `src` format!**), and the file in the blob store was last updated on 2020-12-21, in agreement with the media updated_at timestamp.

camallen:
I looked into this today after seeing some unexpected object paths in the azure `panoptesuploads` storage account (`private` container).

[...]

PR #3540, opened 2021-01-02
Account for case where AWS path is passed to AzureAdapter#put_path

nciemniak:
Existing `AzureAdapter` code does not account for the case where an old AWS path is passed to `put_path` method. Because of this issue, in the case where the export was created in the past, when a user requests an export, the updated export gets saved to the wrong location in the private container. This means that the correct file location is not updated with the new export content. This PR fixes this issue.

Note on tests: only one test was added, but `put_path` tests were rearranged to account for necessary changes to `before` actions. Diff makes it seem like a lot more was changed.

-------

# Review checklist


camallen (commented):
I don't think this will fix the issue for existing medium resources as the existing paths will be reused and this code skipped. In my understanding, all new exports do work with the correct azure file paths which this `put_path` method handles. See below for my thinking as to why this is. 

> Existing AzureAdapter code does not account for the case where an old AWS path is passed to put_path method. Because of this issue, in the case where the export was created in the past, when a user requests an export, the updated export gets saved to the wrong location in the private container. 

On my reading the current export processor code will re-use the db `medium.src` paths to determine which path to upload to the cloud storage location. E.g. 

1. after processing any existing export, upload the resulting file to the `medium.src` existing path without generating a new one via this `put_path` method https://github.com/zooniverse/panoptes/blob/47f7eb0ebd32231da0e9af8dbf55bc19050e4d78/app/models/csv_dumps/dump_processor.rb#L53
2. `medium.put_file` reuses the existing db value to for the cloud object store path https://github.com/zooniverse/panoptes/blob/47f7eb0ebd32231da0e9af8dbf55bc19050e4d78/app/models/medium.rb#L88

This means that the this PR change won't be used for existing export medium resources, instead it'll reuse the aws paths in azure private containers :(

I may be wrong about the above so would appreciate your thoughts on the above. I did look into this before the break and I outlined some thoughts in #3539 on this issue.

nciemniak:
Ahh, darn. Thanks for pointing that out Cam. Yeah, I can see now that this won't fix it, I wrongly assumed that `put_path` was still called to get the upload path in the case of re-requesting exports.

PR #3543, opened 2021-01-04
export media src rewrite

### LINKED ISSUE #3539: Data exports being written to old src path in Azure Blob Storage

nciemniak:
Closes #3539 

Rewrite `src` attribute of export media records in order to resolve https://github.com/zooniverse/panoptes/issues/3539
Reasoning behind this choice explained [here](https://github.com/zooniverse/panoptes/issues/3539#issuecomment-754123093).

The new task will remove `panoptes-uploads.zooniverse.org/{env}/` from the src path. I have tested the rake task locally with manually generated media records, we will also be able to test on staging to confirm that the paths get updated as expected. Since we'll be able to test on staging, separating out the cases for each export type for testing is likely unnecessary, but feel free to comment if you think otherwise.

# Review checklist

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -286,6 +286,37 @@ namespace :migrate do
       end
       puts 'finished supernova src location rewrite'
     end
+
+    # Task to resolve issue where requesting a data export that has an existing export saved
+    # under the AWS path will save the newly generated export to the folder
+    # 'panoptes-uploads.zooniverse.org/{environment}/{media_type}/' rather than to
+    # the folder '{media_type}/' where it should be.
+    desc 'Update src to reflect Azure Blob Storage location for export resources'
+    task rewrite_export_src_attributes: :environment do
+      OLD_PATH_REGEX = /\A(panoptes-uploads.zooniverse.org\/#{Rails.env}\/)?(.+)\z/.freeze
+      export_media_types = %w[
+        project_aggregations_export
+        project_classifications_export
+        project_subjects_export
+        project_workflow_contents_export
+        project_workflows_export
+        workflow_classifications_export
+        Workflow_classifications_export
+      ]
+
+      puts 'starting export media src rewrite'
+      Medium.where(type: export_media_types).find_each.with_index do |medium, index|
+        matches = OLD_PATH_REGEX.match(medium.src)
+        # array index 1 will be nil if src does not start with panoptes-uploads.../production/
+        if matches[1]
+          puts matches[2]
+          # array index 2 is the second capture group (anything after panoptes-uploads.../production/)
+          medium.update_column(:src, matches[2])
+        end
+        puts "progress: #{index} records processed" if index % 1000.zero?
+      end
+      puts 'finished export media src rewrite'
+    end
   end
 
   namespace :subject_workflow_status do
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -301,7 +301,6 @@ namespace :migrate do
         project_workflow_contents_export
         project_workflows_export
         workflow_classifications_export
-        Workflow_classifications_export
       ]
 
       puts 'starting export media src rewrite'
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -305,15 +305,14 @@ namespace :migrate do
       ]
 
       puts 'starting export media src rewrite'
-      Medium.where(type: export_media_types).find_each.with_index do |medium, index|
+      Medium.where(private: true, type: export_media_types).find_each.with_index do |medium, index|
         matches = OLD_PATH_REGEX.match(medium.src)
         # array index 1 will be nil if src does not start with panoptes-uploads.../production/
         if matches[1]
-          puts matches[2]
           # array index 2 is the second capture group (anything after panoptes-uploads.../production/)
           medium.update_column(:src, matches[2])
         end
-        puts "progress: #{index} records processed" if index % 1000.zero?
+        puts "progress: #{index} records processed" if (index % 1000).zero?
       end
       puts 'finished export media src rewrite'
     end


lib/tasks/migrate.rake
```diff
@@ -286,6 +286,37 @@ namespace :migrate do
...
+        workflow_classifications_export
+        Workflow_classifications_export
+      ]
+
+      puts 'starting export media src rewrite'
+      Medium.where(type: export_media_types).find_each.with_index do |medium, index|
+        matches = OLD_PATH_REGEX.match(medium.src)
+        # array index 1 will be nil if src does not start with panoptes-uploads.../production/
+        if matches[1]
+          puts matches[2]
```
camallen, line 312: This log would be noisy, I think remove it.

lib/tasks/migrate.rake
```diff
@@ -286,6 +286,37 @@ namespace :migrate do
...
+    task rewrite_export_src_attributes: :environment do
+      OLD_PATH_REGEX = /\A(panoptes-uploads.zooniverse.org\/#{Rails.env}\/)?(.+)\z/.freeze
+      export_media_types = %w[
+        project_aggregations_export
+        project_classifications_export
+        project_subjects_export
+        project_workflow_contents_export
+        project_workflows_export
+        workflow_classifications_export
+        Workflow_classifications_export
```
camallen, line 304: Wow, is this misspelling really a thing on production & staging? 

If so fancy updating the `src` value to fix the incorrect spelling while we are here.

lib/tasks/migrate.rake
```diff
@@ -286,6 +286,37 @@ namespace :migrate do
...
       puts 'finished supernova src location rewrite'
     end
+
+    # Task to resolve issue where requesting a data export that has an existing export saved
+    # under the AWS path will save the newly generated export to the folder
+    # 'panoptes-uploads.zooniverse.org/{environment}/{media_type}/' rather than to
+    # the folder '{media_type}/' where it should be.
+    desc 'Update src to reflect Azure Blob Storage location for export resources'
+    task rewrite_export_src_attributes: :environment do
+      OLD_PATH_REGEX = /\A(panoptes-uploads.zooniverse.org\/#{Rails.env}\/)?(.+)\z/.freeze
```
camallen, line 296: Ooi, I assume the prod / staging dB doesn't have any incorrect env strings, e.g. production dB has `staging` string prefix? 

If it does we might need to modify the regex to match those strings.

lib/tasks/migrate.rake
```diff
@@ -286,6 +286,37 @@ namespace :migrate do
...
+        project_workflow_contents_export
+        project_workflows_export
+        workflow_classifications_export
+        Workflow_classifications_export
+      ]
+
+      puts 'starting export media src rewrite'
+      Medium.where(type: export_media_types).find_each.with_index do |medium, index|
+        matches = OLD_PATH_REGEX.match(medium.src)
+        # array index 1 will be nil if src does not start with panoptes-uploads.../production/
```
camallen, line 310: Not blocking you can use the match value itself, no match == nil, match returns Match class. 

Perhaps `next if match.nil?`

lib/tasks/migrate.rake
```diff
@@ -286,6 +286,37 @@ namespace :migrate do
...
+        project_classifications_export
+        project_subjects_export
+        project_workflow_contents_export
+        project_workflows_export
+        workflow_classifications_export
+        Workflow_classifications_export
+      ]
+
+      puts 'starting export media src rewrite'
+      Medium.where(type: export_media_types).find_each.with_index do |medium, index|
```
camallen, line 308: Can we use the extra filter here to only find private media resources? To make sure the rake task doesn't touch public resources (subjects etc)

Perhaps
`Medium.where(private: true, type: export_media_types)` ?

camallen (approved):
Minor comments, nothing blocking.

lib/tasks/migrate.rake
nciemniak: Sure, might as well add the additional check.

lib/tasks/migrate.rake
nciemniak: Yeah, it is 😓 there's only one file in that misspelled folder- I might as well move it over to the correct folder and update the `src` attribute via rails console.

lib/tasks/migrate.rake
nciemniak: Ah yes agree. This was for testing, meant to remove.

lib/tasks/migrate.rake
nciemniak: Ive performed a scan of the media records on staging and prod to make sure that we don't have any private media that does not match `panoptes-uploads.zooniverse.org/{environment)` and we are good. Thanks for raising this, was a good thing to check.

PR #3523, opened 2020-12-02
Add Panoptes (production) ingress for www-panoptes.zooniverse.org

shaunanoordin:
## PR Overview

This PR adds a new Kubernetes ingress for `www-panoptes.zooniverse.org`

- This PR is part of a solution for https://github.com/zooniverse/operations/issues/459#issuecomment-726289233
- The `www-panoptes.zooniverse.org` ingress is unique in that it adds a proxy-header to login requests from oAuth apps
- A new Certificate resource is added to the deployment template, _https://_ www-panoptes.zooniverse.org

### Dev Notes

- `rewrite-target` logic isn't used in the new ingress block
- The Certificate resource doesn't need `use-azuredns-solver: "true"` unlike in other current configurations
  - Rules for adding custom NGINX upstream vhost: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-nginx-upstream-vhost
  - Correct syntax for `proxy_set_header Host`: https://github.com/kubernetes/ingress-nginx/issues/5240

### Review checklist

- [ ] First, the most important one: is this PR small enough that you can actually review it? Feel free to just reject a branch if the changes are hard to review due to the length of the diff.
- [ ] If there are any migrations, will they the previous version of the app work correctly after they've been run (e.g. the don't remove columns still known about by ActiveRecord).
- [ ] If anything changed with regards to the public API, are those changes also documented in the `apiary.apib` file?
- [ ] Are all the changes covered by tests? Think about any possible edge cases that might be left untested.

### Status

Ready for review; monitoring the tests to ensure nothing breaks, at minimum.

PR #3546, opened 2021-01-08
Use explicit certificate resource on ingresses

camallen:
Let's encrypt cert expiry notifications led me to not find any cert resources for the panoptes ingresses. Luckily these are still valid and stored in the relevant secrets e.g. https://github.com/zooniverse/panoptes/blob/9f0d0e872be06409d67fdbcf8b6444584d95c9fc/kubernetes/deployment-production.tmpl#L517

Previously this worked due to the use of ingress annotations https://github.com/zooniverse/panoptes/blob/9f0d0e872be06409d67fdbcf8b6444584d95c9fc/kubernetes/deployment-production.tmpl#L507 
and the cert-manager shim would respect the annotations and configure a certificate in the defined secret, https://cert-manager.io/docs/usage/ingress/#how-it-works

However post the cert-manager upgrade these annotations are out of date so the cert manager doesn't recognise these anymore, see https://github.com/zooniverse/operations/pull/502 for details on the cert-manager annotation changes.

So this PR updates the ingress annotations (even though it doesn't use them) and will now explicitly configure certificates for each ingress. Explicit certs allow us to easily find them from the k8s yaml declarations and the `kubectl get certificate` cmd. 

Finally this PR combines the custom `www panoptes` ingress certificate to the main ingress as created in #3523. Post deployment i'll manually cleanup the `www-panoptes-production-tls` certificate and secret resources in k8s.

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -504,7 +504,7 @@ metadata:
   name: panoptes-production-ingress
   annotations:
     kubernetes.io/ingress.class: nginx
-    certmanager.k8s.io/cluster-issuer: letsencrypt-prod
+    cert-manager.io/cluster-issuer: letsencrypt-prod
     nginx.ingress.kubernetes.io/rewrite-target: /$1
     nginx.ingress.kubernetes.io/proxy-buffer-size: "128k"
     nginx.ingress.kubernetes.io/set-real-ip-from: "10.0.0.0/8"
@@ -544,7 +544,7 @@ metadata:
   name: www-panoptes-production-ingress
   annotations:
     kubernetes.io/ingress.class: nginx
-    certmanager.k8s.io/cluster-issuer: letsencrypt-prod
+    cert-manager.io/cluster-issuer: letsencrypt-prod
     nginx.ingress.kubernetes.io/proxy-buffer-size: "128k"
     nginx.ingress.kubernetes.io/set-real-ip-from: "10.0.0.0/8"
     nginx.ingress.kubernetes.io/upstream-vhost: "www.zooniverse.org"
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -552,7 +552,7 @@ spec:
   tls:
   - hosts:
     - www-panoptes.zooniverse.org
-    secretName: www-panoptes-production-tls
+    secretName: panoptes-production-tls-secret
   rules:
   - host: www-panoptes.zooniverse.org
     http:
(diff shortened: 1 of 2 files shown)


zwolf (commented):
I thought that the certs for `*-staging.zooniverse.org` domains were being handled by [this one over in static](https://github.com/zooniverse/static/blob/master/kubernetes/ingress/zooniverse.org.yaml). Is this just to keep it in its own repo with its own secret? If so,  is this PR updating staging to ensure it works as expected, with prod to come later?

camallen:
> I thought that the certs for `*-staging.zooniverse.org` domains were being handled by [this one over in static](https://github.com/zooniverse/static/blob/master/kubernetes/ingress/zooniverse.org.yaml). Is this just to keep it in its own repo with its own secret? If so, is this PR updating staging to ensure it works as expected, with prod to come later?

Not for panoptes at least, note the `panoptes-staging` not `panoptes.staging` https://github.com/zooniverse/static/blob/d986d9e4b6c5bdf63cee3c395f2707c56101cd29/kubernetes/ingress/zooniverse.org.yaml#L15

Here is a view of the staging API cert looks like via the k8s cert manager / ingress. 
``` bash
$ curl -Iv https://panoptes-staging.zooniverse.org
*   Trying 52.186.94.16...
* TCP_NODELAY set
* Connected to panoptes-staging.zooniverse.org (52.186.94.16) port 443 (#0)
......
* Server certificate:
*  subject: CN=panoptes-staging.zooniverse.org
*  start date: Oct 20 12:29:52 2020 GMT
*  expire date: Jan 18 12:29:52 2021 GMT
*  subjectAltName: host "panoptes-staging.zooniverse.org" matched cert's "panoptes-staging.zooniverse.org"
*  issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
*  SSL certificate verify ok.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7fe67b80d600)
> HEAD / HTTP/2
> Host: panoptes-staging.zooniverse.org
> User-Agent: curl/7.64.1
> Accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
< HTTP/2 200 
HTTP/2 200 
< server: nginx/1.17.10
```
The same ingress / cert manager mechanism applies for production.

zwolf:
Ope, def misread `.staging` as `-staging` and I see the existing production cert now, too. Withdrawn!

zwolf (approved):

PR #3552, opened 2021-01-12
Add http_cache parameter to SubjectGroup selection API

### LINKED ISSUE #3551: SubjectGroup selection bug...? - http_cache parameter not recognised
## API bug?

I'm not sure if this is intended or not, but the SubjectGroup selection endpoint rejects the http_cache parameter

e.g. url: `https://panoptes-staging.zooniverse.org/api/subjects/grouped?workflow_id=3412&num_rows=5&num_columns=5&http_cache=true`

result: 422 error, "found unpermitted parameter: http_cache"

The front end's Panoptes JS library sends http_cache=true by default (in fact, not sure if I can switch it off without changing the fundamental library), and http_cache=true is extant on other API calls like to `/api/subjects/queued`

### Status

Unconfirmed if this is a bug or working as intended. If working as intended, front end will need some notable changes to the way API calls are made.

shaunanoordin:
## PR Overview

Closes #3551

- This PR adds "http_cache" to the list of acceptable parameters for the `/subjects/grouped` endpoint.
- The "http_cache=true" parameter is sent by the front end (Panoptes JS library) for pretty much most API calls, so this seems necessary.

### Status

Ready for review, contingent on the assumption that the http_cache wasn't excluded from the list of acceptable parameters on purpose.

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -58,7 +58,7 @@ class Api::V1::SubjectsController < Api::ApiController
     skip_policy_scope
 
     # setup the selector params from user input, note validation occurs in the operation class
-    selector_param_keys = %i[workflow_id subject_set_id num_rows num_columns]
+    selector_param_keys = %i[workflow_id subject_set_id num_rows num_columns http_cache]
     selector_params = params.permit(*selector_param_keys)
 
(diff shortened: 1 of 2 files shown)


camallen (approved):
Marvellous 👍 thanks @shaunanoordin

PR #3568, opened 2021-01-29
Isolate production data export worker pods

camallen:
This has been tested in staging via #3570 and is working ok

Our data exports processes are being interrupted by frequent scaling events on the sidekiq processes. Instead of allowing this interruption due to system demand for normal background job processing rates, we can run 1 deployment pod especially for running data exports via sidekiq. 

This PR will allow data export workers to enqueue jobs to a special queue (and others to maximise our CPU time) set by and ENV var at boot time. 

We will add a special processing sidekiq deployment to process jobs from these special queues. The existing sidekiq deployments will be untouched and won't process the special queues, just the normal ones.

To achieve this we can use cli args to override the sidekiq process config for this special deployment pod and the normal sidekiq deployment pods will use the default values in the config/sidekiq.yml configuration file. 

# Review checklist


kubernetes/deployment-production.tmpl
```diff
@@ -268,6 +269,65 @@ spec:
 ---
 apiVersion: apps/v1
 kind: Deployment
+metadata:
+  name: panoptes-production-sidekiq-dumpworker
+  labels:
+    app: panoptes-production-sidekiq-dumpworker
+spec:
```
camallen, line 276: Add the replicas in to ensure we have two of these running
```suggestion
spec:
  replicas: 2
```

PR #3593, opened 2021-04-16
fix broken exports for annotations derived from FEM dropdown-simple tasks

camallen:
Handle the Fem v2 `dropdown-simple` task annotation format which is derived from PFE lab workflows configured with only 1 dropdown (select). 

This PR adds a new set of v2 classification annotation formatting interface, specifically the V2 code only knows how to format V2 `dropdown-simple` task annotations whilst defaults to the default (v1) annotation formatter for all other tasks (preserve the current behaviour). 

I've refactored the internals of the default annotation formatter to create collaborating objects vs using inheritance of the formatters and simplified the V2 interface using collaborators. I think longer term this will provide more flexibility for extending the exporter to handle the new task annotation schema formats. 

While this PR touches a bit of code, most is for the new v2 annotations formatters and the rest is a refactor to use collaborator objects between v1 and v2 formatters. Also the v1 annotation formatters test suite is pretty comprehensive which helps trust these changes don't break any v1 behaviors. I've added specs for the new V2 formatters.

This is needed for an EC crowds project meeting so i'd like to get this in the wild ASAP. I've used this code locally to test the behaviour and produce the csv export file for the EC project. I'm liaising with Coleman to fine tune the export formats and make sure it meets the aggregation service requirements.

## TODO

- [x] refactor the internal working with delegation / decorators to handle the modified behaviour of v2 annotations vs v1 (most of the behaviour is still v1 based)
- [x] ensure we call the correct formatter class v2 or original based on the classification metadata key  `classifier_version: 2.0`

# Review checklist


CKrawczyk:
Using classification id `321354937` as an example, the classification data when using the python client gives:
```python
{'annotations': [{'task': 'T1', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T2', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T3',  'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T4', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T5',  'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T7', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 4}},
                 {'task': 'T8', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T9', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T10', 'taskType': 'dropdown-simple',  'value': {'option': True, 'selection': 12}},
                 {'task': 'T11', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T13', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T14', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 14}},
                 {'task': 'T15', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T16', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T17', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T19', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T20', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T21', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 14}},
                 {'task': 'T22', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T23', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T25', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 17}},
                 {'task': 'T26', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 14}},
                 {'task': 'T27', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}},
                 {'task': 'T28',  'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 14}},
                 {'task': 'T29', 'taskType': 'dropdown-simple', 'value': {'option': True, 'selection': 18}}]
 }
```

and the current csv dump from the PR gives:

```python
{"annotations": [
    {"task":"T1","value":{"select_label":"Select Circumstance #1"}},
    {"task":"T2","value":{"select_label":"Select Circumstance #2"}},
    {"task":"T3","value":{"select_label":"Select Circumstance #3"}},
    {"task":"T4","value":{"select_label":"Select Circumstance #4"}},
    {"task":"T5","value":{"select_label":"Select Circumstance #5"}},
    {"task":"T7","value":{"select_label":"Select Circumstance #6"}},
    {"task":"T8","value":{"select_label":"Select Circumstance #7"}},
    {"task":"T9","value":{"select_label":"Select Circumstance #8"}},
    {"task":"T10","value":{"select_label":"Select Circumstance #9"}},
    {"task":"T11","value":{"select_label":"Select Circumstance #10"}},
    {"task":"T13","value":{"select_label":"Select Circumstance #11"}},
    {"task":"T14","value":{"select_label":"Select Circumstance #12"}},
    {"task":"T15","value":{"select_label":"Select Circumstance #13"}},
    {"task":"T16","value":{"select_label":"Select Circumstance #14"}},
    {"task":"T17","value":{"select_label":"Select Circumstance #15"}},
    {"task":"T19","value":{"select_label":"Select Circumstance #16"}},
    {"task":"T20","value":{"select_label":"Select Circumstance #17"}},
    {"task":"T21","value":{"select_label":"Select Circumstance #18"}},
    {"task":"T22","value":{"select_label":"Select Circumstance #19"}},
    {"task":"T23","value":{"select_label":"Select Circumstance #20"}},
    {"task":"T25","value":{"select_label":"Select Circumstance #21"}},
    {"task":"T26","value":{"select_label":"Select Circumstance #22"}},
    {"task":"T27","value":{"select_label":"Select Circumstance #23"}},
    {"task":"T28","value":{"select_label":"Select Circumstance #24"}},
    {"task":"T29","value":{"select_label":"Select Circumstance #25"}}]}
```
This `select_label` value is the name of the dropdown task rather than the value that was selected.  Also, aggregation will use the `taskType` value as a way to determine the shape of the classification for all classifier 2.0 classifications.

camallen:
So the FEM now ignores the selected dropdown HTML option `value` (also recorded in the workflow) and instead records the UI HTML select option index to record what the volunteer selects. 

I'll rework the code to handle this now (mirror single / simple task types) but it does introduce and ordering dependency between the FEM UI code HTML elements and the API representation (used in downstream services). I've written up some thoughts in slack, reposted here.

> I was looking into fixing the FEM dropdown task classification exports and noticed the annotation values have changed from the option value in the workflow to the index of the selected value from the UI element.
> 
> This change from the PFE format was a little unexpected but after a chatting with Jim & Shaun they summarised this change as a simpler format that mirrors how the single answer (simple) task annotation format works.
> 
> My only concern with this change is we now have an ordering dependency between the API workflow task option ordering and the UI HTML select option ordering. E.g. the FEM code loads a WF from the API and has wf.task.options = `[{option1, option2, option3}]` but some code between the workflow store (i think) and the UI HTML element wiring happens to reorder that list to say `[{option3, option2, option1}]`. Then when i reconstruct the annotation details API side for say annotation index value 0, the results will be option1 but the UI really showed option3 to the user.
> 
> Historically in PFE land we used the explicit task option value codes (from the API workflow) in the classification annotation values, this ensured we could reconstruct the annotation regardless of how the UI displayed the options.
> 
> My concerns were alleviated as currently the HTML task select ordering is preserved through the map function via https://github.com/zooniverse/front-end-monorepo/blob/af5483c6d4503c46054b6e7c29c5cb80f8b72a0b/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/legacyDropdownAdapter/legacyDropdownAdapter.js#L11-L12
> 
> However I can see a use case request for an ordered dropdown (alphabetically via the translation labels) coming from a project team. Perhaps the (new?) FEM developer doesn’t understand this implicit ordering dependency and thus sorts the option order before rendering in the UI element and thus breaks the ability for downstream services (panoptes exports, caesar, aggregation code etc) to correctly reconstruct the volunteer answers.
> 
> Noting, when I read the ADRs for the dropdown task it didn’t mention that ordering of the task options needs to be preserved to ensure the correct recording of answers.
> 
> I’m not sure what tools in FEM land are using this indexing strategy for annotation values but it’s worth while ensuring folks are aware of the hard dependency with the ordering of the original API representation of the tasks and the corresponding annotation values.

lib/formatter/csv/annotation_for_csv.rb
```diff
@@ -133,66 +140,40 @@ def dropdown_find_selected_option(selects, answer_value)
...
+        workflow_information.at_version
       end
 
-      def get_task_info(subtask)
-        workflow_at_version.tasks.find do |key, task|
-         key == subtask["task"]
-        end.try(:last) || {}
-      rescue
+      # used in combo task to protect against invalid annotations
+      # i.e. bad annotation data from tropical sweden project
```
zwolf, line 170: This comment doesn't add a lot of context--what's a tropical sweden? Is there a better way to phrase this?

lib/formatter/csv/v2/annotation.rb
```diff
@@ -0,0 +1,40 @@
...
+# Overtime this class will evolve behaviour to handle the distinct classification annotation formats
+# if it doesn't these two behaviours can be refactored for collaborating classes
+# to provide csv converstion of the differing annotation formats
+module Formatter
+  module Csv
+    module V2
+      class Annotation
+        attr_reader :classification, :current_annotation, :cache, :default_formatter, :workflow_information, :task_info
+
+        def initialize(classification, annotation, cache, default_formatter=nil)
```
zwolf, line 16: Is allowing `default_formatter` as an arg here for future use? Or for some other purpose? I get that having it in the v2 annotator lets you use the rest of the unchanged task formatters, but it's not being passed explicitly to the v2 annotator anywhere that I can find.

spec/lib/formatter/csv/v2/annotation_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    it 'uses the default annotation formatter' do
+      formatter.to_h
+      expect(default_formatter).to have_received(:to_h)
+    end
+  end
+
+  context 'with a v2 dropdown (simple) task' do
+    # details on what allowCreate wf task values for dropdowns do
```
zwolf, line 36: Does a user-provided selection, `'option' => false` case deserve its own spec?

zwolf (approved):
This is rad, :shipit: . I especially like the WorkflowInformation abstraction. Couple questions, nothing blocking.

lib/formatter/csv/annotation_for_csv.rb
camallen: this has a relevant spec https://github.com/zooniverse/panoptes/blob/5928d94cffb424e68aec480c619271d0783bb0dc/spec/lib/formatter/csv/annotation_for_csv_spec.rb#L196-L210
```suggestion
      # i.e. bad annotation data from tropical sweden project
      # see https://github.com/zooniverse/panoptes/blob/5928d94cffb424e68aec480c619271d0783bb0dc/spec/lib/formatter/csv/annotation_for_csv_spec.rb#L196-L210
```

lib/formatter/csv/v2/annotation.rb
camallen: dependency injection to allow us to easily change the fallback formatter without changing this class interface. It helps in testing as well (i.e. we can inject the double at the build time) - though i didn't do that as i let it build it's default. 

I'm going to leave this in for the extension point for class behaviour

spec/lib/formatter/csv/v2/annotation_spec.rb
camallen: In theory this is impossible in the new classifier as it stands, https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/README.md#dev-notes

So the annotation result would always be `'option' => true`

that said - 👍🏼 I'll add a spec for this anyway as the code exists

PR #3596, opened 2021-04-29
Allow subjects to be unretired

yuenmichelle1:
I believe this is in a good spot. Let me know if I missed cases.  Mainly followed `retire_subjects` path as a guide. 

**Changes**
- creating a new route to unretire subjects by workflow_id.. route is `/api/workflows/YOUR_WF_ID/unretire_subjects` 
-  updated corresponding controller for route action 
- creating operator that calls an UnretirementSubjectWorker
- created UnretirementSubjectWorker that sets the `retired_at` and `retirement_reason` to nil 

**TODO**

- [x] Still getting my feet wet with rspec, but I want to add more tests

# Review checklist

### DIFF
--- a/app/operations/workflows/unretire_subjects.rb
+++ b/app/operations/workflows/unretire_subjects.rb
@@ -1,5 +1,24 @@
 module Workflows
     class UnretireSubjects < Operation
-        
+        validates :workflow_id, presence: true
+
+        integer :workflow_id
+        integer :subject_id, default: nil
+        array :subject_ids, default: [] do
+            integer
+        end
+
+        def execute
+            return if subject_ids.empty?
+            puts "MDY114 WORKFLOW ID #{workflow_id}"
+            SubjectWorkflowStatus.where.not(retired_at: nil).where(workflow_id: workflow_id, subject_id: subject_ids).update_all(retired_at: nil, retirement_reason: nil)
+            RefreshWorkflowStatusWorker.perform_async(workflow_id)
+            NotifySubjectSelectorOfChangeWorker.perform_async(workflow_id)
+        end
+
+        def subject_ids 
+            @cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+        end
+
     end
 end
--- a/app/operations/workflows/unretire_subjects.rb
+++ b/app/operations/workflows/unretire_subjects.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Workflows
   class UnretireSubjects < Operation
     validates :workflow_id, presence: true
@@ -10,11 +12,12 @@ module Workflows
 
     def execute
       return if subject_ids.empty?
+
       UnretireSubjectWorker.perform_async(workflow_id, subject_ids)
     end
 
-    def subject_ids 
-      @cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+    def subject_ids
+      @subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
     end
   end
 end
--- a/app/operations/workflows/unretire_subjects.rb
+++ b/app/operations/workflows/unretire_subjects.rb
@@ -11,9 +11,7 @@ module Workflows
         def execute
             return if subject_ids.empty?
 
-            SubjectWorkflowStatus.where.not(retired_at: nil).where(workflow_id: workflow_id, subject_id: subject_ids).update_all(retired_at: nil, retirement_reason: nil)
-            RefreshWorkflowStatusWorker.perform_async(workflow_id)
-            NotifySubjectSelectorOfChangeWorker.perform_async(workflow_id)
+            UnretireSubjectWorker.perform_async(workflow_id, subject_ids)
         end
 
         def subject_ids 
--- a/app/operations/workflows/unretire_subjects.rb
+++ b/app/operations/workflows/unretire_subjects.rb
@@ -11,13 +11,13 @@ module Workflows
     end
 
     def execute
-      return if subject_ids.empty?
+      return if cached_subject_ids.empty?
 
-      UnretireSubjectWorker.perform_async(workflow_id, subject_ids)
+      UnretireSubjectWorker.perform_async(workflow_id, cached_subject_ids)
     end
 
-    def subject_ids
-      @subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+    def cached_subject_ids
+      @cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
     end
   end
 end
--- a/app/operations/workflows/unretire_subjects.rb
+++ b/app/operations/workflows/unretire_subjects.rb
@@ -1,0 +1,5 @@
+module Workflows
+    class UnretireSubjects < Operation
+        
+    end
+end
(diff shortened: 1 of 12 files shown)


spec/workers/unretire_subject_worker_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
```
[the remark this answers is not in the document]
camallen: these can be fixed by using `# frozen_string_literal: true` at the top of the file. For details https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/

camallen:
Ah - the hound strikes again 🦮. I don't think we mentioned we have a linter check in the CI system here. For the draft phase these are not really important but ideally we address these before merging to have the codebase conform to a normal style. Some can be ignored at times (like method complexity / length) but most defaults are sensible.

FWIW in ruby we use https://rubocop.org/ for linting, you can run it from the cmd line (see `#usage` on that link) or integrate it into your editor (I use VSCode so have this running for fast feeback when i code, https://github.com/misogi/vscode-ruby-rubocop). 

Either way - don't be too put off by the linter feedback.

camallen (commented):
this is looking really good @yuenmichelle1 - the data flow from routes -> controller -> operation -> worker looks good and the implementation seems there 👍🏼. 

It could probably use one or two edge case specs** but it sounds  like this is on your radar anyway.  Nice work 🎉 

** e.g. what happens when i pass in a list params that don't make sense for this operation

app/operations/workflows/unretire_subjects.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Workflows
+  class UnretireSubjects < Operation
+    validates :workflow_id, presence: true
```
camallen, line 5: not blocking, considering this won't be a widely used end point and if used it's most likely used by our cli / client tools.

We can ensure this is a integer by using active model validations,
- https://github.com/AaronLasseigne/active_interaction#validations
- https://apidock.com/rails/v4.2.9/ActiveModel/Validations/HelperMethods/validates_numericality_of

if you add this it might be worth adding a spec for this on the operation
Something like this might work..
```suggestion
    validates :workflow_id, presence: true, numericality: { greater_than: 0, only_integer: true }
```

if you adopt this you may want to do something like this for the subject_ids as well.

app/policies/project_policy.rb
```diff
@@ -21,7 +21,7 @@ class TranslateScope < Scope
         :create_subjects_export,
         :create_workflows_export,
         :create_workflow_contents_export,
-        :retire_subjects, with: WriteScope
+        :retire_subjects, :unretire_subjects, with: WriteScope
```
camallen, line 24: this should probably have associated spec entries for the different authorization levels (types of user), e.g. for collaborator https://github.com/zooniverse/panoptes/blob/61d184848bfe367f4536ddbe3054988c3e734741/spec/policies/project_policy_spec.rb#L177

app/workers/unretire_subject_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class UnretireSubjectWorker
+  include Sidekiq::Worker
+  attr_reader :workflow_id
+
+  sidekiq_options queue: :high
+
+  def perform(workflow_id, subject_ids)
+    @workflow_id = workflow_id
+    return unless workflow_exists?
```
camallen, line 11: not blocking and feel free to ignore. I'd probably avoid the extracting this to a method (and setting up the attribute reader etc) and i'd inline this method, e.g.
```suggestion
    return unless Workflow.exists?(id: workflow_id)
```

spec/controllers/api/v1/workflows_controller_spec.rb
```diff
@@ -734,6 +734,28 @@
     end
   end
 
+  describe '#unretire_subjects' do
+    before do
+      default_request scopes: scopes, user_id: authorized_user.id
+    end
+
+    let(:subject_set_project) { project }
+    let(:subject_set) { create(:subject_set, project: project, workflows: [project.workflows.first]) }
```
camallen, line 743: not blocking - fyi you can setup a subject_set that will create some subjects automatically for you, https://github.com/zooniverse/panoptes/blob/61d184848bfe367f4536ddbe3054988c3e734741/spec/factories/subject_sets.rb#L19
e.g. 
```
let(:subject_set) do 
  create(:subject_set_with_subjects, num_subjects: 1, project: project, workflows: [project.workflows.first])
end

File: spec/operations/workflows/unretire_subjects_spec.rb [hunk 6]
```diff
@@ -0,0 +1,44 @@
...
+
+require 'spec_helper'
+
+describe Workflows::UnretireSubjects do
+  let(:api_user) { ApiUser.new(build_stubbed(:user)) }
+  let(:workflow) { create(:workflow) }
+  let(:subject_set) { create(:subject_set, project: workflow.project, workflows: [workflow]) }
+  let(:subject_set_id) { subject_set.id }
+  let(:subject1) { create(:subject, subject_sets: [subject_set]) }
+  let(:subject2) { create(:subject, subject_sets: [subject_set]) }
```
camallen, line 11: not blocking

same as above with the subject setup - you can setup multiple subjects with one call
```
let(:subject_set) do 
  create(:subject_set_with_subjects, num_subjects: 2, project: workflow.project, workflows: [workflow])
end
let(:subject_1) { subject_set.subjects.first }
let(:subject_2) { subject_set.subjects.last }
```

spec/operations/workflows/unretire_subjects_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  let(:subject_workflow_count) { create }
```
camallen, line 12: i think this can be deleted - I don't think this creates a SubjectWorkflowStatus (note the model name differs from the table name for historical reasons) and isn't used..

spec/workers/unretire_subject_worker_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:set) { sms.subject_set }
+  let(:status) { create(:subject_workflow_status, subject: subject1, workflow: workflow, retired_at: 1.day.ago, retirement_reason: 'other') }
+
+  describe '#perform' do
+    before do
+      allow(RefreshWorkflowStatusWorker).to receive(:perform_async)
+      allow(NotifySubjectSelectorOfChangeWorker).to receive(:perform_async)
+    end
+
+    context 'when subjects are unretireable' do
```
camallen, line 19:
```suggestion
    context 'when subjects are already retired' do
```

camallen (approved):
This is great work @yuenmichelle1 👏🏼 🎩 🎉 

I've left some comments that you can choose to address. My only concerns are minor
- I found was one let block in a spec wasn't used and can be removed
- the project policy spec could do with some extra additions for the `unretired_subjects` action similar to the `retire_subjects` lines.

zwolf (approved):
LGTM too. One thing I'll mention is that Cliff said this back in the original request:

> One specific part of this feature request I want to highlight: in addition to passing a `subject_ids` payload, I'd also like to support passing `subject_set_ids` too. As seen in Cam's script code linked above, most of the time we are looking to unretire all or some subject sets linked to a workflow. A `subject_set`-based option would be **very** helpful!

This PR is complete as-is and can be merged, but for the sake of the usefulness of the feature long term, adding unretirement by subject set ID should be a follow up PR.

Thanks and congrats! :tada: :shipit:

PR #3600, opened 2021-05-06
Unretire subject by subject ids

yuenmichelle1:
**Changes**
- update unretire subjects by subject set id (updating corresponding operation to grab subject ids from subject set id and then send it over to the worker.)

# Review checklist

### DIFF
--- a/app/operations/workflows/unretire_subjects.rb
+++ b/app/operations/workflows/unretire_subjects.rb
@@ -8,16 +8,32 @@ module Workflows
     integer :subject_id, default: nil
     array :subject_ids, default: [] do
       integer
-    end
+    end;
+    integer :subject_set_id, default: nil
+    array :subject_set_ids, default: [] do 
+      integer
+    end;
 
     def execute
       return if cached_subject_ids.empty?
-
+      puts "MDY114 CACHED SUBJECT IDS BEFORE #{cached_subject_ids}"
       UnretireSubjectWorker.perform_async(workflow_id, cached_subject_ids)
     end
 
     def cached_subject_ids
       @cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+      if !subject_set_subject_ids.empty?
+        puts 'MDY114 HITS CHECK'
+        @cached_subject_ids.push(*subject_set_subject_ids)
+      end
+      return @cached_subject_ids
+    end
+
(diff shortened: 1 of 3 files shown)


app/operations/workflows/unretire_subjects.rb
```diff
@@ -17,7 +21,14 @@ def execute
...
 
     def cached_subject_ids
-      @cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+      cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+      cached_subject_ids.push(*subject_set_subject_ids) unless subject_set_subject_ids.empty?
+      @cached_subject_ids = cached_subject_ids
+    end
+
+    def subject_set_subject_ids
+      set_ids ||= Array.wrap(@subject_set_ids) | Array.wrap(@subject_set_id)
```
camallen, line 30: no need to memoize here
```suggestion
      set_ids = Array.wrap(@subject_set_ids) | Array.wrap(@subject_set_id)
```

app/operations/workflows/unretire_subjects.rb
```diff
@@ -17,7 +21,14 @@ def execute
     end
 
     def cached_subject_ids
-      @cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+      cached_subject_ids ||= Array.wrap(@subject_ids) | Array.wrap(@subject_id)
+      cached_subject_ids.push(*subject_set_subject_ids) unless subject_set_subject_ids.empty?
+      @cached_subject_ids = cached_subject_ids
```
camallen, line 26: I think we can simplify this a little bit - keep the `cached_subject_ids` method but add a layer underneath it to handle the subject_id lookups, e.g. something like
```
  def cached_subject_ids
    # note the union | array operator below will allow us to accept both sets of params :)
    @cached_subject_ids ||= subject_ids | subject_set_subject_ids
  end

  def subject_ids
      Array.wrap(@subject_ids) | Array.wrap(@subject_id)
  end
```

camallen (commented):
nice work - some minor comments that might help simplify the code.

camallen (approved):
LGTM 🎉

yuenmichelle1:
@camallen Thanks! Suggested changes made things look super clean and nice.

PR #3616, opened 2021-05-19
tweak sidekiq worker cpu request values

camallen:
setting these too high means the pod can't schedule on our nodes, e.g.
"Warning  FailedScheduling  2m22s  default-scheduler   0/6 nodes are available: 1 node(s) had taint {servicelife: longlife}, that the pod didn't tolerate, 5 Insufficient cpu."

prod sidekiq dumpworker can be set low as it doesn't have an autoscaler that uses these request values

prod sidekiq uses the request values to trigger scaling events, set this lower to allow scheduling but know that this will cause pods to come up and down with load - note these aren't service pods so shouldn't have any impact with the nginx ingress.

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -289,7 +289,7 @@ spec:
           resources:
             requests:
               memory: "2000Mi"
-              cpu: "1000m"
+              cpu: "250m"
             limits:
               memory: "6000Mi"
               cpu: "2000m"
@@ -350,7 +350,7 @@ spec:
           resources:
             requests:
               memory: "500Mi"
-              cpu: "1000m"
+              cpu: "500m"
             limits:
               memory: "2000Mi"
               cpu: "2000m"


camallen:
>  So wouldn't the limits, not the request values, be used to trigger node scaling in the event that the cluster tried to schedule something else?

the AKS cluster autoscaler only schedules new nodes when a pod cannot be scheduled. 

https://docs.microsoft.com/en-us/azure/aks/cluster-autoscaler
> The cluster autoscaler component can watch for pods in your cluster that can't be scheduled because of resource constraints. When issues are detected, the number of nodes in a node pool is increased to meet the application demand.

Note container requests (not limits) determine where a pod can be scheduled, limits are enforced to ensure the pod containers don't consume too many resources. 

> Requests and limits are the mechanisms Kubernetes uses to control resources such as CPU and memory. Requests are what the container is guaranteed to get. If a container requests a resource, Kubernetes will only schedule it on a node that can give it that resource. Limits, on the other hand, make sure a container never goes above a certain value. The container is only allowed to go up to the limit, and then it is restricted.

https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-resource-requests-and-limits

> That is, I thought that the initial scheduling allocated the requests and an HPA will use the limits to determine when to scale, while k8s uses the current usage (bounded by the limits) to decide when to scale.

The HPA modifies the replica count (scale up / down) based on the metrics we've told it to look at - currently avg cpu use on the deployment I believe (so all pods / container in there) as a percentage of the initial container requests. 
https://v1-19.docs.kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#algorithm-details

So in summary 
1. Requests (not limits) are used to figure out where a pod can be scheduled. 
2. If pods can't be scheduled (no available node statifies the pod container request values) then the cluster node autoscaler kicks in
3. new nodes should meet the pod container requests and the pods get scheduled.

PR #3622, opened 2021-06-01
double the number of sidekiq replicas

camallen:
i'm seeing the hpa max out with current traffic volumes (high number of incoming classification)
```
$kubectl describe hpa panoptes-production-sidekiq
...
Reference:                                             Deployment/panoptes-production-sidekiq
Metrics:                                               ( current / target )
  resource cpu on pods  (as a percentage of request):  202% (1011m) / 95%
Min replicas:                                          2
Max replicas:                                          8
Deployment pods:                                       8 current / 8 desired`
```
this PR doubles the current HPA limit to 16 sidekiq pod replicas

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -393,7 +393,7 @@ spec:
     kind: Deployment
     name: panoptes-production-sidekiq
   minReplicas: 2
-  maxReplicas: 8
+  maxReplicas: 16
   targetCPUUtilizationPercentage: 95
 ---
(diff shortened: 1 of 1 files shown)


zwolf (approved):
When are you seeing this? Every time I've checked this it's been sitting around 50% with 2 regular pods and the dumpworker. What symptom presented itself that caused you to look into this? 

Approving now because a higher max isn't a problem if we don't approach it but I'm interested in if performance was degraded or if you just happened to notice that there were a bunch of sidekiq pods.

Issue #3589, opened 2021-04-08
New project resource attribute: authentication invitation


srallen:
There is an authentication invitation that is hard coded right now on PFE associated with the Gravity Spy workflow assignment feature:

<img width="1402" alt="Screen Shot 2021-04-08 at 9 36 03 AM" src="https://user-images.githubusercontent.com/5016731/114045946-0abd9600-984e-11eb-87e5-f788a66d1759.png">

Since we're now generalizing this, I'm thinking a new attribute could be added to the project resource for `authentication_invitation`? This new text would show as another project banner type to appear only for signed out users. I don't think we would need to restrict usage of this to only projects using workflow assignment, but it is likely these projects would use this the most since assignment can't be done if not signed in and I think it does make sense to allow projects to customize their specific authentication invitations.

srallen:
I've implemented the UI with #2196 and once the customized message exists on the project resource, I'll update the component to give preference to that messaging.

PR #3624, opened 2021-06-08
Create 20210602210437_add_auth_invitation_to_project.rb

yuenmichelle1:
Add new field on `project` called `authentication_invitation` relating to issue https://github.com/zooniverse/panoptes/issues/3589

# Review checklist

### DIFF
--- a/db/migrate/20210602210437_add_auth_invitation_to_project.rb
+++ b/db/migrate/20210602210437_add_auth_invitation_to_project.rb
@@ -2,6 +2,10 @@
 
 class AddAuthInvitationToProject < ActiveRecord::Migration
   def change
+    # since PG v11+ we can add a new column and a default at the same time
+    # as it no longer requires a table rewrite 
+    # https://github.com/ankane/strong_migrations#bad-1
+    # https://www.2ndquadrant.com/en/blog/add-new-table-column-default-value-postgresql-11/
     add_column :projects, :authentication_invitation, :text, default: ''
   end
 end
--- a/db/migrate/20210602210437_add_auth_invitation_to_project.rb
+++ b/db/migrate/20210602210437_add_auth_invitation_to_project.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class AddAuthInvitationToProject < ActiveRecord::Migration
   def change
     add_column :projects, :authentication_invitation, :string
--- a/db/migrate/20210602210437_add_auth_invitation_to_project.rb
+++ b/db/migrate/20210602210437_add_auth_invitation_to_project.rb
@@ -1,0 +1,5 @@
+class AddAuthInvitationToProject < ActiveRecord::Migration
+  def change
+    add_column :projects, :authentication_invitation, :string
+  end
+end
--- a/db/migrate/20210602210437_add_auth_invitation_to_project.rb
+++ b/db/migrate/20210602210437_add_auth_invitation_to_project.rb
@@ -2,6 +2,6 @@
 
 class AddAuthInvitationToProject < ActiveRecord::Migration
   def change
-    add_column :projects, :authentication_invitation, :string
+    add_column :projects, :authentication_invitation, :text
   end
 end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1060,7 +1060,7 @@ CREATE TABLE public.projects (
     url_labels jsonb,
     workflow_description text,
     researcher_quote text,
-    authentication_invitation character varying
+    authentication_invitation text
 );
 
 
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -3,7 +3,7 @@
 --
 
 -- Dumped from database version 11.11 (Debian 11.11-1.pgdg90+1)
--- Dumped by pg_dump version 11.11
+-- Dumped by pg_dump version 11.12 (Debian 11.12-0+deb10u1)
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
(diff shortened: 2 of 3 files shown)


db/migrate/20210602210437_add_auth_invitation_to_project.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthInvitationToProject < ActiveRecord::Migration
+  def change
+    add_column :projects, :authentication_invitation, :string
```
camallen, line 3: i'd say we should change the column type to text vs varchar (string) https://stackoverflow.com/questions/3354330/difference-between-string-and-text-in-rails
```suggestion
    add_column :projects, :authentication_invitation, :text
```

db/structure.sql
```diff
@@ -3,7 +3,7 @@
 --
 
 -- Dumped from database version 11.11 (Debian 11.11-1.pgdg90+1)
--- Dumped by pg_dump version 11.11
+-- Dumped by pg_dump version 11.12 (Debian 11.12-0+deb10u1)
```
camallen, line 6: OOI did you dump this with the docker image `pg_dump` and postgres server or using a brew install postgres server and the client tools (pg_dump etc_?

db/migrate/20210602210437_add_auth_invitation_to_project.rb
camallen: Also it's best to set a default on this column to ensure we don't serialize different value types, i.e. `null` or `'possibly empty string or string with data in it'`

Read up on how to do that here https://github.com/ankane/strong_migrations#good-1

Gist is you add the column without a default, set the default and then we can backfill the column values manually. Backfilling can be done in a migration, a rake task or manually on the console. 

We tend to prefer running the data migration / backfill operations via rake tasks so we can test the code, make it resumeable and de-couple from the actual DDL migration (i.e. still migrate the schema even if the data can't be changed), e.g. https://github.com/zooniverse/panoptes/blob/e9001518f410907bd8e87627cd55f659882c4e6e/lib/tasks/migrate.rake#L78-L86

db/migrate/20210602210437_add_auth_invitation_to_project.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthInvitationToProject < ActiveRecord::Migration
```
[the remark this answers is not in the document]
camallen:
```suggestion
# frozen_string_literal: true

class AddAuthInvitationToProject < ActiveRecord::Migration
```

camallen (commented):
This is close - i've left suggestions and comments to get this to a state for merging
1. change the column to a bigger value (string has a shorter length limit)
2. ensure we have a default column value to avoid null and '' values
3. decide / figure out how to back fill the existing table data for the new column

db/structure.sql
yuenmichelle1: Neither? I bashed into my docker container containing the running app, and ran rake db:migrate. I could be wrong, but this change auto-filled when I raked

db/migrate/20210602210437_add_auth_invitation_to_project.rb
yuenmichelle1: I think I see what you mean here. so if i backfill this field, we don't want empty string (`''`)? Should I set it to 'default'?

lib/tasks/migrate.rake
```diff
@@ -455,4 +455,14 @@ namespace :migrate do
       end
     end
   end
+
+  namespace :projects do
+    desc 'Set default value for authentication invitation' do
```
camallen, line 460: This may not be needed anymore after reading that post on the change in PG V11+ behaviour as for rows that don't have that column value, it appears to use the default value when fetching rows and or updating existing rows.
```
Any row that doesn’t have that column must have existed before the table change was made, and uses this value stored in the catalog when the row is fetched.
...
Likewise, any time that a row is updated the column’s value is filled in on the new row version, so over time the proportion of rows using the value fetched from the catalog will decrease.
```

There is now harm in backfilling the data and tbh it could be a useful example of how to run rake tasks etc but it may not be needed now.

I'll leave it up to you to decide to include it or not (and run it via rake task etc).

yuenmichelle1: Thanks for looking into this! I think I'll remove it.

camallen (approved):
LGTM

yuenmichelle1:
@camallen  Off topic, would you happen to know why Travis CI is stuck on "Expected" ? I know it ran with a success when I had this PR in draft mode.

db/migrate/20210602210437_add_auth_invitation_to_project.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAuthInvitationToProject < ActiveRecord::Migration
+  def change
+    add_column :projects, :authentication_invitation, :text, default: ''
```
camallen, line 5:
```suggestion
    # since PG v11+ we can add a new column and a default at the same time
    # as it no longer requires a table rewrite 
    # https://github.com/ankane/strong_migrations#bad-1
    # https://www.2ndquadrant.com/en/blog/add-new-table-column-default-value-postgresql-11/
    add_column :projects, :authentication_invitation, :text, default: ''
```

camallen (commented):
no idea re travis I logged in to travis-ci.com to see if i could restart the build / kick it off but sadly found nothing useful (draft PRs ran). Perhaps it got stuck and this accepting this comment (or adding a new commit of your own) may help get it started again. 

If travis fails I can test manually for you for peace of mind / approval to merge. Longer term we should drop travis for another CI tool, probably GHActions to align with the other team repos.

PR #3625, opened 2021-06-18
ensure we preserve the translated project strings on project copy

camallen:
#3574 introduced project translation copying. However that change had a bug where it would replace the copied translated strings with the primary language strings on all the translations which is useless for translations

This PR changes the way translations are copied and ensure the translated strings stay intact while ensuring the relevant project version records are setup and aligned in the `translation.string_versions` hash.

# Review checklist


yuenmichelle1 (approved):
Sorry for the delay. I wanted to play around and test this on my local. This looks good to me.

camallen:
> Sorry for the delay. I wanted to play around and test this on my local. This looks good to me.

thank you for the quick PR review 🎉

PR #3629, opened 2021-07-01
add per subject_set selection to the designator client

### LINKED ISSUE #3630: "/queued" endpoint seems to ignore ?subject_set_id, for certain workflows
## Functionality Issue

For certain FEM projects with _multiple workflows_ and _multiple subject sets,_ the `/queued` endpoint isn't behaving properly.

- With multi-SSet workflows, the expectation is that I can ping `https://www.zooniverse.org/api/subjects/queued?subject_set_id=85773&workflow_id=16106&http_cache=true` and get ONLY get Subjects from SSet 85773
- However, I'm able to receive Subjects from _every_ Subject Set associated with WF 16106, e.g. SSet 85771 and 85772

_Example: the response from `https://www.zooniverse.org/api/subjects/queued?subject_set_id=85773&workflow_id=16106&http_cache=true` contains results from 85771 (Ultraman series, denoted by the ultraman-___.jpg filenames), 85772 (Transformers series, denoted by the transformers-___.jpg filenames), and 85773 (Overwatch series, denoted by the etc etc etc):_

Screen Shot 2021-07-01 at 16 44 47

### Analysis / Solution

UPDATE: This issue was fixed by @camallen shortly after the initial report. From what I can surmise:

- WF 16106 was initially using _Designator_ to pull Subjects
- This caused it to ignore the ?subject_set_id parameter
- Cam's solution was switch 16106 to use Cellect instead.

WF 16106 is now behaving correctly in regards to the reported issue. 

### Status

UPDATE: Problem & fix identified. It's mostly a "workflow needs to be whitelisted to use Cellect instead of Designator" issue, if I understand correctly.

Currently this issue affects a very small subset of FEM Projects. I'm mostly writing this up so we're aware what the solution is if @eatyourgreens or me encounter the problem again in the future.

camallen:
closes #3630 

Preserve the subject_set_id in the selector client calls to ensure we only select subjects from the specified group

linked to the designator API change in https://github.com/zooniverse/designator/pull/74

# Review checklist

### DIFF
--- a/app/services/designator_client.rb
+++ b/app/services/designator_client.rb
@@ -51,9 +51,9 @@ class DesignatorClient
     end
   end
 
-  def get_subjects(workflow_id, user_id, _group_id, limit)
+  def get_subjects(workflow_id, user_id, subject_set_id, limit)
     url = "/api/workflows/#{workflow_id}"
-    params = { user_id: user_id, limit: limit }
+    params = { user_id: user_id, subject_set_id: subject_set_id, limit: limit }.compact
     request(:get, [ url, params ])
   end
 
--- a/spec/services/designator_client_spec.rb
+++ b/spec/services/designator_client_spec.rb
@@ -59,6 +59,17 @@ RSpec.describe DesignatorClient do
         expect(subject_ids).to eq([1,2,3,4])
       end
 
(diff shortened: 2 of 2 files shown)


shaunanoordin:
Fixes #3630

PR #3631, opened 2021-07-02
add GH actions run tests CI worfklow

camallen:
first attempt at migrate the CI tests run from travis to GH actions 

# Review checklist

### DIFF
--- a/.github/workflows/run_tests_CI.yml
+++ b/.github/workflows/run_tests_CI.yml
@@ -21,13 +21,3 @@ jobs:
         run: docker-compose run -T --rm -e RAILS_ENV=test panoptes bundle exec rake configure:local
       - name: Run all the tests
         run: docker-compose run -T --rm -e RAILS_ENV=test panoptes bundle exec rspec
-  # test:
-  #   needs: setup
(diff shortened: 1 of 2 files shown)


.github/workflows/run_tests_CI.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Run all the tests
+        run: docker-compose run -T --rm -e RAILS_ENV=test panoptes bundle exec rspec
+  # test:
+  #   needs: setup
+  #   runs-on: ubuntu-latest
+  #   env:
+  #     SPECS: spec/controllers/api/v1/[a-m]*.rb
+  #   steps:
+  #     - name: Run tests
+  #       run: docker-compose run -T --rm -e RAILS_ENV=test panoptes bundle exec rspec $SPECS
```
yuenmichelle1, line 31: Do we need the above commented out code?

camallen: nope - i'll remove it.

yuenmichelle1 (approved):
I had one quick question (already posted in a comment). Other than that, it looks like everything seems to be working. So I'm good with it.

PR #3637, opened 2021-07-16
buildkit caching for docker builds

camallen:
adds buildkit caching for the bundler & apt caches to the dockerfile

https://github.com/moby/buildkit/blob/2a4577efabed1b4404e2884ef56873b8c0a42e95/frontend/dockerfile/docs/syntax.md#run---mounttypecache

this allows our dev machines and CI system to re-use the cache across builds and should help speed up the build times of the dockerfile.

# Review checklist


Dockerfile
```diff
@@ -3,7 +3,9 @@ FROM ruby:2.5-slim-stretch
 
 WORKDIR /rails_app
 
-RUN apt-get update && \
+RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
+RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/apt \
```
camallen, line 7: Does the latter `apt-get clean` up operation bust our caches here? 

If so we need to modify the latter install && clean step to leave the apt artifacts in the cache mount location for re-use...

camallen: I think each repo / apt cache should be isolated to avoid any issues with package clashes. I can't really see this happening due to apt internals working but their may be some assumptions about the OS apt cache being isolated fro other OS's, e.g. Ubuntu 18.04 may clash with 14.04 apt caching (relying on the docker base image OS version etc)

eatyourgreens: I don't know much about `apt-get` but my quick reading is that `apt-get clean` will clear `/var/cache/apt/archives`. So, yeah, my initial impression is it would bust the cache here? I haven't tried running this to see if that's true in practice.

eatyourgreens: Having said that, is disk space an issue? I'm not sure what the advantage is of cleaning up after the install.

camallen: it was to remove the cache from the resulting docker image layer. as the cache mount isn't recorded in the layer artefact we don't need the `apt clean` cmd now. in fact it will hinder reusing the cache. good catch :)

eatyourgreens: Other than that, this looks good to me.

eatyourgreens (approved):

PR #3640, opened 2021-07-22
ensure primary language resources are created for template project association resources

camallen:
Ensure we don't break the translations.zooniverse.org app's expected data model when copying a template project's associated resources, e.g. `project.active_workflows`, `project.pages`, project.field_guides`

In the normal `CRUD` API landscape all translated resources will have these primary language translations setup correctly and the translations app relies on them being present. Thus we need to set these up when copying these template project and their translated association resources. 

Thus this PR syncs the default translation (primary language) strings from the copied association resource to the translation. It does not however copy any existing translations on these translated associations (`active_workflows`, etc). Long term this is probably desired but out of scope for this data model fix PR**.

** FWIW this template project copy feature isn't in active use via the API right now but may be as project team and collaborators find the feature useful. We can revisit this if and when they need the full template project translations resources copied as well.

# Review checklist

### DIFF
--- a/app/workers/project_copy_worker.rb
+++ b/app/workers/project_copy_worker.rb
@@ -4,7 +4,7 @@ class ProjectCopyWorker
   sidekiq_options queue: :data_high, lock: :until_executed
 
   def perform(project_id, user_id)
-    ProjectCopier.copy(project_id, user_id)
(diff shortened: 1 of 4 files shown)


yuenmichelle1 (approved):
Sorry for delay. Looks good to me 👍

PR #3651, opened 2021-08-13
Handle unknown subjects in the retirement worker

camallen:
This PR changes the retirement worker to ignore any unknown subject ids passed in. Previously this would raise and error and stop processing the list of subjects we had for retirement. Thus 1 incorrect subject_id could break the retirement list processing at any point. 

I found this issue due to a failed (dead) job in sidekiq and some data debugging found 1 unknown subject id in the list. 

linked HB error https://app.honeybadger.io/projects/40595/faults/38047692

# Review checklist

### DIFF
--- a/app/workers/retire_subject_worker.rb
+++ b/app/workers/retire_subject_worker.rb
@@ -10,6 +10,9 @@ class RetireSubjectWorker
       Array.wrap(subject_ids).each do |subject_id|
         count = subject_workflow_status(subject_id)
         RetirementWorker.perform_async(count.id, true, reason)
+      rescue ActiveRecord::RecordInvalid
+        # need to rescue the error when folks pass in a subject id that doesn't exist
+        # but we still keep processing the whole list of subjects we have
       end
     end
   end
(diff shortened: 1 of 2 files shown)


yuenmichelle1:
This looks good to me. 

One question: would it be considered overkill if we add a test where `subject_ids` is nil or empty? (I'm assuming in those cases `RetirementWorker` would not be expected to perform). Totally fine if you think its an unnecessary.

yuenmichelle1 (approved):

camallen:
> One question: would it be considered overkill if we add a test where `subject_ids` is nil or empty? (I'm assuming in those cases `RetirementWorker` would not be expected to perform). Totally fine if you think its an unnecessary. 

If it's nil or an empty array the `first_or_create!` would error and thus just skip any other work. So I think we're covered for behaviour but maybe we could add a spec. 

Tbh this should only get called after Param checking events in the operation code etc which handles the nil / empty checks https://github.com/zooniverse/panoptes/blob/85af7c8f4afc9ca55632715648d5e39ef459064c/app/operations/workflows/retire_subjects.rb#L19

yuenmichelle1:
Ah got it. I forgot the context of when this is run (through the operation) and was thinking about if someone ran from the `RetireSubjectWorker` straight, which shouldn't be a concern since normal users shouldn't be able to do this anyway. I'm ok with not having the test.

PR #3655, opened 2021-09-09
Add retry mechanism when a medium uploads files to an object store

camallen:
This PR adds a retry mechanism around `put_file` on the Medium resource when uploading to the remote object store. 

When exporting data, the exports are created on the file system and the files have been compressed but when uploading to the object store (s3, azure blob, etc) the `put_file` call fails (e.g. `Faraday::ConnectionFailed: Connection reset by peer`) and the export never finishes and the UX shows broken export requests. 

The problem is outlined in this error https://app.honeybadger.io/projects/40595/faults/78910693

This error is transient thus we can't really predict the failure mode, thus using the retry mechanism seems a reasonable approach to attempt to avoid this error. Longer term having the our network or object store provider not error on upload seems a more reasonable solution. 

# Review checklist

### DIFF
--- a/app/models/csv_dumps/direct_to_s3.rb
+++ b/app/models/csv_dumps/direct_to_s3.rb
@@ -30,6 +30,19 @@ module CsvDumps
       )
     end
 
+    def put_file_with_retry(gzip_file_path, opts={}, num_retries=5)
+      attempts ||= 1
+      put_file(gzip_file_path, opts)
+    rescue UnencryptedBucket => e # do not retry this error
+      raise e
+    rescue => e # rubocop:disable Style/RescueStandardError
+      retry if (attempts += 1) <= num_retries
+
+      # ensure we raise unexpected errors once we've exhausted
+      # the number of retries to continute to surface these errors
+      raise e
+    end
+
     def storage_adapter(adapter='aws')
       return @storage_adapter if @storage_adapter
 
--- a/app/models/csv_dumps/dump_processor.rb
+++ b/app/models/csv_dumps/dump_processor.rb
@@ -50,7 +50,7 @@ module CsvDumps
     end
 
     def write_to_object_store(gzip_file_path)
-      medium.put_file(gzip_file_path, compressed: true)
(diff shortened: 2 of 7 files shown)


yuenmichelle1 (commented):
Everything seems fine to me. I have a quick question for my understanding: 
- https://app.honeybadger.io/projects/40595/faults/78910693 linked in comments errored on `put_file` of `azure_adapter.rb`, is the `put_file` (of `DirectToS3`) in the PR related?

camallen:
> Everything seems fine to me. I have a quick question for my understanding:
> 
>     * https://app.honeybadger.io/projects/40595/faults/78910693 linked in comments errored on `put_file` of `azure_adapter.rb`, is the `put_file` (of `DirectToS3`) in the PR related?

Sadly yes - see the commit message in https://github.com/zooniverse/panoptes/pull/3655/commits/3740cf1c9dbb2f0cdb5c83cd2793010050acda3d
> However as the medium interface is used for DirectToS3 uploader class we need to support this method as it's used in the exporter code.

So all the exporter code re-uses this same interface that now calls `put_file_with_retry` https://github.com/zooniverse/panoptes/blob/3740cf1c9dbb2f0cdb5c83cd2793010050acda3d/app/models/csv_dumps/dump_processor.rb#L53

so DirectToS3 has to implement that method. I chose to add the retry behaviour instead of a method alias that didn't retry at all as it doesn't hurt to retry this on some kind of failure mode connecting to the object store. 

FWIW - DirectToS3 is/was a workaround class (it's really a ducktype of `Medium` but uses a different storage adapter) to write code to a different bucket in S3 vs the default for normal media resources (an encrypted one for private data). We re-use this now to write to a different cloud provider vs the default one setup in `MediaStorage.adapter` (Azure blob). Longer term when we convert the email list exportes / uploaders to write to Azure we can remove the `DirectToS3` class.

yuenmichelle1 (approved):
Ah ok got it. That answers my other lingering question (i.e. why we still have this `DirectToS3` class). 

Yeah, these changes look good. Thanks for walking it through.

PR #3658, opened 2021-09-15
remove server side language content loading

camallen:
This PR removes vestigial code that was used to load translated content via user language preferences, e.g. old *Content models that stored translations before the extraction to translations service. 

Linked to #3099 and prompted by #3656 

All translated content now is handled by the translations service via https://github.com/zooniverse/pandora/

# Review checklist


yuenmichelle1 (approved):

yuenmichelle1:
Not really a comment, more of a question/followup. in response to https://github.com/zooniverse/panoptes/issues/3656, the plan is to pull language preference from the User resource via Panoptes and the api will persist the language preference between sessions. did I understand that correctly?

camallen:
> Not really a comment, more of a question/followup. in response to #3656, the plan is to pull language preference from the User resource via Panoptes and the api will persist the language preference between sessions. did I understand that correctly?

Yes - basically the User Resource becomes a place to store an individuals language preferences that can then be used to modify the UI translation strings. As the data is stored in the API it's consistent across all sessions the user is logged into our platform. 

This persisted language setting can easily be overridden using local / session storage in the unique browser to allow a user to modify the language of a browser by browser basis if that's the sort of behaviour we'd like to offer. 

I.e. volunteer is logged in on two browsers and on
1. Browser 1 they have no local/session store value so uses the `user.langages` values to wire up the UI
2. Browser 2 they have a local/session store value set to `es` to ensure the UI uses Spanish translations where they exist for the UI strings and the custom per project translations content.

PR #3714, opened 2021-10-22
Run data export jobs on set completion

camallen:
builds on #3707 

This PR will now run events when a subject set completes. This PR introduces the following run events when the subject set completes:
1.  create the subject set classifications data export
2. notify the project team the subject set has completed

This PR has abstracted these events to the `SubjectSets::RunCompletionEvents` operation class which is called via the `SubjectSetCompletenessWorker` when a set reaches completion and only if there is a change in the subject set completeness value - i.e. avoid unnecessary update operations that leave the data unchanged.

Finally this PR fixes the lab page export URL for subject set classification data export emails, previously these were linked to the subject set id but now correctly wire up the valid export page lab URL.

~## TODO~
- [x] extract the worker post completeness events code to an operation object to simplify the worker code / specs
- [x] Bring in work from #3719 once that is accepted
- [x] should it use the project owner as the requesting user to send a notification? - leaning to no requesting user here as we have notifications in play already but this may not play well with rate limiting and subsequent requests by research teams (exports may run twice...). Perhaps use a null style object that triggers rate limiting but doesn't send notifications.
- [x] how to best configure this as opt in behaviour via the project configuration (or other means) - ensure the user can't set these fields as it may override our export rate limiting system.

# Review checklist


app/models/project.rb
```diff
@@ -170,10 +170,11 @@ def communication_emails
...
-  def notify_on_subject_set_completion?
-    # default behaviour will be to opt in to subject set completion notifications
-    # but we can by changing this below on the fetch fallback.
-    notify_value = configuration.fetch('notify_on_subject_set_completion', 'false')
-    notify_value.to_s.casecmp('true').zero?
+  def run_subject_set_completion_events?
+    # short term use ENV vars to 'allowlist' projects into this functionality - so we control it
+    # longer term perhaps switch to opt in behaviours via project configuration setting?
+    # or should this be an admin only flag for our control??
+    allow_list_project_ids = ENV.fetch('SUBJECT_SET_COMPLETION_EVENT_PROJECT_IDS', '').split(',').map(&:to_i)
```
camallen, line 177: this could easily be moved to an explicit, admin-only attribute on the project resource that we control via the admin page. TBC

yuenmichelle1 (approved):
I had one minor question on the setting env params, other than that (and merge conflict), it looks good to me.

app/models/project.rb
camallen: As it's currently setup through ENV vars - your understanding is correct. The more i think about this i think we should move it to an admin only `project.run_subject_set_completion_events` boolean attribute (default is false)

Moving it to an admin attribute will allow zoo team members to opt in / out of this functionality via the admin pages / API  and avoid having to ask a dev team member to update some env vars and deploy.

Also longer term it's easy to switch that attribute to non-admin and allow project teams to opt into this behaviour (via lab interface or the API).

I think i should do that before merging... Thoughts?

camallen: I'll merge this for now and follow up with a PR to move this to an admin only project attribute.

I'm mostly concerned with the organizational time debt where a zoo team member has to rope a developer in and merge a PR and deploy etc just to add this feature for a project.

PR #3738, opened 2021-12-01
remove subject group allowlist feature opt in

camallen:
this PR removes the workflow allow list to opt into the subject grid feature. Instead this can be configured by the FEM codebase / lab.

# Review checklist


shaunanoordin (approved):
## PR Review

Prior to this PR, a project that wants to use the "Grid Tool" (technically Subject Group Viewer + Subject Group Comparison Task) needs to have the WF white listed in Panoptes

With this PR, the white list is removed. The only thing required now for a Project Owner to enable the "Grid Tool" is to enable the SGV config + SGC Task on the Project Builder (currently hidden behind an experimental tool).

I'm all for this 👍 This is in light of recent requests (Feb 2022) from Kameswara and Ramanakumar to experiment with the Grid Tool to create something like [Galaxy Zoo: Weird and Wonderful](https://www.zooniverse.org/projects/zookeeper/galaxy-zoo-weird-and-wonderful)

PR #3740, opened 2021-12-02
Update project copy end point

camallen:
This PR is tied to work on the VRO platform where they will want to quickly spin up projects with existing content in them. As such they need to copy a template project but they need the copied project serialized in the json response so they can programatically interact with the newly copied project without having to search through a list of projects they have access to. 

Thus this PR
- adds an operation to call the ProjectCopier instance
- modifies the copier behaviour to be in-band (not async in a worker)
- serializes the newly copied resource via the HTTP 201 (created) response type
- allows the copier to create a new subject set resource during the copy - used for subsequent data uploads in VRO use case

### TODO
- [x] Add the ability to create a new subject set to the project on the copy operation (avoid a sunsequent API call for the manifest upload preparation step)

# Review checklist

### DIFF
--- a/spec/operations/projects/copy_spec.rb
+++ b/spec/operations/projects/copy_spec.rb
@@ -27,7 +27,7 @@ describe Projects::Copy do
 
   it 'raises an error without project param' do
     expect {
-      operation.run!(operation.run!(params.except(:project)))
+      operation.run!(params.except(:project))
     }.to raise_error(ActiveInteraction::InvalidInteractionError, 'Project is required')
   end
 
--- a/spec/operations/projects/copy_spec.rb
+++ b/spec/operations/projects/copy_spec.rb
@@ -34,7 +34,7 @@ describe Projects::Copy do
   it 'raises an error without api_user param' do
     operation = described_class.with({})
     expect {
-      operation.run!(operation.run!(params))
+      operation.run!(params)
     }.to raise_error(ActiveInteraction::InvalidInteractionError, "User can't be blank")
   end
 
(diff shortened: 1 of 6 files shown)


spec/operations/projects/copy_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    copier_double = instance_double(ProjectCopier)
+    allow(copier_double).to receive(:copy)
+    allow(ProjectCopier).to receive(:new).and_return(copier_double)
+    copied_project
+    expect(copier_double).to have_received(:copy)
+  end
+
+  it 'raises an error without project param' do
+    expect {
+      operation.run!(operation.run!(params.except(:project)))
```
yuenmichelle1, line 30: Quick question: is there a reason we do `operation.run!(operation.run!(params.except(:project)))` vs `operation.run!(params.except(:project))`?

yuenmichelle1 (approved):
All this looks good to me. I had one question on the copy test, but it's super minor.

spec/operations/projects/copy_spec.rb
camallen: that's a straight up copy-paste mistake :( 

Luckily that spec still works as the inner `operation.run!(params.except(:project)` raises correctly before the second is executed - fixing. 
```suggestion
      operation.run!(params.except(:project))
```

spec/operations/projects/copy_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  it 'raises an error without project param' do
+    expect {
+      operation.run!(params.except(:project))
+    }.to raise_error(ActiveInteraction::InvalidInteractionError, 'Project is required')
+  end
+
+  it 'raises an error without api_user param' do
+    operation = described_class.with({})
+    expect {
+      operation.run!(operation.run!(params))
```
camallen, line 37:
```suggestion
      operation.run!(params)
```

PR #3741, opened 2021-12-03
Enforce 10k limit SubjectSetImport manifest uploads

camallen:
This PR ensures we don't allow folks to create SubjectSetImport manifests over a row count limit and thus limit the number of subject records a user imports to a project at any given time (note: all subjects created by this import process are externally hosted).

It does this by counting the incoming manifest data rows before creating a SubjectSetImport resource, thus avoiding the poor UX of erroring while processing and orphaning an resource in an error state (there are no updates on that resource). 

Sadly this means we download the manifest twice**, once on validation check before we create / save the subject set import resource and once when processing. I think that's ok as they shouldn't be very large and the second download is robust via the worker retry mechanisms. 

** We can't pass a tmp download file to the worker system as it runs in a different ruby vm, we could save this file as a resource attachment but I think it's simpler in code / resources to pay the download cost again.

### TODO

- [x] Extract the manifest counter and limit validations to an operation class and cleanup the controller specs (should have done this first)

# Review checklist

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -17,7 +17,8 @@ module Api
       Subjects::Selector::MissingSubjectSet,
       Subjects::Selector::MissingSubjects,                 with: :not_found
     rescue_from ActiveRecord::RecordInvalid,               with: :invalid_record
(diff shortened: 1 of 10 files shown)


yuenmichelle1 (approved):
LGTM

PR #3754, opened 2021-12-16
track when a user has seen a subject

camallen:
This PR modifies the behaviour of how we track the subject ids a user has seen in the UserSeenSubject record. Specifically we now touch an existing uss record when we update the tracked subject_ids, this allows us to use the USS model as a proxy for when they've last seen a subject on a workflow.

Also this will switch to an explicit `first_or_create!` method to ensure we are still creating the record when we can't find it in the db.

# Review checklist


yuenmichelle1 (approved):
LGTM

PR #3760, opened 2022-01-06
increase the client body payload max size

camallen:
some classification payloads may be bigger than the k8 ingress default nginx payload size - http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#proxy-body-size

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -589,6 +589,7 @@ metadata:
     cert-manager.io/cluster-issuer: letsencrypt-prod
     nginx.ingress.kubernetes.io/rewrite-target: /$1
     nginx.ingress.kubernetes.io/proxy-buffer-size: "128k"
+    nginx.ingress.kubernetes.io/proxy-body-size: 20m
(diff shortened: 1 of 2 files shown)


shaunanoordin (approved):
## PR Review

Currently, attempting to submit a "large Classification" to Panoptes will cause the FEM Classifier (and possibly the PFE Classifier, though this has to be confirmed) to silently fail. The Panoptes API will respond with a 413, but the user on FEM will not know the submission failed (AND the failing classification will be perpetually queued to be re-submitted, in the background, but this is a separate issue)

This PR attempts to test a quick workaround to the issue, by increasing the acceptable limit for a "large Classification"

- Currently, a "large Classification" is any HTTP Request with a body size above *1MB* (estimation)
  - In our tests, we were able to successfully submit a Classification with a 945kb body , but failed to submit a Classification with a 1.2MB body.
  - Testes we performed by submitting the text of a random novel to a simple Workflow with a single Text Task.

### Status

Changes LGTM

PR #3761, opened 2022-01-07
avoid reporting kinesis payload error for specific projects

camallen:
This PR introduces a change to avoid reporting kinesis publishing errors for some known projects via an 'ignore' list. 

This change came about as some projects don't use the downstream services (caesar etc) - these we can avoid noisy error reports by not reporting. Note all other kinesis publishing errors will be reported and we report all errors for projects that are not on the 'ignore' list

This will be useful to keep the error signal to noise ratio high for debugging possible issue with very large transcription payloads and our k8s nginx ingress.

# Review checklist

### DIFF
--- a/app/workers/publish_classification_worker.rb
+++ b/app/workers/publish_classification_worker.rb
@@ -23,10 +23,14 @@ class PublishClassificationWorker
   end
 
   def publish_to_kinesis!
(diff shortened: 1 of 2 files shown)


yuenmichelle1 (approved):
LGTM

PR #3788, opened 2022-02-15
auto export nasa email lists

camallen:
This PR adds the ability to auto export the nasa email list from the user email subscription settings.

# Review checklist

### DIFF
--- a/app/workers/emails_export_worker.rb
+++ b/app/workers/emails_export_worker.rb
@@ -4,6 +4,7 @@ class EmailsExportWorker
 
   sidekiq_options queue: :data_medium
   BETA_DELAY = 15.minutes.freeze
+  NASA_DELAY = 30.minutes.freeze
   PROJECT_SPREAD = 1.hour.freeze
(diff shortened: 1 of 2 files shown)


lcjohnso (approved):
LGTM!  What else, if anything, needs to be done to make this operational?  Just start calling the job as part of the automated daily dump, or is that included in this PR via edit to `app/workers/emails_export_worker.rb`?

Also worth noting that this builds on original PR https://github.com/zooniverse/panoptes/pull/3421.

camallen:
> What else, if anything, needs to be done to make this operational? Just start calling the job as part of the automated daily dump, or is that included in this PR via edit to `app/workers/emails_export_worker.rb`?

Nothing - this will be run automatically in prod via the scheduled jobs config (every day @ 3am) https://github.com/zooniverse/panoptes/blob/6ef1579c7f9030cd76b8b01722744670bce4f338/app/workers/emails_export_worker.rb#L9

PR #3812, opened 2022-03-29
handle any case of #priority subject metadata

camallen:
allow project owners leeway to get the #priority metadata case wrong and still have the same behaviour of correctly specifying the subject selection ordering

This PR changes the case sensitive metadata key lookup to an insensitive one. It relies on the the `jsonb_each_text` function to extract the keys and values from all subject metadata and then we can do a lower case comparison on the `#priority` metadata key. 
https://www.postgresql.org/docs/11/functions-json.html

# Review checklist

### DIFF
--- a/app/workers/subject_metadata_worker.rb
+++ b/app/workers/subject_metadata_worker.rb
@@ -29,12 +29,13 @@ class SubjectMetadataWorker
 
   def run_ar5_update
     update_sms_priority_sql =
-      <<-SQL
+      <<-SQL.squish
         UPDATE set_member_subjects
-        SET    priority = CAST(subjects.metadata->>'#priority' AS NUMERIC)
+        SET    priority = CAST(value AS NUMERIC)
         FROM   subjects
+        cross join jsonb_each_text(metadata)
         WHERE  subjects.id = set_member_subjects.subject_id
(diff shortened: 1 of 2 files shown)


camallen:
an alternative setup could detect this metadata key on `subjects` `create` end point and then ensure it's downcased for storage and consistent use in other parts of this system and downstream services. 

Perhaps around here https://github.com/zooniverse/panoptes/blob/0e2b183319651baaca85a314a0f8d7c173ed73a1/app/controllers/api/v1/subjects_controller.rb#L166-L179

That might be a better way to approach this problem, especially considering downstream services may be looking for this attribute to be downcased, https://github.com/zooniverse/subject-set-search-api/blob/125c2e9812aa3cdc13f85abbd6948b8f48409e39/subject-set.py#L28-L31

eatyourgreens:
There's JS code in the subject viewer, which expects `priority` to be lowercase too.
https://github.com/zooniverse/front-end-monorepo/blob/9530ecd8dd125a517f07101dd736689d3529c4c8/packages/lib-classifier/src/store/SubjectStore/Subject/Subject.js#L93-L99

camallen:
closing in favour of #3815

PR #3813, opened 2022-03-29
use set display name for naming the export file

### LINKED ISSUE #3810: Subject Set Exports: CSV Filename Incomplete
When downloading a new subject set level classification export (see #3719), the output filename assigned on download is:
`-classifications.csv`

Expected behavior: filename is downcased + no spaces version of the subject set display name with `-classifications.csv` appended to the end:
`terraaqua20210301to20210430-classifications.csv` for Subject Set 97255

It looks like @camallen might have already been looking into this in #3731, but doesn't seem like it is currently working as expected.

camallen:
fixes #3810 

Ensure we set the subject set display name as part of the file export name via the the content-disposition header for per subject set classification exports. 

This PR adds the SubjectSet resource to the lookup types for setting the content disposition and in turn ensures the value is set on the uploaded object store resource to be used when downloading. 

# Review checklist

### DIFF
--- a/app/models/csv_dumps/finds_medium.rb
+++ b/app/models/csv_dumps/finds_medium.rb
@@ -47,7 +47,7 @@ module CsvDumps
 
     def content_disposition
       case resource
-      when Workflow
+      when Workflow, SubjectSet
         name = resource.display_name.parameterize
       when Project
         name = resource.slug.split("/")[1]
--- a/spec/models/csv_dumps/finds_medium_spec.rb
+++ b/spec/models/csv_dumps/finds_medium_spec.rb
@@ -61,5 +61,18 @@ describe CsvDumps::FindsMedium do
       file_name = "#{name}-#{type}.#{ext}"
       expect(medium.content_disposition).to eq("attachment; filename=\"#{file_name}\"")
     end
+
+    context 'when the resource is a subject-set' do
+      let(:resource) { create :subject_set }
+
+      it 'should update the medium content_disposition' do
+        finder.medium
+        medium.reload
+        type = medium.type.match(/\Aproject_(\w+)_export\z/)[1]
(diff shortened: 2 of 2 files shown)


lcjohnso (approved):
LGTM! Thanks for the quick resolution, @camallen !

One Q: I’m assuming that display names are sanitized (like project names are?) so that they will produce valid export file names. Is that a safe assumption?

camallen:
> One Q: I’m assuming that display names are sanitized (like project names are?) so that they will produce valid export file names. Is that a safe assumption?

It should be as it uses `paramterize` method https://apidock.com/rails/ActiveSupport/Inflector/parameterize 
Also it's the same as we do for the workflow exports and we've had no complaints on those.

eatyourgreens:
Are there any limits on subject set names? Here's one I've set up from a British Library catalogue file, with the name: [_Glass Plate Negative Collection: documentary photographs of Siberian indigenous peoples and cultures [Late 19th to Early 20th Century]_](https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify/workflow/3599/subject-set/5018?env=staging&demo=true)

camallen:
> Are there any limits on subject set names?

there is a hard limit and it is very large.... that name should be fine, this is what the filename would look like as an export
`glass-plate-negative-collection-documentary-photographs-of-siberian-indigenous-peoples-and-cultures-late-19th-to-early-20th-century-classifications.csv`

``` ruby
> t
 => "Glass Plate Negative Collection: documentary photographs of Siberian indigenous peoples and cultures [Late 19th to Early 20th Century]" 

> t.parameterize
 => "glass-plate-negative-collection-documentary-photographs-of-siberian-indigenous-peoples-and-cultures-late-19th-to-early-20th-century"
```

PR #3814, opened 2022-03-29
Actions: Production migration & deploy

zwolf:
Templates and workflows for production deploy, migration, and task running. **Migration and deploy workflows are separate, unlike staging**. Production migration will trigger when the workflow detects a push to the `production-migrate` tag, deploys on push to `production-release`. Both of those tags are updateable via chatops and the jobs themselves (and tasks) can also be run manually.

Jenkinsfile deleted.

# Review checklist


camallen (approved):
LGTM - i think we need to remove the prod db k8s tmp file https://github.com/zooniverse/panoptes/blob/master/kubernetes/job-db-migrate-production.tmpl as it's added here not renamed.

PR #3834, opened 2022-04-26
fix metadata reserved keys transform

### LINKED ISSUE #3832: Updates to Subjects error when update params do not contain metadata
Recently, the `SubjectsController#build_update_hash` method was updated to include a call to a new method:

https://github.com/zooniverse/panoptes/blob/e3949292f9510224cb3688bb362250bed7967d3a/app/controllers/api/v1/subjects_controller.rb#L178

for the purpose of ensuring that a specific set of metadata keys were downcased before an update, so that metadata cases like `#Priority`, would act as expected:

https://github.com/zooniverse/panoptes/blob/e3949292f9510224cb3688bb362250bed7967d3a/app/controllers/api/v1/subjects_controller.rb#L203-L209

This method assumes that `params[:metadata]` exists, which is not always true. We received a ticket in which a user was trying to update a subject's locations (adding a frame to an existing subject) via the python client. The request that this generated failed and a `NoMethodError: undefined method 'transform_keys!' for nil:NilClass` error was logged.

camallen:
fixes  #3832

fixes a bug introduced in #3815 - this PR will now skip the metadata reserved key transform when metadata params are not present in the update action.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -201,6 +201,8 @@ class Api::V1::SubjectsController < Api::ApiController
   end
 
   def downcase_reserved_metadata_keys(params)
+    return unless params.key?(:metadata)
+
     params[:metadata].transform_keys! do |key|
       if RESERVED_METADATA_KEYS.include?(key.downcase)
         key.downcase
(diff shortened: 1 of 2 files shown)


lcjohnso:
FYI @zwolf -- Cam put in this bug fix this morning.

PR #3866, opened 2022-06-15
Rails 5.0 update Tests fixes for EventsController, SMS model, and CatchAPIJSONParseErrors

yuenmichelle1:
Moving forward I'll probably just do 1-2 specs at a time for manageable PRs. 

Taking care of tests within EventsControlller, SetMemberSubject model, and  CatchAPIJSONParseErrors 

- [x] rspec ./spec/controllers/api/events_controller_spec.rb:88 # Api::EventsController using json #create with an invalid json message should return 422

- [x] rspec ./spec/models/set_member_subject_spec.rb:162 # SetMemberSubject with seen data :unseen_for_user_by_workflow when the user has seen all the workflow subjects should return an empty set
 fix: method expects id to be parameters, prior was sending the whole object

- [x] rspec './spec/middleware/catch_api_json_parse_errors_spec.rb[1:3:2:1:1]' # CatchApiJsonParseErrors call errors with ParseError when the call is JSON behaves like a json error response should return an rack response array

- [x] rspec './spec/middleware/catch_api_json_parse_errors_spec.rb[1:3:2:1:4]' # CatchApiJsonParseErrors call errors with ParseError when the call is JSON behaves like a json error response should return the body as stringified JSON

- [x] rspec './spec/middleware/catch_api_json_parse_errors_spec.rb[1:3:2:1:3]' # CatchApiJsonParseErrors call errors with ParseError when the call is JSON behaves like a json error response should return a application/vnd.api+json content-type

- [x] rspec './spec/middleware/catch_api_json_parse_errors_spec.rb[1:3:2:1:2]' # CatchApiJsonParseErrors call errors with ParseError when the call is JSON behaves like a json error response should return the correct status

- [x] rspec './spec/middleware/catch_api_json_parse_errors_spec.rb[1:3:2:1:5]' # CatchApiJsonParseErrors call errors with ParseError when the call is JSON behaves like a json error response should match the msg

- [x] rspec ./spec/middleware/catch_api_json_parse_errors_spec.rb:45 # CatchApiJsonParseErrors call errors with ParseError when the call is not JSON should reraise the error

Describe your change here.

# Review checklist


spec/controllers/api/events_controller_spec.rb
```diff
@@ -87,7 +87,7 @@ def overridden_params(new_params)
 
         it "should return 422" do
           allow(subject).to receive(:create_params)
-            .and_raise(ActionDispatch::ParamsParser::ParseError.new('test', 'test'))
```
yuenmichelle1, line 90: `ActionDispatch::ParamsParser::ParseError` will be depecrated after 5.0 
See : https://guides.rubyonrails.org/5_0_release_notes.html
https://stackoverflow.com/questions/47942780/migrating-from-rails-4-to-rails-5-using-actiondispatchrequest-parameter-parse

TODO: Update `application.rb` to use middleware by instead of sending as strings send over the class i.e. `CatchJsonApiParseErrors` vs `"CatchApiJsonParseErrors"`

spec/models/set_member_subject_spec.rb
```diff
@@ -152,15 +152,15 @@
         let(:subject_ids) { [] }
 
         it "should return the all the worflow set_member_subjects" do
-          expect(SetMemberSubject.unseen_for_user_by_workflow(user, workflow)).to match_array(smses)
```
yuenmichelle1, line 155: method expects a user Id and workflow Id. when sending over just the object, the query that expects something along the lines of `WHERE user.id=user_id`,comes up `WHERE user.id=NULL`. 

I believe it has something to do along the lines with https://github.com/rails/arel/issues/531, or something similar. 
(It seems that this may be broken until we get to rails 5.2 [where we send an object when an id is expected)

POSSIBLE TODO (in a followup PR) -
update https://github.com/zooniverse/panoptes/blob/b9e1e605285e01c15d9e2e45b1515a190989a11e/lib/subjects/set_member_subject_selector.rb#L26-L31 reference to send over `user.id` and `workflow.id`

spec/controllers/api/events_controller_spec.rb
```diff
@@ -87,7 +87,7 @@ def overridden_params(new_params)
...
+            .and_raise(JsonSchema::ValidationError.new)
```
camallen, line 90: Not blocking - i think this whole `event_controller.rb` can go - it was originally made to integrate with our old API events to keep counts up to date for existing Ouroboros projects. 

now that isn't running anymore we can remove this code. 

perhaps a follow up PR or add to a task pile for the rails 5.0 upgrade.

camallen (approved):
LGTM - 👍🏼 nice work.

spec/middleware/catch_api_json_parse_errors_spec.rb
```diff
@@ -1,72 +1,71 @@
...
+    let(:error) {
+      # raise an error here to simulate the HTTP request data parsers blowing up
+      # https://github.com/rails/rails/commit/b3d41eae4b138d6c9d38bd9c1cbe033802c0e21d
+      # to match the deprecated params parsing error behaviour
+      # via last know error $! (https://til.hashrocket.com/posts/da62981e47-ruby-)
+      begin
+        # this would really be a specific parser error in reality, by Rails 5.1 should be ActionDispatch::Http::Parameters::ParseError
+        raise StandardError
+      rescue StandardError
+        return ActionDispatch::ParamsParser::ParseError.new('test', 'test')
```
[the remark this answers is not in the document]
yuenmichelle1: ah you know what, i forgot i caught this yesterday. params are expected in rails 4.2 but deprecated (but not breaking if added) in 5.0

PR #3871, opened 2022-06-16
switch instructions to use the BUNDLE_GEMFILE env var for env dual boot

camallen:
This PR adds instructions for running the switching to rails 5 install of the app vs using the next bin cmd. For some reason outside of docker the next shell script doesn't work, thus we can avoid it an override the gemfile env var to determine the version of next gems. 

# Review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,13 +101,25 @@ https://www.youtube.com/watch?v=6aCfc0DkSFo
 
 #### Install the gems via next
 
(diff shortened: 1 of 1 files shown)


yuenmichelle1 (approved):
> For some reason outside of docker the next shell script doesn't work, thus we can avoid it an override the gemfile env var to determine the version of next gems.

Got it. I was wondering why we would use this vs `next __etc_` cmds. Makes sense.  👍

PR #3875, opened 2022-06-17
convert query scope to list of AR records for cache_key lookup

camallen:
Ensure we use the AR model cache_key vs the new Rails 5 AR Relation query sql cache_key (as our sql can vary slightly between the get etag and other update / delete operations).

convert the query scope to an array of active record objects (vs an ActiveRecord::Relation scope) to ensure we generate the cache key based on the underlying Active Record objects and thus have a stable cache key https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-cache_key as opposed to the ActiveRecord::Relation which depends on the query sql https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-cache_key

Note: This is a little tricky and we need to make sure we get this right for the way we do IF Match precondition checks that our UI code relies upon to ensure we can modify and delete resources, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match

Happy to jump on call to discuss and approach this problem a different way :) Please do read the code comments and links above for details on why i chose this implementation. 
# Review checklist


app/controllers/api/v1/media_controller.rb
```diff
@@ -17,7 +17,7 @@ def schema_class(action)
   def index
     if has_one_assocation?
       #generate the etag here as we use the index route for linked media has_one relations
-      headers['ETag'] = gen_etag(controlled_resources)
+      headers['ETag'] = gen_etag(controlled_resources.to_a)
```
camallen, line 20: We have to be careful about converting a large number of db scope records to an array here as the scope will not have any pagination modifiers added until it's fed through the serializer code, thus the .to_a may result in a large number of objects being pulled from db and allocated in ruby land. 

TODO: check if the `has_one_association?` is a self limiting scope modifier and thus ensures we won't have problems with large scope array conversions.

app/controllers/concerns/medium_response.rb
```diff
@@ -1,7 +1,7 @@
 module MediumResponse
   def medium_response(medium)
     scope = Medium.where(id: medium)
-    headers['ETag'] = gen_etag(scope)
+    headers['ETag'] = gen_etag(scope.to_a)
```
camallen, line 4: Check we don't use this on index routes with the ability to convert large scopes to arrays

app/controllers/api/v1/media_controller.rb
camallen: this is safe as the `has_one` association constraint means we can only load 1 media resource into an array.

app/controllers/concerns/medium_response.rb
camallen: this is safe as the `medium_response` is only used for export resources - i've renamed this module / method for clairty

camallen:
https://github.com/zooniverse/panoptes/pull/3875#issuecomment-1161864752

For clarity - these controller params used in send operations are constrained via our custom routes DSL for specifying the allowed `media_name` types. https://github.com/zooniverse/panoptes/blob/1c81e8c5eea5c4626dca64ca7718d066e68ed67a/config/routes.rb#L83
and 
https://github.com/zooniverse/panoptes/blob/1c81e8c5eea5c4626dca64ca7718d066e68ed67a/lib/routes/json_api_routes.rb#L34

Thus these params are controlled method args and are safe to use with the `Object.send` method. 

Other solutions are welcome here too :)

lib/json_api_controller/precondition_check.rb
```diff
@@ -28,9 +28,20 @@ def precondition_error_msg
...
-      current_etag = gen_etag(query)
+      # convert the query scope to an array of active record objects (vs an ActiveRecord::Relation scope)
+      # to ensure we generate the cache key based on the underlying Active Record objects
+      # and thus have a stable cache key
+      #
+      # Note: the original SHOW scope query may differ slighly from the UPDATE scope query
+      # and thus the resulting query cache key is different and we get a different etag value here
+      # vs the provided one.
+      #
+      # https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-cache_key
```
camallen, line 39:
```suggestion
      # https://api.rubyonrails.org/v5.0/classes/ActiveRecord/Integration.html#method-i-cache_key
```

lib/json_api_controller/precondition_check.rb
```diff
@@ -28,9 +28,20 @@ def precondition_error_msg
...
+      # as opposed to the ActiveRecord::Relation which depends on the query sql
+      # https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-cache_key
```
camallen, line 41:
```suggestion
      # https://api.rubyonrails.org/v5.0/classes/ActiveRecord/Relation.html#method-i-cache_key
```

yuenmichelle1:
I didnt notice this before but it looks like there are 2 new tests that are failing (I'm assuming that are due to these changes). These new tests do not look like they were part of the original failing batch of failing tests lookin back at this PR https://github.com/zooniverse/panoptes/pull/3374 and are currently not failing in master branch.

camallen:
> I didnt notice this before but it looks like there are 2 new tests that are failing (I'm assuming that are due to these changes). These new tests do not look like they were part of the original failing batch of failing tests lookin back at this PR #3374 and are currently not failing in master branch.

I'll look into these but i think they are unrelated and are intermittent failures by the look. If they are unrelated we can merge this PR and fix these in another PR

PR #3877, opened 2022-06-21
ensure we use strings param keys

camallen:
change controller request specs to use string keys as symbol keys aren't parsed correctly

# Review checklist

### DIFF
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -46,7 +46,7 @@ describe TokensController, type: :controller do
       end
 
       context "when supplying valid user credentials" do
-        let(:valid_creds) { params.merge!(login: owner.login, password: owner.password) }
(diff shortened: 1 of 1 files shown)


spec/controllers/tokens_controller_spec.rb
```diff
@@ -46,7 +46,7 @@
       end
 
       context "when supplying valid user credentials" do
-        let(:valid_creds) { params.merge!(login: owner.login, password: owner.password) }
+        let(:valid_creds) { params.merge!('login' => owner.login, 'password' => owner.password) }
```
yuenmichelle1, line 49: not blocking but do we also want https://github.com/zooniverse/panoptes/blob/1c81e8c5eea5c4626dca64ca7718d066e68ed67a/spec/controllers/tokens_controller_spec.rb#L30 to also have string keys ?

camallen: good catch - this would effectively not supply credentials so the test works. I'll update to make sure it properly tests not allowing invalid credentials

PR #3882, opened 2022-06-23
Pg selection spec rails5

camallen:
This PR fixes the postgresql available SetMemberSubject Active Record query where the bind params are incorrectly assigned. I tried a few solutions and left notes as comments in the code. 

We've got a rails 4 / 5 conditional branch here to keep the single DB round trip for rails 4 but allow the rails 5 version to use the 2 db round trips but maintain correctness. 

Longer term rails 5.1 may fix this and we can switch back to the single db round trip code that works in rails 4.2

# Review checklist

### DIFF
--- a/lib/subjects/postgresql_selection.rb
+++ b/lib/subjects/postgresql_selection.rb
@@ -23,13 +23,46 @@ module Subjects
     def available
       query = Subjects::SetMemberSubjectSelector.new(workflow, user).set_member_subjects
       subject_set_ids = if workflow.grouped
-                            # respect the user if they want to select from a training set
-                            opts[:subject_set_id]
-                          else
-                            # default mode: do not select from training sets
-                            workflow.non_training_subject_sets.pluck(:id)
-                          end
-      query.where(subject_set_id: subject_set_ids)
+                          # respect the user if they want to select from a training set
+                          Array.wrap(opts[:subject_set_id])
+                        else
+                          # default mode: do not select from training sets
+                          workflow.non_training_subject_sets.pluck(:id)
+                        end
+      # Handle a bug when the rails 5.0 AR scope bind params merge in rails 5.0 (works find in 4.2)
+      #
+      # failed fix 1 - ensure the subject_set_id param clause is before the id clause
+      #
+      # SetMemberSubject.where(subject_set_id: subject_set_ids, id: query).select(:id)
+      #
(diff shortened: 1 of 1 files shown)


lib/subjects/postgresql_selection.rb
```diff
@@ -23,13 +23,50 @@ def selection_strategy
...
-                            workflow.non_training_subject_sets.pluck(:id)
-                          end
-      query.where(subject_set_id: subject_set_ids)
+                          # respect the user if they want to select from a training set
+                          Array.wrap(opts[:subject_set_id])
+                        else
+                          # default mode: do not select from training sets
+                          workflow.non_training_subject_sets.pluck(:id)
+                        end
+      if Gem::Version.new(Rails.version) > Gem::Version.new('5.0')
```
camallen, line 32: we could look at only running this code for rails 5.0 and allow 5.1 to run the old code?
```suggestion
      if Gem::Version.new(Rails.version) > Gem::Version.new('5.0') && Gem::Version.new(Rails.version) < Gem::Version.new('5.1')
```

lib/subjects/postgresql_selection.rb
```diff
@@ -23,13 +23,50 @@ def selection_strategy
...
-                          end
-      query.where(subject_set_id: subject_set_ids)
+                          # respect the user if they want to select from a training set
+                          Array.wrap(opts[:subject_set_id])
+                        else
+                          # default mode: do not select from training sets
+                          workflow.non_training_subject_sets.pluck(:id)
+                        end
+      if Gem::Version.new(Rails.version) > Gem::Version.new('5.0') && Gem::Version.new(Rails.version) < Gem::Version.new('5.1')
+        # Handle a bug when the rails 5.0 AR scope bind params merge in rails 5.0 (works find in 4.2)
```
yuenmichelle1, line 33:
```suggestion
        # Handle a bug when the rails 5.0 AR scope bind params merge in rails 5.0 (works fine in 4.2)
```

yuenmichelle1 (approved):
Yikes, it was definitely interesting to see the failed fixes. Do we want to put a TODO comment or create some sort of reminder of this so we can remove this when we bump to 5.1+?

camallen:
> Yikes, it was definitely interesting to see the failed fixes. Do we want to put a TODO comment or create some sort of reminder of this so we can remove this when we bump to 5.1+?

The gem version ranges should reactivate the old code for rails 5.1, so this 'fix' should only be around for the rails 5.0 version

PR #3884, opened 2022-06-28
Rails Next canary deployment

zwolf:
Adds a new Action that builds and pushes the docker image, then deploys it. 

The build takes `file: Dockerfile.rails-next` to specify the image should be built with Rails 5. 

The push takes `tags: ${{ github.sha }}-next,latest-next` to (🤞) differentiate between the two different images being built off of the same github SHA. (This could be matrix'd in future but I wanted this first go to be very obvious). 

The canary template includes the `-next` in the image tag, so it deploys the rails 5 image in the canary deployment. The canary ingress is pointed at the canary staging app, and shares the `panoptes-staging.zooniverse.org` host with the regular staging ingress. However, the canary ingress includes the annotations that will match and route traffic that includes the `Canary-Testing-Opt-In: always` header. 

also:
* removes redundant azure env vars
* removes the "azure" part of all the canary names
* updates the beta ingress annotations

# Review checklist


camallen (approved):
LGTM - I think https://github.com/zooniverse/panoptes/pull/3885 will have to come in for the app to build / boot correctly for staging canary.

Also I'd probably have split the staging update into 1 PR and followed up with a production one post successful deploy of staging canary setup (and ingress changes). 

But let's :shipit: and see how this all works and then iterate on any required changes.

zwolf:
I'll wait for #3885, but it looks like there's also failing tests. I reran them a few times in case they were intermittent but it's the same tests with the same values.

zwolf:
Reran the tests a bunch of times, no luck on getting them to pass. Looks like these failing specs are addressed in https://github.com/zooniverse/panoptes/pull/3882, so this can merge after that and 3885 are.

PR #3894, opened 2022-07-05
update flipper gem

camallen:
Updates flipper from v0.17.2 to v0.24.1

Need to change the deprecated API, intializers and memoizing middleware setup (automatic now) in line with the latest releases, see https://github.com/jnunemaker/flipper/blob/master/Changelog.md#upgrading for details. 

### TODO
- [x] Update the rails 4 lockfile to use this gem version as well
- [x] Ensure the flipper UI system works as expected

#### Flipper features to remove
- [x] `cellect_controller_caching` - https://github.com/zooniverse/panoptes/pull/3898

# Review checklist


spec/spec_helper.rb
```diff
@@ -32,6 +32,8 @@
 
   config.filter_run_excluding disabled: true
 
+  config.raise_errors_for_deprecations!
```
camallen, line 35: Remove this once we've surfaced all the deprecations

yuenmichelle1 (approved):
LGTM. it definitely will be excited not to see the `Flipper::Middleware::Memoizer appears to be running twice. Read how to resolve this at https://github.com/jnunemaker/flipper/pull/523` messages anymore

PR #3902, opened 2022-07-07
update sprockets and autofixer gems and update bundler in /docs

yuenmichelle1:
After rack update `bundle exec middleman build --clean` errored due to some scss rendering issues with middleman-sprockets. 
See : https://github.com/slatedocs/slate/issues/1197
~~update~~ downgraded middleman-sprockets gem  and udpated middleman-autoprefix

# Review checklist


docs/Gemfile
```diff
@@ -3,8 +3,8 @@ source 'https://rubygems.org'
 
 # Middleman
 gem 'middleman', '~>4.2'
-gem 'middleman-autoprefixer', '~> 2.7'
-gem 'middleman-sprockets', '~> 4.1'
+gem 'middleman-autoprefixer', '~> 3.0'
+gem 'middleman-sprockets', '~> 4.0'
```
camallen, line 7: Not blocking - Ideally we only fix the gems we need to and remove all these version constraints, thus allowing bunder to determine which set of dependencies are allowed (and easily update them). 

If we did this and lock `gem 'middleman-sprockets', '~> 4.0'` then autoprefixer could have updated without intervention on our end.

camallen (approved):
LGTM - the state of sass and ruby looks a bit of a crap shoot

PR #3903, opened 2022-07-07
update deprecated testhelpers to test::controllerhelpers

yuenmichelle1:
Update for Deprecated `Devise::TestHelpers`
`DEPRECATION WARNING: [Devise] including `Devise::TestHelpers` is deprecated and will be removed from Devise.
For controller tests, please include `Devise::Test::ControllerHelpers` instead.`
# Review checklist


spec/spec_helper.rb
```diff
@@ -8,14 +8,16 @@
 require "rspec/rails"
 require "sidekiq/testing"
 require 'flipper/adapters/memory'
+require 'mock_redis'
+require 'database_cleaner'
```
camallen, line 12: OOI - were these new includes needed for this PR to work? They should be available via the bundler setup but leave them in if it's breaks without them.

camallen (approved):
LGTM

spec/spec_helper.rb
yuenmichelle1: Oh! I did need this locally, but noticed that tests were running in GH action without needing this. I'll remove.

PR #3906, opened 2022-07-08
before_filter replaced with before_action

yuenmichelle1:
`DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead.`
`DEPRECATION WARNING: skip_before_filter is deprecated and will be removed in Rails 5.1. Use skip_before_action instead.`
`DEPRECATION WARNING: prepend_before_filter is deprecated and will be removed in Rails 5.1. Use prepend_before_action instead.`
`DEPRECATION WARNING: after_filter is deprecated and will be removed in Rails 5.1. Use after_action instead.`

# Review checklist

### DIFF
--- a/app/controllers/api/v1/field_guides_controller.rb
+++ b/app/controllers/api/v1/field_guides_controller.rb
@@ -8,7 +8,7 @@ class Api::V1::FieldGuidesController < Api::ApiController
 
   schema_type :json_schema
 
-  before_filter :set_language_if_missing, only: [:index]
+  before_action :set_language_if_missing, only: [:index]
 
   private
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,5 +1,6 @@
 class SessionsController < Devise::SessionsController
-  skip_before_action :verify_authenticity_token, only: [:destroy], if: :json_request?
+  skip_before_action :verify_authenticity_token, 
+    if: -> { json_request? && action_name == "destroy" }
   after_action :set_csrf_headers, only: %i[create destroy]
   after_action :set_csrf_headers, only: :new, if: :json_request?
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,7 +1,7 @@
 class SessionsController < Devise::SessionsController
   skip_before_action :verify_authenticity_token, only: [:destroy], if: :json_request?
-  after_filter :set_csrf_headers, only: [:create, :destroy]
-  after_filter :set_csrf_headers, only: :new, if: :json_request?
+  after_action :set_csrf_headers, only: [:create, :destroy]
+  after_action :set_csrf_headers, only: :new, if: :json_request?
 
   def new
     respond_to do |format|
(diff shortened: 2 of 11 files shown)


app/controllers/api/v1/field_guides_controller.rb
```diff
@@ -8,7 +8,7 @@ class Api::V1::FieldGuidesController < Api::ApiController
 
   schema_type :json_schema
 
-  before_filter :set_language_if_missing, only: [:index]
+  before_action :set_language_if_missing, only: [:index]
```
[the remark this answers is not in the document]
yuenmichelle1: https://github.com/rubocop/rubocop-rails/issues/343 I ran into this as well. I think this sniff can be "ignored" since `index` is a known controller method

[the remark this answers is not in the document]
camallen: Feel free to 'resolve' any comments you don't want to fix. Alternatively if we don't want this cop enforced at all then we can globally turn it off.

app/controllers/sessions_controller.rb
```diff
@@ -1,7 +1,7 @@
 class SessionsController < Devise::SessionsController
-  skip_before_filter :verify_authenticity_token, only: [:destroy], if: :json_request?
-  after_filter :set_csrf_headers, only: [:create, :destroy]
-  after_filter :set_csrf_headers, only: :new, if: :json_request?
+  skip_before_action :verify_authenticity_token, if: -> { json_request? && action_name == 'destroy' }
```
camallen, line 2: OOI - did the `only: [:destroy]` option change as the API for these callbacks has changed? I.e. we inlined the only option as outlined in https://api.rubyonrails.org/v5.1.0/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_normalize_callback_options

app/controllers/sessions_controller.rb
```diff
@@ -1,7 +1,7 @@
...
+  after_action :set_csrf_headers, only: %i[create destroy]
+  after_action :set_csrf_headers, only: :new, if: :json_request?
```
camallen, line 4: do we have to inline the `only: :new` to the if block param as well here?

camallen (approved):
LGTM - i'm not sure exactly why the callback options have changed to inline in a `skip_before_filter` but not the `before_action`. It's good to have consistent use of these callbacks.

app/controllers/sessions_controller.rb
yuenmichelle1: I believe `only: [:destroy]` is still valid syntax. Though per https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/IgnoredSkipActionFilterOption, `if` option will be ignored.

yuenmichelle1: Actually do we want `if` to be ignored? if so, we can ignore hound sniff and I can revert it back.

app/controllers/sessions_controller.rb
camallen: > We can change if we want to keep consistent

Yes please - consistency of how we are using these dsl actions is desired

app/controllers/sessions_controller.rb
camallen: no my reading of the original syntax is we want this this action to be skipped only if the actions is a destroy action and only if it's a json request so we want both checks to run. 

What this skip action is doing is avoiding rails in built CSRF checks / errors for json API calls on logout actions, https://github.com/zooniverse/panoptes/pull/1424

PR #3910, opened 2022-07-14
Remove 0 replicas limit

zwolf:
Removing this line re-enables deployment of the production canary.

# Review checklist

### DIFF
--- a/kubernetes/deployment-production-canary.tmpl
+++ b/kubernetes/deployment-production-canary.tmpl
@@ -5,6 +5,7 @@ metadata:
   labels:
     app: panoptes-production-canary-app
 spec:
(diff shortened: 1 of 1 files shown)


kubernetes/deployment-production-canary.tmpl
```diff
@@ -5,7 +5,6 @@ metadata:
   labels:
     app: panoptes-production-canary-app
 spec:
-  replicas: 0
```
camallen, line 8: Preferably we set this as `replicas: 1` instead of removing so it's clear 1 is running and easy to set it back to 0 when we don't need it.

camallen (approved):

PR #3911, opened 2022-07-18
Prep staging upgrade to 5.0

yuenmichelle1:
# Review checklist

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -7,6 +7,14 @@ on:
   workflow_dispatch:
 
 jobs:
+  build_and_push_image:
+      name: Build and Push Image
+      uses: zooniverse/ci-cd/.github/workflows/build_and_push_image.yaml@main
+      with:
+        repo_name: panoptes
+        commit_id: ${{ github.sha }}
+        latest: true
+  
   deploy_production:
     name: Deploy to Production
     uses: zooniverse/ci-cd/.github/workflows/deploy_app.yaml@main
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -13,7 +13,6 @@ jobs:
       with:
         repo_name: panoptes
         commit_id: ${{ github.sha }}
-        latest: true
   
   deploy_production:
     name: Deploy to Production
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -13,6 +13,7 @@ jobs:
     with:
       repo_name: panoptes
       commit_id: ${{ github.sha }}
+      file: Dockerfile.rails-next
       latest: true
 
   db_migration_staging:
--- a/kubernetes/deployment-staging.tmpl
+++ b/kubernetes/deployment-staging.tmpl
@@ -126,7 +126,7 @@ spec:
     spec:
       containers:
         - name: panoptes-staging-app
-          image: ghcr.io/zooniverse/panoptes:__IMAGE_TAG__-next
+          image: ghcr.io/zooniverse/panoptes:__IMAGE_TAG__
           resources:
             requests:
               memory: "500Mi"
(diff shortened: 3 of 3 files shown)


camallen (commented):
If we deploying from the `next` image that is built by the staging canary deploy action we run link the canary action to this deploy action and that carries race condition risks of coupling between this action running before the canary build action event (thus attempts to deploy an image that doesn't exist etc).

Rather we want the staging actions to build their own image based on the rails 5 docker file and deploy that one without coupling to the canary deploy (that might be rails 6 at some point)

Rather we can add the specific dockerfile to staging deploy build step via the use of a custom input `file: Dockerfile.rails-next` to the staging build and push action at https://github.com/zooniverse/panoptes/blob/040394c309db79d08a5ca9b87f1e61ceebf76e7c/.github/workflows/deploy_staging.yml#L13-L16

I've noted we need another input in our custom resusable ci/cd action and added it in https://github.com/zooniverse/ci-cd/pull/31. That PR will need to be merged and then we can switch these around. 

Using that singular change means we don't have to edit the k8s deploy templates for staging, rather we just build the underlying image slightly differently (rails 5 vs 4 dockerfile).

zwolf:
Production deploys use the image built during the staging deploy in order to not pay the cost of building the same thing twice for no reason. So that build step would need to be moved to the production deploy action, meaning further deviating from our existing process.

We talked about the race condition you describe yesterday. If the staging deploy action fails because the rails 5 image hasn't finished ('tag not found'), it can be rerun via the UI. This doesn't seem too onerous to me while keeping the current process intact, which means less reversion later. Having our staging and production deploys be so fundamentally different is a short term requirement for this kind of project, but it shouldn't need to be kept like this for long and we shouldn't pay the cost of building the same thing twice in perpetuity to support this.

I can see the problem with building into _this_ process that kind of race condition, though, so I'll leave it up to @yuenmichelle1 to decide what makes the most sense.

camallen:
> if we update the deploy-staging github workflow, what would need to be updated in deploy-production workflow for this to not affect our production deploys?

You'd need to build the production image independently from the staging image which is how we used to do it (used build caching to speed it up). Specifically you'd need to add an image build and push job before the deploy job in https://github.com/zooniverse/panoptes/blob/040394c309db79d08a5ca9b87f1e61ceebf76e7c/.github/workflows/deploy_production.yml#L10

```
# Mimic the staging setup (without latest tag) in https://github.com/zooniverse/panoptes/blob/040394c309db79d08a5ca9b87f1e61ceebf76e7c/.github/workflows/deploy_staging.yml#L10-L16
build_and_push_image:
    name: Build and Push Image
    uses: zooniverse/ci-cd/.github/workflows/build_and_push_image.yaml@main
    with:
      repo_name: panoptes
      commit_id: ${{ github.sha }}
```

Personally i don't mind which way you approach this. To me, adding a known race condition / failure mode to a deploy tool is not a good approach as it's introducing a known bug that may break deploys and then alerts folks who will have to pay some attention to it / most likely start ignoring over time - that seems like a bad approach to CD tooling.

yuenmichelle1:
> it's introducing a known bug that may break deploys and then alerts folks who will have to pay some attention to it / most likely start ignoring over time

^ thats a fair point. it looks like https://github.com/zooniverse/ci-cd/pull/31 has been merged so i will update the deploy staging and deploy production github action

.github/workflows/deploy_production.yml
```diff
@@ -7,6 +7,14 @@ on:
   workflow_dispatch:
 
 jobs:
+  build_and_push_image:
+      name: Build and Push Image
+      uses: zooniverse/ci-cd/.github/workflows/build_and_push_image.yaml@main
+      with:
+        repo_name: panoptes
+        commit_id: ${{ github.sha }}
+        latest: true
```
camallen, line 16: Don't set the latest on production builds - this should be reserved for master branch builds. Instead rely on the GH sha id.
```suggestion
```

camallen (approved):

PR #3912, opened 2022-07-21
isolate rails 5 staging deploy from prod deploys

camallen:
Follow up to #3911 - after more reflection on the staging rails 5 deploy changes I believe we introduced another possible deployment bug in #3911.

Re using the GH sha id after a production deploy would build a new image and update the container registry at that tag (prod would have a rails 4 gemfile build). The means that the existing staging deploy at the same sha id would be overwritten and would not include the rails 5 gem versions built for the staging deploy, rather it would have the production rails 4 versions :(

This would only happen if the production tag and the GH master branch head were the same (i.e. no changes on master) and the staging system redeployed to a node without the existing rails 5 staging image on it, thus it would pull down the prod rails 4 version and boot staging mode up. 

While this may not happen as the deploys are stable we run the risk of our staging testing env 'switching' out from under us which is not what we want. 

To avoid this scenario, this PR adds a tag suffix to the staging image builds and when we deploy production we don't overwrite the previously built staging rails 5 image tag. Specifically we add a tag suffix `-next` to the built staging images to isolate the staging rails 5 deploys from the prod rails 4 deploys.

Finally this PR also removes the staging canary setup from deploying / running as we no longer need this canary while we have the staging system running on rails next v5 version. I can extract this to it's own PR if folks want. 

# Review checklist

### DIFF
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -12,7 +12,7 @@ jobs:
     uses: zooniverse/ci-cd/.github/workflows/build_and_push_image.yaml@main
     with:
       repo_name: panoptes
-      commit_id: ${{ github.sha }}
+      commit_id: ${{ github.sha }}-next
       file: Dockerfile.rails-next
       latest: true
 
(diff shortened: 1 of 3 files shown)


yuenmichelle1 (approved):
LGTM

PR #3921, opened 2022-08-17
Update to Rails 5.0

yuenmichelle1:
Prep Production to Rails 5.0 Upgrade

- Update Gemfile to use rails 5.0 (and updated Gemfile.lock)
- Update Dockerfile to use ruby 2.6

# Review checklist


.github/workflows/run_tests_CI.yml
```diff
@@ -27,18 +27,13 @@ jobs:
           - Gemfile
           - Gemfile.next
         ruby:
-          - 2.5
+          - 2.6
```
camallen, line 30: you only need one of these

Dockerfile
```diff
@@ -1,5 +1,4 @@
-# debian stretch has libjemalloc1 https://packages.debian.org/stretch/libjemalloc1
-FROM ruby:2.5-slim-stretch
+FROM ruby:2.6-slim-buster
```
camallen, line 1: as noted in the comment - we want to keep jemalloc v1 around and that package is available under the debian stretch version 

jemalloc is used for controlling ram growth / memory allocations for our app (especially our worker nodes). 

I'd suggest not removing this and instead using the 2.6 stretch image https://hub.docker.com/_/ruby/tags?page=1&name=2.6-slim

related jemalloc PR https://github.com/zooniverse/panoptes/pull/3088

Dockerfile
```diff
@@ -9,14 +8,12 @@ RUN apt-get update && apt-get -y upgrade && \
         # git is required for installing gems from git repos
         git \
         # libjemalloc1 (v3) provides big memory savings vs jemalloc v5+ (default on debian buster)
-        libjemalloc1 \
         libpq-dev \
-        nodejs \
+        postgresql-client-11 \
```
camallen, line 12: Wondering why we need the explicit client v11 here vs the libpq-dev package.. What is the difference between libpq-dev and the new pg client? 

Also why did we drop node support...isn't that needed for the asset precompilation? will this image build correctly without it?

Dockerfile
```diff
@@ -9,14 +8,12 @@ RUN apt-get update && apt-get -y upgrade && \
...
         tmpreaper \
         && \
         apt-get clean
 
-ENV LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so.1
```
camallen, line 19: this is for dynamically switching the ruby memory allocator at run time - if we switch back to jemalloc v1 we will need this back.

Gemfile
```diff
@@ -45,11 +45,7 @@ gem 'pg_search'
 gem 'puma', '~> 5.6.4'
 gem 'pundit', '~> 2.2.0'
 gem 'rack-cors', '~> 1.0', require: 'rack/cors'
-if next?
```
camallen, line 48: don't remove the `next?` branch - we'll need this to keep the application dual bootable for the next upgrade. Instead just update the gem versions inside the brances, e.g. 
```
if next?
 gem 'rails', '~> 5.0'
else
 gem 'rails', '~> 5.0'
end

camallen (commented):
the production dockerfile is different to the dev ones, specifically around the `jemalloc` memory allocator stuff. 

We can switch this to a new version or try a different approach via MALLOC_ARENA_MAX for default one (see reading links in https://github.com/zooniverse/panoptes/pull/3088) but i think that's a change we want to do independent of this ruby and rails version upgrades. 

Also we need to check the nodejs and pg client differences as 
1. node is needed for the asset compilation step
2. pg client is normally only needed for dev setup (psql access / dumping schemas etc)

So i think this PR needs another cycle of work to get it ready for deployment. FWIW Ideally we get this PR merged and then switch the staging setup backup to work on this dockerfile (not the next one) so we test it's working before trying to deploy to production.

yuenmichelle1:
> So i think this PR needs another cycle of work to get it ready for deployment. FWIW Ideally we get this PR merged and then switch the staging setup backup to work on this dockerfile (not the next one) so we test it's working before trying to deploy to production.

I think I see what you're saying. but just to be sure. initially my thought was to mimic Dockerfile.rails-next into Dockerfile to get rails 5 upgrade up, but that's off base. In this case, it sounds like all we need to update within Dockerfile is the ruby version to 2.6-slim-stretch.(2.5-slim-stretch would have an issue on install step ` E: Unable to locate package postgresql-client-11`) 

Once this PR is OK'd, in a separate PR, we can update `deploy_staging.yml` (remove file input and remove -next from commit ids)

Let me know if I'm off base.

camallen:
> In this case, it sounds like all we need to update within Dockerfile is the ruby version to 2.6-slim-stretch.(2.5-slim-stretch would have an issue on install step E: Unable to locate package postgresql-client-11)

Yep - as it's a staging / production image we don't need the `postgresql-client-11` package installed, that's only for dev work like importing the schema etc.

zwolf (approved):
:shipit: 

edit: w/ the caveat that we're gonna update the staging action and deploy with the same files & test first, then deploy prod.

PR #3934, opened 2022-08-25
downgrade sidekiq and redis in order for sidetiq to enqueue jobs

yuenmichelle1:
redis >4.1 .exists method returns an integer instead of a boolean which is an issue for sidetiq since it checks exists boolean to enqueue jobs. (if integer, sidetiq never enqueues job)
See more info on Slack thread here : https://zooniverse.slack.com/archives/C0DTP3L2K/p1661452113456909?thread_ts=1661411214.673549&cid=C0DTP3L2K

# Review checklist


zwolf (approved):
Another option may be to revert to the old behavior via a flag in the initializer: 

https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#450

> Redis#exists now returns an Integer by default, as warned since 4.2.0. The old behavior can be restored with Redis.exists_returns_integer = false.

I'm fine with this route, though, with the understanding that we've gotta replace sidetiq ASAP.

PR #3938, opened 2022-08-29
Replace Sidetiq with Sidekiq-cron

yuenmichelle1:
# Review checklist


spec/support/schedulable.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+shared_examples 'is schedulable' do
+  let(:job) { Sidekiq::Cron::Job.new(name: 'cron_job_name', cron: cron_sched, class: class_name) }
+
+  it 'gets queued on enqueued_times' do
+    enqueued_times.each { |enqueued_time|
```
zwolf, line 7: Mixing block syntax here is kind of awkward, I suggest using `do/end` here

zwolf (approved):
lgtm! :shipit:

PR #3940, opened 2022-09-01
Update Gemfile.next with Rails 5.1

yuenmichelle1:
The original plan was to take care of the first two bullet points and then triage out the remaining tests. After taking care of first 2 bullet points, there were about 10 or so remaining failing tests, which didnt feel like a lot to triage. 

Related reading articles: 
https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-5-0-to-5-1.html
https://eileencodes.com/posts/actioncontroller-parameters-now-returns-an-object-instead-of-a-hash/
https://hackernoon.com/how-i-upgraded-rails-to-version-515-and-lived-to-tell-the-tale-2fhy30sn
https://www.bigbinary.com/blog/controller-actions-default-no-content-in-rails-5-if-template-is-missing
https://stackoverflow.com/questions/4632271/render-nothing-true-returns-empty-plaintext-file
https://stackoverflow.com/questions/36175098/post-via-redirect-is-deprecated-and-will-be-removed-in-rails-5-1

### Changes

- Controller Tests:  `params` and `headers` keys are now needed in Rails 5.1. See: https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-5-0-to-5-1.html (Point 6 in article). **_[This was the biggest change that touches many files  ]_**
- `ActionController::Parameters` Now returns `Object` instead of a `HashWithIndifferentObject` (See: https://eileencodes.com/posts/actioncontroller-parameters-now-returns-an-object-instead-of-a-hash/). This affects `resource_ids.rb` as well as `updatable_resource.rb`, `projects_controller.rb#copy`
- Default response in Rails 5 is `204 No Content` (See: https://www.bigbinary.com/blog/controller-actions-default-no-content-in-rails-5-if-template-is-missing and https://stackoverflow.com/questions/4632271/render-nothing-true-returns-empty-plaintext-file). `head` is no preferred over `render: nothing`.  This affects changes in `workflows_controller.rb` within `retire_subjects` and `unretire_subjects`, `check_resources_exist_spec.rb`, `updatable_resoure_spec.rb`
- `post_via_redirect` has been removed in 5.1. Affects `devise_support.rb`
- ActiveRecord should not been passed into `.find`. We need to send `id` vs record with `.find` method. Affects `user_added_to_project_mailer_worker_spec.rb`, 'project_request_mailer_spec.rb`
- Removed tests related to unused http_cacheable; these tests only had setup data but no examples/assertions. 
- https://stackoverflow.com/questions/12598435/update-all-or-update-attribute-doesnt-update-the-column Update weird hack on `project_spec.rb` instead of inspecting `full_project.active_workflows`  with `.inspect` to get workflows to be loaded, we reload the workflows.

# Review checklist


spec/lib/formatter/csv/subject_spec.rb
```diff
@@ -93,7 +93,7 @@
 
     context "with a subject that has no location metadata" do
       it "should match the db ordered subject_locations array" do
-        allow_any_instance_of(Medium::ActiveRecord_Associations_CollectionProxy)
+        allow_any_instance_of(Medium.const_get('ActiveRecord_Associations_CollectionProxy'))
```
camallen, line 96: this seems strange that you have to access the collection proxy through a const get lookup... do you have any details on this?

spec/models/project_spec.rb
```diff
@@ -337,14 +337,14 @@
...
-      full_project.active_workflows.inspect
+      #update_all sends an update directly to the database - it won't update the report models you have in memory, so we need to reload models
+      full_project.active_workflows.map(&:reload)
       expect(full_project.active_workflows)
-        .to receive(:inject)
-        .and_call_original
-
+      .to receive(:inject)
+      .and_call_original
+
```
camallen, line 347:
```suggestion

```

spec/serializers/user_serializer_spec.rb
```diff
@@ -73,6 +73,8 @@
 
   describe "private user attributes" do
     let(:result) do
+      # save user so that User.where/User.find can find zooniverse id
+      user.save
```
camallen, line 77: user should have been created (`create(:user)`) by the factory and should be persisted already (have an id for lookup). this seems strange.

camallen (approved):
i think we can merge this an iterate...my brain is melted reviewing a PR this big...probably not my best review work.

spec/lib/formatter/csv/subject_spec.rb
yuenmichelle1: I think it has to do with this https://github.com/rails/rails/blob/v5.1.7/activerecord/lib/active_record/relation/delegation.rb#L8-L24
we cannot access `ActiveRecord_Associations_CollectionProxy` in the way that we did

camallen: can you use this syntax `Medium::CollectionProxy` instead of the `const_get` one? It seems to resolve to the same AR proxy class.

lib/subjects/postgresql_selection.rb
```diff
@@ -29,44 +29,8 @@ def available
...
-        #
-        # fixes the subject set id bind param but still fails on the limit bind param later on :(
 
-        # Working solution
-        #
-        # run two queries
-        # one with with a pluck on the query select to pull the avaialble SMS IDs
-        # and then apply the subject set id filter to create a new simple scope
-        # that we pass onto to the selector strategy class which then applies the limit bind params etc
         SetMemberSubject.where(id: query.pluck(:id), subject_set_id: subject_set_ids).select(:id)
```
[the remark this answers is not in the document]
yuenmichelle1: Unfortunately, i tried the other solutions mentioned and it still fails in 5.1. we could keep what we had but update the check to run `if Gem::Version.new(Rails.version) > Gem::Version.new('5.0') && Gem::Version.new(Rails.version) < Gem::Version.new('5.2')`.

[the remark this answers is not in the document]
camallen: cool - that makes sense then - perhaps leave these comments in and the Gem Version comparison switch to remind us to try this later...

fwiw - i looked into this again and i think we're using too many layers of query merging. TBH this code could do with a refactor but that's a lot of work and using 2 trips here simplifies the resulting query and our understanding of it so that's probably a win for us maintaining this code.  

As you were :) ii

spec/serializers/user_serializer_spec.rb
yuenmichelle1: This one was weird. We had a failing test https://github.com/zooniverse/panoptes/blob/master/spec/serializers/user_serializer_spec.rb#L87-L94 where `result[:zooniverse_id]` is nil. 

This is because https://github.com/zooniverse/panoptes/blob/e98eb1a80f247fbef465cd1e952b755eb6ef2071/spec/serializers/user_serializer_spec.rb#L76 we call `User.where(id: user.id)`. 

**Weird behavior is that when we either `user.reload`, `User.find(user.id)`, or `User.where(id:user.id)`, the result of `user.zooniverse_id` is `nil`. _BUT_ before the `user` is reloaded (or queried), `user` had a `zooniverse_id`  from the `create(:user)` & https://github.com/zooniverse/panoptes/blob/e98eb1a80f247fbef465cd1e952b755eb6ef2071/app/models/user.rb#L58.** 

My guess is that resulting user from the `after_create` is not saved in test.

camallen: this is due to the `after_create` model callback (always a terrible idea to use these IMHO**)

A simple fix would be to set the zooniverse_id to what you expect on create, i.e. on line 6 setup the factory with the `zooniverse_id` already set to a random uniq value so it doesn't do anything on the callback
```
let(:user) { create(:user, migrated: migrated, sign_in_count: sign_in, zooniverse_id: "zoo-id-#{SecureRandom.random_number(100)}") }
```

** callbacks modify the model state without explicit action which you may not expect, id’ rather see this logic extracted out to a service object / operation style class that does this work for you…TBH this whole fix for ouroboros flow is terrible and should have been extracted to a service object from the start for use in the controller

PR #3942, opened 2022-09-06
disable the Gemfile.next CI tests when there is no upgrade happening

camallen:
As we aren't testing any gem upgrades (e.g. rails 5.1) the Gemfile and Gemfile.next are the same - thus the CI tests for Gemfile.next don't need to be run. Lets save on GH compute and test CI runtime by not running these tests till we need them.

When we create a PR to upgrade a gem using the Gemfile.next technique (e.g. rails 5.1 in #3940) we will need to re-enable these CI Gemfile.next test runs by removing this commented line

# Review checklist

### DIFF
--- a/.github/workflows/run_tests_CI.yml
+++ b/.github/workflows/run_tests_CI.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         gemfile:
           - Gemfile
-          - Gemfile.next
+          # - Gemfile.next
         ruby:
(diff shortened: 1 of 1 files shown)


yuenmichelle1 (approved):
>  Lets save on GH compute and test CI runtime by not running these tests till we need them.

Fair point. LGTM after the branch gets updated to be in sync with master.

PR #3944, opened 2022-09-06
Update schedule.yml scheduled workers with the correct queues

yuenmichelle1:
Adding queues to what they were configured to when we were using sidetiq. 

Side by Side comparison: 
Sidetiq: https://panoptes.zooniverse.org/sidekiq/sidetiq
Sidekiq-cron: https://panoptes-staging.zooniverse.org/sidekiq/cron

# Review checklist


camallen (approved):
LGTM

PR #3945, opened 2022-09-06
allow unknown user ids to trigger the congestion rate limiter

camallen:
This PR is a fix for the automated subject set export functionality added in #3714

Specifically the issue is the congestion rate limiter raises an error when running and stops all jobs linked to it. this issue is due to the special internal user id in https://github.com/zooniverse/panoptes/blob/581ad45e99bc88826dba32aa4978533c21a6d8c9/app/operations/subject_sets/run_completion_events.rb#L26-L29

As we use a special internal user (id=-1) to internally trigger the classification so we can't raise and AR not found error, rather in this PR we catch this error and use the special id to setup congestion rate limiter key and carry on with the rate limiter middleware / job. 

# Review checklist

### DIFF
--- a/app/workers/concerns/rate_limit_dump_worker.rb
+++ b/app/workers/concerns/rate_limit_dump_worker.rb
@@ -22,17 +22,17 @@ module RateLimitDumpWorker
   module ClassMethods
     def congestion_enabled?(requester_id)
       # if the user is missing, should only happen via the rails console
-      return false if requester_id.blank?
+      # false if the user is a special skip rate limit user
+      return false if requester_id.blank? || Panoptes::RateLimitDumpWorker.skip_rate_limit_user_ids.include?(requester_id)
 
-      skip_user_ids = Panoptes::RateLimitDumpWorker.skip_rate_limit_user_ids
-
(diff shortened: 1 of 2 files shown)


app/workers/concerns/rate_limit_dump_worker.rb
```diff
@@ -22,17 +22,17 @@ module RateLimitDumpWorker
...
-      if skip_user_ids.include?(requester_id)
-        # false if the user is a special skip rate limit user
-        false
-      else
-        # false if user is admin, true if not admin
-        !User.find(requester_id).is_admin?
-      end
+      user = User.find(requester_id)
+      # false - disable congestion if user is admin
+      # true - emable congestion if they are a normal user
```
yuenmichelle1, line 30:
```suggestion
      # true - enable congestion if they are a normal user
```

yuenmichelle1 (approved):

app/workers/concerns/rate_limit_dump_worker.rb
camallen: thanks :)

PR #3948, opened 2022-09-20
Bump active-model-serializers to 0.10.13

yuenmichelle1:
Related to https://github.com/zooniverse/panoptes/pull/3947/files

Pulling out logic related to dual boot to separate concerns. 

### Changes:
 From AMS 0.10.0rc2 to 0.10.13, serializer.each_asociation as well as serializer.type has been removed. Replaced with .associations and .json_key respectively.
Serializer.associations returns association instance which is an enumerable that can be iterated. .json_key returns in our case ._type defined in each serializer class.

Defining `type` attr for each serializer that extends ActiveModel::Serializer.

Adding fixture based on what was received when tailing kinesis stream and adding appropriate test on classification_serializer_spec.rb to test output of .serialize comes out in the expected format of our kinesis stream.

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ end
 source 'https://rubygems.org'
 
 gem 'active_interaction', '~> 3.8.3'
-gem 'active_model_serializers', '0.10.0.rc2' # Event stream
+gem 'active_model_serializers' # Event stream
 gem 'activerecord-import', '~> 1.4'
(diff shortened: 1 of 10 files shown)


camallen (approved):
LGTM - Amazing 🎉 

:shipit: to staging for testing on the staging kinesis and downstream staging systems 👏🏼

Issue #2492, opened 2017-10-16
Modify project completion metric to use classification counts


camallen:
As per the comment from PFE - https://github.com/zooniverse/Panoptes-Front-End/issues/4093#issuecomment-336896345

Perhaps we should summate the workflow metrics according to the prefs now set instead of using the default retired ratio.

mrniaboc:
I'm absolutely behind this 100%. If I was the kind of person who wrote in all caps, I would be right now.

camallen:
The current calculation doesn't take into account the proportion of each subjects subject counts, instead we just average out the the sum of the active workflow percentages. https://github.com/zooniverse/Panoptes/blob/fddc6e38dd4d1d61f3b767b2b3904d5496286a8a/app/workers/calculate_project_completeness_worker.rb#L42

If i have 2 active workflows:
1. workflow 1 = 1 retired / 5 subjects = 0.2
0. workflow 2 = 9 retired / 10 subjects = 0.9

The current calc is (0.2 + 0.9) / 2 = 0.55, it is actually 10/15 = 0.66. This needs to change to either just the classification counts or we weight the percentages coming from each workflow, e.g. 0.2 of 5/15 and 0.9 of 10/15.

camallen:
Also noting that classification counts may be impacted by early retirement from caesar as well, though it's probably more accurate than what we have.

mrniaboc:
The current set-up continues to confuse researchers and volunteers alike. Got this message from Jules earlier today:

"Just a quick query. The stars on the Solar Stormwatch II homepage has always remained at 0%. It's currently at 37% if you click on the link but not everyone does and a demoralising 0% is all they see. Is this fixable? If not could it be removed just leaving the link? I've had a look to see if the SSW team could do anything but it doesn't look like it."

We have a progress bar by classification count option in the stats page, we really should at lest allow the option for the homepage!

camallen:
Just noting that classifications count can also be problematic if the team changes the retirement limit hald way through a data processing run, https://www.zooniverse.org/talk/17/864027?comment=1426076

mrniaboc:
True, but not as problematic as everyone wondering why projects show 0%
complete all the time even though they have thousands of classifications.

mrniaboc:
This still continues to confuse researchers. Today it was Ed Hawkins, the PI of the Weather Rescue projects. He has released three projects and knows his stuff, but still couldn't figure out why his progress bar was not increasing.

[...]

PR #3950, opened 2022-09-26
switch to proportional project completeness metric

camallen:
linked to #2492 

this PR switches the project completeness metric from the average of all workflow completion metrics to a proportional metric derived from the number of subjects in a workflow. 

```
E.g. for a project with 2 active workflows this new behaviour results in 
workflow_1 = { completeness: 0.5, subjects_count 100, retired_subjects: 50 } 
workflow_2 = { completeness: 0.9, subjects_count 900, retired_subjects: 810 } 

workflow_1.proportional_completeness = 0.5 (100 / 1000) = 0.05
workflow_2.proportional_completeness = 0.9 * (900 / 1000) = 0.81
project.completeness = 0.86 - more accurate

# under the old average metric we would have
(0.5 + 0.9) / 2 = 0.7 
project.completeness = 0.7 - not accurate
```

This is because the average metric doesn't factor the proportion of the subjects that each workflow is responsible for and can return inaccurate values depending on how workflows and subject sets are setup across the project. 

switching to a proportional metric that factors in the number of subjects each workflow has to ensure we have a more accurate completeness metric as we take the subject proportions into account as well. 

Not only that but we cover edge cases as reported on talk https://www.zooniverse.org/talk/17/2614180?comment=4277958 This project had a completed workflow(s) and one training workflow with 0 subjects (training subjects don't count) but a derived completeness metric of 0.5 `((1.0 + 0.0) / 2.0)` where the average calculation failed and thus this project would never 'pause'. 

With this new calculation this project would have a completeness calculation of 1.0 `1.0 + 0.0` where the training workflow has a 0.0 proportional value so no effect on the overall completeness of the project. 

# Review checklist

### DIFF
--- a/app/workers/calculate_project_completeness_worker.rb
+++ b/app/workers/calculate_project_completeness_worker.rb
@@ -42,11 +42,10 @@ class CalculateProjectCompletenessWorker
     # return the sum of the proportional completeness metrics
     project.active_workflows.sum(0.0) do |active_workflow|
       # as each workflow has a proportion of the total subjects
-      # we can use this proportion to weight the each workflow's completeness metric to the total for the project
+      # we can weight each workflow's completeness metric to the total for the project
       subject_proportion = active_workflow.subjects_count.to_d / project_subjects_count
 
-      # the summated proportion metrics will be clamped to the 0.0..1.0 range
-      # by the fraction of subjects / total subjects
+      # the summated proportion metrics will be clamped to the 0.0..1.0 range by subjects / total subjects fraction
       active_workflow.completeness * subject_proportion
     end
   end
--- a/app/workers/calculate_project_completeness_worker.rb
+++ b/app/workers/calculate_project_completeness_worker.rb
@@ -37,10 +37,18 @@ class CalculateProjectCompletenessWorker
   end
 
   def project_completeness
-    return 0.0 if no_active_workflows?
+    return 0.0 if no_active_workflows? || no_linked_subjects?
 
-    completenesses = project.active_workflows.pluck(:completeness)
-    completenesses.sum / completenesses.size.to_f
+    # return the sum of the proportional completeness metrics
+    project.active_workflows.sum(0.0) do |active_workflow|
+      # as each workflow has a proportion of the total subjects
+      # we can use this proportion to weight the each workflow's completeness metric to the total for the project
+      subject_proportion = active_workflow.subjects_count.to_d / project_subjects_count
+
+      # the summated proportion metrics will be clamped to the 0.0..1.0 range
+      # by the fraction of subjects / total subjects
+      active_workflow.completeness * subject_proportion
+    end
   end
 
   def workflow_completeness(workflow)
@@ -50,7 +58,7 @@ class CalculateProjectCompletenessWorker
 
     retired_subjects = workflow.retired_subjects_count
     total_subjects = workflow_subjects_count
-    (0.0..1.0).clamp(retired_subjects / total_subjects.to_f)
+    (0.0..1.0).clamp(retired_subjects / total_subjects.to_d)
   end
 
   def project_columns_to_update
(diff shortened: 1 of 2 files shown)


app/workers/calculate_project_completeness_worker.rb
```diff
@@ -50,7 +57,7 @@ def workflow_completeness(workflow)
 
     retired_subjects = workflow.retired_subjects_count
     total_subjects = workflow_subjects_count
-    (0.0..1.0).clamp(retired_subjects / total_subjects.to_f)
+    (0.0..1.0).clamp(retired_subjects / total_subjects.to_d)
```
yuenmichelle1, line 60: Wanted to double check this, is there ever a case `total_subjects` is ever not a whole number? Are we using `to_d` in this case for precision of the calculation `retired_subjects / total_subjects`?

yuenmichelle1 (approved):
This calculation looks good to me. I had one or two quick question/s on one thing just for my own understanding.

lcjohnso (approved):
Good to see this improvement implemented.  Fixing the bug where training-only workflows would keep projects active is a great win.

However, it is worth noting clearly that the project completion metric, which is visualized via the primary progress bar on the project homepage, is still based on retirement count (not classification count).  There are good technical reasons for this, and I'm not recommending any changes in scope here, but it is worth acknowledging that this falls short of the full aspirations of #2492.

Finally: I'm interested in knowing the answer to `to_d` versus `to_f` usage, and whether the failure of `to_d` for nil/null cases will cause any issues.

app/workers/calculate_project_completeness_worker.rb
camallen: yes - I switched to .to_d `BigDecimal` class to avoid the float precision issues in the resulting calculations and kept it consistent in the file (i.e. avoid mixing .to_f and .to_d)

Details outlined here https://www.rubyguides.com/2016/07/numbers-in-ruby/  in the `Float Imprecision` section.

camallen: > Wanted to double check this, is there ever a case total_subjects is ever not a whole number? 

Workflow.`subjects_count` wraps the db attribute `real_set_member_subjects_count` at https://github.com/zooniverse/panoptes/blob/2fea4d095ca46436d8faac466baa8f732e7bd531/app/models/workflow.rb#L138 and that has a not null default 0 set on the db attribute, https://github.com/zooniverse/panoptes/blob/2fea4d095ca46436d8faac466baa8f732e7bd531/db/structure.sql#L1964

So that value should always be a whole number that is set by the workflow counters as per https://github.com/zooniverse/panoptes/blob/2fea4d095ca46436d8faac466baa8f732e7bd531/app/counters/workflow_counter.rb#L17

camallen:
> but it is worth acknowledging that this falls short of the full aspirations of #2492.

Noted - i haven't set that issue to close on this PR, rather i linked this PR to that issue to ensure it was available in the GH UI for reference. #2492 will still be open with the nuances of how to improve the completion metric.

> Finally: I'm interested in knowing the answer to to_d versus to_f usage, and whether the failure of to_d for nil/null cases will cause any issues.

https://github.com/zooniverse/panoptes/pull/3950#discussion_r981962420 i don't think it should but if `.to_d` proves problematic i'll revisit this to fix to handle the invalid inputs to the calculations

NomadPurple:
Good to see the change.
It did rather oddly bring a random selection of old projects out of Paused onto the Active board. I think they are gradually returning to Paused.
There are five remaining African American Civil War Soldiers, Weather Rescue, Frogsong, NfN Invertebrate Time Machine and Project Plummage.
I will check tomorrow to see if they move to Pause status, off the Active board.

NomadPurple:
I was hoping the overnight process would remove the Paused projects from the Active board but this is not the case.
NfN Labs & Rainfall Rescue have appeared from Pause to Active to Pause to Active at least twice. Both should be in Paused.
Currently Weather Rescue has gone back to Paused......will it reappear?
All these projects have at least 4 (most many more) separate workflows, all are 100% completed and the problem definitely appeared when other active projects moved to the new completion calc.  Could it be a rounding issue?

camallen:
@NomadPurple yep you are correct in that this was a precision issue from using floats. See the solution in https://github.com/zooniverse/panoptes/pull/3954/files with test data from real projects, it adds rounding on the summated value to a known precision (4 digits by default) to avoid the loss of precision.

NomadPurple:
All the projects seem to have made their ways to Pause permanently :)

camallen:
thanks for the follow up @NomadPurple 👏🏼

NomadPurple:
I've been monitoring for a while and now we have projects going into Paused before all the subjects are complete.
Currently Cedar Creek has moved to Paused with 100% completeness metric but Total subjects 315,778 Completed subjects 315,765.
It seems to happen when projects are very very close to full completeness and usually volunteers continue to classify, but occasionally it takes hours.
It's not good because it encourages volunteers to classify on Paused projects.

PR #3952, opened 2022-09-29
switch back to float calculation for completeness metrics

camallen:
related to the changes in #3950 switch back to `.to_f` metric representation for the workflow completion and the summed metric to avoid loss of precision via `.to_d`

This fixes the issue with completed projects not moving to the paused state as reported in https://www.zooniverse.org/talk/14/2620044?comment=4287231&page=1

# Review checklist


yuenmichelle1 (approved):
lgtm

PR #3953, opened 2022-09-29
update project completeness worker, set completeness to be of decimal for precisness

yuenmichelle1:
Due to floating point operations being inexact, when project_completeness gets calculated to `0.9999999999999999.to_f` (with 16 9s) will keep it as is, but  `0.99999999999999999.to_f` will return to `1.0`.  However, `0.9_w/_16_9s.to_f.to_i` will return 0 and cause a small number of paused projects to become active. 

# Review checklist


yuenmichelle1:
Closing this does not help for projects whose completeness calculate to close to 1 but not 1 (eg 0.9999999999999996) and should be paused

PR #3954, opened 2022-09-30
round the project completeness metric to avoid FP precision issues

camallen:
Follow up to #3950 and #3952 and specifically https://github.com/zooniverse/panoptes/pull/3950#issuecomment-1263051646

This PR changes the way we summate project completeness metrics to avoid issues with FP precision. 

Sometimes the calculation loses / accretes precision due to floating point arithmetic. To avoid this issue we can simple round the summated value to a significant  number of digits (4 by default). This PR provides an env var to change the rounding precision at runtime as needed to avoid code changes.

# Review checklist

### DIFF
--- a/app/workers/calculate_project_completeness_worker.rb
+++ b/app/workers/calculate_project_completeness_worker.rb
@@ -1,9 +1,11 @@
-require 'retirement_schemes'
+# frozen_string_literal: true
 
 class CalculateProjectCompletenessWorker
   include Sidekiq::Worker
   using Refinements::RangeClamping
   attr_reader :project
+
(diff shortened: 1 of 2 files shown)


NomadPurple:
We now have projects going early (before all subjects are retired) into completion. Please see the following Zootalk thread
https://www.zooniverse.org/talk/17/3046186?page=1&scrollToLastComment=true
It seems to be a rounding issue as all the projects are very close to completion.
My suggestion/question having looked at https://github.com/zooniverse/panoptes/blob/master/app/workers/calculate_project_completeness_worker.rb 
is why use % completion as the trigger to send workflows/projects into Paused. Wouldn't it avoid these rounding issues if we took a step back and used retired_subjects = total_subjects instead?

PR #3961, opened 2022-10-07
fix strong params nested array objects being unpermitted

### LINKED ISSUE #3960: Project Builder: Workflow Steps cannot be updated
## API issue?

Recorded on 6 Oct 23:00 BST
Affects: FEM WFs with Transcription Tasks, possibly others

When the Project Builder attempts to send an update to the API, to change a Workflow's Steps, the data isn't fully digested (possibly by the API?). As a result, any Workflow that uses Steps (mostly WFs with Transcription Tasks) stop working.

For example, if I send...

```
PUT /api/workflows/3631

body: {
  workflows: {
    steps: [[ "S0", { "taskKeys": [ "T0", "T1" ] } ]]
  }
}
```

...the API will return...

```
workflow: {
  steps: [[ "S0", {} ]]
}
```

### Original Report

@snblickhan reported on Slack that a [Transcription WF on Corresponding with Quakers](https://fe-project.zooniverse.org/projects/rachaelsking/corresponding-with-quakers/classify/workflow/22387) was crashing.

This WF 22387 was copied (via Project Builder's "Copy Workflow" function) from WF 22208.

@mcbouslog determined that the problem was due to 22208 having "incomplete steps". (i.e. you have step keys `S1`, but empty `{}` step values)

### Replication

- Open the project builder for any project with FEM Lab + Transcription Tasks enabled, e.g. https://master.pfe-preview.zooniverse.org/lab/1966/workflows/?env=staging
- Create a new Workflow.
- Add a Transcription Task.
- Observe the WF that was updated in response, in your Network Tab.
  - Expected: the `steps` data sent on the PUT (update) request should be reflected in the updated WF.
  - Actual: the `steps` data in the updated WF is missing data.

Example:

_On a newly-created Workflow, I create a Transcription Task, generating a PUT request._

Screenshot 2022-10-06 at 23 00 23

Screenshot 2022-10-06 at 22 59 52

_The API, however, responds with steps that's missing data_

Screenshot 2022-10-06 at 23 00 09

### Status

This is an **urgent issue** for any project FEM project with Transcription Tasks.

Impact:
- No new Transcription WFs can be properly created for FEM. (WF will immediately crash on FEM Classifier)
- Editing existing WFs with Transcription Tasks may cause problems to that WF. 
- Any Transcription WF created/edited while this bug exists will need to be manually fixed, after the bug is sorted?

Possible workaround:
- for each Transcription WF affected, edit the WF resource's `.steps` to the correct value, **directly via the database.** (untested)

camallen:
related to #3113 and closes #3960 

The client payload for steps attribute contains nested arrays with json objects. Strong params fails to permit these via the explicit allowlist via `permit!` in our params validator in https://github.com/zooniverse/panoptes/blob/8cfcace49491a4017bfcd017cb95fa0a85f39bcd/lib/json_api_controller/json_schema_validator.rb#L42

This is a rails fraemwork issue that was fixed in 5.2+ https://github.com/rails/rails/pull/23650 so i've backported that fix and added specs on this. 

# Review checklist

### DIFF
--- a/lib/gem_ext/gem_ext.rb
+++ b/lib/gem_ext/gem_ext.rb
@@ -1,3 +1,5 @@
 require_dependency 'gem_ext/doorkeeper/application'
 require 'gem_ext/doorkeeper/server'
 require 'gem_ext/doorkeeper/client_credentials_creator'
+
+require 'gem_ext/rails/strong_parameters'
--- a/lib/gem_ext/rails/strong_parameters.rb
+++ b/lib/gem_ext/rails/strong_parameters.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# backport the strong params nested array fix for permit!
+# that is fixed in 5.2+ https://github.com/rails/rails/pull/32593/
+# landed in https://github.com/rails/rails/blob/v5.2.8.1/actionpack/CHANGELOG.md#rails-521-august-07-2018
+if Gem::Version.new(Rails.version) < Gem::Version.new('5.2')
+  module ActionController
+    class Parameters
+      def permit!
+        each_pair do |key, value|
+          Array.wrap(value).flatten.each do |v|
+            v.permit! if v.respond_to? :permit!
+          end
+        end
+
+        @permitted = true
+        self
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


yuenmichelle1 (approved):
LGTM nice. This looks like its working. I tested a bit on my local. I wasn't sure how crazy `steps` would look like coming from the FE so I tested the following:
- if `steps` had an array of an array (original bug)
-  if `steps` had an array of an array of an array 
- if `steps` had an array of an array of an array and the object containing `taskKeys` had another object inside that had a key with an array value with an object in that array. 

**Long story short, I don't think `steps` will ever get as crazy as I've described, but update now reads anything nested in `steps`/nested array objects in general. 👍**

PR #3962, opened 2022-10-10
Refactor pg selection code to use CTEs instead of subqueries

camallen:
This PR switches the pg subject selection code** from using nested subqueries to find the available pools of subjects to using postgres [CTE](https://www.postgresql.org/docs/current/queries-with.html) queries. This removes the need for custom AR version scoping code and should be faster (benchmarks needed!)

It uses a new PG query [extensions gem ](https://github.com/georgekaraszi/ActiveRecordExtended#common-table-expressions-cte) in place of an older union query gem - it provides a suite of new features and is compatible with rails 5+ and is maintained. 

**used when external subject selector service ([designator](https://github.com/zooniverse/designator/)) is not available

In the benchmarking below it appears that the CTE version is slightly quicker (use total not real values) but there doesn't seem to be a decrease in performance which is key for me. 

### Benchmarks (to be take with a grain of salt)

*All queries were run with subject data loaded already, i.e. i removed the results that loaded the ~225K subjects from disk*

#### Current deployed version
```
##### Current master with hot data paths
### WF: 2363
# usual priority project wf subjects (~2K)
# with priority selection

### logged in user with no seen subjects
#       user     system      total        real
#  0.073161   0.010550   0.083711 (  1.018071)

### user with some seen subjects
#       user     system      total        real
#  0.014965   0.000000   0.014965 (  0.370547)

### no user
#       user     system      total        real
#  0.017332   0.002949   0.020281 (  0.386075)

### WF: 13693
# good sized wf subjects (~225K)
# with random selection

### user with no seen subjects
#       user     system      total        real
#  0.462185   0.020308   0.482493 (  3.011191)

### user with some seen subjects
#       user     system      total        real
#  0.347239   0.003387   0.350626 (  2.128591)

### user with seens subjects
#       user     system      total        real
#  0.330617   0.011642   0.342259 (  2.019306)

### no user
#       user     system      total        real
#  0.330270   0.007683   0.337953 (  2.114776)
```

#### Refactored branch
```
## Refactor branch with hot data paths
### WF: 2363
# usual priority project wf subjects (~2K)
# with priority selection

### logged in user with no seen subjects
#       user     system      total        real
#  0.053747   0.003275   0.057022 (  0.779420)

### user with some seen subjects
#       user     system      total        real
#  0.005348   0.000000   0.005348 (  0.168495)

### no user
#       user     system      total        real
#  0.002275   0.000134   0.002409 (  0.165507)

### WF: 13693
# good sized wf subjects (~225K)
# with random selection

### user with no seen subjects
#       user     system      total        real
#  0.032827   0.006103   0.038930 (  1.355070)

### user with some seen subejcts
#       user     system      total        real
#  0.006971   0.000000   0.006971 (  0.810077)
 
### user with seens subjects
#       user     system      total        real
#  0.010655   0.000000   0.010655 (  0.790179)

### no user
#       user     system      total        real
#  0.006159   0.000000   0.006159 (  0.780472)
```

## TODO 
- [x] Run benchmark test queries on new and old code

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ source 'https://rubygems.org'
 gem 'active_interaction', '~> 3.8.3'
 gem 'active_model_serializers' # Event stream
 gem 'activerecord-import', '~> 1.4'
-gem 'active_record_union', '~> 1.3.0'
+gem 'active_record_extended'
 gem 'aws-sdk', '~> 2.10'
 gem 'azure-storage'
 gem 'azure-storage-blob'
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,8 @@ source 'https://rubygems.org'
 
 gem 'active_interaction', '~> 3.8.3'
 gem 'active_model_serializers' # Event stream
-gem 'activerecord-import', '~> 1.4'
 gem 'active_record_extended'
+gem 'activerecord-import', '~> 1.4'
 gem 'aws-sdk', '~> 2.10'
 gem 'azure-storage'
 gem 'azure-storage-blob'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -42,8 +42,10 @@ GEM
       activemodel (>= 4.1, < 7.1)
       case_transform (>= 0.2)
       jsonapi-renderer (>= 0.1.1.beta1, < 0.3)
-    active_record_union (1.3.0)
-      activerecord (>= 4.0)
+    active_record_extended (1.4.0)
+      activerecord (>= 5.0, < 6.1)
+      ar_outer_joins (~> 0.2)
+      pg (< 2.0)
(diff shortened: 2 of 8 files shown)


yuenmichelle1 (approved):
Apologies for the delay; was reading up on CTEs and getting context. Honestly this does look cleaner and definitely more readable. 

In terms of performance, I wonder if these only get better as we update postgres to 12+
https://medium.com/alan/ctes-optimization-8c082efc47ef. 

Focusing on the "Potential Dire Consequences" of ^ the above article, esp first point 

> Indexes cannot be used efficiently after a CTEs, in JOINs or WHERE clauses for instance

it does not look like this affects us in our CTE use case, so I'm good with this refactor.

PR #3966, opened 2022-10-12
explicit avoid code injection via media_route params

camallen:
media routes declare allowed polymorphic resources for use in routes and these are constrained to certain resource names, https://github.com/zooniverse/panoptes/blob/5b0cbf938e1e33830ec4204778aa0c6c9f5030bc/lib/routes/json_api_routes.rb#L34

This PR avoids any possible code injection we avoid using those params and instead rely on an allow list in the media controller. 

Longer term that means if we add new media resources in routes we have to declare them in this controller file as well to ensure they work and not error. 

# Review checklist

### DIFF
--- a/app/controllers/api/v1/media_controller.rb
+++ b/app/controllers/api/v1/media_controller.rb
@@ -10,6 +10,16 @@ class Api::V1::MediaController < Api::ApiController
 
   schema_type :json_schema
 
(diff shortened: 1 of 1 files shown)


yuenmichelle1 (approved):
LGTM

PR #3967, opened 2022-10-13
avoid clobbering a project's existing workflows when updating links

camallen:
related to https://github.com/zooniverse/Panoptes-Front-End/pull/6212#discussion_r994067886

This PR ensures we use an array form for the project's workflow links when using the `update_links` controller paths and the relation manager. Thus we avoid clobbering a project's existing workflow links as we never want to lose existing workflows on a project. 

FWIW - I can't see a use case where a client would want to clobber the existing workflow links but we will see if any bug reports come in and Hyrum's law kicks in https://www.hyrumslaw.com/

# Review checklist


yuenmichelle1 (approved):
LGTM

PR #3968, opened 2022-10-13
touch the resource we are linking to on update_links

camallen:
This PR introduces a change to controller `update_links` behaviours where we ensure always modify the `resource.updated_at` value. 

When we `save` the resource we need to ensure we modify the cache key (cache bust) on the linking resource so we generate a new serialization response that includes the update_link operations results. As such we run `resource.touch` if the save operation had no changes (no update issued) and we modify the `updated_at`.

# Review checklist

### DIFF
--- a/spec/support/updatable.rb
+++ b/spec/support/updatable.rb
@@ -81,24 +81,32 @@ RSpec.shared_examples "supports update_links" do
   let(:updated_resource) { resource.reload }
   let(:stringified_test_relation_ids) { Array.wrap(test_relation_ids).map(&:to_s) }
   let(:linked_resources) { updated_resource.send(test_relation) }
-
-  before(:each) do
-    default_request scopes: scopes, user_id: authorized_user.id
-    params = {
+  let(:params) do
+    {
       link_relation: test_relation.to_s,
       test_relation => test_relation_ids,
       resource_id => resource.id
     }
-    post :update_links, params
   end
 
-  it 'should update any included links' do
+  before do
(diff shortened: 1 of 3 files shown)


spec/support/updatable.rb
```diff
@@ -81,25 +81,33 @@ def self.extract_linked_resource_ids(linked_resource)
...
   end
 
-  it 'should retain pre-existing links' do
+  it 'retains pre-existing links' do
+    post :update_links, params
     expect(linked_resources.map(&:id)).to include(*old_ids)
   end
+
+  it 'updates the cache key on the resource' do
+    # this is so the serializer resonse cache is busted and the links includes the newly added resource
```
yuenmichelle1, line 108:
```suggestion
    # this is so the serializer response cache is busted and the links includes the newly added resource
```

spec/support/updatable.rb
```diff
@@ -117,18 +125,22 @@ def self.extract_linked_resource_ids(linked_resource)
...
   end
 
-  it "should create a new linked_resource" do
-    linked_resource
+  it 'creates a new linked_resource' do
     expect{ update_via_links }.to change { linked_resource.class.count }.by(1)
   end
+
+  it 'updates the cache key on the original resource' do
+    # this is so the serializer resonse cache is busted and the links includes the newly added resource
```
yuenmichelle1, line 143:
```suggestion
    # this is so the serializer response cache is busted and the links includes the newly added resource
```

yuenmichelle1 (approved):
lgtm

PR #3969, opened 2022-10-13
remove regex testing for relation association includes

camallen:
This PR modifies the V1 Adapter Event stream serializer to avoid the use of escaping in regexes, instead break the use case into simpler methods with string comparisons

The code in this class is largely unused and can be refactored down even further (i've done this and might PR it after this lands), however it's largely untested but we do have high level tests on output formats so that's useful. Also this code is really hard to understand, I believe it's a recursive formatter that loops over the associations and data nestings to derive the format that we serialize via the restpack gem...I think this was a lift and shift operation from an earlier version of AMS that matched our pre v1 JSON-API format, https://github.com/zooniverse/panoptes/pull/1293

# Review checklist


lib/serialization/v1_adapter.rb
```diff
@@ -109,21 +109,33 @@ def attributes_for_serializer(serializer, options)
...
+    end
+
+    def check_nested_assoc(assoc)
+      include_opt = @options[:include]
+      include_opt = include_opt.split(',') if include_opt.is_a?(String)
+
+      # check if any of the included associations match the prefix path of the association
+      # i.e. ${association}.path
+      nested_assoc_pattern = "#{assoc}."
+      include_opt.any? { |included_assoc| included_assoc.starts_with?(nested_assoc_pattern) }
```
yuenmichelle1, line 138: I'm curious to know in what cases this is ever true. (i.e.
 ``` 
nested_assoc_pattern = "#{assoc}."
include_opt.any? { |included_assoc| included_assoc.starts_with?(nested_assoc_pattern)
```

yuenmichelle1 (approved):
LGTM I had one quick question. I wanted to do a quick sanity check and run this against the serializer specs with breakpoints so i can see what was going on and it looks like one of the checks (nested assoc check) never came up.

lib/serialization/v1_adapter.rb
camallen: it's not that i could find - this was what i was referencing when i said
> The code in this class is largely unused and can be refactored down even further (i've done this and might PR it after this lands), however it's largely untested but we do have high level tests on output formats so that's useful.

I have another branch where i remove the unused code and refactor but I didn't have any trust in it as we don't have any tests on this code...so I left it with this refactor for the security issue fix. 

I can push that code up as a draft PR if you fancy looking at it and we can move forward if you think it's worthwhile?

PR #3975, opened 2022-10-17
Rails 5.1 remove unnecessary test of optimistic locking

yuenmichelle1:
# Review checklist


camallen (approved):
LGTM 

However I think this whole file can go as it is not testing any api functionality.

yuenmichelle1:
> LGTM
> 
> However I think this whole file can go as it is not testing any api functionality.

That's fair. The only test that is left in this

> LGTM
> 
> However I think this whole file can go as it is not testing any api functionality.

That's fair. Viewing the remaining test in that file, it does not look meaningful in any way. So I'm ok with deleting this test file altogether.

PR #3976, opened 2022-10-18
Rails 5.1 - update serve_static_files to public_file_server.enabled

yuenmichelle1:
DEPRECATION WARNING: `config.serve_static_files` is deprecated and will be removed in Rails 5.1.
Please use `config.public_file_server.enabled = false` instead.

# Review checklist


camallen (approved):
:shipit:

PR #3993, opened 2022-11-04
Rails 5.2 next upgrade

yuenmichelle1:
Start of Running Rails 5.2 in next dual boot

Learning from previous mistakes: This PR will only focus on updating the Rails gem in next version and seeing how many specs fail and any Deprecated Warnings. 

What's nice is that app boots up in 5.2 but it does look like several tests fail. 

### **TODO:**   
(in follow up PRs)

- [ ] `DEPRECATION WARNING: Schema+ PG Indexes is deprecated for ActiveRecord 5.2 and up. Please see the README.md for more details. (called from <top (required)> at /rails_app/config/application.rb:14)`
- [ ] Fix specs (divvy and conquer as a team) 
- [ ] See: https://jetrockets.com/blog/rails-5-2-changes-in-callbacks : it looks like we have references to `attribute_changed?` (eg. `email_changed?`) as well as the `.changed?` methods. This behavior possibly has changed in Rails 5.2, so we may want to replace with its equivalent methods.

### **Notes**
- It looks like some tests are failing due to specs that test callbacks https://jetrockets.com/blog/rails-5-2-changes-in-callbacks. 
- Other tests are failing due to frozen string: see https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do. This looks like it is as easy as replacing `<<` with `+=`, but I could be wrong. 

### **Reading materials:** 
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2
https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-5-1-to-5-2.html
https://guides.rubyonrails.org/5_2_release_notes.html#upgrading-to-rails-5-2
https://jetrockets.com/blog/rails-5-2-changes-in-callbacks
https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do
https://www.bigbinary.com/blog/rails-adds-support-for-recyclable-cache-keys
https://www.bigbinary.com/blog/activerecord-relation-cache-key

# Review checklist


camallen (approved):
PR is still a draft but LGTM

PR #3997, opened 2022-11-07
Remove memory-eating cache from iNat API interface

zwolf:
Keeping the observations cached seemed like a good idea at the time, but the iNat API responses are so huge that holding on to them was causing the memory usage to spike and `exit code 137` the pod. Tests showed that without the cache, mem usage plateaus around 400-500MB but still rises very slowly. Further testing with extremely large datasets (150k+) is required to make sure that it doesn't hit the 700MB limit on staging. If it does, options are "get better at ruby memory management" or "bump the staging dumpworker pod memory limit to compensate", depending on how egregious it is. 

# Review checklist

### DIFF
--- a/lib/inaturalist/api_interface.rb
+++ b/lib/inaturalist/api_interface.rb
@@ -7,12 +7,12 @@ module Inaturalist
     require 'json'
 
     # Set maximum imported subjects, or no limit with -1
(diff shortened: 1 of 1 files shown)


yuenmichelle1 (approved):
LGTM

PR #3998, opened 2022-11-08
Rails 5.2 test fixes 2

yuenmichelle1:
### **Fixes for middleware and models specs (though some changes may actually fix more than the focused test suites)** 
See: https://github.com/zooniverse/panoptes/actions/runs/3420941929/jobs/5696414771 for reference on failing tests. 

### **Changes**

- `ActionDispatch::ParamsParser::ParseError` is deprecated and  replaced with `ActionDispatch::Http::Parameters::ParseError` (see https://github.com/rails/rails/commit/e16c765ac6dcff068ff2e5554d69ff345c003de1 or https://qiita.com/ledsun/items/4e3431de080c97796817#rails-52%E3%81%AE%E5%A4%89%E6%9B%B4 for details ) 
- `cache_key` coming from the resource is now a frozen string. in order to append cache_key to get our extended_cache_key we need to send `create_append_cache_key` a dupped cache_key string. [Related https://www.bigbinary.com/blog/rails-adds-support-for-recyclable-cache-keys] 
 -  Update project.rb check if `attribute_changed?` to `saved_change_to_attribute?` (eg. `beta_requested_changed?` => `saved_beta_requested_changed?`
 - Update user_group_spec  to create `owned` resource to fixed `owner.rb` spec (checks whether `user_group.owns?` the resource. `user_group.owned_resources` comes out as empty when `owned` is not created, could not figure out a way to stub in order for `user_group.owned_resources` to include the acl of resource other than to create the resource.. this might be a small transaction since `owned` only looks to be used in owner specs. **any advice on a workaround as opposed to the `create(:project)` for owned would be great**
 

**Reading**

- https://github.com/rails/rails/commit/e16c765ac6dcff068ff2e5554d69ff345c003de1
- https://qiita.com/ledsun/items/4e3431de080c97796817#rails-52%E3%81%AE%E5%A4%89%E6%9B%B4
- https://edgeguides.rubyonrails.org/5_2_release_notes.html
- https://www.bigbinary.com/blog/rails-adds-support-for-recyclable-cache-keys
- https://jetrockets.com/blog/rails-5-2-changes-in-callbacks

# Review checklist


spec/models/user_group_spec.rb
```diff
@@ -5,7 +5,7 @@
   let(:named) { user_group }
   let(:activatable) { user_group }
   let(:owner) { user_group }
-  let(:owned) { build(:project, owner: owner) }
+  let(:owned) { create(:project, owner: owner) }
```
yuenmichelle1, line 8: I couldn't figure out a workarond to get `user_group.owned_resources` to come out as expected in order for the `owner.rb` shared specs to work  other than to create the `owned` resource. 

Any tips in order to avoid create would be great

app/models/concerns/extended_cache_key.rb
```diff
@@ -32,7 +32,7 @@ def cache_key
     cache_key = super
     associations_cache_key = compound_association_cache_keys.join("+")
     methods_cache_key = compound_method_cache_keys.join("")
-    create_append_cache_key(cache_key, methods_cache_key, associations_cache_key)
+    create_append_cache_key(+cache_key, methods_cache_key, associations_cache_key)
```
camallen, line 35: TIL about this syntax - nice! 
```suggestion
    # dup the frozen string https://ruby-doc.org/core-2.6/String.html#method-i-2B-40
    create_append_cache_key(+cache_key, methods_cache_key, associations_cache_key)
```

FWIW - this is the sort of link / comment that ideally would be included in the commit message as well, https://github.com/zooniverse/panoptes/pull/3998/commits/d52f664109668b19c4e69c57b078ec8cce596554

spec/models/user_group_spec.rb
camallen: Instead of creating the project for every spec in this file you can instead do it for just the failing 'is an owner' ones to save on the spec cost setup
```
it_behaves_like "is an owner" do
  before { owned.save }
end
```

camallen (approved):
LGTM

spec/models/user_group_spec.rb
yuenmichelle1: ooh yep that works. thanks!

PR #4000, opened 2022-11-09
Rails 5.2 test fixes t-z controllers and replacing deprecated changed?/attribute_changed? for after_save callbacks. 

yuenmichelle1:
**Changes**

- Update `lib/formatter/csv/classification.rb` to use `merge!` instead of `reverse_merge!`. In Rails 5.2, Hash#reverse_merge! behavior changes ( See https://github.com/rails/rails/pull/28077
 and https://guides.rubyonrails.org/5_2_release_notes.html#active-support-notable-changes) 
- Update `user.rb`, `find_or_create.rb` and `user_info_scrubber_spec.rb` to use `saved_changes?`/`saved_change_to_attribute?` as opposed to deprecated `changed?`/ `attribute_changed` https://www.fastruby.io/blog/rails/upgrades/active-record-5-1-api-changes.html (only applies to any `after_save` callbacks, `before_save` callbacks allow `changed?` and `attribute_changed?`)

# Review checklist


app/operations/user_project_preferences/find_or_create.rb
```diff
@@ -10,7 +10,7 @@ def execute
         upp.email_communication = user.project_email_communication
       end
 
-      if upp.new_record? || upp.changed?
+      if upp.new_record? || upp.saved_changes?
```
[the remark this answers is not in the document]
yuenmichelle1: Ah yeah. I guess I could have left this as `.changed?` but `.has_changes_to_save?` reads better. 

> perhaps why we have to re-assess where this has been applied in the upgraded code so far
Agreed I wrote a list of where this change was applied (as well as other things I would like to test before we move to prod)

- [ ] versioning.rb Check the save_version (change on L36)
- [ ] app/models/user.rb L311: check set_zooniverse_id behavior. (change on L311)
- [ ] app/models/project.rb L118 & L120 check that project notifications are sent if recently beta requested or if recently launch requested 
- [ ] (Unrelated to saved_changes, but I wanted to ensure cache_key returns as expected) extended_cache_key.rb

lib/json_api_controller/updatable_resource.rb
```diff
@@ -29,11 +29,10 @@ def update_links
...
         add_relation(resource, relation, params[relation])
-        if resource.changed?
-          resource.save!
-        else
-          resource.touch
-        end
+        # as this resource may not have changed but the linked resources may have
+        # ensure we modify the upated_at timestamp to cache bust this resource
+        resource.updated_at = Time.zone.now
+        resource.save!
```
zwolf, line 35: I thought #touch was just sugar for `resource.update_attributes(:updated_at, Time.zone.now)` or basically what this looks like it does. Does this do something different?

zwolf (approved):
lgtm, just one non-blocking question

lib/json_api_controller/updatable_resource.rb
yuenmichelle1: `.touch` ran into issues with a callback where the resource's owner could not be read. 
tbh (non sequitur) i have found touch to be finnicky since rails 5 (maybe even before). https://github.com/rails/rails/issues/34768, we see this issue a bit in SubjectSetSubjectCounterWorker

camallen: the original issue what the `.save!` call was not running as the `.changed?` returned false. That conditional was added to avoid 2 round trips to the db on each `update_links` action, i.e.
- one if if we need to save and `updated_at` will change and 
- another if the resource hasn't changed but we want to cache bust via touching `updated_at`

I couldn't see an API interface to detect the the associated records may have changed and instead of running both the save and the touch (2 db round trips) i figured manually modifying the `updated_at` on the original resource was a quick fix for both use cases.

FWIW i don't think `.touch` if finicky and we don't use `counter_caches` on our records, rather we're hitting long running counter jobs and the resource changes under us. If that's the case we should reload them before saving where these stale object errors surface.

PR #4003, opened 2022-11-11
Add uniqueness constraint to iNatImportWorker jobs

zwolf:
More than one job per set of args shouldn't be allowed to run at once. This is difficult behavior to spec, see https://github.com/mhenrixon/sidekiq-unique-jobs#uniqueness

> It is recommended to leave the uniqueness testing to the gem maintainers. 

Because of how the controller gives back a 200 immediately and the job is queued, SubjectSetImport is created, all in the background with no ability to track progress at the moment, so it's important that that request is idempotent. I'll give it a spin with a small set on staging to make sure.

also included an unrelated extra spec I thought of 🙃

# Review checklist


yuenmichelle1 (approved):
👍

PR #4016, opened 2022-11-16
Use the v2 of the iNat API

zwolf:
Using version 2 of the iNat API allows for the response to be customized via the "fields" parameter. This param is JSON encoded as RISON and indicates the only fields, including nested objects, that should be included in the response. This will reduce the memory footprint of large iNat imports. Hopefully it's enough.

edit: Here's the iNat API v2 docs for reference: https://api.inaturalist.org/v2/docs/

# Review checklist


yuenmichelle1 (approved):
This looks fine to me. One thing we should take care of (kinda off the topic of using v2 api) is we should handle the case if the inat api sends us an error, otherwise I think the worker might get stuck in a loop.

zwolf:
The sidekiq worker is set to not retry at all (`retry: 0`). If the iNat API throws, say, a 500, Faraday itself will retry the request twice, then a `Faraday::ServerError` gets raised and the job will straight up die. Failed imports just get logged, but if the whole API is complaining the job should probably just quit. It should be the same for other types of errors that Faraday raises, too. I did have a look at that code and realized that it didn't make sense to catch an error just to immediately re-raise it, though.

PR #4026, opened 2022-11-21
add validation to medium resource 

yuenmichelle1:
update tests and add validation to medium resource to allow specific content types 

# Review checklist

### DIFF
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -17,6 +17,7 @@ class Medium < ApplicationRecord
   EXPORT_MEDIUM_TYPE_REGEX = /\A(project|workflow)_[a-z_]+_export\z/i
 
   validate :validate_content_type
+  validate :validate_export_content_type, if: :export_medium
 
   def self.inheritance_column
     nil
@@ -99,14 +100,15 @@ class Medium < ApplicationRecord
     raise e
   end
 
-  def validate_content_type
-    errors.add(:content_type, 'Invalid file type') if BLOCKED_CONTENT_TYPES.match?(content_type)
-
-    export_medium = type.match?(EXPORT_MEDIUM_TYPE_REGEX)
-    validate_export_medium if export_medium
+  def export_medium
+    type.match?(EXPORT_MEDIUM_TYPE_REGEX)
   end
 
-  def validate_export_medium
+  def validate_content_type
+    errors.add(:content_type, 'Invalid file type') if BLOCKED_CONTENT_TYPES.match?(content_type)
+  end
+
+  def validate_export_content_type
     errors.add(:content_type, "Content-Type for exports must be one of #{ALLOWED_EXPORT_CONTENT_TYPES.join(", ")}") unless ALLOWED_EXPORT_CONTENT_TYPES.include?(content_type)
   end
 
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -35,7 +35,7 @@ class Medium < ApplicationRecord
     'text/plain': 18,
     'video/mp4': 19,
     'video/mpeg': 20
-}.freeze
+  }.freeze
 
   EXPORT_MEDIUM_TYPE_REGEX = /\A(project|workflow)_[a-z_]+_export\z/i
 
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -107,7 +107,7 @@ class Medium < ApplicationRecord
   end
 
   def validate_export_medium
-    errors.add(:content_type, "Content-Type must be one of #{ALLOWED_EXPORT_CONTENT_TYPES.join(", ")}") if !ALLOWED_EXPORT_CONTENT_TYPES.include?(content_type)
+    errors.add(:content_type, "Content-Type for exports must be one of #{ALLOWED_EXPORT_CONTENT_TYPES.join(", ")}") unless ALLOWED_EXPORT_CONTENT_TYPES.include?(content_type)
   end
 
   private
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -35,7 +35,7 @@ class Medium < ApplicationRecord
     'text/plain': 18,
     'video/mp4': 19,
     'video/mpeg': 20
-  }
+}.freeze
 
   EXPORT_MEDIUM_TYPE_REGEX = /\A(project|workflow)_[a-z_]+_export\z/i
 
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -12,6 +12,7 @@ class Medium < ApplicationRecord
   before_destroy :queue_medium_removal, unless: :external_link
 
   ALLOWED_EXPORT_CONTENT_TYPES = %w(text/csv).freeze
+  ALLOWED_CONTENT_TYPES = %r{jp(e)?g|gif|png|ico|txt|mp(3|4)|webm|og(a|g|m|v|x)|spx|opus|pdf|ttf|tar|gz|tgz|bz2|tbz2|zip|csv|mp(e)?g|svg|plain|m4(a|b)}
   EXPORT_MEDIUM_TYPE_REGEX = /\A(project|workflow)_[a-z_]+_export\z/i
 
   validate do |medium|
@@ -20,6 +21,8 @@ class Medium < ApplicationRecord
       medium.errors.add(:content_type, "Content-Type must be one of #{ALLOWED_EXPORT_CONTENT_TYPES.join(", ")}")
     end
   end
+
+  validate :validate_content_type
 
   def self.inheritance_column
     nil
@@ -100,6 +103,12 @@ class Medium < ApplicationRecord
     # ensure we raise unexpected errors once we've exhausted
     # the number of retries to continute to surface these errors
     raise e
+  end
+
+  def validate_content_type
+    if !ALLOWED_CONTENT_TYPES.match?(content_type)
+      errors.add(:content_type, 'Invalid file type')
+    end
   end
 
   private
--- a/app/models/medium.rb
+++ b/app/models/medium.rb
@@ -12,7 +12,7 @@ class Medium < ApplicationRecord
   before_destroy :queue_medium_removal, unless: :external_link
 
   ALLOWED_EXPORT_CONTENT_TYPES = %w(text/csv).freeze
-  ALLOWED_CONTENT_TYPES = %r{jp(e)?g|gif|png|ico|txt|mp(3|4)|webm|og(a|g|m|v|x)|spx|opus|pdf|ttf|tar|gz|tgz|bz2|tbz2|zip|csv|mp(e)?g|svg|plain|m4(a|b)}
+  ALLOWED_CONTENT_TYPES = %r{jp(e)?g|gif|png|ico|txt|mp(3|4)|webm|og(a|g|m|v|x)|spx|opus|pdf|ttf|tar|gz|tgz|bz2|tbz2|zip|csv|mp(e)?g|svg|plain|m4(a|b)|json}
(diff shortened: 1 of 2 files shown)


app/models/medium.rb
```diff
@@ -102,6 +105,10 @@ def put_file_with_retry(file_path, opts={}, num_retries=5)
     raise e
   end
 
+  def validate_content_type
+    errors.add(:content_type, 'Invalid file type') unless ALLOWED_CONTENT_TYPES.match?(content_type)
```
camallen, line 109: any reason this regex doesn't match the whole content type? 

i think i could smuggle unallowed media resources in if they contain an allowed substring, e.g. `text/html-json` would pass this test and allow the media resource to be saved...it may fail to set an actual path on the validation via https://github.com/zooniverse/panoptes/blob/85142aa1bc5c57f8ec0164174aa7583a885bcbb9/app/models/medium.rb#L35 though..

still this should be more robust.

app/models/medium.rb
```diff
@@ -12,6 +12,7 @@ class MissingPutFilePath < StandardError; end
   before_destroy :queue_medium_removal, unless: :external_link
 
   ALLOWED_EXPORT_CONTENT_TYPES = %w(text/csv).freeze
+  ALLOWED_CONTENT_TYPES = /jp(e)?g|gif|png|ico|txt|mp(3|4)|webm|og(a|g|m|v|x)|spx|opus|pdf|ttf|tar|gz|tgz|bz2|tbz2|zip|csv|mp(e)?g|svg|plain|m4(a|b)|json/
```
camallen, line 15: Did you consider inverting this check to a block list of non-allowed content-types vs an inclusive allow list? 

Allow lists are harder to maintain as they are more restrictive but can provide finer controls. 

If we do the allow list I think this regex should match what we are testing, i.e the list in and no more.  I would remove the spx, opus, ttf, tar, gz, bz2, tbz2, zip etc. Also i would ensure we test the content types in full not substrings (can smuggle by faking content types, e.g. `test/html-json` will pass).

Further we could switch from a regex to an dict (O(1) lookup) or an array inclusion (O(N) lookup), e.g. 
```
allowed_content_types = {
'application/json': 1,
....
}
if allowed_content_types.key?(content_type)
  # passed
else 
  # failed
  errors.add....
end
```

eatyourgreens: This is a list of file extensions, which is slightly different from Content-Type eg. `.txt` files have Content-Type `text/plain`. 

I’m not sure if this is validating the file name for an allowed extension or the Content-Type header for an allowed type and sub-type.

camallen: For us to host the media resources the API expects the mime types to be set on creation of the media resource, e.g. for subjects https://panoptes.docs.apiary.io/#reference/subjects/subject-collection/create-a-subject

those are then validated here in the Medium resource so what we will be validating against is the mime type here stored in the `content_type` attribute on that model. These in turn set the `content-type` header that the blob storage system will return via the HTTP request to fetch the media resource from the CDN domain. 

So we aren't validating file extensions here and why i'm worried about the this simple regex allow invalid mime types to pass.

camallen: FWIW the server never knows the file extension of the media resource, instead the client tells the API it want's to upload a media resource of content_type X (e.g. `image/jpeg`) and the api stores this media resource in the DB, creates a uniq id for blob storage and returns a writable blob storage URL to the client so it can upload the file to the remote blob storage system. 

At no point does the API see the file that will be uploaded, so we can't inspect the contents of the file etc. This would only be possible post the upload by the client.

eatyourgreens: In PFE, I modified the Panoptes file upload to allow any image, audio or video types plus a limited set of text and application types.
https://github.com/zooniverse/Panoptes-Front-End/pull/6260

eatyourgreens: I could probably have chosen better names for these variables but this is the code that checks Content-Type to see if a file can be uploaded.
https://github.com/zooniverse/Panoptes-Front-End/blob/0a7990d11e3d90600d8043593069ee2704dab3c5/app/lib/put-file.js#L1-L8

app/models/medium.rb
```diff
@@ -21,6 +22,8 @@ class MissingPutFilePath < StandardError; end
     end
   end
 
+  validate :validate_content_type
```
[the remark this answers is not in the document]
yuenmichelle1: no reason. just a missed opportunity to refactor.  we can merge the two validations into `validate_content_type` if thats what you're suggesting.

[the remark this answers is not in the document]
camallen: It's slightly confusing that we have a generic validate method and a specific one for this use case however both seem related.

worth refactoring and perhaps use a generic validate method that calls out the various specific validation methods

app/models/medium.rb
yuenmichelle1: > Did you consider inverting this check to a block list of non-allowed content-types vs an inclusive allow list?
> Allow lists are harder to maintain as they are more restrictive but can provide finer controls.

I'm fine either way. IMO when I was thinking through this at least, blocklists seemed a bit "harder" since it felt like there would be many things to block (I could be wrong). Basing on this list of types: https://gist.github.com/tylerlee/53609bff1346cebf8f0a85b6be29a88e it felt like there would be too much to block. 

> If we do the allow list I think this regex should match what we are testing, i.e the list in and no more. I would remove the spx, opus, ttf, tar, gz, bz2, tbz2, zip etc. 

Sounds good. I was going by what was mentioned in slack as the regex test, but that apparently was incorrect. 

> Also i would ensure we test the content types in full not substrings (can smuggle by faking content types, e.g. test/html-json will pass).

Ah, yeah thats a good test to add.

app/models/medium.rb
```diff
@@ -12,14 +12,11 @@ class MissingPutFilePath < StandardError; end
   before_destroy :queue_medium_removal, unless: :external_link
 
   ALLOWED_EXPORT_CONTENT_TYPES = %w(text/csv).freeze
+  BLOCKED_CONTENT_TYPES = %r{html|javascript|css|spx|opus|ttf|tar|gz|tgz|bz2|tbz2|zip}
```
camallen, line 15: Is there a way to add `exe` files to this list as well? 

I think i prefer an explicit list of mime types here instead of a regex, e.g. %w[text/html... etc)
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

app/models/medium.rb
```diff
@@ -102,6 +99,17 @@ def put_file_with_retry(file_path, opts={}, num_retries=5)
     raise e
   end
 
+  def validate_content_type
+    errors.add(:content_type, 'Invalid file type') if BLOCKED_CONTENT_TYPES.match?(content_type)
+
+    export_medium = type.match?(EXPORT_MEDIUM_TYPE_REGEX)
+    validate_export_medium if export_medium
```
camallen, line 106: Not blocking but i don't really like how this is lumped in here though i can see it's logically validating the content type.

I think extracting this to it's own custom validation makes more sense
```
validate :validate_content_type
validate :validate_media_content_type if: :export_medium
```

camallen (approved):

app/models/medium.rb
yuenmichelle1: ah i must have misunderstood what you meant initially on https://github.com/zooniverse/panoptes/pull/4026#discussion_r1029982988.  i'm ok with updating.

app/models/medium.rb
yuenmichelle1: The more I wrote explicit BLOCKED_CONTENT_TYPES/mime types that should be blocked, the more I convinced myself that I prefer an ALLOW_LIST (with explicitly listed out mime types). Here is my reasoning: 

1. explicitly listed blocked lists are harder to catch someone trying to fake content types (eg. your `test/html-json`)...though I question if smuggling here is an issue since I was under the assumption Azure Blob Storage "handles this". (By "handles this" - I mean for eg. if someone uploads a text/html under a faked mime type azure will return the type as the true mime type to panoptes for medium upload anyway.)
2. Allow lists are stricter, which I believe is something we want. (But maybe that's just my preference)
3.  It would keep it consistent with what UI validation is doing which uses an Allow list as well. *

Point 3 was probably the biggest selling point that pushed me to this decision.
I'm sorry to ask of you again @camallen . Do you mind reviewing newest changes? I appreciate your feedback.

camallen: Not a problem for another review @yuenmichelle1 

2 & 3. yes they are - if you are happy with this then it's 👍🏼 from me. The main issue with an allow list is whenever a client wants a new mime type we need to code it up and deploy. Apart from that i think consistency between apps is a great idea.

Re point 1. The client submits the mime type to the api for what file it wants to upload. We then 
1. create a media resource in the db with the supplied mime type (it's it's allowed) 
2. we then request a blob storage URL that allows the client to put a file to that blob storage location on our blob storage system - note this URL has a time expiry so it's not allowed to write forever to that blob storage location. 
3. the client then uploads directly to the blob storage provider using the time limited put URL
4. the media resource can now provide a get URL for the blob storage location for reading in clients

So at no point does the blob storage system (azure / s3 etc) validate the mime / content type of the user supplied files. We trust the client provides the correct one and we set the content type on the blob storage system accordingly https://github.com/zooniverse/panoptes/blob/84de373bf03296d75d62e4f80a6857439f82fce4/app/services/media_storage/azure_adapter.rb#L59

Note we also use the correct file extension for the supplied content type when we create the blob storage path, https://github.com/zooniverse/panoptes/blob/84de373bf03296d75d62e4f80a6857439f82fce4/app/services/media_storage/azure_adapter.rb#L28

so we need to validate the content types are ok for us either from the block or allow list approach.

app/models/medium.rb
```diff
@@ -12,14 +12,12 @@ class MissingPutFilePath < StandardError; end
   before_destroy :queue_medium_removal, unless: :external_link
 
   ALLOWED_EXPORT_CONTENT_TYPES = %w(text/csv).freeze
+  ALLOWED_CONTENT_TYPES = %w[application/json application/pdf audio/aac audio/midi audio/x-midi audio/mp3 audio/mp4 audio/x-m4a audio/mpeg audio/wav image/jpeg image/gif image/png image/tiff image/x-icon image/svg+xml text/csv text/plain video/mp4 video/mpeg].freeze
```
camallen, line 15: perhaps add a comment and a link to the UI code to note that it's consistent and allow folks to look it up easily.

camallen: using an array here is a O(N) at worst search operation for each media resource we want to validate. Moving to a hash would be an O(1) constant lookup.

I think the current list size is ok but something to consider if this list was to grow bigger. 

Also you could sort by most common types (image etc) to decrease the search cost but i'd probably just use a hash to speed it up and forget about it.

app/models/medium.rb
```diff
@@ -102,6 +100,18 @@ def put_file_with_retry(file_path, opts={}, num_retries=5)
     raise e
   end
 
+  def export_medium
+    type.match?(EXPORT_MEDIUM_TYPE_REGEX)
+  end
+
+  def validate_content_type
+    errors.add(:content_type, 'Invalid file type') unless ALLOWED_CONTENT_TYPES.include?(content_type)
```
camallen, line 108: this `include?` is O(N) at worst - see comment before about the hash vs array data structure.

camallen, line 108: should we supply the list of allowed mime types (or link to code) in the error, helpful error messages that include details avoid the client having to guess what the valid mime types are.

camallen (approved):

app/models/medium.rb
yuenmichelle1: ah i see. thanks for the context 👍

PR #4042, opened 2022-12-03
avoid cloberring static assets from new pods only

camallen:
The `commit_id.txt` on some pod deploys was showing inconsistent results (i.e. old deploy sha values) even though the pod contained the new sha value in the commit_id.txt baked into the container. It appears the commit_id.txt file on the shared volume between nginx and panoptes api containers (panoptes api app pods) was sometimes being overwritten with a file from the old containers during deploys and thus leaving us with the possibly inconsistent state.

This PR attempts to fix this issue by now only copying the static assets to the shared nginx volume files system if they are newer, i.e. from the newer pod deployment and not a scaling / restart event.

This is achieved this by using the `cp -u` switch to avoid clobbering new pod static assets with old ones due to race condition on the pod lifecycle event hook runs https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution. 

> -u, --update                 copy only when the SOURCE file is newer than the destination file or when the destination file is missing

Note we don't control the number of times this hook can be called which may have been an oversight when setting this up
> Hook delivery is intended to be at least once, which means that a hook may be called multiple times for any given event, such as for PostStart or PreStop. It is up to the hook implementation to handle this correctly.

https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-delivery-guarantees

# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -195,7 +195,7 @@ spec:
           lifecycle:
             postStart:
               exec:
-                command: ["/bin/bash", "-c", "cp -R /rails_app/public/* /static-assets"]
+                command: ["/bin/bash", "-c", "cp -Ru /rails_app/public/* /static-assets"]
         - name: panoptes-production-nginx
           image: zooniverse/nginx
           resources:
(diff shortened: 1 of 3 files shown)


yuenmichelle1 (approved):
LGTM

PR #4049, opened 2022-12-06
remove deprecated schema_plus_pg_indexes

yuenmichelle1:
# Review checklist


camallen (approved):
I think the only issue i have with this change is our old migrations will no break if we re-ran the `RAILS_ENV=test rails db:migrate` task. 

Ideally this change (or a follow up PR) archives our old migrations for reference** (perhaps in a different directory) and allows the db migration task to run again. 

I'm lazy and I like to reference how to run migrations from our old ones vs searching etc...if folks don't find value in this then we can just drop them and rely on the `structure.sql` file for the canonical state of the db and write migrations from scratch in future.

camallen:
> is that related to this change? I think our old migrations would break now if we re-ran from scratch because we don't have our migrations tagged with the version of Rails (ie.
> `class MigrationName < ActiveRecord::Migration[4.2]` vs most of our migrations `class OtherMigrationName < ActiveRecord::Migration`)

I thought they would run but it's worth testing via setting up a fresh test db and migrating it 
```
RAILS_ENV=test rails db:drop
RAILS_ENV=test rails db:create
RAILS_ENV=test rails db:structure:load
RAILS_ENV=test rails db:migrate
```


>  I'm fine with moving our old migrations to a different directory, I'm unsure what to name it and possibly where to put it in order to not confuse anyone.

perhaps next to migrations but called `old-migrations` ?

> I feel like I might have missed something, I'm trying to think of scenarios other than maybe in a local dev environment where we we would have to run db migration task all over from the beginning vs db setup which I think (?) db setup falls back on structure.sql.

Agreed - we should be using the canonical `strucutre.sql` (`schema.rb`) representation and using migrations for new changes to that.

zwolf:
> perhaps next to migrations but called old-migrations ?

I don't understand what moving older migrations to a new folder accomplishes, it seems confusing. Are new migrations going to go in the regular location, then need to be periodically tested for their ability to be run and possibly moved? What does that test look like, if the standard dev & test practice is to load the db schema from structure.sql, where the changes already exist?

yuenmichelle1:
@camallen 

**Re: testing `db:migrate` on a fresh db.** 

1.  
I tried this before without `db:stucture:load`, i.e. running
```
RAILS_ENV=test rails db:drop
RAILS_ENV=test rails db:create
RAILS_ENV=test rails db:migrate
```
and received an error on `db:migrate` saying 
```
Caused by:
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateDatabase < ActiveRecord::Migration[4.2]
```

But if you add `db:structure:load` to the process then `db:migrate` will run. 



2. 
I think @zwolf posed an interesting question though that we may want to discuss if we are introducing `old-migrations`. 

_Mainly: When should we move newer migrations to the `old-migrations` folder?_ 

My thinking is if we move migrations to  an `old-migrations` folder is to move the migrations that we currently have to `old-migrations`, and once the `db/migrate` folder has maybe 20+ files, we can think about moving more to `old-migrations`.

PR #4051, opened 2022-12-12
archive old migrations, move to old_migrations folder

yuenmichelle1:
# Review checklist


zwolf (approved):
¯\_(ツ)_/¯ I'm not sure how I feel about "old" being the primary descriptor here and the redundancy of `/migrate/old_migrations` ('archived', perhaps?) but I'm not gonna lose any sleep over it.

camallen (approved):
I think I prefer the directory `db/migrate/archive/` vs `db/migrate/old_migrations/` for the directory but it doesn't worry me much. Ref https://codewithrails.com/clean-up-db-migrations

PR #4062, opened 2022-12-20
Doorkeeper cleanup worker update to batch deletes

yuenmichelle1:
Really tricky to see if this fixes `ActiveRecord::StatementInvalid: PG::QueryCanceled: ERROR: canceling statement due to statement timeout` errors  from cleanup worker. I believe it does (and we probably can be convinced more in staging). Reason why I believe it does: within local testing: with batching we split a longer query of DELETE query to chunks of a bit faster queries.

Testing locally with 4000 rows

**(without batching)**
`DEBUG -- : [primary]  Doorkeeper::AccessToken Destroy (37.4ms)  DELETE FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."refresh_token" IS NULL AND (created_at + interval '1 second' * expires_in < clock_timestamp())`
**37.4ms**

 **with batching** 

` DEBUG -- : [primary]   (1.4ms)  SELECT  "oauth_access_tokens"."id" FROM "oauth_access_tokens" WHERE (revoked_at IS NOT NULL) ORDER BY "oauth_access_tokens"."id" ASC LIMIT 1000`

` DEBUG -- : [primary]  Doorkeeper::AccessToken Destroy (6.0ms)  DELETE FROM "oauth_access_tokens" WHERE (revoked_at IS NOT NULL) AND "oauth_access_tokens"."id" IN (etc ids of the 1000 pulled from previous query)` 
**1.4 + 6.0 ms** about that * 4 

`

# Review checklist

### DIFF
--- a/lib/gem_ext/doorkeeper/access_cleanup.rb
+++ b/lib/gem_ext/doorkeeper/access_cleanup.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Doorkeeper
   class AccessCleanup
     def cleanup!(keep_old_refresh_tokens_for=14)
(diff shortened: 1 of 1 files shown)


camallen (approved):
:shipit: LGTM

PR #4075, opened 2023-01-10
upgrade to sidekiq 6

yuenmichelle1:
https://www.mikeperham.com/2019/09/03/welcome-to-sidekiq-6.0/

Upgrading to sidekiq 6. One of the biggest breaking changes is 

>  Remove the daemonization, logfile and pidfile command line arguments and sidekiqctl binary.

which I don't believe we are using anyway. 

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ end
 gem 'ranked-model', '~> 0.4.8'
 gem 'restpack_serializer', git: 'https://github.com/zooniverse/restpack_serializer.git', branch: 'panoptes-api-version', ref: 'cef0969cef'
 gem 'scientist', '~> 1.6.3'
-gem 'sidekiq', '~> 5.2.5'
+gem 'sidekiq', '~> 6.0'
 gem 'sidekiq-congestion', '~> 0.1.0'
 gem 'sidekiq-cron'
 gem 'sidekiq-unique-jobs', '~> 6.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -418,11 +418,11 @@ GEM
       nokogiri (>= 1.12.0)
     scientist (1.6.3)
     shellany (0.0.1)
-    sidekiq (5.2.10)
-      connection_pool (~> 2.2, >= 2.2.2)
+    sidekiq (6.0.7)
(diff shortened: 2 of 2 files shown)


camallen (approved):
LGTM - any reason we're not going to the latest v6 minor version? https://github.com/mperham/sidekiq/blob/main/Changes.md#658

also it might be good to switch to the JSON log format as well https://github.com/mperham/sidekiq/wiki/Logging#output-format
```
Sidekiq.configure_server do |config|
  config.logger.formatter = Sidekiq::Logger::Formatters::JSON.new
end
```

yuenmichelle1:
> LGTM - any reason we're not going to the latest v6 minor version? https://github.com/mperham/sidekiq/blob/main/Changes.md#658

We hit some breaking changes by going to latest version (v6.5.8). See:  https://github.com/zooniverse/panoptes/pull/4021.

```
TypeError: Job is not a class
/home/runner/work/panoptes/panoptes/vendor/bundle/ruby/2.6.0/gems/sidekiq-unique-jobs-6.0.25/lib/sidekiq_unique_jobs/sidekiq_unique_ext.rb:50:in `<module:Sidekiq>'
/home/runner/work/panoptes/panoptes/vendor/bundle/ruby/2.6.0/gems/sidekiq-unique-jobs-6.0.25/lib/sidekiq_unique_jobs/sidekiq_unique_ext.rb:5:in `<top (required)>'
```

Probably because by 6.2.2 of Sidekiq: 
> Rename internal API class Sidekiq::Job to Sidekiq::JobRecord [#4955]

(I'm betting this was not a breaking change for Sidekiq, but if used with `sidekiq-unique-jobs` then is a breaking change). 

But we could probably go as high as 6.2.1, then update sidekiq-unique-jobs, then update sidekiq to latest v6 version. 

I'll try that out.

yuenmichelle1:
> I'll try that out.

So.. 6.0.7 is the highest I can go right now without having to muddle around with middleware. 
Sidekiq 6.1.0 introduces a new change: 
> Remove rack-protection, reimplement CSRF protection [#4588]

which causes the `/sidekiq` route to crash with an Argument Error: invalid base64.  I haven't looked too much into a fix just yet.

PR #4083, opened 2023-01-16
update sidekiq and sidekiq-unique-jobs

yuenmichelle1:
Some changes in Sidekiq UI:
- New Tabs: `Locks`, `Expired Locks` and `Changelogs` , Removed tab `Unique Digests` (replaced with `Changelogs`) 
- TODO (on merge and deploy): Last time I updated this was when we were first upgrading to Rails 5.0 and ran into high increases in non-web transactional throughput in new relic. See: https://zooniverse.slack.com/archives/C97H6UG2G/p1662044308032679. I don't know how to test that this is not a repeat of last time locally. 

Screen Shot 2023-01-16 at 1 41 16 PM

Screen Shot 2023-01-16 at 1 41 25 PM

Screen Shot 2023-01-16 at 1 42 11 PM

# Review checklist


config/initializers/sidekiq.rb
```diff
@@ -6,14 +6,25 @@ def self.redis_url
```
yuenmichelle1, line 6: Adding middleware per: https://github.com/mhenrixon/sidekiq-unique-jobs#add-the-middleware

config/initializers/sidekiq.rb
```diff
@@ -6,14 +6,25 @@ def self.redis_url
...
 Sidekiq.configure_server do |config|
   config.redis = { url: SidekiqConfig.redis_url }
+
+  config.client_middleware do |chain|
+    chain.add SidekiqUniqueJobs::Middleware::Client
+  end
+
   config.server_middleware do |chain|
     chain.add Sidekiq::Congestion::Limiter
+    chain.add SidekiqUniqueJobs::Middleware::Server
```
camallen, line 23: I think this should probably go ahead of the `Sidekiq::Congestion::Limiter` as that particular middleware only runs server side and relies on redis stored sets for limiting, https://github.com/parrish/sidekiq-congestion. As such the values in the redis sets will eventually time out etc even with no other jobs runs etc as they use redis key TTL values, see https://github.com/parrish/Congestion/pull/1

This means that if the client middleware of UniqueJobs runs and say the `Sidekiq::Congestion::Limiter` middleware stops the server callback running it may leave the lock state in an unexpected state. 

Rather i'd put preference on the `SidekiqUniqueJobs::Middleware::Client` to run first / always and then allow the `Sidekiq::Congestion::Limiter` to run 

Details on middleware ordering - https://github.com/mperham/sidekiq/wiki/Middleware#middleware-ordering

Happy to expand on this on a call or here as well.

camallen (approved):
LGTM

config/initializers/sidekiq.rb
yuenmichelle1: ah this was something i did question (the ordering that is) when i was configuring the middleware for SidekiqUniqueJobs. 

thanks for context.

yuenmichelle1:
Tested out a `until_and_while_executing` lock as well using INatImportWorker as a test: 
See: 
Screen Shot 2023-01-17 at 2 45 08 PM

looks to be working 👍

PR #4094, opened 2023-01-25
Rails 6.0 in next version

yuenmichelle1:
Awesomely, all tests pass and app boots up just fine in next dual boot. There will be followup PRs to address Deprecation warnings though:

- [ ] on autoloaded constants

```
DEPRECATION WARNING: Initialization autoloaded the constants Configurable, DesignatorClient, RedirectUriValidator, Serialization, Serialization::PanoptesRestpack, CachedSerializer, ACLSerializer, NoCountSerializer, FilterHasMany, OwnerLinkSerializer, EventStreamSerializers::ProjectSerializer, EventStreamSerializers::UserSerializer, EventStreamSerializers::WorkflowSerializer, EventStreamSerializers::SubjectSerializer, Refinements, Refinements::RangeClamping, TasksVisitor, MediaStorage::AbstractAdapter, and TalkApiClient.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Configurable, for example,
the expected changes won't be reflected in that stale Module object.

`config.autoloader` is set to `classic`. These autoloaded constants would have been unloaded if `config.autoloader` had been set to `:zeitwerk`.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /rails_app/config/environment.rb:5)
A server is already running. Check /rails_app/tmp/pids/server.pid. 
```

- [ ] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (2 levels) in <module:JSONApiRender> at /rails_app/app/controllers/concerns/json_api_render.rb:11)```

- [ ] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (2 levels) in <top (required)> at /rails_app/spec/support/api_response.rb:3)```

- [ ] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (4 levels) in <top (required)> at /rails_app/spec/controllers/home_controller_spec.rb:33)```

- [ ] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (4 levels) in <top (required)> at /rails_app/spec/controllers/home_controller_spec.rb:19)```

- [ ]  ```DEPRECATION WARNING: update_attributes is deprecated and will be removed from Rails 6.1 (please, use update instead) (called from block (3 levels) in <top (required)> at /rails_app/spec/workers/workflows_dump_worker_spec.rb:23)```

# Review checklist


camallen (commented):
LGTM - great the app works on rails 6.0 👏🏼 

perhaps break those deprecation warnings into issue(s) and we can collaborate on getting them fixed.

Issue #4097, opened 2023-01-26
Rails 6 Upgrade Task: Deprecation Warning: Rails 6.1 will return Content-Type header without modification. (home_controller, workflows_dump_worker)


yuenmichelle1:
- [x] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (4 levels) in <top (required)> at /rails_app/spec/controllers/home_controller_spec.rb:33)```

- [x] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (4 levels) in <top (required)> at /rails_app/spec/controllers/home_controller_spec.rb:19)```

- [x]  ```DEPRECATION WARNING: update_attributes is deprecated and will be removed from Rails 6.1 (please, use update instead) (called from block (3 levels) in <top (required)> at /rails_app/spec/workers/workflows_dump_worker_spec.rb:23)```

PR #4098, opened 2023-01-27
Rails 6.0 Upgrade Task: update content_type to media_type in home_controller and change update_attributes to update in workflows_dump_worker_spec

yuenmichelle1:
https://github.com/zooniverse/panoptes/issues/4097

Describe your change here.

# Review checklist


spec/controllers/home_controller_spec.rb
```diff
@@ -16,7 +16,7 @@
       end
 
       it "should default to html" do
-        expect(response.content_type).to eq("text/html")
+        expect(response.media_type).to eq('text/html')
```
yuenmichelle1, line 19: I question whether we still need these tests, knowing that (per Deprecation warning: 
`Rails 6.1 will return Content-Type header without modification.`). Regardless I updated them.

camallen: It's probably worth checking how rspec sets up the get request, specifically the `content_type` header. If it's a sound default then we can probably skip this test, perhaps useful https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec#requesting-a-json-response

spec/workers/workflows_dump_worker_spec.rb
```diff
@@ -20,7 +20,7 @@
         tasks: tasks, pairwise: !workflow.pairwise,
         grouped: !workflow.grouped, prioritized: !workflow.prioritized
       }
-      workflow.update_attributes(updates)
+      workflow.update(updates)
```
camallen, line 23: did this one sneak into this PR? happy for it to stay

camallen (approved):
LGTM

spec/workers/workflows_dump_worker_spec.rb
yuenmichelle1: oops very possible that i lumped it in since i was on the spec file anyway.

PR #4107, opened 2023-02-06
Rails 6.0 in Canary

yuenmichelle1:
Update kubernetes template replicas to 1 for staging canary to test rails 6.0 on canary

Describe your change here.

# Review checklist

### DIFF
--- a/kubernetes/deployment-staging-canary.tmpl
+++ b/kubernetes/deployment-staging-canary.tmpl
@@ -5,7 +5,7 @@ metadata:
   labels:
     app: panoptes-staging-canary-app
 spec:
-  replicas: 0
+  replicas: 1
   selector:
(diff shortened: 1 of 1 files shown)


camallen (approved):
:shipit:  

TBH this type of changes doesn't need a review from a maintainer of the repo. Should we relax the branch restrictions for cases like this to speed up the dev cycle?

yuenmichelle1:
> :shipit:
> 
> TBH this type of changes doesn't need a review from a maintainer of the repo. Should we relax the branch restrictions for cases like this to speed up the dev cycle?

i'm ok with that. 👍 I'm not sure if there is a branch protection rule for these type of changes, but I think I have permissions to just bypass without reviewer if needed.

Issue #4104, opened 2022-10-27
Workflow get request returns unexpected workflow after unlinking subject set from workflow


lcjohnso:
[EDIT 2023-02-03 per @mcbouslog ] Originally opened in Panoptes-Front-End. Transferred to panoptes because we believe this might be a workflow caching issue. With a successful request to unlink a subject set from a workflow, the db reflects the link destroyed, but an immediate subsequent get request returns a workflow that still includes the subject set link (unexpected). The same workflow get request 5 minutes later does not include the subject set link (as expected).

Original comment as opened in PFE...

## Expected behavior
If you unlink a subject set on the project builder's workflow editor page, the set should no longer be linked and the tick box remain unchecked.

## Current behavior
If you unlink a subject set on the project builder's workflow editor page, then reload the page (within ~5 min), the tick box is reticked.  However, if you try to untick the box again, a console error appears: `Couldn't find SubjectSet with 'id'=73874 [WHERE "subject_sets_workflows"."workflow_id" = $1]`

When simultaneously tracking subject_sets_workflows entries in the Panoptes DB, it appears that the untick action successfully unlinks the subject set, however the cached state of the page has not been updated so the tickbox is re-checked on reload.  Interesting, the caching not only affects the project builder web interface, but also the state of the workflow `links` field from the Python client.

Also: it appears that if a second subject set is linked, the state is updated -- i.e., the unlinking of the first subject set is correctly reflected on the workflow editor page (and Client/API response) after reload.

## Steps to replicate
1. Link a subject set via a workflow editor page
2. Unlink the subject set
3. Reload the page -- the unlinked subject set's tick box will be ticked again

## Additional Comments
It is possible the issue is not one having to do with PFE, but rather Panoptes in how it is caching responses (as evidenced by similar behavior shown by Python Client) -- worth investigating that backend aspect early on.

yuenmichelle1:
This could be due to the Rails upgrades that have been undergone. 
See: https://www.bigbinary.com/blog/rails-5-allows-updating-without-updating-timestamps

Cached_serializer utilizes resource's `updated_at` for `serializer_cache_key`. See https://github.com/zooniverse/panoptes/blob/4698c5d6e41a6c627a8b1d927f49671d931ddffd/app/serializers/concerns/cached_serializer.rb#L18

We did see something similar in `update_links` where we had to `touch` the resource  [Rails 4.2] (now in Rails 5 updated to set the `updated_at` time and saving the resource. 
See: 
[...]

PR #4108, opened 2023-02-09
Bug Fix: update destroy_links to update updated_at time of resource

yuenmichelle1:
Related to: https://github.com/zooniverse/panoptes/issues/4104

# Review checklist

### DIFF
--- a/lib/json_api_controller/updatable_resource.rb
+++ b/lib/json_api_controller/updatable_resource.rb
@@ -45,7 +45,7 @@ module JsonApiController
       resource_class.transaction do
         destroy_relation(resource, relation, params[:link_ids])
         # as this resource may not have changed but the linked resources may have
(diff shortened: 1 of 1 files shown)


lib/json_api_controller/updatable_resource.rb
```diff
@@ -44,6 +44,10 @@ def destroy_links
       resource = controlled_resources.first
       resource_class.transaction do
         destroy_relation(resource, relation, params[:link_ids])
+        # as this resource may not have changed but the linked resources may have
+        # ensure we modify the upated_at timestamp to cache bust this resource
```
camallen, line 48: fix my spelling mistake that was copied over 🙊 
```suggestion
        # ensure we modify the updated_at timestamp to cache bust this resource
```

camallen (approved):
LGTM :shipit:

PR #4109, opened 2023-02-10
turn off AR query cache for classification exports

camallen:
this PR attempts to reduce the memory of large classification exports by using the `ActiveRecord::Base.uncached` method to disable the query cache for exports. https://api.rubyonrails.org/classes/ActiveRecord/QueryCache/ClassMethods.html#method-i-uncached

Note: find in batches does already ignore the query cache, https://github.com/rails/rails/issues/31175#issuecomment-345259529 however any queries in the subsequent block will be cached for the duration of the export. 

If this proves useful on the large classification exports then it may be useful across all exports as we don't reuse the cache AR query cache across batches (hence why it's already disabled for `find_in_batches`). 

If it doesn't prove useful we can revert this pr

Some related reading:
- https://github.com/codetriage/CodeTriage/pull/1157
- https://github.com/rails/rails/issues/31175
- https://hustle.bizongo.in/the-adventures-of-sidekiq-part-ii-wip-163b0f438e4e

# Review checklist

### DIFF
--- a/app/models/csv_dumps/classification_scope.rb
+++ b/app/models/csv_dumps/classification_scope.rb
@@ -11,11 +11,23 @@ module CsvDumps
 
     def each
       read_from_database do
-        completed_resource_classifications.find_in_batches do |batch|
-          subject_ids = setup_subjects_cache(batch)
-          setup_retirement_cache(batch, subject_ids)
-          batch.each do |classification|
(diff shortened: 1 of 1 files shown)


yuenmichelle1 (approved):
Looks fine to me. Worth trying out.

Issue #4096, opened 2023-01-26
Rails 6 Upgrade Task: DEPRECATION WARNING: Content-Type header without modification (json_api_render, api_response)


yuenmichelle1:
- [ ] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (2 levels) in <module:JSONApiRender> at /rails_app/app/controllers/concerns/json_api_render.rb:11)```

- [ ] ```DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead. (called from block (2 levels) in <top (required)> at /rails_app/spec/support/api_response.rb:3)```

yuenmichelle1:
I could be wrong on this but I take this deprecation warning as us not being able to set content-type header anymore. which looks like what we are doing in: 
https://github.com/zooniverse/panoptes/blob/27033fb31b550ea9c5db5806019792e19638ecf7/app/controllers/concerns/json_api_render.rb#L11

@zwolf  @camallen  @lcjohnso

camallen:
that content type should only be set for the API routes so removing it from the application controller makes sense to me. 

that api vnd content type should be set correctly on incoming requests and validated here, https://github.com/zooniverse/panoptes/blob/27033fb31b550ea9c5db5806019792e19638ecf7/app/controllers/api/api_controller.rb#L11

yuenmichelle1:
I believe #4135 closes this issue.

PR #4114, opened 2023-02-17
Rails 6.0 issues on Signing out from PFE

yuenmichelle1:
change order of `format` `respond_to` on session destroy. (PFE uses `Accept */*` header when calling `session_controller`'s `destroy` via `DELETE` which causes `ActionController::RespondToMismatchError` errors, since in Rails `respond_to` will take the first format as priority. https://github.com/rails/rails/issues/40023#issuecomment-673675138). 

it seem we have done this in other controllers (i.e. prioritizing format.json vs format.html), eg. `passwords_controller`.

but still worth checking to see if we hit this error elsewhere. 

(EDIT: Testing  if there are still similar issues: Will append below as I test/see issues)

**RELATED ISSUES FOUND WHILE TESTING** 
- [ ] It does look like updating password from profile settings page also has issues. 

it does look like https://github.com/zooniverse/panoptes/issues/4096 this deprecation warning might have something to do with these issues. 

# Review checklist

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -19,8 +19,8 @@ class SessionsController < Devise::SessionsController
 
   def destroy
     respond_to do |format|
-      format.html { super }
       format.json { destroy_from_json }
+      format.html { super }
     end
   end
 


camallen (approved):
LGTM - to me this means the default for any `Accept: */*` request will now be JSON response not HTML? 

> PFE uses `Accept: */*` header when calling session_controller's destroy via DELETE which causes

Why is PFE not using a JSON accept header request here?! that seems like an error from the PFE client

eatyourgreens:
`auth.signOut()` should be specifying JSON. Here are the headers that it’s supposed to send with the DELETE request. 

We can definitely fix the API client if it’s sending the wrong headers. 

https://github.com/zooniverse/panoptes-javascript-client/blob/143a06cb527d973392d0842e9b943febabe8d59c/lib/config.js#L67-L70

yuenmichelle1:
> `auth.signOut()` should be specifying JSON. Here are the headers that it’s supposed to send with the DELETE request.
> 
> We can definitely fix the API client if it’s sending the wrong headers.
> 
> https://github.com/zooniverse/panoptes-javascript-client/blob/143a06cb527d973392d0842e9b943febabe8d59c/lib/config.js#L67-L70

That is strange...by the looks of the javascript client it should send `application/json` as accept header..but that is not what is displayed. 
Screen Shot 2023-02-18 at 7 06 44 PM

For now, I will merge this change since it should not change overall behavior (and fix the incoming Rails 6 version of our app), but still looking into issues with new password setting.

eatyourgreens:
I don’t know if this is related, but FrontDoor strips `Accept` headers from requests on cached routes. Which basically means we can’t use content negotiation and CDN caching, I think. 
https://github.com/MicrosoftDocs/azure-docs/issues/86858

eatyourgreens:
Definitely an API client bug.
https://github.com/zooniverse/panoptes-javascript-client/issues/197

PR #4128, opened 2023-03-02
update active_interaction to latest version

yuenmichelle1:
See: https://github.com/AaronLasseigne/active_interaction/blob/main/CHANGELOG.md#upgrading-1
There's been some breaking changes from 3.8.3 -> 4.0 which will affect our Operations. 

FWIW, I did some testing on local and things look fine. The change is tiny, but I wanted a second or third set of eyes to ensure I'm not missing anything. Or if anyone had context on why we had a pinned version for this. (I'm assuming its because we were on Rails 4 for a while). 

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ end
 
 source 'https://rubygems.org'
 
-gem 'active_interaction', '~> 3.8.3'
+gem 'active_interaction'
 gem 'active_model_serializers' # Event stream
 gem 'active_record_extended'
 gem 'activerecord-import', '~> 1.4'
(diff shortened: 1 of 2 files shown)


camallen:
> Or if anyone had context on why we had a pinned version for this. (I'm assuming its because we were on Rails 4 for a while).

Looks like it was pinned originally and depenabot upgrade it to it's highest allowed major version
https://github.com/zooniverse/panoptes/commit/c61a49f981287706cb123173ba08b97ae7ca6ae2

camallen (approved):
I think the gem changelog doesn't have any surprises for us as we don't seem to be using the features with breaking changes. 

LGTM :shipit: for testing on staging and we can always roll back.

PR #4134, opened 2023-03-07
Revoke doorkeeper token on signout

yuenmichelle1:
Related: https://github.com/zooniverse/how-to-zooniverse/issues/319

Related PJC PR: https://github.com/zooniverse/panoptes-javascript-client/pull/201

# Review checklist


app/controllers/sessions_controller.rb
```diff
@@ -35,6 +35,7 @@ def create_from_json
   end
 
   def destroy_from_json
+    revoke_access_tokens! if doorkeeper_token
```
camallen, line 38: Perhaps clarify what this method will do - we're revoking all tokens that this user has for the token's linked application, i.e. PFE or other client application.
```suggestion
    revoke_all_user_application_access_tokens! if doorkeeper_token
```

camallen: You're testing for the token existence here and in the revoke method (via safe & operator), perhaps simplify the logic here and push the boolean existence test into the method call?

Personally i'd leave it and remove the safe operator as method calls aren't free so avoiding them is better!

app/controllers/sessions_controller.rb
```diff
@@ -52,4 +53,15 @@ def set_csrf_headers
     response.headers['X-CSRF-Param'] = request_forgery_protection_token.to_s
     response.headers['X-CSRF-Token'] = form_authenticity_token.to_s
   end
+
+  def revoke_access_tokens!
+    application_id_to_revoke = doorkeeper_token&.application_id
```
camallen, line 58: the doorkeeper token should exist here due to the check on token existence before you call this method, either remove the safe operator or add a guard clause first thing to this method
```
return unless doorkeeper_token
```

app/controllers/sessions_controller.rb
```diff
@@ -52,4 +53,15 @@ def set_csrf_headers
...
+    return unless application_id_to_revoke
+
+    user = User.find(doorkeeper_token&.resource_owner_id)
```
camallen, line 61: more safe operators! these should be done once via the guard clause / conditional before the method is called.

app/controllers/sessions_controller.rb
```diff
@@ -52,4 +53,15 @@ def set_csrf_headers
...
+    Doorkeeper::AccessToken.revoke_all_for(
```
camallen, line 62:
```suggestion
    # revoke all tokens that this user owns for the client application this token is linked to, e.g. PFE, FEM, classrooms etc. 
    Doorkeeper::AccessToken.revoke_all_for(
```

camallen (approved):
LGTM - minor comments on the guard clauses and safe operator & usage. 

Nice work 👏🏼

PR #4139, opened 2023-03-09
Rails 6.1   api_response.rb tests fix

yuenmichelle1:
`response.content_type` check returns `application/vnd.api+json; version=1; charset=utf-8` vs `application/vnd.api+json; version=1`
# Review checklist


spec/support/api_response.rb
```diff
@@ -1,6 +1,6 @@
 shared_examples "an api response" do
   it "should return the correct content type" do
-    expect(response.content_type).to eq("application/vnd.api+json; version=1")
+    expect(response.content_type).to include('application/vnd.api+json; version=1')
```
camallen, line 3: Is this missing the `; charset=utf-8` suffix?

camallen (approved):
LGTM if your happy with tests

spec/support/api_response.rb
yuenmichelle1: Yeah. `response.content_type` returns `application/vnd.api+json; version=1` in 6.0 but will return `application/vnd.api+json; version=1; charset=utf-8` in 6.1

PR #4169, opened 2023-03-29
Sign out revoke tokens in a worker

yuenmichelle1:
Related to latency in sign out. Newly added revoking all user's relevant client app tokens, cause slow sign out (especially if user has many tokens...eg. opening up many tabs of the same client). We move the revoking to a worker to hopefully speed up the process. 

# Review checklist

### DIFF
--- a/spec/workers/revoke_tokens_worker_spec.rb
+++ b/spec/workers/revoke_tokens_worker_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe RevokeTokensWorker do
+  let(:worker) { described_class.new }
+  let(:user) { create(:user) }
+  let(:oauth_app) { create(:non_confidential_first_party_app, owner: user) }
+  let(:user_token) { create(:access_token, application_id: oauth_app.id, resource_owner_id: user.id) }
+  let(:user_other_token) { create(:access_token, application_id: oauth_app.id, resource_owner_id: user.id) }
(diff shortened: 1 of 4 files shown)


spec/workers/revoke_tokens_worker_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:oauth_app) { create(:non_confidential_first_party_app, owner: user) }
+  let(:user_token) { create(:access_token, application_id: oauth_app.id, resource_owner_id: user.id) }
+  let(:user_other_token) { create(:access_token, application_id: oauth_app.id, resource_owner_id: user.id) }
+  let(:user_revoked_token) { create(:access_token, application_id: oauth_app.id, resource_owner_id: user.id) }
+  let(:another_user) { create(:user) }
+  let(:other_user_token) { create(:access_token, application_id: oauth_app.id, resource_owner_id: another_user.id) }
+  let(:other_oauth_app) { create(:non_confidential_first_party_app, owner: user) }
+  let(:other_app_token) { create(:access_token, application_id: other_oauth_app.id, resource_owner_id: user.id) }
+
+  before do
```
zwolf, line 17: Instead of this block (excepting the method call), if you need the resource immediately you can use `let!(` to directly invoke the create right then instead of lazily the first time its accessed.

zwolf (approved):
No reason this can't be in a worker, though I am surprised the lookup/delete takes long enough to cause problems. I made one little suggestion, otherwise lgtm.

PR #4174, opened 2023-03-31
remove uglifier gem

yuenmichelle1:
Rails 6 uses ES6 as a standard and no longer supports Uglifier. See: https://www.mintbit.com/blog/rails-5-6-upgrade-es6-uglifier-bug. 
https://github.com/zooniverse/panoptes/actions/runs/4577101400
With Rails 6, default compiler is Webpacker, so it probably is safe to drop uglifier altogether. Why this didn't get caught in building staging Canary...is a mystery. :/ 

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,6 @@ gem 'sidekiq-unique-jobs'
 gem 'standby'
 gem 'stringex', '~> 2.8'
 gem 'strong_migrations'
-gem 'uglifier', '~> 4.2'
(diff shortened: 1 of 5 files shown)


zwolf (approved):
Agreed that this should have been caught by the canary build. Does seem like this should be expected and is probably required. If this looks good on the canary, then the 6.1 deploy to staging PR can get updated and we'll take it from there.

PR #4199, opened 2023-05-10
update membership serializer to send roles

yuenmichelle1:
Part of UserGroups Efforts. Memberships route does not send roles (which we may need to read to check if a user is logged in). 

Note: Due to CachedSerializer, we may have to Rails.cache.clear first. (In the chance there has been a call to memberships that has been cached. Cached responses from previous calls prior to this change will not show roles. )

# Review checklist


zwolf (approved):
LGTM!

PR #4210, opened 2023-06-13
Stats add new stats visibility column to user groups

yuenmichelle1:
Adding stats_visibility column to user groups for new stats service. 

# Review checklist


zwolf (approved):
LGTM. The post-migration task will need to be run manually via GHA against the cluster, so let me know when you wanna throw it against staging first. The runner executes `bundle exec rake [input]`, so that input will need to include the namespace.

PR #4211, opened 2023-06-13
Add collaborators to project communications

lcjohnso:
The recipient list for project-level emails (i.e., notifications for SubjectSet completion events) were previously just the project owner and the "communications" role (a once-used project role that is not currently configurable in project builder). This PR adds collaborators to that recipient list to widen access to project notifications -- especially given that owner accounts are not always actively monitored.


lcjohnso:
PR may need test updates? See https://github.com/zooniverse/panoptes/pull/2626 for relevant tests in that similar-focused PR.

zwolf (approved):
LGTM :shipit:

PR #4216, opened 2023-06-21
Add infinite PG timeout to migrations

zwolf:
Allow extremely long running migrations to run extremely long. Current PG_STATEMENT_TIMEOUT is 65000, and the wait time on the job is set to 24h, but the PG statement should be allowed to run as long as necessary. Ideally not that long, but best to not time it out regardless.

Based on https://github.com/zooniverse/caesar/pull/1467/files. 

# Review checklist


yuenmichelle1 (approved):
🎲

PR #4218, opened 2023-06-23
Allow admins to update settings in project preferences

zwolf:
Typical API auth is skipped for the ProjectPreferencesController's #update_settings method. It's done manually instead, checking a `project.owners_and_collaborators` includes the api_user. This also skips checking if a user is a global admin, though. This PR adds that check and allows admins to update any project's settings to allow them to assist with project building.

# Review checklist


lcjohnso (commented):
Great -- thanks for this! Looks good, but Michelle should review for code accuracy.

yuenmichelle1 (approved):
LGTM

PR #4249, opened 2023-09-29
Tweak admin permissions for ProjectPreferences update_settings

lcjohnso:
Tweaking https://github.com/zooniverse/panoptes/pull/4218 -- run `is_admin?` check on `api_user` instead of `user` so that an admin flag is required in the request, not just that the user has admin privlidges.  This allows for non-admin use of API endpoint by Zoo team members. 

My initial commit is going to break the spec, and I could use assistance in correcting the test.  Not sure if there's a standard case / method of testing for admin flag use.


lcjohnso:
Closing in favor of #4257

PR #4257, opened 2023-10-17
Require admin flag for admin UPP requests 

zwolf:
Admin UPP requests only reflect on whether a user is an admin in the database and does not check whether they included the admin flag in the request, signaling intention to act as one.

This PR uses the `api_user.is_admin?` method directly instead, which checks both `user.admin?` and whether the flag was set in the request. The spec needed just the flag set in the request params.

I though this would be a bigger lift so I started my own branch but it turned out to be as simple as https://github.com/zooniverse/panoptes/pull/4249 so this will supersede that one (thanks @lcjohnso!)

 
# Review checklist


lcjohnso (approved):
LGTM!

Issue #4238, opened 2023-08-25
Field guide's attached images not included in project copy


lcjohnso:
## Current Behavior

When the `ProjectCopier` ([link](https://github.com/zooniverse/panoptes/blob/master/app/workers/project_copy_worker.rb#L7)) is run, the field guide's attached images are not copied over. Tutorial images images are copied over (see https://github.com/zooniverse/panoptes/blob/master/lib/project_copier.rb#L13) but field guide images are not.

## Desired Behavior

Add field guide attached images to the project copier, possibly via the `INCLUDED_ASSOCIATIONS` target list.

zwolf:
Fixed by https://github.com/zooniverse/panoptes/pull/4258

lcjohnso:
Reopening: while media resources are cloned (good), the media item IDs in the field guide's `items` field remain unchanged, so the linked icon images still do not appear in the [lab](https://www.zooniverse.org/lab/30326/guide) for the cloned project or in the [field guide on the project](https://www.zooniverse.org/projects/american-prairie/cameras-for-conservation-ey/about/research) (bad).

Example on prod: [Cameras for Conservation](https://www.zooniverse.org/projects/american-prairie/cameras-for-conservation) / [Cameras for Conservation EY](https://www.zooniverse.org/projects/american-prairie/cameras-for-conservation-ey)
Project Source: ID=24811, https://www.zooniverse.org/lab/24811
Project Destination: ID=30326, https://www.zooniverse.org/lab/30326
Field Guide IDs = 3023 and 3140 for source and destination, respectively (note that both instances `items` values are the same)

attn: @zwolf (not high priority, but still wish this would work better)

Issue #4251, opened 2023-10-04
Talk Users are not created with project copy


lcjohnso:
**Current Behavior:** `ProjectCopier` ([link](https://github.com/zooniverse/panoptes/blob/master/app/workers/project_copy_worker.rb#L7)) does not create Talk roles along with Panoptes project roles, creating a situation where the owner/collab or any project team member cannot moderate or take other non-public actions on their project Talk. 

Example Error: When the project owner tries to add collaborators via the Lab, a generic error message is displayed.
![image](https://github.com/zooniverse/panoptes/assets/2006101/6b5fb0ee-b22b-4ecb-9c17-d4bc25e8a6ec)

However, a look at the network tab shows that the error specifically has to do with Talk permissions, resulting in 401 responses to requests to create new Talk roles for collaborators (likely because the logged in user, the project owner, didn't have admin permissions to add other roles).
![image (1)](https://github.com/zooniverse/panoptes/assets/2006101/308ffc1b-7623-4cc5-adbd-7df0e442de35)

Also: even after the error message in the Lab, after reload the Panoptes aspects of collaborator roles all work correctly -- hence the explanation that the error is _solely_ to do with Talk roles.

**Desired Behavior:** Create Talk roles along with Panoptes roles when `ProjectCopier` runs.

**Potential Solution:** add Talk client call for role creation to `ProjectCopier`, executing actions such as those described in [the Stack Overflow entry](https://stackoverflowteams.com/c/zooniverse/questions/350) regarding project ownership migration:
```
talk_client = TalkApiClient.new
talk_client.roles.create(name: "admin", user_id: project.owner.id, section: "project-#{project.id}")
````

See also: another `ProjectCopier` issue = https://github.com/zooniverse/panoptes/issues/4238

zwolf:
Fixed by https://github.com/zooniverse/panoptes/pull/4258

PR #4258, opened 2023-10-18
Project copying fixes: Talk roles and field guide images

zwolf:
Fixes a couple project copying bugs. 

* Includes Field Guide attached images in project deep copies.
* Creates Talk admin roles for copied projects. Uses the same TalkAdminCreateWorker that fires on project creation.

# Review checklist

### DIFF
--- a/lib/project_copier.rb
+++ b/lib/project_copier.rb
@@ -36,6 +36,9 @@ class ProjectCopier
       # to keep the translations system working with these copied resources
       setup_associated_primary_language_translations(copied_project)
 
+      # Creates Talk roles via background worker
+      TalkAdminCreateWorker.perform_async(copied_project.id)
+
       # return the newly copied project
       copied_project
     end
--- a/lib/project_copier.rb
+++ b/lib/project_copier.rb
@@ -23,7 +23,7 @@ class ProjectCopier
     # Should this all be wrapped in a transaction?
     # to ensure we rollback an sub resource creations,
     # e.g. inband primary lang strings for the associated resources....
-    Project.transaction(requires_new: true) do
+    copied_project = Project.transaction(requires_new: true) do
       copied_project = copy_project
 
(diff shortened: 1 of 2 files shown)


lib/project_copier.rb
```diff
@@ -35,6 +36,9 @@ def copy
       # to keep the translations system working with these copied resources
       setup_associated_primary_language_translations(copied_project)
 
+      # Creates Talk roles via background worker
+      TalkAdminCreateWorker.perform_async(copied_project.id)
```
yuenmichelle1, line 40: I can't remember. Is this a safe operation to run a sidekiq job within a `transaction`?

zwolf: Hm, probably not if the worker is reliant on the mutated state of the db, which this would be. I'll move the call outside the block. 👍

lib/project_copier.rb
```diff
@@ -10,7 +10,8 @@ class ProjectCopier
     :tagged_resources,
     :avatar,
     :background,
-    { active_workflows: %i[tutorials attached_images] }
+    { active_workflows: %i[tutorials attached_images] },
+    { field_guides: %i[attached_images] }
```
yuenmichelle1, line 14: question on this: 
https://github.com/zooniverse/panoptes/pull/4258/files#diff-36605a4c75522dc2019f50020273c1e6b97ae1472374d4877b39eb81526c1d7fR7 (L7) pulls in a shallow copy of field_guides. 

My understanding is that this format `{ field_guides: %i[attached_images] }` is a deeper includes (I'm taking that as it includes a shallow copy of field_guides + anything deeper you tell it to include.) 

Am I off base in thinking that if we can remove L7 since we have L14 [this new line]?

zwolf: Yeah, this is redundant, good catch.

lcjohnso:
Closes #4238 and #4251

PR #4270, opened 2023-12-06
Project Copier: tweaks to initialization

lcjohnso:
Add additional fields to `EXCLUDE_ATTRIBUTES` so that they are not carried over to newly-copied project from original project.


yuenmichelle1 (changes requested):
Hey Toyosi! 

I like your idea of checking that the new copied projects have excluded attributes as whatever the defaults are, but I think the original project also had these attributes as defaults. (View factory [here](https://github.com/zooniverse/panoptes/blob/master/spec/factories/projects.rb) to see what attributes were set for the original project). 
Because of that, I don't think we are truly testing that these attributes are actually not being copied over. 

I think a better test would be to set the excluded attributes of the original project and then check that the copied project's excluded attributes does not equal that of the original project.  (Then we don't have to  add the secondary test about `launched_row_order` and `beta_row_order`)

Maybe something like:
```
it 'does not copy over excluded attributes' do
  project_with_excluded_keys = create(:full_project, classifications_count: 3, classifiers_count: 2, launch_date: Date.yesterday, completeness: 0.5, activity: 1, lock_version: 8)

   other_copied_project = described_class.new(project_with_excluded_keys.id, copyist.id).copy

    ProjectCopier::EXCLUDE_ATTRIBUTES.each do |attr|
        expect(other_copied_project[attr]).not_to eq(project_with_excluded_keys[attr])
    end
end
```


Side note: 
If you did want to check for defaults in addition, I would suggest using a hash as opposed to an array so we don't have to use an index to keep track of what attribute we're comparing.

spec/lib/project_copier_spec.rb
```diff
@@ -41,6 +41,23 @@
       expect(copied_project.configuration['source_project_id']).to be(project.id)
     end
 
+    it 'sets the value of each excluded attribute to their default' do
+      # N.B, to be updated once ProjectCopier::EXCLUDE_ATTRIBUTES values are updated aswell
+      default_values = [0, 0, nil, 0.0, 0, 0]
+      # excluding launched_row_order and beta_row_order as they are primary keys and don't have a fixed default value
```
yuenmichelle1, line 47: Unrelated, but I dont think `launched_row_order` and `beta_row_order` are primary keys. 

I think for projects the only primary key is `id`. BUT I do think `launched_row_order` and `beta_row_order` are indexes of project. (See: [here](https://github.com/zooniverse/panoptes/blob/b91a4c27344c4cf89592e52c1d3977e3c29d966e/db/structure.sql#L3205) and  [here](https://github.com/zooniverse/panoptes/blob/b91a4c27344c4cf89592e52c1d3977e3c29d966e/db/structure.sql#L3233) )

yuenmichelle1:
> Thanks, this makes alot of sense. I initially thought to compare the attributes between both projects but felt to compare with the defaults. But this works well especially with the launched_row_order and beta_row_order indexes

Awesome! Glad it made sense :) @Tooyosi  Whenever you are ready, feel free to re-request my review and I'll take a look at it.

PR #4272, opened 2023-12-20
Add valid_email & confirmed_at to /me-only attributes

zwolf:
Include User model attributes `valid_email` and `confirmed_at` in serialized user objects via the `/me` route. These attributes can then be used to indicate an invalid email or that an email address needs confirmation on the user's profile/settings page.

# Review checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,6 @@ ADD ./Gemfile /rails_app/
 ADD ./Gemfile.lock /rails_app/
 
 RUN bundle config --global jobs `cat /proc/cpuinfo | grep processor | wc -l | xargs -I % expr % - 1`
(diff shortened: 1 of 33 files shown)


yuenmichelle1 (approved):
Looks good. Do we want to add a test on users_controller for `#me` method tests?

zwolf:
Private attrs are specced out already here: https://github.com/zooniverse/panoptes/blob/42e4ec17c5685a1123ff49659e550c149b7d3ae2/spec/serializers/user_serializer_spec.rb#L74-L94

The date comparison failed with a nanosecond-scale difference between the database & object values, hence the explicit setting of the confirmation attrs here.

PR #4280, opened 2024-02-07
implement project preference read settings endpoint

Tooyosi:
Describe your change here.
Endpoint to expose read permissions for Project Preferences

# Review checklist

### DIFF
--- a/spec/controllers/api/v1/project_preferences_controller_spec.rb
+++ b/spec/controllers/api/v1/project_preferences_controller_spec.rb
@@ -197,4 +197,45 @@ RSpec.describe Api::V1::ProjectPreferencesController, type: :controller do
       end
     end
   end
+
+  describe '#read_settings' do
+    let!(:project) { create(:project, owner: authorized_user) }
+    let!(:upp) { create(:user_project_preference, project: project) }
+    let(:run_generic_read) { get :read_settings, params: { project_id: project.id, format: :json } }
+    let(:unauthorised_user) { create(:user) }
+    let(:run_unauthorised_user_read) { get :read_settings, params: { project_id: project.id, user_id: unauthorised_user.id, format: :json } }
+
+    describe 'genetic preferences' do
+      before(:each) do
+        default_request user_id: authorized_user.id, scopes: scopes
+        run_generic_read
+      end
+
+      it 'responds with a 200' do
+        expect(response.status).to eq(200)
+      end
+  
+      it 'returns the correct response data' do
+        json_response = JSON.parse(response.body)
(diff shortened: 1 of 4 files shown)


spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -197,4 +197,56 @@
...
+
+  describe '#read_settings' do
+    let!(:project) { create(:project, owner: authorized_user) }
+    let!(:upp) { create(:user_project_preference, project: project) }
+    let(:run_generic_read) { get :read_settings, params: { project_id: project.id, format: :json } }
+    let(:run_invalid_project) { get :read_settings, params: { project_id: 500, format: :json } }
+    let(:unauthorised_user) { create(:user) }
+    let(:run_unauthorised_user_read) { get :read_settings, params: { project_id: project.id, user_id: unauthorised_user.id, format: :json } }
+
+    describe 'genetic preferences' do
```
zwolf, line 209: geneRic 🧬

lcjohnso:
> The two different before_actions set either the resource instance var (update) or the project UPP scope (read). Both would need to raise an exception if the user is unauthorized, since reading these should be restricted to project owners/collabs and admins. I’m not sure if a user themself should be able to read their own UPP, @lcjohnso?

We do not currently allow access to a user's own UPP resources, and I see no persuasive reason to enable that permission. Unless it is more difficult to exclude rather than include user read permissions, proceed with granting read permissions to project owner/collabs only.

spec/controllers/api/v1/project_preferences_controller_spec.rb
zwolf: ^^ typo

spec/controllers/api/v1/project_preferences_controller_spec.rb
```diff
@@ -234,13 +235,17 @@
...
+      it 'only fetches settings of owned project' do
         default_request user_id: unauthorised_user.id, scopes: scopes
-        run_unauthorised_user_read
+        get :read_settings, params: { project_id: project.id, user_id: unauthorised_user.id, format: :json }
+        expect(response.status).to eq(404)
       end
 
-      it 'only fetches settings of owned project' do
-        expect(response.status).to eq(404)
+      it 'only fetches settings of the authorized user' do
```
zwolf, line 244: For clarity, this isn't 'authorized' user as much as it is 'specified' user, as it's the user_id specified in the params. This is unrelated to that user's authorization since users can't obtain their own UPPs anyway.

PR #4291, opened 2024-02-21
Adjust User group search to mimic user search

yuenmichelle1:
Part of User Groups stats effort. See: FE Stats Project Board here: https://github.com/orgs/zooniverse/projects/36/views/2?pane=issue&itemId=41487549

Want is to mimic user_group admin search to mimic that of user search. (But in our case instead of searching by `user.login` we search by `user_group.display_name` (not a unique field)). 

Compared to the last migration where we added a `tsv` column. https://github.com/zooniverse/panoptes/blob/master/db/migrate/archive/20151210134819_add_tsvector_to_projects.rb there are a couple differences: 
- with the addition of strong_migrations gem post this ^ 2015 migration, we now have to add `safety_assured` on the creation of sql trigger
- I chose to separate each operation (create the column, add it as a generalized inverted index, create the sql trigger) into its own migration file. 

# Review checklist


yuenmichelle1:
Pinging @mcbouslog so that he and I can test against https://github.com/zooniverse/Panoptes-Front-End/pull/6968 once this change is merged into staging

zwolf (approved):
LGTM

Issue #4278, opened 2024-01-26
Reset email validity on change of email address


lcjohnso:
If a user changes their email address, for example via submitting a new address via https://www.zooniverse.org/settings/email, I propose the following actions should be taken:
1. reset email validity (`valid_email`)
    - default to `"valid_email": true`
2. ~~reset email confirmation fields (`confirmed_at`, `confirmation_token`, `confirmation_sent_at`)~~
    - ~~default to `"confirmed_at": null`~~
3. ~~automatically resend confirmation email to trigger verification of new email address~~

UPDATE: we are OK keeping people confirmed as long as they've confirmed at any point. Scope of issue changed to be **only** `valid_email` reset.

lcjohnso:
Zach's suggestion for investigation: follow the chain of actions / workers that current sends "your email has changed" email to the user".  Adding this "change `valid_email` to default = true" action might be a natural thing to add to an existing worker, or as a secondary action triggered along with the "email changed" mailer.

PR #4292, opened 2024-02-22
set valid_email to true on user email change

Tooyosi:
Describe your change here.

# Review checklist

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -52,6 +52,7 @@ class Api::V1::UsersController < Api::ApiController
       super do |user|
         if user.email_changed?
           update_email_user_ids << user.id
+          user.update_attribute(:valid_email, true)
         end
       end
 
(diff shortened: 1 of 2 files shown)


lcjohnso:
Closes #4278

zwolf (changes requested):
The spec needs to be modified a bit to test this behavior correctly. `valid_email` is not set in the user factory, so the database default of 'true' is used when a new user is instantiated. This is why the included test passes even though the user wasn't reloaded from the database: the `valid_email` attribute was true to begin with.

You need to create a new user with an invalid email in your spec:

`let(:invalid_user) { create(:user, valid_email: false) }`

Then, when you check that the attribute was changed in your spec:

`expect(user.reload.valid_email).to be_truthy`

The controller logic looks good, so this way you know that you started with an invalid email and can confirm that it was successfully changed.

zwolf (approved):
LGTM!  For future reference, this is being done instead of using Devise's Reconfirmable module to avoid requiring users who have fixed or changed their email address to confirm. Since the initial confirmation is specifically to weed out spam, this is sufficient.

Issue #4248, opened 2023-09-29
Owner/collab read permissions for Project Preferences


lcjohnso:
**Context:** Thanks to work originating with Gravity Spy and Leveling Up / workflow selection, the capability to _**edit**_ the `settings` field in a user's project preferences was granted to project owners and collaborators, e.g. so they can set `workflow_id: "1934"` to assign a user to a specific "level" according to workflow_id. 

**Problem:** While edit / write permissions are granted, there is currently no way for project owner / collab to _**read**_ project preferences given our current permissions model.  Part of this is by design because the data contained in the UserProjectPreferences resource is mixed: some fields are user-facing (i.e., project-specific email subscription, user preferences / properties), some fields are project-facing (i.e., assigned workflow).

I agree that a project owner/collab should have **read & write permission** for the ProjectPreferences.settings field.  However, whether owner/collab should have read permissions for other fields is not clear.  Fields include: 
- `email_communication`: project-specific mailing list opt-in/out
- `preferences`: FE-facing settings & data, like `selected_workflow` (for default behavior of PFE "Classify" button), minicourse and tutorial progress & completion flags.
- `legacy_count`: Ouroboros-era classification count storage
- `activity_count`: deprecated Ouroboros-era classification count storage (was not final solution; use `legacy_count` instead) 

It would probably be OK to expose any/all of the other ProjectPreference fields to owner/collabs.  In that case a more standard access policy could be used.  Otherwise, access to only the `settings` field, or a subset of all fields, would likely require a field-specific workaround like was used for write permissions.

Links:
- Initial Panoptes PRs for write permissions: #1903, #1914, #1919 
- Related panoptes-python-client code: https://github.com/zooniverse/panoptes-python-client/blob/master/panoptes_client/project_preferences.py

zwolf:
> It would probably be OK to expose any/all of the other ProjectPreference fields to owner/collabs.

Is this the feature you'd like to see added? For a given project, instead of "owners/collabs can write project_preferences.settings only", the policy should be "owners/collabs can read and write project_preferences as though they were the user"? I don't see an issue with that, the only other contents are the ones you described above as well as some counts (activity_count, activity_count_by_workflow).

PR #4299, opened 2024-02-29
4248 ownercollab read permissions for project preferences

Tooyosi:
Describe your change here.

# Review checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem 'p3p', '~> 2.0'
 gem 'panoptes-client'
 gem 'pg', '~> 1.4'
 gem 'pg_search'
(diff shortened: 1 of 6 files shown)


yuenmichelle1:
Related to Issue: https://github.com/zooniverse/panoptes/issues/4248

yuenmichelle1 (approved):
Awesome. Changes look good. 👍 

Not blocking, but I would add a test to ensure other attributes are not being serialized into the response for read/update_settings. 
(Eg. 'read_settings response does not serialize irrelevant attributes') Something along those lines. 

Pinging @zwolf  so he's aware of changes made. (@zwolf  no added work for you here, unless you wanted to add anything).

PR #4328, opened 2024-05-15
Clear unsubscribe token on user deactivation

yuenmichelle1:
Follow up on conversation on Slack thread found here: https://zooniverse.slack.com/archives/C06DCM0V9/p1713893304800329

Scrubbing unsubscribe_token on user deactivation . 

 - Also added rake task to scrub unsubscribe_token for already deactivated users

# Review checklist


zwolf (approved):
TIL that our unsubscribe tokens are deterministic and use Rails' built in verifier! Neat. LGTM.

PR #4331, opened 2024-05-17
append current time to copied project display name

Tooyosi:
Describe your change here.

# Review checklist

### DIFF
--- a/lib/project_copier.rb
+++ b/lib/project_copier.rb
@@ -11,7 +11,7 @@ class ProjectCopier
     launched_row_order
     beta_row_order
   ].freeze
-  
+
   INCLUDE_ASSOCIATIONS = [
     :tutorials,
     :pages,
@@ -61,7 +61,9 @@ class ProjectCopier
     copied_project = project_to_copy.deep_clone include: INCLUDE_ASSOCIATIONS, except: EXCLUDE_ATTRIBUTES
     copied_project.owner = user
 
+    current_timestamp = Time.now.utc.strftime("%Y-%m-%d %H:%M:%S")
     copied_project.display_name += ' (copy)' if user == project_to_copy.owner
+    copied_project.display_name += " #{current_timestamp}"
 
     copied_project.assign_attributes(launch_approved: false, live: false)
     # reset the project's configuration but record the source project id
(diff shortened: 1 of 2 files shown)


lib/project_copier.rb
```diff
@@ -61,7 +61,9 @@ def copy_project
     copied_project = project_to_copy.deep_clone include: INCLUDE_ASSOCIATIONS, except: EXCLUDE_ATTRIBUTES
     copied_project.owner = user
 
+    current_timestamp = Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')
     copied_project.display_name += ' (copy)' if user == project_to_copy.owner
+    copied_project.display_name += " #{current_timestamp}"
```
lcjohnso, line 66: Following method used in workflow_copier.rb (see [workflow_copier.rb#L22](https://github.com/zooniverse/panoptes/blob/master/lib/workflow_copier.rb#L22)), and removing `if user == owner` (that doesn't seem necessary since _all_ workflows will have a unique `copy` tag added.

```suggestion
    copied_project.display_name += " (copy: #{Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')})"
```

spec/lib/project_copier_spec.rb
```diff
@@ -21,8 +21,9 @@
       expect(new_copy.display_name).to include('(copy)')
     end
 
-    it 'has matching attributes' do
-      expect(copied_project.display_name).to eq(project.display_name)
+    it 'appends current timestamp to display_name' do
+      allow(Time).to receive(:now).and_return(Time.utc(2024, 5, 17, 12, 0, 0))
+      expect(copied_project.display_name).to eq("#{project.display_name} 2024-05-17 12:00:00")
```
lcjohnso, line 26: If suggestion above is used, then here is the fix for the spec.

```suggestion
      expect(copied_project.display_name).to eq("#{project.display_name} (copy: 2024-05-17 12:00:00)")
```

lcjohnso (changes requested):
Minor recommendation for copied project display_name.

lcjohnso (approved):
LGTM

lcjohnso:
FYI: you can choose to "commit suggestion" from review comments when they are available -- that might be faster than creating those edits / commits on your own.

Tooyosi:
> FYI: you can choose to "commit suggestion" from review comments when they are available -- that might be faster than creating those edits / commits on your own.

Yeah, i'll do going forward. Just wanted to reconfirm with the tests locallt

PR #4333, opened 2024-05-17
Querying Memberships Fix: ensure distinct memberships when querying for ones own public memberships

yuenmichelle1:
Bug found while working on User Groups Features 

Correspondence on issue found in this Slack thread here: https://zooniverse.slack.com/archives/C010QAPB67J/p1715888434468079?thread_ts=1715813389.436219&cid=C010QAPB67J

**TL;DR; of the issue is that when doing a `GET` request to /memberships (either via `/memberships?user_id=_user_group_id=_` or  via `/memberships/id`) to one's own public membership, response responds with duplicates.** 

- The reason is that when querying memberships (in membership policy) we add a union of any public memberships. This leads to duplicates when one is querying one's own public membership. 
- Because of this duplicates issue, we cannot deactivate memberships through `DELETE` because `ETag` headers will never match the precondition (due to response of `GET` request having dupes). Which results in `412 Precondition Failed` error. 

Changes: 
- Ensure distinct memberships returned in query 
- Updating membership policy specs to ensure duplicates are not return using a `contains_exactly` rspec matching array method vs `includes`  match

# Review checklist

### DIFF
--- a/app/policies/membership_policy.rb
+++ b/app/policies/membership_policy.rb
@@ -20,7 +20,7 @@ class MembershipPolicy < ApplicationPolicy
       when :show, :index
         public_memberships = scope.where(identity: false, state: Membership.states[:active])
                                .joins(:user_group).merge(UserGroup.where(private: false))
-        Membership.union_all(query, public_memberships)
+        Membership.union_all(query, public_memberships).distinct
       else
         query
       end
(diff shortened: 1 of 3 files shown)


zwolf (approved):
LGTM, just a comment on the readability of those long argument lists.

PR #4338, opened 2024-05-30
Allow zoo admins to view join token when querying user group

yuenmichelle1:
Part of User Groups Feature: 
Allow zooniverse admins to view join token when querying user group 

- Updated User Group Serializer `include_join_token?` check to check if current_user `is_admin?`

Pinging @mcbouslog  so he is aware of this upcoming change. 

# Review checklist


Tooyosi (approved):
LGTM

PR #4339, opened 2024-05-31
Remove references to aggregation attribute on Workflows

zwolf:
The workflows table has an `aggregation` attribute that's going to collide with the new relation to the aggregations table. As far as I can tell, the only use for this was to indicate whether associated aggregation resources were "public" or "private". 

This PR includes `aggregation` in the Workflow model's `ignored_columns` and removes references to this column from elsewhere. A subsequent PR will include the migration to remove the column from the database. Because ActiveRecord caches attributes at runtime (and `aggregation` is serialized on workflow resources), this will prevent exceptions between deployment and migration. I also needed to do some short circuiting in the policy specs, but they're all getting removed/rewritten in the batch agg PR.

Mostly this is to satisfy strong_migrations and stick to best practices. The migration PR will be a quick one to review/deploy before the batch agg PR merges.

# Review checklist


zwolf:
Seems most sensible to just deprecate the AggregationsController now in advance of https://github.com/zooniverse/panoptes/pull/4303. This PR should fix the issues causing those specs to fail, and that PR fully refactors the AggregationsController and its spec.

spec/policies/aggregation_policy_spec.rb
```diff
@@ -87,7 +87,7 @@
           expect(resolved_scope).to include(public_aggregation)
         end
 
-        it 'includes aggregations from owned private projects' do
+        xit 'includes aggregations from owned private projects' do
```
yuenmichelle1, line 90: what are the plans with skipped tests? Mainly asking because I can see this easily being forgotten in history as skipped and never addressed again. 

If the plans are to delete these tests, maybe we can do that now? Or maybe the batch aggregation PR will take care of these skipped tests removals/give these specs more context so that we can un-skip the tests. 

Maybe adding a comment or something before the skipped tests to explain what the plan is would be helpful. BUT if this is quickly mitigated by the followup PR, them we can leave as is.

yuenmichelle1 (approved):
Looks fine. I had one quick comment, but definitely not blocking.

spec/policies/aggregation_policy_spec.rb
zwolf: Everything related to aggregations is being refactored by the batch aggregation PR. These bits also had to be skipped here because the notion of a private project was handled by the attribute that this removes. I'll remove them if you prefer, either way it's all getting removed when that PR merges (hopefully very soon!).

PR #4352, opened 2024-06-25
Batch Aggregation mailer

zwolf:
If the params to AggregationsController#update include a change to status, send an email to the associated user that includes the links to the output files if the run was successful. "Complete" here means either succeeded or failed.

# Review checklist

### DIFF
--- a/app/workers/aggregation_completed_mailer_worker.rb
+++ b/app/workers/aggregation_completed_mailer_worker.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class AggregationCompletedMailerWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_high
+
+  def perform(agg_id)
+    aggregation = Aggregation.find(agg_id)
+    AggregationCompletedMailer.aggregation_complete(aggregation).deliver
+  end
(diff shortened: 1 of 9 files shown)


app/workers/aggregation_completed_mailer_worker.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AggregationCompletedMailerWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :data_high
+
+  def perform(agg_id)
+    aggregation = Aggregation.find(agg_id)
+    AggregationCompletedMailer.aggregation_complete(aggregation).deliver
```
yuenmichelle1, line 10: I think `.deliver` might be a deprecated method replaced by `deliver_now` or `deliver_later`

yuenmichelle1 (approved):
Changes look fine. I have one minor comment.

app/workers/aggregation_completed_mailer_worker.rb
zwolf: apparently so! Every other mailer uses it, but I'll just change it here and leave that refactor for later.

Issue #4364, opened 2024-07-16
A copied workflow with survey task does not show images linked to survey task


mcbouslog:
**Summary:** The survey task uses workflow_attached_images, but it appears the workflow copier does not duplicate those images when copying a workflow. This results in a working task & workflow (references to URLs from original workflow remain, can be accessed by the task), but the user cannot edit the images and the task editor interface does not work as expected.

## Expected behavior
- copy workflow with survey task
- go to new workflow, survey task, "Images" tab of survey task editor
- see list of images per task object `images` property

## Current behavior
- do not see list of images per task object `images` property
- because images are not linked to new workflow resource
- to volunteer new survey task loads as expected, because urls in task object `images` property exist (as long as original workflow exists?)
- but project collaborator can't see/delete images in survey task editor "Images" tab

## Steps to replicate
- https://master.pfe-preview.zooniverse.org/lab/1634/workflows/3817?env=staging

## Additional information
- Linked Talk Post: https://www.zooniverse.org/talk/18/3378978
- Possible PFE-based solution (but since API is already responsible for copying, best to make change to Panoptes): when copying a workflow, could link attached images I think [here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/actions/workflow.js#L23) with `.update({ links.attached_images: [workflowData.links.attached_images.ids] })` as per [PJC link docs](https://github.com/zooniverse/panoptes-javascript-client/tree/main/lib/json-api-client#working-with-links), or something along those lines

lcjohnso:
@mcbouslog -- Issue is transferred to Panoptes, and yes seems related to original copier PR. I'll note that we've also recently adjusted what gets duplicated or not for project copier, so perhaps @Tooyosi can take a look at this issue.

PR #4369, opened 2024-07-29
[issue-4364] add attached_images association to workflow copier

Tooyosi:
Describe your change here.
Solves for this [issue](https://github.com/zooniverse/panoptes/issues/4364)

# Review checklist

### DIFF
--- a/lib/workflow_copier.rb
+++ b/lib/workflow_copier.rb
@@ -10,13 +10,17 @@ class WorkflowCopier
     finished_at
   ].freeze
 
+  INCLUDE_ASSOCIATIONS = [
+    :attached_images
+  ].freeze
+
   def self.copy_by_id(workflow_id, target_project_id)
     source_workflow = Workflow.find(workflow_id)
     copy(source_workflow, target_project_id)
   end
 
   def self.copy(source_workflow, target_project_id)
-    copied_workflow = source_workflow.deep_clone(except: EXCLUDE_ATTRIBUTES)
+    copied_workflow = source_workflow.deep_clone(except: EXCLUDE_ATTRIBUTES, include: INCLUDE_ASSOCIATIONS)
     copied_workflow.project_id = target_project_id
     copied_workflow.active = false
     copied_workflow.display_name = "#{copied_workflow.display_name} (copy: #{Time.now.utc})"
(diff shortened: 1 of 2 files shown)


yuenmichelle1 (approved):
Overall change looks ok to me. 👌🏼 

It looks like 
☑️  it passed the spec https://github.com/zooniverse/panoptes/blob/master/spec/lib/workflow_copier_spec.rb#L35-L38 which checks to make sure non-excluded attributes are being included. 
☑️  and it passes your new spec that attached_images be included
☑️ passes the spec that any excluded attributes are set to defaults for the copied workflow. 

OOC, do you know why `deep_cloneable` gem's `#deep_clone` is not including attached images? It feels like it should without us having to explicitly include it. But current behavior in panoptes has proved otherwise. 🤔  (I havent looked too much into this, but was wondering if you have.) @Tooyosi

Tooyosi:
> Overall change looks ok to me. 👌🏼
> 
> It looks like ☑️ it passed the spec https://github.com/zooniverse/panoptes/blob/master/spec/lib/workflow_copier_spec.rb#L35-L38 which checks to make sure non-excluded attributes are being included. ☑️ and it passes your new spec that attached_images be included ☑️ passes the spec that any excluded attributes are set to defaults for the copied workflow.
> 
> OOC, do you know why `deep_cloneable` gem's `#deep_clone` is not including attached images? It feels like it should without us having to explicitly include it. But current behavior in panoptes has proved otherwise. 🤔 (I havent looked too much into this, but was wondering if you have.) @Tooyosi


Yeah, my understanding is it has to be included for it to be copied along, here: https://github.com/moiristo/deep_cloneable?tab=readme-ov-file#association-inclusion

yuenmichelle1 (changes requested):
Super sorry I should have caught this sooner when looking through spec.

Are we certain that `workflow.attached_images` in spec has media (i.e. are we certain `workflow.attached_images` is not an empty array?)  ? 

I worry that the way  we define `media` with `let` and how `let` gets lazily evaluated, that the media is actually not getting associated with the `workflow` until invoked (i.e. until we specifically call out `media` eg. `puts media`). Maybe `let!`  or instead setting/creating `media` within the test `it copies the attached_images` would get media associated to workflow?

Tooyosi:
> Super sorry I should have caught this sooner when looking through spec.
> 
> Are we certain that `workflow.attached_images` in spec has media (i.e. are we certain `workflow.attached_images` is not an empty array?) ?
> 
> I worry that the way we define `media` with `let` and how `let` gets lazily evaluated, that the media is actually not getting associated with the `workflow` until invoked (i.e. until we specifically call out `media` eg. `puts media`). Maybe `let!` or instead setting/creating `media` within the test `it copies the attached_images` would get media associated to workflow?

Thanks for picking this, somehow it seems like it was good originally but i can see the issue. I have moved the creation into the specific test case and checking the count as both items would not be the exact same records (i.e different id, created date e.t.c)

yuenmichelle1 (approved):

PR #4374, opened 2024-08-07
[issue-4351] add logic to notify old user email of email change

Tooyosi:
Describe your change here.

# Review checklist


app/mailers/user_info_changed_mailer.rb
```diff
@@ -1,9 +1,10 @@
 class UserInfoChangedMailer < ApplicationMailer
   layout false
 
-  def user_info_changed(user, info)
+  def user_info_changed(user, info, previous_email=nil)
     @user = user
-    @email_to = user.email
+    @recipient_emails = [user.email]
+    @recipient_emails << previous_email if previous_email.present? && info == 'email'
```
zwolf, line 7: `previous_email.present?` can be shortened to just "if previous_email"

Also I think this line can be moved to inside the `when "email"` block in the case statement below as they both have the same condition.

app/controllers/api/v1/users_controller.rb
```diff
@@ -47,17 +47,11 @@ def me
...
-        end
-      end
-
-      update_email_user_ids.each do |user_id|
-        UserInfoChangedMailerWorker.perform_async(user_id, "email")
+    prev_email = user.email
+    super do |user|
+      if user.email_changed?
+        user.update(valid_email: true)
+        UserInfoChangedMailerWorker.perform_async(user.id, 'email', prev_email)
```
zwolf, line 54: I think this will reintroduce the race condition that the tap/array-with-a-single-id thing seeks to avoid. The save in `update` happens after the block is yielded, so when the worker is called before the `super` block is finished, the user hasn't been updated in the db and the email will go to the previous address. 

Moving the worker call outside of the super block should be sufficient, assuming that `user.email_changed?` is still accurate after the save. It's possible that "changed?" info is lost after the transaction, though, and perhaps that's why the `update_email_user_ids.each` exists. Either way, it's gotta be outside the super block in order to ensure that the new address is saved while you pass along the old one.

zwolf (changes requested):
Couple things to address here. Adding an additional test for "when changing email / also sends an email to the old address" would confirm behavior, so I suggest adding that also.

app/controllers/api/v1/users_controller.rb
Tooyosi: I get now. The user.email_changed? is not very accurate outside the save so i'm passing it through a variable now

Tooyosi:
> Couple things to address here. Adding an additional test for "when changing email / also sends an email to the old address" would confirm behavior, so I suggest adding that also.

I captured this behaviour in the `sends an email to the new and old address if user is valid` case

zwolf:
This behavior looks good. I think if you're introducing changes to the interfaces for both UserInfoChangedMailer and UserInfoChangedMailerWorker that you should add a test for each that includes the `previous_email` being passed around. This will prove that the worker gets it in its list of arguments during the update and that the mailer is being passed an array with both emails in it.

spec/workers/user_info_changed_mailer_worker_spec.rb
```diff
@@ -3,9 +3,18 @@
 RSpec.describe UserInfoChangedMailerWorker do
   let(:project) { create(:project) }
   let(:user) { create(:user) }
+  let(:prev_pass) { 'oldpassword@example.com' }
```
zwolf, line 6: I think this should this be `prev_email` here and below?

spec/workers/user_info_changed_mailer_worker_spec.rb
```diff
@@ -4,8 +4,17 @@
   let(:project) { create(:project) }
   let(:user) { create(:user) }
 
-  it 'should deliver the mail' do
-    expect{ subject.perform(user.id, "password") }.to change{ ActionMailer::Base.deliveries.count }.by(1)
+  context 'email delivery' do
```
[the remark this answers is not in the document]
zwolf: a good idea, if just to shut the hound up later

zwolf (approved):
One more small comment, non blocking. LGTM! :shipit:

Issue #3494, opened 2020-10-25
Remove deprecated email list exporter


zwolf:
Previous use of mailout1 email server (running X, hosted on linode) made use of CSV exports of email lists. Even after migration from AWS to Azure, we continued to export lists to secure S3 buckets. Previously this issue was focused on updating the panoptes email list exports uploader to use the azure media storage adapter (move away from S3; see S3 specific code [here](https://github.com/zooniverse/panoptes/blob/master/app/models/csv_dumps/direct_to_s3.rb).

Now, in the listmonk era where list syncing is performed by https://github.com/zooniverse/listmonk-sync, there is no need to run these email list exports.

Proposed action: halt daily export job (sidekiq cron job) and remove s3 export code (optional, but probably good move to remove aws cli dependencies?)

lcjohnso:
Closed by https://github.com/zooniverse/panoptes/pull/4451

PR #4451, opened 2025-03-25
remove deprecated mail exporter files and cron job

Tooyosi:
Describe your change here.

Removes files related to deprecated email list exporter.
Solves for this [issue](https://github.com/zooniverse/panoptes/issues/3494)
# Review checklist


zwolf (approved):
lgtm. I don't see a pressing reason so include the AwsAdapter in this deletion either, but if we do delete it eventually then it could take all the AWS gems with it.

lcjohnso:
Data in zooniverse-exports s3 bucket (path = s3://zooniverse-exports/[production,staging,test]/email_exports/*) have been deleted, completing this cleanup.

Issue #4435, opened 2025-02-25
Improve long-running user_seen_subject updates


lcjohnso:
During 24-25 panoptes database migration era (single->flexible), the user_seen_subjects table was identified as the source for two issues:
1. Sidekiq job backups: we would occasionally see a series of long-running (~1 min) user_seen_subject workers that would backup the job queue. This led to us scaling up minimum resource limits for sidekiq and adding an alternative pod type to keep queues moving.
2. Bloat: the user_seen_subjects table was observed to have a large, and rapidly increasing size on disk -- from Nov 22 to Feb 17, table size increased from 3026 GB to 5180 GB; migrated table size is 3.2 GB. 

Hypothesis: individual rows in user_seen_subjects are getting HUGE. Example: an individual user has a `subject_ids` array (one of the table's columns) that contains 1.4 million classification ID entries (13 MB when exported to CSV), reflecting a highly active user classifying on a single, persistent workflow (here, Daily Minor Planet's "Validation" workflow). These enormous rows explain the sluggishness of updates (lots of data to write as additional classification IDs are appended to `subject_ids` array) as well as the bloat (perhaps these data blobs are not cleaned up after update).

We should consider API / technical solutions that could alleviate this issues, including but not limited to restructuring that table & `subject_ids` field. Not sure whether the current subject_ids column array is a workable solution given the large data volumes.

PR #4468, opened 2025-05-16
UserSeenSubjects Optimization attempt - grouping updates to the same record together

yuenmichelle1:
Attempt to address: https://github.com/zooniverse/panoptes/issues/4435

Issue: 
pre-flex server Azure Postgres migration, we were experiencing high volumes of UserSeenSubjectsWorker enqueued in Sidekiq queue, which led to sidekiq queues being backed up with multiple UserSeenSubjectsWorker jobs; essentially slowing down any other job that belongs to the same queue. Possible reason for high volumes of UserSeenSubjectsWorker jobs is due to workflows having quick tasks and volunteers classifying those quick tasks at a quick rate (eg. Daily Minor Planet). 
This could lead to latency due to lots of dead tuples due to updates to the same record within a short timeframe. 

To mitigate, we introduce grouping of UserSeenSubjectsWorker jobs that update the same record utilizing sidekiq-grouping https://evilmartians.com/opensource/sidekiq-grouping. 

# Review checklist


zwolf (approved):
This looks good to me. 5 seconds is pretty quick, so it may not be obvious outside of testing that this is getting used. But that's a fine place to start tweaking.

Also If you want to change the rubocop settings to extend/remove the 15 line limit it complains about, all good

PR #4475, opened 2025-05-25
BatchAgg: update filenames for mailer

lcjohnso:
As part of May 2025 updates to Batch Aggregation, edit the mailer to use updated URLs for output files.

Linked to: https://github.com/zooniverse/aggregation-for-caesar/pull/821, https://github.com/zooniverse/panoptes-python-client/pull/328


lcjohnso:
Before merging & deploying: I recommend we delete existing aggregations to start fresh regarding new file upload locations. There are only 7 existing entries on prod, so seems acceptable to delete. We could also move to new expected locations in the `production` storage container (I did one of these for testing). Something to keep in mind.

zwolf (approved):

Issue #3366, opened 2020-06-23
Improve behavior for failed data exports


lcjohnso:
**Current behavior:** Exports tend to fail silently, leaving a never-changing `Export is being generated.` message on the Project Builder's Data Export page.  Project teams are left uncertain whether they should wait (and for how long) before requesting a new export.

**Proposed behavior** (_rough sketch; feedback and refinement welcome_): Pending exports timeout after N hours (24 for workflows and subjects, longer for classifications) and/or after a liveness check on a relevant Sidekiq process comes back false.  After timeout or registered failure, clear `Export is being generated.` message.  Bonus: dead/failed jobs are re-tried.

lcjohnso:
Reopen stale issue!

camallen:
FWIW I figured out why our exports don't retry. It's due to our use of the https://github.com/parrish/sidekiq-congestion. When the exports are stopped and sidekiq re-queued in redis, sidekiq then tries to re-schedule them (retry) and the congestion middleware stops the job from running as it fails the congestion configs. 

We need to modify the way the sidekiq congestion middleware works to have a jobs in flight identifier (that is consistent across the failures / restarts) and check for allow these  in-flight jobs to restart without testing the underlying congestion data / configs.

lcjohnso:
Current status (May 2025): Out of 19785 exports, 356 have `state: creating` metadata even though no export jobs are running, 8 of which are from the past two months.

Based on these numbers, the need for this enhancement isn't high. Since the database migration in early 2025, it seems like most jobs are completing successfully and in reasonable durations (<6 hours). In other words, due to the lack of failures or very long duration export creation jobs, project teams are much less confused about failure vs. still running and whether or not to make a new request and try the export again.

PR #4483, opened 2025-07-21
implement logic to requeue export jobs and update status via cron job

Tooyosi:
Describe your change here.
Solves for this [issue](https://github.com/zooniverse/panoptes/issues/3366)
# Review checklist

### DIFF
--- a/kubernetes/deployment-production.tmpl
+++ b/kubernetes/deployment-production.tmpl
@@ -199,7 +199,7 @@ spec:
               exec:
                 command: ["/bin/bash", "-c", "cp -Ru /rails_app/public/* /static-assets"]
         - name: panoptes-production-nginx
-          image: zooniverse/nginx
+          image: ghcr.io/zooniverse/docker-nginx
           resources:
             requests:
               memory: "50Mi"
--- a/kubernetes/deployment-staging-canary.tmpl
+++ b/kubernetes/deployment-staging-canary.tmpl
@@ -60,7 +60,7 @@ spec:
               exec:
                 command: ["/bin/bash", "-c", "cp -Ru /rails_app/public/* /static-assets"]
         - name: panoptes-staging-nginx
-          image: zooniverse/nginx
+          image: ghcr.io/zooniverse/docker-nginx
           resources:
             requests:
               memory: "100Mi"
--- a/kubernetes/deployment-staging.tmpl
+++ b/kubernetes/deployment-staging.tmpl
@@ -188,7 +188,7 @@ spec:
               exec:
                 command: ["/bin/bash", "-c", "cp -Ru /rails_app/public/* /static-assets"]
         - name: panoptes-staging-nginx
-          image: zooniverse/nginx
+          image: ghcr.io/zooniverse/docker-nginx
           resources:
             requests:
               memory: "25Mi"


Tooyosi:
> My understanding of the goal here is based on #3366: to manually retry jobs that have died without updating their state in panoptes. They’re stuck with “state: creating” and get lost.
> 
> * What is the reasoning behind manually reprocessing the sidekiq dead and retry queues? When a job raises an unhandled exception, sidekiq-status will set it’s status to “failed”—but it’s retries will be automatically handled by sidekiq. This should also be true when a process gets a shutdown signal (and sidekiq-status is set to “interrupted”): the job is pushed back into redis by sidekiq and picked up again by a different process. So these jobs will be or have been retried by sidekiq automatically. They’re able to “successfully” fail to complete for some reason (misconfiguration, error in decoding the task or an annotation, etc) and ideally they would also have raised errors in honeybadger. This PR seems like it will loop these jobs infinitely.
> * I am curious what happens when jobs that kill the pod they’re running on (usually via OOM/137). Does the job status in redis get updated automatically somehow via sidekiq-status? Or does it get stuck on “working” the same way its associated media gets stuck on “creating”? If so, then it’ll never get requeued. You could test this by running a SimulateLongJob in a local worker and killing the pod/process or otherwise terminating the job, then checking the status via sidekiq API in another pod.
> * The metadata update that a medium gets when an export is successfully finished is `medium.metadata["state"] = 'ready’` . But this PR uses `completed` in the medium metadata if it’s got a sidekiq-status of “complete” (or if it doesn’t have a job ID). That might have other consequences, such as rendering the PFE page that lists available exports to download, for instance, so `STATE_COMPLETED` should probably be `ready`. That said, if the sidekiq-status is “complete” but the medium was pulled via a metadata state of “creating|requeued|failed|interrupted”, then it seems like sidekiq has gotten confused and we shouldn’t overwrite the existing medium status.
> * The ‘workflow_classifications_export’ job type has an explicit expiration time of 36 hours. But project classification exports are a superset (all workflows) so they will take even longer than a single workflow and so should use the longer expiration time also. However, the sidekiq-status middleware is configured with an expiration time of 24 hours. I think this means that sidekiq-status will stop tracking these jobs after 24 hours, meaning that the longer 36 hour internal expiration for classification exports won’t actually be used.
> * It might be cool to add the web UI component of sidekiq status: https://github.com/kenaniah/sidekiq-status?tab=readme-ov-file#adding-the-web-interface
> * There’s a lot of logging going on and I think it’ll be pretty noisy, both in the logs and in the code. Pod logs are kind of ephemeral right now so this isn’t a big deal, though. Do note that the app log level is set to `:info` , so the debug messages won’t be printed.
> 
> Very happy to chat about this PR whenever you’re available!

- The rationale behind reprocessing the dead job is for the 'Bonus: dead/failed jobs are re-tried.' part of the original issue. Might be an overkill since jobs would get retried by sidekiq. Especially with the `:retrying` status. Maybe this pr can just be limited to setting the status as failed if they exceed the time period?
- I'll surely check this out
- For the first part of the third comment, i'll change the status to ready. I wanted to confirm this during review aswell. The pr might change significantly aswell maybe we discuss this on the call?
- I used the 36 hours as a placeholder value since the issue says `longer for classifications`.  Is this a good way to check for classifications? If yes i'll adjust the sidekiq-status middleware.
- I'll look into this
- I can reduce the logs aswell

PR #4498, opened 2025-09-12
Improve worker performance of project-level subject exports

Tooyosi:
improve project level subject exports by implementing caching of subject information

Describe your change here.

# Review checklist

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM ruby:2.7-slim-buster
+FROM ruby:2.7-slim
 
 WORKDIR /rails_app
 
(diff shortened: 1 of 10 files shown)


lib/formatter/csv/subject.rb
```diff
@@ -74,11 +81,11 @@ def workflow_ids
       end
 
       def locations
-        {}.tap do |locs|
-          subject.ordered_locations.each_with_index.map do |loc, index|
-            locs[index] = loc.get_url
-          end
-        end.to_json
+        locs = {}
```
zwolf, line 84: Is this change just for style?

zwolf (commented):
Couple questions, specifically about that uncached method, otherwise LGTM. 

I started to write a script that benchmarked the subject export internals so we'd have some numbers ton this, but it's real fiddly to manage locally and ended up being a timesink on top of the review. I think this can go to staging and we can do some testing there.

lib/formatter/csv/subject.rb
Tooyosi: Not entirely, I eliminated the `.map` used since that would give a new array and we really don't need that.  We are relying on the `each_with_index` alone which is more appropriate. And since we are modifying the new loc object within the loop, we no longer need the `{}.tap`

Issue #4497, opened 2025-09-10
Data export fetching ambiguity due to project name change


lcjohnso:
**Expectation:** when a project team requests a project-level data export (e.g., the subjects export) and that export previously existed, the new export CSV will overwrite the previous results, the updated_at timestamp will update, and the metadata.state will change to "ready" when available.

**Problem:** if the project has changed names, the query to the media table will return more than one result due to changes in the `filename` variable in the content_disposition column that is based on the project name (use of this: the filename replaces the hash filename when downloaded). Unfortunately, in some cases the first result is the _older / defunct file_, meaning teams _do not have access_ to the desired, most recent file.

**Example:** for DELVE Dwarf Galaxy Question (project_id = 22365) has two project_subjects_export rows in the media table due to project name change -- OLD filename = "delve-dwarfquest-for-the-public-subjects.csv" vs. CURRENT filename = "delve-dwarf-galaxy-quest-milky-way-neighbors-subjects.csv". 
Panoptes DB Query = `select * from media where linked_id=22365 and linked_type='Project' and type = 'project_subjects_export'`

**Possible Solution:** prevent the creation of multiple rows in the media table for a single type of data export by improving the identification and overwriting of existing rows when new versions of

PR #4499, opened 2025-09-16
clean up duplicate data exports

Tooyosi:
Describe your change here.
Solves for this [issue](https://github.com/zooniverse/panoptes/issues/4497).

Change finds and updates the latest export and deletes any old dupicates
# Review checklist

### DIFF
--- a/app/operations/create_or_update_medium.rb
+++ b/app/operations/create_or_update_medium.rb
@@ -16,12 +16,37 @@ class CreateOrUpdateMedium < Operation
 
   def execute
     media['metadata']["state"] = 'creating'
-    if medium = object.public_send(type)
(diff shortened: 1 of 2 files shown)


Tooyosi:
> I’m a little concerned about the use of .with_lock in the operation. A locked project, workflow, or subject set row can’t be updated via the API or by other automated processes like counters, completeness calcs, version bumps, etc. Seems like the goal is to avoid multiple active attempts to create a new export media. but even in this unlikely case, these would just get cleaned up on subsequent attempts.
> 
> If there’s a risk of blocked & failed updates leading to worker retries or errors from the API, even if it's also unlikely, I’m not sure the lock is necessary.

I agree, i have removed with_lock from the operation

PR #4518, opened 2025-10-23
upgrade doorkeeper and fix broken rails 6.1 issues

Tooyosi:
Describe your change here.
- Upgrading as doorkeeper 4.4 is [not compatible with rails 7](https://git.thape.com.cn/rails/doorkeeper/-/blob/main/CHANGELOG.md#553)
- My aim is to make the upgrades compatible with both rails 6 and 7

**Changes Made**
- Update `Doorkeeper::OAuth::Helpers::ScopeChecker.valid?` with keyword args to match new signature. [See](https://github.com/doorkeeper-gem/doorkeeper/blob/ae2a7a53115b43906ef3ac2f03d1e5f0f927cd65/lib/doorkeeper/oauth/helpers/scope_checker.rb#L39)

- Eager load lib paths
- Add `ClientCredentialsRequest` monkey patch to match
- Add `client_credentials_request` to make parameters hash to match doorkeeper expectation. [See](https://github.com/doorkeeper-gem/doorkeeper/blob/be58d6bffa3890a0f533b31da88763df541db686/lib/doorkeeper/oauth/client_credentials_request.rb#L37)
- Change `ClientCredentialsRequest` to `ClientCredentials`. [See](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.8.2/lib/doorkeeper/oauth/client_credentials/creator.rb)
- Add `error_response` monkeypatch to match 401 error response on invalid gramt/resource owner
# Review checklist

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -4276,7 +4276,7 @@ ALTER TABLE ONLY public.users
 -- PostgreSQL database dump complete
 --
 
-\unrestrict TAYyLkEhW6QLBZicY9koLE4tJDz53MbApYi64pEpOaQcXttYCgyvKlxA6JnSQZY
+
 
 SET search_path TO "$user", public;
 
@@ -4552,5 +4552,4 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20250326191749'),
 ('20250530191528'),
 ('20251027120000');
-
 


yuenmichelle1 (commented):
Per our BE chat yesterday and at @lcjohnso 's request from yesterday's BE chat, I will let Zach do the formal review. 

⚠️ One thing to note are the big changes in doorkeeper from 4 to 5 (should note these changes as we move this into staging and prod and heavily test)
See: https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions#from-4x-to-5x

^ One change you might need with this PR is a migration: `rails generate doorkeeper:confidential_applications`

lcjohnso:
> Per our BE chat yesterday and at @lcjohnso 's request from yesterday's BE chat, I will let Zach do the formal review.

@yuenmichelle1  sorry, this doesn’t match my intention or statement yesterday. I asked Toyosi to tag Zach for courtesy and visibility, but that you should remain primary reviewer (due to your context as reviewer for other upgrade PRs). However, if you are not comfortable reviewing, please flag and we can assign to Zach if necessary.

yuenmichelle1 (approved):

PR #4521, opened 2025-11-03
rename app constants and files to match rails 7 zentiwerk constant

Tooyosi:
- rename app constants and files to match rails 7 zentiwerk constant. Solves this [error](https://github.com/zooniverse/panoptes/actions/runs/18981435155/job/54214723512?pr=4520)
- replace `.keys` with `attribute_names` as the former breaks on rails 7. See [here](https://api.rubyonrails.org/v7.0/classes/ActiveModel/Errors.html#method-i-attribute_names)

Describe your change here.

# Review checklist


Tooyosi:
> Looks fine. I think the docs you linked for the replacement of keys to attribute_names is a rails 8 doc as opposed to a change log or rails 7 doc.

My bad, i updated the comment

PR #4530, opened 2025-11-14
add project filter by languages

yuenmichelle1:
add project filter by languages
- add gin index on projects configuration to speed up query 

# Review checklist


Tooyosi:
Looks good, i'm curious as to how it would work with multiple languages configured. Maybe we can have a test for that? Would it return projects that has atleast one of the languages or strictly the configured languages?

PR #4531, opened 2025-11-15
Fix memcached incr errors for uploaded_subjects_count; normalize cache to raw numeric

Tooyosi:
Solves [this Honeybadjer error](https://app.honeybadger.io/projects/40595/faults/125201173#notice-trace).

The issue occurs because in Rails 7, there is a stricter requirement that any values passed to Rails.cache.increment must have been written with `raw: true`. If the value was written using `Rails.cache.write` or `fetch` without `raw: true`, increment will fail silently and return nil. See [here](https://api.rubyonrails.org/v7.1/classes/ActiveSupport/Cache/MemCacheStore.html#:~:text=Incrementing%20a%20non,nil) , [here](https://github.com/rails/rails/blob/9c29f7076348c881a7ab25bd436f7e0b49df8797/activesupport/lib/active_support/cache/mem_cache_store.rb#L110) and [here](https://github.com/rails/rails/blob/9c29f7076348c881a7ab25bd436f7e0b49df8797/activesupport/lib/active_support/cache/mem_cache_store.rb#L127)

Also see [here](https://github.com/rails/rails/issues/44787).

I tried to normalise in some way the existing cache values to match rails 7 requirements

Describe your change here.

# Review checklist

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -355,5 +355,4 @@ class User < ApplicationRecord
   def subject_count_cache_expiry
     ENV.fetch("UPLOADED_SUBJECTS_COUNT_CACHE_EXPIRY", 1.hour).to_i.seconds
   end
-
 end


app/models/user.rb
```diff
@@ -348,11 +349,60 @@ def favorite_collections_for_project(project_id)
...
     @subjects_count_cache_key ||= "User/#{id}/uploaded_subjects_count"
   end
 
   def subject_count_cache_expiry
     ENV.fetch("UPLOADED_SUBJECTS_COUNT_CACHE_EXPIRY", 1.hour).to_i.seconds
   end
+
+  def normalize_subjects_count_cache!
+    key = subjects_count_cache_key
+    current = Rails.cache.read(key) || Rails.cache.read(key, raw: true)
```
yuenmichelle1: OOC, why are we reading cache one with `raw:true` and one without?

yuenmichelle1 (commented):
I have a couple questions on these changes. 


It does feel like we could probably get away with adding `raw: true` on our cache fetch and it should be good. 

eg. in user.rb  (I have not tested this suggestion but it might work?)

```
def uploaded_subjects_count
    count = Rails.cache.fetch(subjects_count_cache_key, expires_in: subject_count_cache_expiry, raw: true) do
      DatabaseReplica.read('read_user_uploaded_subjects_counts_from_replica') do
        Subject.where(upload_user_id: id).count
      end
    end
    count.to_i
  end
```

Then `.cache.increment()` should actually increment cache as expected. 

The one thing we might have to worry about is previously cached `User/_id_/uploaded_subjects_count` will be of different format, BUT I think cache gets cleared on a new deploy. 

Otherwise we can probably run `Rails.cache.clear`* with caution

Tooyosi:
> I have a couple questions on these changes.
> 
> It does feel like we could probably get away with adding `raw: true` on our cache fetch and it should be good.
> 
> eg. in user.rb (I have not tested this suggestion but it might work?)
> 
> ```
> def uploaded_subjects_count
>     count = Rails.cache.fetch(subjects_count_cache_key, expires_in: subject_count_cache_expiry, raw: true) do
>       DatabaseReplica.read('read_user_uploaded_subjects_counts_from_replica') do
>         Subject.where(upload_user_id: id).count
>       end
>     end
>     count.to_i
>   end
> ```
> 
> Then `.cache.increment()` should actually increment cache as expected.
> 
> The one thing we might have to worry about is previously cached `User/_id_/uploaded_subjects_count` will be of different format. We could version the cache key `User/_id_/v2/uploaded_subjects_count`.
> 
> Otherwise we can probably run `Rails.cache.clear`* with caution.

I thought about this and ideally should be fine. Reason why i did the extra step to fetch without the `raw` is the case when the invalid cache were set in rails 6.1 and we deploy the rails 7 changes. My thought is to fix them this way and maybe when upgrading to 7.2 we can remove most of this and use only `raw: true`. If this makes sense

PR #4550, opened 2026-02-02
prevent devise email sending to non valid email users

Tooyosi:
Describe your change here.
Utilizing [this method](https://www.rubydoc.info/github/plataformatec/devise/Devise%2FModels%2FAuthenticatable:send_devise_notification) by devise to prevent emails sent to users with `valid_email: false`

# Review checklist


lcjohnso:
Looks brilliant! @zwolf can you perform a technical review on this PR?

zwolf (approved):
Just one small question, LGTM!

Issue #4567, opened 2026-03-12
Enable search for organizations


lcjohnso:
We do not have a search bar capable of matching and directing an individual to a particular organization page.

Two use cases where org search can be useful:

1. Primary: As a Zoo team member, I want to be able to find and load an org's admin page so I can toggle the org to "visible" (equivalent to launch approved) so it appears on zooniverse.org/projects. Currently I have to page through five pages of results on https://www.zooniverse.org/admin/organization-status to find the organization of interest (or had the URL directly if I know the org slug), and adding search (like we provide for users and projects on the admin page) would be a significant benefit.
2. Secondary: as a volunteer, I want to search and find a specific organization while on the zooniverse.org/projects. Note: this is not currently needed because there is only one page of orgs, but planning for the eventuality that we add pagination to orgs and search may be the easier path to accessing a particular org.

PR #4589, opened 2026-04-08
Add search for organizations

Tooyosi:
Describe your change here.

Solves this [issue](https://github.com/zooniverse/panoptes/issues/4567)

Add search support to the organizations index endpoint.

Organizations can now be found by:
- exact display name
- fuzzy display name match for queries with 3+ characters

## Changes

- add custom organization search logic in `OrganizationsController`
- override the inherited default free-text search so the organization-specific exact-first behavior is used
- add `pg_search` support to `Organization`
- add `tsv`, trigger, backfill, and indexes for full-text/trigram search
- add controller specs for exact, fuzzy, and visibility-related search cases
- document the new `search` param for organizations

# Review checklist


db/migrate/20260323120200_create_sql_trigger_to_update_organizations_tsv_vector.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateSqlTriggerToUpdateOrganizationsTsvVector < ActiveRecord::Migration[6.1]
```
yuenmichelle1:
```suggestion
class CreateSqlTriggerToUpdateOrganizationsTsvVector < ActiveRecord::Migration[7.2]
```

db/migrate/20260323120100_add_index_tsv_to_organizations.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AddIndexTsvToOrganizations < ActiveRecord::Migration[6.1]
```
yuenmichelle1:
```suggestion
class AddIndexTsvToOrganizations < ActiveRecord::Migration[7.2]
```

db/migrate/20260323120000_add_tsv_to_organizations.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddTsvToOrganizations < ActiveRecord::Migration[6.1]
```
yuenmichelle1:
```suggestion
class AddTsvToOrganizations < ActiveRecord::Migration[7.2]
```

spec/controllers/api/v1/organizations_controller_spec.rb
```diff
@@ -49,7 +49,42 @@
...
+
+          expect(json_response['organizations'].first['id']).to eq(exact_match.id.to_s)
+        end
+
+        it 'matches display name case-insensitively' do
+          get :index, params: { search: exact_match.display_name.upcase }
+
+          expect(json_response['organizations'].first['id']).to eq(exact_match.id.to_s)
+        end
+
```
yuenmichelle1: Can some of these specs be replaced by the shared_examples 'filter by display_name'?

yuenmichelle1 (changes requested):

yuenmichelle1 (approved):

PR #4591, opened 2026-04-13
upgrade docs ruby version to 3.0

Tooyosi:
Describe your change here.

# Review checklist


lcjohnso (changes requested):
Hi @Tooyosi -- Having just done similar updates for the parallel docs setup on Caesar during documentation day, can you bump the container to use `ruby:3.2-slim` instead? In addition, we should also run an update of Gems & test the local build.

lcjohnso (approved):
bundle updated multiple gems, pushed resulting lock file

PR #4594, opened 2026-04-15
fix project filtering bug and add multi org projects to documentation

Tooyosi:
Describe your change here.

## Summary

- Filter projects by `organization_id` through the new `organization_projects` join table
- Avoid `DISTINCT` on the projects index query so `organization_id` filtering works with sorting
- Document that projects can be linked to multiple organizations
- Add project docs for the `organizations` link and `organization_id` filter

## Why

Projects can now belong to multiple organizations through `organization_projects`. The previous project filter joined organizations and used `DISTINCT`, which caused PostgreSQL errors when sorting project results:

Error from staging logs
```
[3c3f403e-85c1-4b2b-adba-916f6e09322a] PG::InvalidColumnReference (ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ..._state" = 0 AND "organizations"."id" = 3 ORDER BY "projects"...
```

# Review checklist

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -889,7 +889,8 @@ Resources related to _Panoptes Projects_.
 ## Project [/projects/{id}{?include,display_name,approved,beta,live}]
 A Project is a collection of subjects and task workflows that a
 volunteer performs on subjects. The project also holds supplementary
-text describing the tasks volunteers perform.
+text describing the tasks volunteers perform. A project can be linked to
+multiple organizations.
 
 It has the following attributes:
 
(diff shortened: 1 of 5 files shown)


yuenmichelle1 (approved):
OH. The overridden `serialize_page` method uses `pluck(:id)` which results in a sql query that will look something like like: `select project.ids from projects etc `  

that plus using `.distinct`  in your organizations filter query will result in a sql query that will look something like:  `select distinct project.ids from projects` 
which will cause issues when you have a sort /`order by` a different field that is not the fields selected (`id`). 

ok. lgtm

PR #4599, opened 2026-04-23
Enable PKCE via Doorkeeper

zwolf:
Enables Proof Key for Code Exchange (PKCE) for the Authorization Code flow via Doorkeeper. This will allow more secure login from mobile, CFE/IFEs, and other SPAs.

* adds the requisite fields to the OAuth grants table
* limits PKCE challenges to S256 (plain not supported)
* adds the new hidden fields to the overridden Doorkeeper view (only if used)

# Review checklist


lcjohnso:
@zwolf -- Can you clarify: by enabling PKCE support, is its use _required_ for the Authorization Code flow, or is it _optional_? I would prefer if the support is optional, so as not to disrupt any current usage / implementations of the flow w/o PKCE, as identifying and communicating relevant affected projects / people may be difficult.

zwolf:
Nothing changes with the current flows. With this PR, S256-based PKCE just becomes an option that gets used when the request includes a `code_challenge` in the auth request.

yuenmichelle1 (approved):

PR #4625, opened 2026-05-20
Create subject counts on collection

yuenmichelle1:
Create a counter_cache column (`subjects_count`) in order to filter out collections by number of subjects.
https://github.com/zooniverse/panoptes/issues/4577

TODO after merge, deploy, migration 

- [ ] Follow up PR to filter collections by min_subjects (PR: #4634)
- [ ] Run Rake task to backfill collections subjects_count

# Review checklist

### DIFF
--- a/lib/tasks/collections.rake
+++ b/lib/tasks/collections.rake
@@ -4,7 +4,7 @@ namespace :collections do
     Collection.in_batches(of: 1_000) do |batch|
       collection_ids = batch.pluck(:id)
 
-      counts = CollectionSubject
+      counts = CollectionsSubject
                .where(collection_id: collection_ids)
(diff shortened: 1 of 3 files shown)


lib/tasks/collections.rake
```diff
@@ -0,0 +1,22 @@
+namespace :collections do
+  desc 'Backfill subjects_count on collections'
+  task backfill_subjects_count: :environment do
+    Collection.in_batches(of: 1_000) do |batch|
+      collection_ids = batch.pluck(:id)
+
+      counts = CollectionSubject
```
Tooyosi: I think this should be `CollectionsSubject`

Tooyosi (approved):

PR #4634, opened 2026-06-02
Filter collections by min num of subjects

yuenmichelle1:
Follow up to: https://github.com/zooniverse/panoptes/pull/4625 
Addresses: https://github.com/zooniverse/panoptes/issues/4577 - Filtering out collections to show collections that have more than either 1,2, or 3 subjects. 

- Implement counter_cache on collection_subjects to count collection's `subjects_count`
- Add a filter param on collections `?min_subjects` which will query collections and filter out any collections with less than min_subjects. 
     - Eg. `https://www.zooniverse.org/api/collections?min_subjects=1` (should return collections that have more than one subject) cc @mcbouslog  
   - Update `#update` when subjects get linked so that AR callbacks are being called. 
   - Overwrite `destroy_relation` and `add_relation` when links are an array type to ensure counter_cache updates and resource is up to date/not stale before further saves . 
   
   
###  For context of our collections:  :

**Total rows: 434,478 rows of collections**

```
Query: select collections.*, count(subject_id) from collections inner join collections_subjects on collections.id = collections_subjects.collection_id group by collections.id

~ 4.7 s
```

**Collections with > 1 subjects: 237,935 rows (434,478 - 237,935 ‎ = 196,543 collections with 1 subject)**

```
select collections.*, count(subject_id) from collections inner join collections_subjects on collections.id = collections_subjects.collection_id group by collections.id having count(subject_id) > 1
Query 2.5 s
```

**Collections > 2 subjects:  191,827 rows (434,478 - 191,827 ‎ = 242,651 collections with 1 or 2 subjects; 242,651 - 196,543 = 46,108 collections with 2 subjects)**

```
select collections.*, count(subject_id) from collections inner join collections_subjects on collections.id = collections_subjects.collection_id group by collections.id having count(subject_id) > 2
Query 4.3 s
```
   
# Review checklist

### DIFF
--- a/lib/tasks/collections.rake
+++ b/lib/tasks/collections.rake
@@ -4,7 +4,7 @@ namespace :collections do
     Collection.in_batches(of: 1_000) do |batch|
       collection_ids = batch.pluck(:id)
 
-      counts = CollectionSubject
+      counts = CollectionsSubject
                .where(collection_id: collection_ids)
                .group(:collection_id)
                .count
--- a/lib/tasks/collections.rake
+++ b/lib/tasks/collections.rake
@@ -5,9 +5,9 @@ namespace :collections do
       collection_ids = batch.pluck(:id)
 
       counts = CollectionSubject
-              .where(collection_id: collection_ids)
-              .group(:collection_id)
-              .count
+               .where(collection_id: collection_ids)
+               .group(:collection_id)
+               .count
 
       collections_to_update = collection_ids.map do |id|
         [id, counts[id] || 0]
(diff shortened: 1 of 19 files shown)


lib/tasks/collections.rake
```diff
@@ -0,0 +1,22 @@
+namespace :collections do
+  desc 'Backfill subjects_count on collections'
+  task backfill_subjects_count: :environment do
+    Collection.in_batches(of: 1_000) do |batch|
+      collection_ids = batch.pluck(:id)
+
+      counts = CollectionSubject
```
Tooyosi: I think this should be CollectionsSubject

Tooyosi:
LGTM, just the comment on the rake file

Tooyosi (approved):

PR #4643, opened 2026-06-15
add new drawing annotation for exports

Tooyosi:
Describe your change here.
Adds  drawing annotation export support for drawing annotations  in classification exports.

## Changes

- Added `Formatter::Csv::V2::DrawingAnnotation`
- Updated the v2 annotation dispatcher to handle workflow tasks with `type: "drawing"`
- Uses `toolIndex` to look up the workflow tool label
 
# Review checklist


zwolf (approved):
Looks great! :shipit: Thought I had some questions but then I read the specs some more and did some console sanity checks, LGTM!

lcjohnso (approved):

